PR #11, opened 2018-01-26
[BUG] Fix appveyor

ryan-roemer:
23femk

- Fix script escaping for Windows
- Only run jest on `src` directory
- Exclude tests from npm publishing


ryan-roemer:
@kenwheeler -- As long as this passes all CIs, I think this is in final form and ready for review.

Issue #16, opened 2018-01-30
Performance optimization: stitching together queries


jamesplease:
Cool library, @kenwheeler ! I'm still checking out the code to see all of the features. I appreciate the simple approach you've taken here, particularly in comparison to the complexity of Apollo's source (I ❤️  Apollo, don't get me wrong! 🙂 )

---

One thing I noticed is that if a user passes in 10 queries, [10 requests are made](https://github.com/FormidableLabs/urql/blob/55de396e3797f8c68aef968334dd0405dbb1c866/src/components/client.tsx#L216-L217). I was thinking that it'd be neat if the queries could be stitched together.

Apollo and Relay can accomplish features like this by breaking the query down into an AST, which is powerful, but probably a little heavy-handed for what this library is doing. Luckily, the GraphQL syntax is (currently, at least) simple enough that there should be a way to do this robustly without an AST by making a few reasonable assumptions.

If you're :+1: to this idea, I'd be happy to put together a proof of concept.

---

**What lead me to this idea?**

I noticed that urql caches at the query level, rather than the root query level. i.e.;

```
{
  queryA
  queryB
}
```

creates one entry in the cache, and not two.

I was going to open a similar issue to this one, suggesting that a lightweight function could be written that splits up the query for caching purposes. But then I realized that because the component accepts an array of queries, that maybe that was the recommended approach instead.

I wonder how hard it'd be to write these two functions: one that joins root queries, and the other that splits them apart.

What do you think?

Thanks for reading! :v:

kenwheeler:
I would love a proof of concept for this!

jmeas:
> This comment is mostly about the effects of this update on request deduplication (described in #30 ) , which is because that is more likely to get merged before this does because I already have the work done there.

Should #30 land before this, then the implementation of `fetch-dedupe` there will need to be updated for this feature.

Instead of passing _one_ request key to `fetch-dedupe`, it will need to support `n` request keys, which represent all of the "sub-requests" contained in the "wrapping" HTTP request.

Once you pass all of the keys into fetch-dedupe, it will loop through its store of keys and compute difference. When the difference is 0, then the request contains new information and needs to be made.

---

An advanced optimization would be to _remove_ the queries that are already in flight from the request, and then pull them out of the cache once all of the requests return, but this would require a bit of tricky code to make it work seamlessly. I wonder if Apollo does that... :)

kitten:
[...]

PR #31, opened 2018-02-03
WIP: Add support for request deduplication

### LINKED ISSUE #30: Feature idea: deduplication of requests
By default, Apollo dedupes requests that contain the same root queries using [`apollo-link-dedup`](https://www.apollographql.com/docs/link/links/dedup.html).

urql already generates [a key for the response cache](https://github.com/FormidableLabs/urql/blob/39d81e2e09de08d3f6e5c4ad9e0d0287efecd555/src/modules/client.ts#L122), and that same key could be used for request deduplication. When response caching and request deduplication use the same underlying key, then your deduplication is as granular as your cache, which is cool 😎 

So, for instance, right now `urql` could deduplicate the _exact same query_, but if #16 is ever implemented, then it could dedupe requests like the following:

```
{
  QueryA
}
```

```
{
  QueryA
  QueryB
}
```

I'm working on a lib called `fetch-dedupe` that abstracts `apollo-link-dedupe`'s functionality into a generic-purpose lib, but for now [here's the code](https://github.com/jmeas/react-request/blob/57a404bf8f6c660004c9f9a084bffaeaef3b436b/src/fetch-dedupe.js).

---

If you were to copy pasta that file into urql, then the diff is actually pretty small to implement request deduplication:

```diff
- fetch(this.url, {
+ fetchDedupe(this.url, {
  body,
  headers: { 'Content-Type': 'application/json' },
  method: 'POST',
  ...fetchOptions,
- })
+ }, {
+   requestKey: hash,
+   responseType: 'json'
+ });
- .then(res => res.json())
```

although you might also want to add a new config that lets folks skip the deduplication, similar to Apollo's.

I'd PR a preview but I've never written TypeScript :v:

> Note: some modifications would likely need to be made to `fetchDedupe` to support sub-request body deduplication as described in #16

jamesplease:
Will resolve #30 

---

**Todo**

- [ ] Make sure it actually works. I didn't test this integration
- [ ] ~~Add typings for `fetch-dedupe`~~
- [ ] Fix/add more typings to the other files. I was just guessin 🙈 
- [ ] ~~Possibly remove unused exports from `fetch-dedupe`~~
- [ ] Update documentation
- [ ] Add tests

---

**What are those other methods from fetch-dedupe for?**

> Update: this section is irrelevant now that fetch-dedupe has been externalized

- `clearRequestCache`:  mostly useful for testing
- `isRequestInFlight`: allows you to provide callback props to consumers letting them know when an actual request will be sent. For instance, [`beforeFetch`](https://github.com/jmeas/react-request#beforefetch)/[`afterFetch`](https://github.com/jmeas/react-request#afterfetch) from `react-request` are powered by this function.
  The value of these methods is that they allow users to synchronize the response with an external cache system. For instance, if someone wanted to throw a normalized resource layer on top of urql, they would want to do that _only_ when the network requests resolve, and not `n` times for `n` deduped requests 🙂


src/modules/fetch-dedupe.ts
```diff
@@ -0,0 +1,93 @@
+// This is a cache of in-flight requests. Each request key maps to an
```
ryan-roemer: Is this file a copy-and-paste of pre-existing work?

If so, can we get a link here of the original, decide if the license is acceptable, and the perhaps add an extra comment header here indicating origin right here.

ryan-roemer: So far, I've just found @jmeas ' _own_ work :) https://github.com/jmeas/react-request/blob/8a57c09be6c3c75df0c6c21f7f72587d01380877/src/fetch-dedupe.js -- but let's get a confirmation and make it explicit in either case.

Thanks!

jmeas: It is from another lib of mine, [request-request](https://github.com/jmeas/react-request/blob/master/src/fetch-dedupe.js), which is licensed under MIT. As the author of the code, I'm OK with no extra header comments here muddying up the source if that's alright with FormidableLabs.

I plan to open sauce the `fetch-dedupe` piece as a separate module, ([npm placeholder here](https://www.npmjs.com/package/fetch-dedupe)) but I haven't gotten around to it. It'll also be MIT licensed.

Whenever that happens (probably in the next 1-2 weeks), this could be updated to depend on that npm module if that is preferable.

ryan-roemer: @jmeas -- Thanks for the response and that sounds fine without extra header comments.

src/modules/client.ts
```diff
@@ -132,13 +134,16 @@ export default class Client {
...
             method: 'POST',
             ...fetchOptions,
+          }, {
+            requestKey: hash,
+            responseType: 'json',
+            dedupe: dedupeRequest
           })
-            .then(res => res.json())
             .then(response => {
               if (response.data) {
```
jmeas: Amusingly, this line needs to become `response.data.data` 🙂 

The `response` argument here isn't a Response object, but is instead the parsed _body_ of a response that happens to have a `data` key. `fetchDedupe` returns you the actual response body with a new attribute, `data`, that is the parsed body. Which is how we end up with `response.data.data`. Awesome

jmeas:
Just an update: `fetch-dedupe@0.1.0` has been published. Would you like me to use that dependency instead of copy pasta'ing the code? I don't have a preference.

kenwheeler:
Ok gonna give this a spin today

kenwheeler:
@jamesplease should we dedupe by default?

jamesplease:
@kenwheeler , I think that request deduplication makes sense to be the default.

Apollo [enables deduplication by default](https://www.npmjs.com/package/apollo-link-dedup). I know less about Relay, but my understanding of its network layer is that it also effectively acts as a dedupe layer (although I think it works a little differently than this). Anyway, all of this is to say that there is a good chance that users of Urql will be familiar with this behavior if they have Relay/Apollo experience.

My personal use-case for deduping is primarily for what I call "reference data." Reference data are the resources that I fetch on the client, but that is only read, and not written to. For instance, a list of languages or a list of countries. About half of the resource types that I fetch in my apps fall into this category. I make components that give me this data in a render prop, so that anywhere that I need, say, countries, I can just do:

```js
<Countries>
  {({ countries, loading, failed }) => {
    console.log('Got some countries!');
  }}
</Countries>
```

`Countries` could be a query component that fetches on mount. If 4 sections of my app need the country data on page load, then 4 of these components will be mounted at the exact same time. Without request deduping, that means 4 identical requests are kicked off. With request deduping, it is 1 request.

---

As an aside: the next version of fetch-dedupe, 3.0 ([PR here](https://github.com/jamesplease/fetch-dedupe/pull/30)), will support response caching. That means, if you would like, you could swap out Urql's response caching for fetch-dedupe's. The first version will be "simple caching" where it caches at the request level, but the version after that, 3.1, will support caching at the sub-request level which would support #16

kitten:
@jamesplease We're planning a new architecture for gql operations, which are similar to Apollo's links. They're called exchanges and I'm currently working on them on the `next` branch.

Part of that work is a `dedupExchange`, which deduplicates in-flight requests. This is what you're also working on here, correct? I'd greatly appreciate it if you could take a look and check whether this solves the problem as well :+1: https://github.com/FormidableLabs/urql/blob/next/src/modules/dedup-exchange.ts

jamesplease:
Cool! If you’re borrowing inspiration from Apollo’s links, then you should have this covered. I don’t have 🕙 to review it just now but I’ll take a look sometime in the future.

Thanks @kitten :v:

PR #32, opened 2018-02-04
Enable --declaration emit for TypeScript.

### LINKED ISSUE #25: Types
Hey, can we publish typescript types for this package? 
I saw it is written in Typescript, so this should be easy. Using it in a Typescript code base not having types is a bummer.

Maybe we can just publish the raw `.ts` files but am not to confident about this option.

DanielRosenwasser:
Fixes #25.

Here's the summary of what I did just so you feel more confident accepting this PR.

First, I added a new `tsconfig.dts.json` that extends the usual config, but turns emit back on. Building with `tsconfig.dts.json` will emit `.d.ts` files. Note this also produces `.js` files, but I've updated the `.npmignore` to ignore anything other than `.d.ts` files.

To get `--declaration` working, first I turned on `--declaration` in your `tsconfig.json` (as opposed to just in your `tsconfig.dts.json`, so that the editor can tell you about `--declaration` errors when they happen). I also had to turn `--allowJs` off since it's currently not a supported combo. Finally, I added a hack to get `--declaration` emit working with `noUnusuedLocals` (without needing a `// @ts-ignore`).

Lastly, I added a a `types` field to `package.json` pointing at the produced `index.d.ts`, so that when TypeScript can find the entry point for your types.

I think in the future we'd like to make the process easier (@mhegazy @bowdenk7), since it's slightly more complex with Babel, but I think that's okay. Hopefully this helps the next person who needs to figure it out!

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "main": "lib/index.js",
   "module": "es/index.js",
   "jsnext:main": "es/index.js",
+  "types": "types/index.d.ts",
   "scripts": {
     "coverage": "jest --coverage",
     "clean-lib": "rimraf lib",
@@ -14,7 +15,9 @@
     "clean-es": "rimraf es",
     "build-es": "npm run clean-es && builder run build-babel -- -d es",
     "watch-es": "watch \"npm run build-es\" src/ -d",
-    "build": "builder concurrent --buffer build-lib build-es",
+    "clean-types": "rimraf types",
+    "build-types": "npm run clean-types && tsc -p ./tsconfig.dts.json",
+    "build": "builder concurrent --buffer build-lib build-es build-types",
     "demo:start:graphql-server": "node example/src/server/index.js",
     "demo:start:app":
       "webpack-dev-server --hot --inline --config \"example/webpack.config.js\"",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -14,7 +15,9 @@
     "clean-es": "rimraf es",
     "build-es": "npm run clean-es && builder run build-babel -- -d es",
     "watch-es": "watch \"npm run build-es\" src/ -d",
-    "build": "builder concurrent --buffer build-lib build-es",
+    "clean-types": "rimraf types",
+    "build-types": "npm run clean-types && tsc -p ./tsconfig.dts.json",
```
ryan-roemer: Does this still work if the command is instead:

```js
 "build-types": "npm run clean-types && tsc -p tsconfig.dts.json",
```

(The `./` is problematic for windows cmd users IIRC)

kenwheeler: it should, trying locally

DanielRosenwasser: As a Windows user, it wasn't problematic for me, but you are the Johnny Cash of your dot-slash.

ryan-roemer: @DanielRosenwasser -- I didn't actually check your user profile until Ken told me -- and it's fair to say that any of your command suggestions should be good to go cross-platform-wise unequivocally 😄 

I must say that

> Johnny Cash of your dot-slash

is truly a priceless gem that I hope enters our lexicon with a lot more regularity 😆

PR #35, opened 2018-02-04
Remove dependency on experimental TypeScript build.

DanielRosenwasser:
Doesn't look like you're using pipeline; feel free to turn this down, but I'd rather not accidentally break your build from deleting a branch. ;)


kenwheeler:
My man. 💯🔥👏

PR #48, opened 2018-02-18
Use loose mode for babel-transforms

Andarist:
This outputs less bytes & generally should be more performant, although it is not entirely spec-compliant (I'd say we shouldn't really care much about those edge cases).


kenwheeler:
@ryan-roemer do we have a policy on loose mode?

ryan-roemer:
@kenwheeler -- Let's check what the min+gz difference is, and I'll look up current best practice on loose mode.

ryan-roemer:
Here's a diff with `-Naurw` flags applied (aka ignores whitespace) in `lib`: https://gist.github.com/ryan-roemer/b59b87feaf12e80677bb508ee54ca039

.babelrc.js
```diff
@@ -0,0 +1,26 @@
...
+    '@babel/proposal-pipeline-operator',
+    [
+      'module-resolver',
+      {
+        extensions: ['.js', '.jsx', '.ts', '.tsx'],
+        root: ['.'],
+      },
+    ],
+    test && './coverage-fix.js',
+    cjs && '@babel/transform-modules-commonjs',
```
ryan-roemer: We should remove this and instead above do:

```
['@babel/env', { modules: cjs, loose }],
```

Or, perhaps more consistently refactor to:

```js
const modules = BABEL_ENV === 'commonjs' || test;
// STUFF
['@babel/env', { modules, loose }],
```


More generally, I try and always steer folks away from taking a dependency in a root `.babelrc` on something that is **not** explicitly in `ROOT/package.json`. And this one isn't -- https://github.com/FormidableLabs/urql/blob/master/package.json#L70-L79

Instead we're rely on `node_modules` flattening which can break in odd cases at odd times because something that works with relying on a flattened location is just "accidentally" working :)

ryan-roemer:
@kenwheeler -- It looks like [react](https://github.com/facebook/react/blob/master/.babelrc#L12) and [redux](https://github.com/reactjs/redux/blob/master/.babelrc#L13) have some amount of `loose` going on, so at a high level seems fine. Why don't you take a look at my diff above for any gotchas, and then I'd say we're otherwise cool to merge this once the code changes I've requested come in...

.babelrc.js
Andarist: I'd used to do this - steering `modules`, but lately I've ditched it, because `transform-modules-commonjs` supports `loose` mode too (notice that I havent used `loose` for separate `transform-modules-commonjs`), but its gain is IMHO minimal, while at the same time it "breaks" spreading namespaced imports - `__esModule` becomes enumerable. I agree that this is not the best pattern in general, but I have found myself doing it from time to time.

Good note on flattening thing though. I can adjust my PR in 2 ways:
- using `modules` option conditionally
- adding `modules-commonjs` as explicit dependency.

Just let me know which one do u prefer

Andarist:
@ryan-roemer I've left the question regarding requested changes.

.babelrc.js
ryan-roemer: Ah, that makes sense. Can you:

- [ ] Add `@babel/transform-modules-commonjs` as a `devDependencies`
- [ ] Capture the gist of your explanation here in a code comment above this line in the `.babelrc` so that we have more context for future folks who pass by this line of code.

?

Thanks!

Andarist:
@ryan-roemer adjusted the PR, please let me know if there is anything left to do :)

PR #49, opened 2018-02-21
Only emit declaration files with TypeScript

DanielRosenwasser:
In TypeScript 2.8 (and in TypeScript 2.7.2, but low-key) [we added the `emitDeclarationOnly` flag](https://github.com/Microsoft/TypeScript/pull/20735). This flag basically says "I only want to create `.d.ts` files", which is basically perfect for libraries that use Babel to emit JS.

C.C. @mhegazy (and thanks @nojvek!)

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,7 +1,7 @@
 /*
 !/lib
 !/es
-!/types/**/*.d.ts
+!/types
 **/tests/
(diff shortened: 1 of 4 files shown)


kenwheeler:
Niiiiiice

PR #53, opened 2018-03-13
Expose client in Connect child function

tptee:
This PR exposes the client to `Connect`'s child function in case child components want to manually fetch from the client.


ryan-roemer:
Snapshots need to be updated.

PR #55, opened 2018-03-14
Added static hoisting for ConnectHOC

Johan-dutoit:
Static hoisting FTW.

This uses https://github.com/mridgway/hoist-non-react-statics, which does a great job!

### DIFF
--- a/src/components/connect-hoc.tsx
+++ b/src/components/connect-hoc.tsx
@@ -1,4 +1,5 @@
 import React, { Component } from 'react';
+import hoistStatics from 'hoist-non-react-statics';
 import Connect from '../components/connect';
 import { IMutation, IQuery } from '../interfaces/index';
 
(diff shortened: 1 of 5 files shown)


morajabi:
This is amazing! @Johan-dutoit

src/components/connect-hoc.tsx
```diff
@@ -17,22 +18,25 @@ export interface IHOCProps {
 
 function ConnectHOC(opts?: IHOCProps | ((_) => IHOCProps)) {
   return (Comp: any) =>
-    class extends Component {
-      constructor(props) {
-        super(props);
+    hoistStatics(
```
kitten: Small nit; I'd prefer to name the class and to wrap it (i.e. call `hoistStatics`) on return.

Should clean this file up a little tbh 😅

PR #62, opened 2018-04-07
use urql without using components

patcito:
add some instructions to README to use urql without components

### DIFF
--- a/README.md
+++ b/README.md
@@ -410,7 +410,19 @@ mutation($id: ID!) {
   { id: 5 }
 );
 ```
+### use urql without using components
 
+While the goal of this project is to work nicely with react, sometimes you just need to do a mutation or a query without using any components. In this case you can call [client.executeQuery](src/components/client.tsx#L188) directly.
+
+Example:
+
+```javascript
+let variables = {"q": "Did I do that?"}
+let skipCache = true //set this to true if you don't need cache
+client.executeQuery(query(myQuery, variables), skipCache).then(function(data) {
+  console.log(data);
+});
+```
 ## TODO
 
 * [ ] Server Side Rendering


README.md
```diff
@@ -410,7 +410,19 @@ mutation($id: ID!) {
   { id: 5 }
 );
 ```
+### use urql without using components
```
morajabi: I think it's better to change the sentence case to this: "Use Urql without Using Components" (Use title.sh if you're not a native English speaker like me)

README.md
```diff
@@ -410,7 +410,19 @@ mutation($id: ID!) {
...
+
+Example:
+
+```javascript
+let variables = {"q": "Did I do that?"}
+let skipCache = true //set this to true if you don't need cache
+client.executeQuery(query(myQuery, variables), skipCache).then(function(data) {
+  console.log(data);
+});
+```
```
morajabi: This snippet needs some refactoring as well, e.g. change `let` to `const`. Also, we don't need to have a dummy variables object here, we have already documented `query()`. 

Although I don't know if the core team members goal is to document this, as it can be used as a bad practice, most of the times.

README.md
patcito: I didn't capitalize because the previous titles are not capitalized either (mutation, query etc).

morajabi: Those are the API ref, this is a title. 

Look at these: 
> What is urql
> Why this exists
> How its different (typo here tho! @kenwheeler)
> Install
> Getting Started

kenwheeler:
@morajabi I'm going to have our marketing team review the readme for grammar before the next minor release

kenwheeler:
@patcito thanks.homie!

PR #64, opened 2018-04-09
Improve/Fix TS types

jaredpalmer:
This PR does a few things:

- Introduces a new generic TS type `UrqlProps<Data, Mutations>` that works pretty much like Formik's `FormikProps<Values>` and React Router 4's `RouteComponentProps<Params>`, (i.e. you use it to properly type the children of the render prop function). 
- Fixes the example to use the new generic
- Adds detailed comments around the generic for improved editor integration for Urql x TS consumers.

What does it look like?

```tsx
// ...

export interface ITodoQuery {
  todos: Array<{ id: string; text: string }>;
  user: {
    name: string;
  };
}

export interface ITodoMutations {
  addTodo: (input: { text: string }) => void;
  removeTodo: (input: { id: string }) => void;
}

const Home: React.SFC<{}> = () => (
  <Connect
    query={query(TodoQuery)}
    mutation={{
      addTodo: mutation(AddTodo),
      removeTodo: mutation(RemoveTodo),
    }}
    children={({
      cache,
      loaded,
      data,
      addTodo,
      removeTodo,
      refetch,
    }: UrqlProps<ITodoQuery, ITodoMutations>) => {
      return (
        <div>
          {!loaded ? (
            <Loading />
          ) : (
            <TodoList todos={data.todos} removeTodo={removeTodo} />
          )}
          <TodoForm addTodo={addTodo} />
          <button type="button" onClick={() => refetch({})}>
            Refetch
          </button>
          <button
            type="button"
            onClick={refetch.bind(null, { skipCache: true })}
          >
            Refetch (Skip Cache)
          </button>
        </div>
      );
    }}
  />
);

```

More improvements can be made to the type signature `ICache`'s methods.

### DIFF
--- a/src/components/connect.tsx
+++ b/src/components/connect.tsx
@@ -3,7 +3,7 @@ import { ICache, IClient, IMutation, IQuery } from '../interfaces/index';
 import ClientWrapper from './client';
 import { Consumer } from './context';
 
-export interface IConnectProps<Data = {}, Mutations = {}> {
+export interface IConnectProps<Data, Mutations> {
   children: (props: UrqlProps<Data, Mutations>) => ReactNode; // Render prop
   query?: IQuery | IQuery[]; // Query or queries
   mutation?: IMutation; // Mutation map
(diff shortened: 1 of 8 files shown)


src/components/connect.tsx
```diff
@@ -60,7 +60,9 @@ export type UrqlProps<Data, Mutations = {}> = {
   refreshAllFromCache: () => void;
 } & Mutations;
 
-export default class Connect extends Component<IConnectProps> {
+export default class Connect<Data = {}, Mutations = {}> extends Component<
```
jaredpalmer: This is how the TS team suggested I do the default in Formik. It improves the error stack when TS users forget to pass `UrqlProps` a generic for their data.

kenwheeler:
You're the man!

PR #67, opened 2018-04-30
Swap out uuid for counter based subscriptions

kitten:
I'm not sure why uuid was used to determine IDs for subscriptions on
the client. Instead we can use a create-broadcast like approach, where
we just count up the number of created (not active) subscriptions as
IDs.

Note: `uuid` is still necessary as a `devDependency` for the example as far as I can tell.

### DIFF
--- a/src/modules/client.ts
+++ b/src/modules/client.ts
@@ -1,5 +1,3 @@
-import uuid from 'uuid/v4';
-
 import { ICache, IClientOptions, IMutation, IQuery } from '../interfaces/index';
 import { gankTypeNamesFromResponse } from '../modules/typenames';
 import { hashString } from './hash';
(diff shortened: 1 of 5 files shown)


src/modules/client.ts
```diff
@@ -95,14 +95,14 @@ export default class Client {
...
+  ): number {
     // Create an identifier, add callback to subscriptions, return identifier
-    const id = uuid();
+    const id = this.subscriptionSize++;
     this.subscriptions[id] = callback;
     return id;
   }
 
-  unsubscribe(id: string) {
+  unsubscribe(id: number) {
```
kitten: @kenwheeler Couldn't we also adopt the pattern of returning an unsubscribe callback from the `subscribe` method? I'd love the simplicity of applying that.

PR #79, opened 2018-05-15
Improved query subscription updates

kitten:
It should be possible to provide a `query` and a `subscription` and to describe how to update the query's result with a subscription's result. In this PR this new behaviour is introduced and the updates will be described using the `updateSubscription` function.

After an event from the subscription, the ClientWrapper will also invalidate the query's cache.


kitten:
@kenwheeler let's do `v0.3.0-next2` please? 🎊

PR #83, opened 2018-05-22
Fix appveyor.

ryan-roemer:
Had an error wherein the right node version never got installed... Also add node10.


ryan-roemer:
Confirmed -- node6 installs node6

```
Installing node 6.14.2 (x86)...
```

PR #87, opened 2018-05-25
fix: executeMutation output

brajabi:
The Promise should be resolved with the whole object including `{ data, errors }`.

### DIFF
--- a/src/modules/client.ts
+++ b/src/modules/client.ts
@@ -147,7 +147,7 @@ export default class Client {
 
   executeMutation$(
     mutationObject: IQuery
-  ): Observable<IExchangeResult['data']> {
+  ): Observable<IExchangeResult> {
     // Create hash key for unique query/variables
     const { query, variables } = mutationObject;
     const key = hashString(JSON.stringify({ query, variables }));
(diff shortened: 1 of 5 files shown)


src/modules/client.ts
```diff
@@ -147,7 +147,7 @@ export default class Client {
 
   executeMutation$(
     mutationObject: IQuery
-  ): Observable<IExchangeResult['data']> {
+  ): Observable<IExchangeResult> {
```
kitten: This will also need to change in more locations. There should be a client interface and the client wrapper might also refer to it a couple of times

kitten:
Referring to my comment above, some types will need to be changed :)

Some tests also need to change but the unit tests themselves will be able to help you with that. Please refer to Travis and “yarn test” locally

brajabi:
While fixing tests and adding new cases, noticed `executeMutation` already throws in case of GraphQL errors with status 2xx 🤔

 
Maybe this part is throwing? https://github.com/FormidableLabs/urql/blob/next/src/modules/client.ts#L174
a
If it's the case, probably we don't need this breaking change, although that throwing isn't the best way :)

kitten:
@brajabi Anything I can do to help you move this along, so we can evaluate whether the change makes sense? :+1:

morajabi:
Is there anything we can do to land this? :)

PR #89, opened 2018-06-11
Fix server example 'todo' query

SkaterDad:
Found a single `=` instead of `==` in the example todo resolver.

### DIFF
--- a/example/src/server/schema.js
+++ b/example/src/server/schema.js
@@ -50,7 +50,7 @@ const resolvers = {
       return store.todos;
     },
     todo: (root, args, context) => {
-      return store.todos.find(a => a.id == args.id);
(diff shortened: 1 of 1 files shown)


example/src/server/schema.js
```diff
@@ -50,7 +50,7 @@ const resolvers = {
       return store.todos;
     },
     todo: (root, args, context) => {
-      return store.todos.find(a => (a.id = args.id));
+      return store.todos.find(a => a.id == args.id);
```
kitten: Nice catch! Can you change this to a triple equal though please? :)

SkaterDad: You bet!  I mistakenly thought you were using == in other spots.

PR #72, opened 2018-05-08
Do not merge: Next Release

kitten:
**Do not merge please!**

This is just here to collect some comments on the changes and have an overview over all changes 👍 

### Current list of changes

- **Exchanges!** :100:💱 A link-like abstraction to process operations independently from the client
- In-flight request deduplication (via `dedupExchange`): Deduplicates in-flight requests via the operation's key
- Extracted the main-caching logic into `cacheExchange`
- Added cancellation support to `httpExchange`
- Added `CombinedError` that holds `graphQLErrors` and `networkError`
- Added `subscription` prop to `ClientWrapper` and a matching helper
- Added subscription support via `subscriptionExchange` (transport agnostic)
- Added `transformExchange` option to client
- Pass observables all the way down to make cancellation work on unmount

PR #92, opened 2018-06-21
Support injecting custom fetch() implementation

phpnode:
Hi! I needed to be able to use a custom transport for urql and this seems like the least intrusive way to add it. This PR adds an optional `fetch` field to `IClientOptions`, it defaults to the global `fetch()`. 

For my particular use case (running graphql in a webworker) it would actually be better to be able to inject a `makeRequest()` method that abstracts over fetch completely but that's a bigger change and would conflict with `fetchOptions` so I left it out.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ package-lock.json.*
 bundles
 dist
 .rpt2_cache
+.DS_Store
--- a/src/interfaces/client-options.ts
+++ b/src/interfaces/client-options.ts
@@ -2,6 +2,7 @@ import { ICache } from './cache';
 
 export interface IClientOptions {
   url: string;
+  fetch?: (input?: Request | string, init?: RequestInit) => Promise<Response>;
   fetchOptions?: object | (() => object);
   cache?: ICache;
   initialCache?: object;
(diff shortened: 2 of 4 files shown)


kitten:
I think we discussed it before, @kenwheeler, and decided againt it 😅  I proposed it before as well. Let's see if we can land this though :+1:

Unfortunately though it'd need to be built against the `next` branch as that's what is planned to go out as v0.3 and it refactors how all queries work using a new "exchange pipeline". More info here: https://github.com/FormidableLabs/urql/pull/72

phpnode:
@kitten happy to retarget against the `next` branch if there's chance of this landing, will wait for @kenwheeler's input before doing so though

kitten:
@phpnode Btw, pending @kenwheeler's input, your original use-case is probably better implemented as an alternative exchange though maybe?

> For my particular use case (running graphql in a webworker)

Since the `next` branch enables a "custom query pipeline" you could also use that to write an alternative `http-exchange`: https://github.com/FormidableLabs/urql/blob/next/src/modules/http-exchange.ts

kenwheeler:
I'm cool with it. As long as it defaults to global fetch.

kitten:
@phpnode Alright then :+1: it'd be cool to retarget this to `next` and make the change in the `http-exchange` instead

Issue #86, opened 2018-05-25
Queries fail when fetchOptions headers overwritten


ar1a:
I have overriden my fetch options like so

```
const client = new Client({
  url: 'http://localhost:4000/',
  fetchOptions: () => {
    const token = window.localStorage.getItem('token');
    console.log(token);
    if (token) {
      return {
        headers: {
          Authorization: `Bearer ${token}`
        }
      };
    }
    return {};
  }
});
```
And once token isn't null, it makes requests at plain/text instead of application/json. This is tracked down to https://github.com/FormidableLabs/urql/blob/1564d4d713059f5acf25412a7d334209d4d7b102/src/modules/client.ts#L196 in which it spreads over headers. 

I fixed this clientside by adding 
```
          'Content-Type': 'application/json'
``` 
to my fetchOptions, but this could be fixable by deep merging fetchOptions

kitten:
It could be fixed by doing that but I’d say this is 100% expected behaviour. Since deep merging is quite unnecessary and can make it hard to override options correctly it might make more sense to just add an option for “headers”

cc @kenwheeler

cross311:
@kitten Would it be an option to pass in the default fetch options so spread operator can be used on the modified properties?

Looking at the code looks like the defaults are hard coded, so moving up their definition up a bit (minus body) to be passed into fetchOptions if it is a function.  Then continue to do shallow spread afterwords to keep it backwards compatible.

I am game for doing a pull to take a stab at this.

kenwheeler:
@cross311 that seems like a solid enhancement. 💯

ar1a:
right, this should be fixed then according to #96?

kitten:
@ar1a yep, thank you for pushing this change 👍 in retrospect it makes a lot of sense and I should’ve been less criticism. 

Also thanks @cross311 for taking care of it.

Hopefully we can land all changes soon ✨

PR #96, opened 2018-06-28
Maintain base headers, even if fetchOptions provided

cross311:
In response to issue #86

### DIFF
--- a/src/modules/client.ts
+++ b/src/modules/client.ts
@@ -14,7 +14,7 @@ import { cacheExchange } from './cache-exchange';
 import { dedupExchange } from './dedup-exchange';
 import { defaultCache } from './default-cache';
 import { hashString } from './hash';
-import { defaultHeaders, httpExchange } from './http-exchange';
+import { httpExchange } from './http-exchange';
 
 export default class Client {
   url: string;
@@ -22,7 +22,7 @@ export default class Client {
   subscriptionSize: number; // Used to generate IDs for subscriptions
   cache: ICache; // Cache object
   exchange: IExchange; // Exchange to communicate with GraphQL APIs
-  fetchOptions: RequestInit | ((defaults: RequestInit) => RequestInit); // Options for fetch call
+  fetchOptions: RequestInit | (() => RequestInit); // Options for fetch call
   subscriptions: { [id: string]: (ClientEvent) => void }; // Map of subscribed Connect components
 
   constructor(opts?: IClientOptions) {
@@ -91,7 +91,7 @@ export default class Client {
     return {
       fetchOptions:
         typeof this.fetchOptions === 'function'
-          ? this.fetchOptions({ headers: defaultHeaders })
+          ? this.fetchOptions()
           : this.fetchOptions,
       skipCache: !!skipCache,
       url: this.url,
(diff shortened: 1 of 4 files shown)


src/modules/http-exchange.ts
```diff
@@ -23,6 +23,10 @@ const createAbortController = () => {
   return new AbortController();
 };
 
+export const defaultFetchOptions: RequestInit = Object.freeze({
```
kitten: This doesn't significantly change anything as we already have `fetchOptions`, so I think it'd not make sense to customise this per call of fetch as you can just override most stuff. So letting the `httpExchange` override using `fetchOptions` and then introducing `defaultFetchOptions` doesn't add much value.

Also I think it's not worth using `Object.freeze` here.

Anyway, can we introduce defaults for `headers` instead as that's what we wanted to go for originally :)

cross311: @kitten So just do default headers.

As for the freeze, since it is being passed into a function on the outside a developer could do a non-immutable change (ie not using spread operator) and it would have side effects.  To prevent that someone from making that mistake, I froze it so they would get a runtime exception.

src/modules/http-exchange.ts
```diff
@@ -23,6 +23,13 @@ const createAbortController = () => {
   return new AbortController();
 };
 
+// default headers to be used in functional fetch options in 'client.ts'
+// freezing the object so that a consumer of the library does not
+// use a mutable change on headers causing unexpected side effects on other http calls.
+export const defaultHeaders: {} = Object.freeze({
```
kitten: I'm not sure whether this sacrifices some compat or pulls in another polyfill. Either might be bad :/ Gotta check first. But I think it'd be save to just call `fetchOptions` with a fresh default object every time as that's just as save

src/modules/http-exchange.ts
```diff
@@ -46,7 +53,7 @@ export const httpExchange = (): IExchange => operation => {
 
     fetch(url, {
       body,
-      headers: { 'Content-Type': 'application/json' },
+      headers: defaultHeaders,
```
kitten: So sorry for being unclear in the issue and the comment before :(

I'm not sure whether I see the benefit of: `fetchOptions: ((defaults: RequestInit) => RequestInit);` where the input are defaults. I believe we should just keep spreading it in as below (`...fetchOptions`) and below it add:

```js
headers: {
  'Content-Type': 'application/json',
  ...fetchOptions.headers
}
```

I see the intent of deduping it there, but since case-sensitivity doesn't matter, I'd expect the eventual XHR/fetch to dedupe it.

`(defaults: RequestInit) => RequestInit` is nice in theory but I'm not sure it adds a lot of value. Any thoughts?

src/modules/client.ts
```diff
@@ -91,7 +91,7 @@ export default class Client {
     return {
       fetchOptions:
         typeof this.fetchOptions === 'function'
-          ? this.fetchOptions()
+          ? this.fetchOptions({ headers: defaultHeaders })
```
kitten: Let's maybe move this into `http-exchange` so that _if_ we want to go for defaults as an argument we can reconstruct the object every time and don't import `defaultHeaders` here as indicated in the next comment

cross311: Given statements above it makes sense.

src/modules/http-exchange.ts
cross311: I agree.  Seems much cleaner as well and does not put the responsibility on the library user.

cross311:
@kitten Thanks for the great feedback.  I think this is what you were looking for.  I removed any other changes made.  Change looks nice and clean, and users of the library will not have anything new to understand.

PR #97, opened 2018-06-28
nit. use function default parameter

ral51:
This is very minor. Use default function parameter for vars argument. Feel free to reject if you think this is not needed.


kitten:
It’d be great to target this towards the next branch instead since that’s what’ll go out as 0.3 :)

That being said though, we do want to allow null as an input (or falsely values that is) so this changes the behaviour and is a breaking change. Sorry about that, but thank you for taking the time to take a look 👍

PR #102, opened 2018-07-06
Add react router example

kasbah:
This is a copy of the existing example with react-router, some routes and parameter capture added. 

It seems to work fine but one strange thing is the webpack dev server can't deal with a non `/` entry. You have to go to `/` first, you can't start off with `/about/:id`. 

I mainly made this because I was trying to figure out if some issues I was having while using this with [after.js](https://github.com/jaredpalmer/after.js) were issues with react router or after.js (turns out neither, more on that later) so it's no problem if you don't want to add this here. 

Not sure if I am supposed to PR `master` or `next` but this seems to work merged with either.


kitten:
Thank you for your hard work first of all! We've been working hard on `v1` in the meantime and it has now dropped on `master`.

This release unfortunately means new APIs, new behaviour and new examples (multiple ones! :tada:)

So I'd be happy to accept rebased examples for the new version. On the other hand it'd be great to follow something like `next`'s naming convention for examples. That being said, specifically for this example I'm not quite sure whether `react-router` would add any values for the examples? It's completely separate after all.

For now I'll close this together with your other PR, #106. Sorry for not reacting for a while! And thanks again!

PR #104, opened 2018-07-07
Export formatTypeNames for SSR

kasbah:
I have found a way to do SSR with after.js. See https://github.com/jaredpalmer/after.js/pull/153. Mainly it's powered by this function: 

```js
  static async getInitialProps({urql}) {
    if (urql) {
      const q = formatTypeNames(query(QUERY));
      return urql.executeQuery(q).then(props => {
        props.loaded = true;
        return props;
      });
    }
  }
```

So I need `formatTypeNames` for that otherwise the cache/store made server-side isn't using the same queries.

This way of doing SSR probably also works with [next.js](https://github.com/zeit/next.js) as after.js is modelled after that and next.js uses a similar `getInitialProps` method.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -16,5 +16,6 @@ export * from './modules/dedup-exchange';
 export * from './modules/error';
 export * from './modules/http-exchange';
 export * from './modules/subscription-exchange';
+export { formatTypeNames } from './modules/typenames';
 
 export * from './interfaces/index';


kasbah:
I guess a better way of resolving this (at least before stuff around #48 is worked out) would be to make `executeQuery` add typenames. How would you feel about that?

kitten:
@kasbah I think I've seen this on Twitter before and it's a super creative solution, albeit it being a little incomplete :+1:

Anyway, we can avoid this PR, as I have some exciting change for a future version (not scheduled for v0.3) here: https://github.com/FormidableLabs/urql/pull/98

This moves **all caching logic** to the `cacheExchange` that will be introduced in `0.3`. This not only means that the components will have no caching concerns anymore, but also that you can use `executeQuery` for the entire equivalent, cached query.

So feel free to give it a go and let me know how it's working out for you :+1: (you can also DM on Twitter `@_philpl`)

The next logical step for your project/example would probably be to create `urql-next` or `urql-after` and make an HOC like this that accepts the serialised cache from initialProps and creates an urql client from it on the client: https://github.com/zeit/next.js/blob/canary/examples/with-apollo/lib/with-apollo-client.js

For now, I'll close this PR, but let me know what you think :tada:

kasbah:
Ok, makes sense. I like the simplicity of the current implementation but I haven't delved into subscriptions yet so may gain some appreciation for these changes later. I tried to switch to urquelle but it gives me a http 500 when dropped into urql's place (but everything works when I switch back to the v3-beta). 

```
 networkError: Error: Internal Server Error
    at ./node_modules/urquelle/es/modules/http-exchange.js:19:1
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7),
```

Does this mean that `client.executeQuery` will still create a cache without `__typename` for v0.3 though? Would you accept a PR to fix that in the short-term? 

Thanks for the tips on the HOC by the way I hadn't thought of that.

PR #109, opened 2018-07-13
Append typenames via visitor and graphql-ast-types

imranolas:
Updates the typenames utilities to use `graphql/language/visit` and `graphql-ast-types`

### DIFF
--- a/src/interfaces/exchange.ts
+++ b/src/interfaces/exchange.ts
@@ -3,10 +3,17 @@ import Observable from 'zen-observable-ts';
 import { CombinedError } from '../modules/error';
 import { IOperation } from './operation';
 
+type Scalar = string | number | boolean;
+
+export interface IExecutionData {
+  __typename?: string;
+  [field: string]: IExecutionData | Scalar | Scalar[] | IExecutionData[];
+}
+
 // Adapted from: https://github.com/graphql/graphql-js/blob/ae5b163d2e6c124107fa0971f6d838c8a7d29f51/src/execution/execute.js#L105-L114<Paste>
 export interface IExecutionResult {
   errors?: Error[];
-  data?: object;
+  data?: IExecutionData;
 }
 
 export interface IExchangeResult {
--- a/src/interfaces/exchange.ts
+++ b/src/interfaces/exchange.ts
@@ -3,11 +3,11 @@ import Observable from 'zen-observable-ts';
 import { CombinedError } from '../modules/error';
 import { IOperation } from './operation';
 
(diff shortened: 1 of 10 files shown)


src/interfaces/exchange.ts
```diff
@@ -3,10 +3,17 @@ import Observable from 'zen-observable-ts';
 import { CombinedError } from '../modules/error';
 import { IOperation } from './operation';
 
+type Scalar = string | number | boolean;
+
+export interface IExecutionData {
+  __typename?: string;
```
kitten: This seems to cause issues as other types in the library aren’t adapted for it.

This type should be fine as long as we can let the client component retune the result (separate PR?)

But let’s not couple the entire exchange’s type to “__typename” as another cache exchange or otherwise might not be relying on this :) e.g. a schema based cache

src/modules/typenames.ts
```diff
@@ -1,63 +1,32 @@
...
 import { IQuery } from '../interfaces/index';
 
-const TYPENAME_FIELD: FieldNode = {
-  kind: 'Field',
-  name: {
-    kind: 'Name',
-    value: '__typename',
-  },
-};
+const TYPENAME_FIELD: FieldNode = field(name('__typename'));
```
kitten: Noice ❤️

src/modules/typenames.ts
```diff
@@ -1,63 +1,32 @@
...
-        return (
-          selection.kind === 'Field' &&
-          (selection as FieldNode).name.value === '__typename'
-        );
-      });
-
-      if (!exists) {
-        selectionSet.selections.push(TYPENAME_FIELD);
-      }
+function hasTypename(node: SelectionSetNode): boolean {
```
kitten: That’s an awesome addition 👌

src/modules/typenames.ts
```diff
@@ -68,21 +37,24 @@ export function formatTypeNames(query: IQuery) {
...
 
-export function gankTypeNamesFromResponse(response: object) {
+export function gankTypeNamesFromResponse(response: IExecutionData) {
   const typeNames = [];
   getTypeNameFromField(response, typeNames);
   return typeNames.filter((v, i, a) => a.indexOf(v) === i);
 }
 
-function getTypeNameFromField(obj: object, typenames: string[]) {
+function isObject(obj): obj is IExecutionData {
```
kitten: This typecheck is not strictly correct as it’s casting to the type containing the typename field, but the actual typename check is outside of it.

Then again if we remove said type and contain it to this file only it’s not as relevant

PR #115, opened 2018-11-11
Fix setState calls for unmounted components

### LINKED ISSUE #114: Connect causes setState warning after unmount
Hey 👋 

I had a `Connect` component for a mutation inside another `Connect` component which was querying the same type. The mutation caused a cache invalidation and the child `Connect` to unmount. The result is the following React warning in the console:

screen shot 2018-11-11 at 13 33 27

I checked the source code and I think the `Connect` component calls `setState` regardless of whether or not the component is still mounted:

https://github.com/FormidableLabs/urql/blob/70f2c882948171f9d6acee41e2584622d790e59f/src/components/client.tsx#L259-L261

cangoektas:
Fixes #114 

Let me know if you need anything to be changed!

### DIFF
--- a/src/components/client.tsx
+++ b/src/components/client.tsx
@@ -46,8 +46,10 @@ export default class UrqlClient extends Component<IClientProps, IClientState> {
   mutations = {}; // Stored Mutation
   typeNames = []; // Typenames that exist on current query
   subscriptionID = null; // Change subscription ID
+  _isMounted = false; // Used to avoid setState warning after unmounting
 
   componentDidMount() {
+    this._isMounted = true;
     this.formatProps(this.props);
   }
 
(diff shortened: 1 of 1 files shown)


kitten:
Hiya, I'll unwrap this more in the future in the original issue, but I'll post a small summary here first :)

Urql will quite soon be going through a refactor that develops it into a proper solution as a Graphql client for React.

Work has already begun on the next branch and additionally here: https://github.com/FormidableLabs/urql/blob/3b9c2ede23cec47a9accb363e5cc7bf4891045c4/src/components/client.tsx

The biggest change is the introduction of exchanges, similar Apollo's links. This means that in the future version all behaviour will be controlled by Observables and observable factories.

So in that branch the unsubscription takes care of this issue. Hope you understand and sorry for not making this clear while these planned changes were still in flux :)

Apologies and feel free to ask any questions regarding the future changes 🙌

PR #119, opened 2018-11-23
Do no merge: Refactor cache promises (WIP)

andyrichardson:


src/exchanges/dedupe.ts
```diff
@@ -0,0 +1,30 @@
+import { Observable, of } from 'rxjs';
+import { partition, merge, flatMap, tap } from 'rxjs/operators';
+import { Exchange, ExchangeResult, Operation } from '../types';
+
+export const dedupeExchange = (): Exchange => {
```
kitten: - `operation` should not be shared, so `Map` can instead use `operation.key`
- `forward` is called too late (inside `flatMap`)
- maybe use `mergeMap` instead of the "obsolete" `flatMap` alias

Overall, we don't need to share the actual observables (see second point); Because it's an ongoing stream of operations/results we can take advantage of that and add the following two operations:

- On `ops$` do a continuous filter by active operations; can use `Set`
- On the results (`forward(filteredOps$)`) tap and remove operation keys from `Set`

src/exchanges/fetch.ts
```diff
@@ -0,0 +1,16 @@
+import { Exchange } from '../types';
+import { mergeMap } from 'rxjs/operators';
+
+export const fetchExchange = (): Exchange => () => ops$ =>
+  ops$.pipe(
```
kitten: Quick recap: This can mostly use the old implementation (at least the factory) without any modifications except the error behaviour that's emitting error events. It's important we keep `CombinedError` around.

src/lib/typenames.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  for (const prop in obj) {
+    if (typeof obj[prop] !== 'object') {
+      continue;
+    }
+
+    const value = obj[prop];
+    typenames =
+      value.__typename !== undefined
+        ? new Set([...typenames, value.__typename])
```
kitten: Oh no; Recreating `Set` can be really expensive

src/lib/typenames.ts
```diff
@@ -0,0 +1,99 @@
...
+        : getTypeNameFromField(value, typenames);
+  }
+
+  return typenames;
+}
+
+export function gankTypeNamesFromResponse(response: object) {
+  const typeNames = getTypeNameFromField(response);
+  return [...typeNames].filter((v, i, a) => a.indexOf(v) === i);
```
kitten: Unnecessary spread

src/util/setupJest.ts
```diff
@@ -0,0 +1,6 @@
+import * as adapter from 'enzyme-adapter-react-16';
+import * as enzyme from 'enzyme';
+
+(global as any).fetch = jest.fn();
```
kitten: I'd do this separately in the test that needs this, so we can reset the mock before each test run

PR #147, opened 2018-12-03
use ordered imports

andyrichardson:


kitten:
This is actually a rule we've consciously removed before among others :) It doesn't add any value and is basically just noise. If we only order local vs external imports that'd be sweet, but otherwise let's please remove it :)

PR #155, opened 2018-12-10
Fix back-to-front ternary in Connect example

raquelxmoss:
Hi folks, I noticed a back-to-front ternary while I was reading the docs, here's a fix for it 🙂

### DIFF
--- a/README.md
+++ b/README.md
@@ -347,7 +347,7 @@ Example:
 <Connect
   query={query(MyQuery)}
   children={({loaded, data}) => {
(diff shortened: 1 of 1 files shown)


kitten:
Hiya! Sorry for not reacting to this for a while. We've been working on the `v1` version for a while now which has just landed in `master` and will be available and announced as soon as the docs are ready :)

Unfortunately this means that this example has been removed for now, but we'd always appreciate a read of the upcoming docs, which we're writing right now: #181

PR #156, opened 2018-12-18
V0 Hooks | Add experimental API for supporting hooks

blainekasten:
This PR is an experiment in adding an API to support React 16.7 hooks. It is currently exported as `unstable_` in case anything with the spec changes between the alpha and the release. 

The new APIs are as follows:

```js
// useQuery returns an object of network lifecycle values and methods
const { loading, fetching, data, error, refetch } = useQuery(`{ todos { id } }`);

// useMutation returns a function which casts a mutation and returns a promise of the results 
const addTodo = useMutation(`mutation($text: string) { addTodo(text: $text) { id, text } }`, optionalArgs);

// useSubscription is like useQuery without the refetch api
const { loading, fetching, data, error } = useSubscription(`subscription() { todosAdded() { id, text } }`);
```

**Main Changes:**
https://github.com/FormidableLabs/urql/pull/156/files#diff-a7439654f284c715a081aff9841d4a4b
https://github.com/FormidableLabs/urql/pull/156/files#diff-1991fb8ceff143e71300699cd3d9e78a
https://github.com/FormidableLabs/urql/pull/156/files#diff-c11084be6600cdace1d019aba38f6ae8

_Other changes_
- Changed `IExchangeResult` to take a Generic for it's data structure to improve typing. 
  - but lazily added a bunch of `any`'s in other parts of the codebase.
- use react.createContext when available over the polyfill.
     
Things todo in order to ship this feature:

- [ ] decide if this API design is desirable. A potentially controversial aspect is not needing to wrap strings in query/mutation calls.
- [ ] decide if we want to ship this early under the `unstable_` flag or wait until react 16.7 ships
- [ ] decide if this can be merged prior to the v1 rewrite
- [ ] add unit tests _after_ the ecosystem supports hooks (currently can't work)
- [ ] add integration tests


src/components/context.tsx
```diff
@@ -2,9 +2,11 @@ import createReactContext, {
   ConsumerProps,
   ProviderProps,
 } from 'create-react-context';
-import { ComponentClass } from 'react';
+import { createContext, ComponentClass } from 'react';
 
-const context = createReactContext({});
+export const context = createContext
+  ? createContext({})
+  : createReactContext({});
```
blainekasten: hooks does not work with `create-react-context`. So this ternary allows the runtime to use native context if it exists on the users react version.

blainekasten: This is probably something that should ship in master anyhow.

kitten: Isn't `create-react-context` supposed to do that automatically? 🤔 https://github.com/jamiebuilds/create-react-context/blob/master/src/index.js#L5

PR #157, opened 2018-12-19
V1 Hooks | Add experimental API for supporting hooks

blainekasten:
_This branch is a rewrite of #156 in the v1 architecture._

This PR is an experiment in adding an API to support React 16.7 hooks. It is currently exported as `unstable_` in case anything with the spec changes between the alpha and the release.

The new APIs are as follows:

```js
// useQuery returns an object of network lifecycle values and methods
const { loading, fetching, data, error, refetch } = useQuery(`{ todos { id } }`);

// useMutation returns a function which casts a mutation and returns a promise of the results 
const addTodo = useMutation(`mutation($text: string) { addTodo(text: $text) { id, text } }`, optionalArgs);
```
     
Things todo in order to ship this feature:

- [ ] decide if this API design is desirable. A potentially controversial aspect is not needing to wrap strings in query/mutation calls.
- [ ] decide if we want to ship this early under the `unstable_` flag or wait until react 16.7 ships
- [ ] decide if we want to merge this as part of v1 or pre-v1.
- [ ] add integration tests with cypress
- [ ] support subscriptions. v1 currently doesn't have any support for subscriptions.

### DIFF
--- a/src/lib/hook-mutation.ts
+++ b/src/lib/hook-mutation.ts
@@ -1,0 +1,33 @@
+// @ts-ignore
+import { useContext, useEffect, useState } from 'react';
+import { context } from '../components/context';
+import { Client, ExchangeResult, MutationHook } from '../types';
+
+export function useMutation<Args>(
+  mutationQuery: string,
+  passedVars: object = {}
+): MutationHook<Args> {
+  const client = (useContext(context) as any) as Client;
+
+  return variables =>
+    new Promise((resolve, reject) => {
+      const inst = client.createInstance({
+        onChange: result => {
+          if (result.error) {
+            reject(result.error);
+            return;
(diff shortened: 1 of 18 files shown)


src/lib/hook-mutation.ts
```diff
@@ -0,0 +1,33 @@
...
+
+          resolve(result);
+        },
+      });
+
+      inst.executeMutation({
+        query: mutationQuery,
+        variables: {
+          ...passedVars,
+          ...(variables as Object),
```
parkerziegler: Are you sure we want to use capital O "Object" here? I think you probably want small o `object`: https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html

PR #160, opened 2018-12-28
V1 subscriptions

blainekasten:
# Adds Subscriptions to V1

This API is mostly inline with v0 with a few changes.

1. `transformExchange` becomes `forwardSubscription(operation, observable): { unsubscribe: () => void}` in ClientOptions.
2. Connect takes an array of subscriptions instead of just one.
3. Changes the `updateSubscription` callback fn to have a `type` parameter as it's first argument, which is the name of the subscription (Allowing the user to differ between multiple subscription responses).

How to use:

```js
import { createClient } from 'urql';
import { SubscriptionClient } from 'subscriptions-transport-ws';

const subscriptionClient = new SubscriptionClient('ws://localhost:3000', {});

const client = createClient({
  url: 'http://localhost:3001/graphql',
  forwardSubscription(operation, observer) {
    return subscriptionClient.request(operation).subscribe({
      next(data) {
        observer.next({ operation, data, error: null });
      },
      error(error) {
        observer.error({ operation, data: null, error });
      },
    });
  }
});
```

## Internally

Internally a few notable things are happening
1. Exchanges are given `clientOptions` as a new key in the argument.
1. `ClientInstance` gets an `executeSubscription` method
1. `ClientInstance` gets an `executeUnsubscribeSubscription` method
1. `ClientInstance` configuration has a new callback method specifically for subscriptions: `onSubscriptionUpdate`
1. a new exchange is created for subscriptions
1. `operationName` internally are switched from a string to an `enum`

## other stuff

1. Backend is updated to support subscriptions


src/lib/client.ts
```diff
@@ -1,4 +1,4 @@
-import { Observable, Subject, Subscription } from 'rxjs';
+import { Observable, Subject, Subscription as RxSubscription } from 'rxjs';
```
andyrichardson: Such a shame the GraphQL folks decided to go with the term `Subscription`.

To avoid conflicts, would it make sense to use the term `SubscriptionQuery` to refer to a GraphQL Subscription?

blainekasten: I'm indifferent. Just needs a unique name somehow. I'm fine with SubscriptionQuery and can make that change.

blainekasten: Actually. I think it'd be sort of nice to keep the external API using Subscription and renaming rx internally. The reason being that we have types for `Query` `Mutation` and adding a `SubscriptionQuery` would be off convention.

example/src/app/home.tsx
```diff
@@ -17,13 +22,33 @@ export interface ITodoMutations {
...
     query={createQuery(TodoQuery)}
     mutations={{
       addTodo: createMutation(AddTodo),
       removeTodo: createMutation(RemoveTodo),
     }}
+    subscriptions={[
+      createSubscription(TodoAdded),
+      createSubscription(TodoRemoved),
+    ]}
+    updateSubscription={handleSubscription}
```
andyrichardson: I see what you're going for here. I think we could simplify this by declaring the subscription handling logic when calling createSubscription.

Something like this maybe?

```tsx
// Connect prop
subscriptions={[
  createSubscription(TodoAdded, handleTodoAdded),

// Outside of component
const handleTodoAdded = (response: TodoAddition, data: any[]) => 
  [...data, { id: response.id, text: response.text }]
```

blainekasten: I do like that. I was just mirroring the API from v0 for now. @kitten what do you think?

src/lib/client.ts
andyrichardson: Good point - how about `GraphqlQuery` `GraphqlMutation` and `GraphqlSubscription`?

This would prevent anyone creating their own exchanges from having to deal with the same conflict/confusion.

blainekasten: I like that direction actually. Good point about the exchange typings. Since the data shapes are an internal implementation, should we name them `UrqlQuery` `UrqlMutation` `UrqlSubscription` ?

example/src/app/index.tsx
```diff
@@ -1,11 +1,40 @@
...
+  createClient,
+  dedupeExchange,
+  fetchExchange,
+  cacheExchange,
+  createSubscriptionExchange,
+} from '../../../src/index';
 import Home from './home';
 
+import { SubscriptionClient } from 'subscriptions-transport-ws';
+const subscriptionClient = new SubscriptionClient(
```
andyrichardson: Would creating the websocket client inside the subscriptionExchange work? Ideally the complexity of the websocket communication would be abstracted.

I'm thinking this way we can remove the need for users to have to configure their websocket connection and subscription exchange. In the case where no subscriptions are created within the app, no websocket connection would be made.

```tsx
// ..
import { WebSocket } from 'ws'; // Or whatever websocket library we use

// ..
export const subscriptionExchange: Exchange = ({ forward }) => {
  /** Store of subscription operations */
  let operations = new Map<string, Operation>();
  /** Store of response subjects */
  let subjects = new Map<string, Subject<ExchangeResult>>();
  /** Websocket connection */
  let socket: WebSocket | undefined;

  /** Create connection on first subscription operation if not found */
  const getConnection = (target: string) => {
    if (socket !== undefined) {
      return socket;
    }

    socket = new WebSocket(target);

    socket.on('message', (data: any) => {
      const subscriptionName = getSubscriptionNameFromData(data); // Somehow map response to subscription id?
      const sub = subjects.get(subscriptionName);

      sub.next({ data, operation: operations.get(subscriptionName)}); // Send response back to client instance
    });

    return socket;
  }

  const useSubscription = (operation: Operation) => {
    const connectedSocket = getConnection(operation.context.websocketUrl);
    const subscriptionName = getSubscriptionNameFromOperation(operation); // Somehow parse subscription for unique id

    if (!operations.has(subscriptionName)) {
      operations.set(subscriptionName, operation);
      subjects.set(subscriptionName, new Subject<ExchangeResult>());
    }

    connectedSocket.send(operation.query); // Send query over socket
    return subjects.get(subscriptionName).asObservable(); // Return stream (new data will be pushed by socket.on('message'))
  }

  return ops$.pipe(
    mergeMap(operation => operation.operationName !== 'subscription' ? forward(of(operation)) : useSubscription(operation))
  )
}
```

src/lib/client.ts
andyrichardson: Even better!

kitten: I'm a little put off by names that reference the library name 😅 generally that'd indicate some coupling of the API concept and the internals of the library itself. For instance, what if emotion renamed "styled()" to "emotion()"?

The GraphQLMutation naming is good, but the spelling of GraphQL might make it prone to typos?

example/src/app/home.tsx
kitten: I would regard all these creators as potential functions that might be removed actually if we're planning support for graphql-tag. So depending on them might make things harder.

Also coupling this logic to the creator might not make much sense if this transformation function assumes that there's other queries and mutations that need reconciling. If we assume that these creators might be abstracted away by users (as queries and other things like this are shared) having this transformer on the factory function might give off the impression that it's related to the subscription and not to the other pieces of data

example/src/app/index.tsx
kitten: Sorry for leaving a third rather negative comment 😅

The HTTP spec part of GraphQL is so narrow (unintentionally) that there's not really an alternative to it. Thus all http / fetch logic across all graphql client libraries end up looking roughly the same. The same cannot be said about websockets.

Since the GraphQL spec makes no assumptions about transport layers there's no standard to graphql over websockets. This means that the transport implementation is specific and coupled to how the server implents it. The Apollo libraries have corresponding libs for the server side and client side for subscriptions.

So the motivation behind the pre-v1 api was to keep this agnostic so any transport layer implementation can be used for subscriptions. It wouldn't make sense to couple this unless we assume a specific implementation on the server-side, which is probably a job for another library.

example/src/app/home.tsx
blainekasten: Those are amazing points. Let's leave it as is. it'll make the migration from v0-v1 simpler for subscriptions since the API is a near mirror.

example/src/app/index.tsx
andyrichardson: Wait... So Graphql subscription/websocket implementations are not standardized (yet)?

I'd definitely still prefer if we can keep exchanges plug an play rather than having to create them.

Maybe we could pass send and receive callbacks to the client and then forward them through to the exchange - that way we only 'instantiate'/configure via the client.

blainekasten: Couple thoughts:

1. yes, they are non-standard, so @kitten is right in his synopsis.
2. if we were to try and bundle as much in this library as possible we could potentially bloat the lib for non-subscription users (majority?).
3. Isn't the point of exchanges to be configurable? I think putting the subscription exchange in the defaultExchanges might be unwise since it's not as common a use case and increases bytes.

andyrichardson: I think exchanges are useful but being able to abstract the concept for core functionality, including subscriptions, would be useful to most. 

The point of exchanges (as I see it) is for extensibility - to take our implementation of _urql_ and tweak/add to it as you please. Something more vanilla such as subscriptions probably doesn't need the involvement of our extensibility interface?

How about the following - I think it addresses the concerns mentioned but while keeping things simple and consistent:

 * Default exchanges are: _subscription, dedupe, cache, fetch_.
 * Users configure default exchanges at _client_ creation (as currently is the case with fetch config).
 * Subscription config:
   * Users configure their own communication client (addresses non-standard problem prevents aforementioned additional bloat) and provide utility functions to the _urql_ client for sending and receiving websocket communication.
   * The _subscription exchange_ throws a runtime error (or console error message in development) if a subscription operation is called without the required config options being supplied to the _client_.

blainekasten: Frankly I'll defer to you and @kitten. I think there are valid arguments on both sides and I don't feel like the right person to make that decision since i'm less involved with the codebase as a whole.

If you guys can agree on something I'll make whatever change that is.

blainekasten:
Just pushed a big change to handle subscription prop changes and unmounting to properly dispose of subscription connections.

src/lib/client.ts
kitten: All in all, I'd really go with just `Subscription`, `Query`, and `Mutation` as our Urql component/internals naming. Rx + Urql in a single file is userland and less likely overall to cause naming conflicts. Let's think about what naming makes sense for us first and foremost

blainekasten: Okay. Decided. Leaving it as is.

example/src/app/index.tsx
blainekasten: Cool. I'm changing this then to make a subscriptionExchange part of the defaultExchanges, and adding an API option to client init called `forwardOperation`

andyrichardson:
I like the `forwardSubscription` idea, To simplify things and remove the need for mutation, can we do something like this?

```
createClient({
  url: 'http://localhost:3001/graphql',
  onSubscription: (payload: string) => subscriptionClient.request(payload),
});
```

This way the user can return an observable/subject for the graphql request.

I would have some questions about this approach:
 * Your example looks to use a new connection for each subscription query. Is that the case, and if so, is that the common practice for most graphql subscription/websocket clients? 
 * Are we handling connection states such as an the closing of a connection?

src/types.ts
```diff
@@ -121,7 +150,7 @@ export interface Operation extends Query {
   /** A unique identifier. */
   key: string;
   /** The type of operation. */
-  operationName: string;
+  operationName: OperationType;
```
andyrichardson: 👍

src/lib/typenames.ts
```diff
@@ -72,6 +72,19 @@ function addTypenameToDocument(doc: DocumentNode) {
...
+  const hasSubscriptionQuery = doc.definitions.some(
+    (definition: OperationDefinitionNode) =>
+      definition.operation === 'subscription'
+  );
+
+  // Subscriptions aren't allowed to query for more than one of anything per the spec.
+  // @see https://facebook.github.io/graphql/draft/#sec-Single-root-field
+  // This prevents a subscription from getting a typename injection which would cause the above error
+  if (hasSubscriptionQuery) {
+    return print(doc);
```
kitten: We probably still want to add __typename to anything but the root query fiel set though, since itd be hard to implement solid and more complex caching strategies otherwise

src/lib/client.ts
```diff
@@ -103,33 +117,79 @@ export const createClient = (opts: ClientOptions): Client => {
...
+      subscriptionSubject
+        .pipe(
+          pipeExchanges(
+            opts.exchanges !== undefined ? opts.exchanges : defaultExchanges,
+            subject
+          )
+        )
+        .subscribe(response => {
+          instanceOpts.onSubscriptionUpdate({
+            // @ts-ignore
```
andyrichardson: Are these ignores WIP?

src/types.ts
```diff
@@ -99,10 +117,21 @@ export interface StreamUpdate {
...
+  error?: ExchangeResult['error'];
+}
+
 /** Arguments for creating a client instance. */
 export interface CreateClientInstanceOpts {
   /** A callback function for when changes occur. */
   onChange: (data: StreamUpdate) => any;
+
+  /** A callback function for when subscriptions return new data. */
+  onSubscriptionUpdate: (data: SubscriptionStreamUpdate) => any;
```
andyrichardson: Is there any way of determining which subscription the data is being returned from?

I think it would be simpler to just pass a subscription back from the `executeSubscription` function?

```tsx
client.executeSubscription(subscription).subscribe((data) => console.log(`subscription ${subscription} returned data ${data}`))
```

src/lib/typenames.ts
blainekasten: word. I'll try and figure that out.

src/lib/client.ts
blainekasten: ah yeah. Just because I haven't typed `response`. can do

src/types.ts
blainekasten: I could, but I was trying to follow the general pattern of the library. There may be downsides to this approach where each call site needs to manage the state. Whereas when the client handles it it's centralized.

Determing which subscription is actually easy because of the shape of the data.

`const subscriptionName = Object.keys(data)[0]`

blainekasten:
> I like the forwardSubscription idea, To simplify things and remove the need for mutation, can we do something like this?
>
> ```js
> createClient({
>   url: 'http://localhost:3001/graphql',
>   onSubscription: (payload: string) => subscriptionClient.request(payload),
> });
> ```
> This way the user can return an observable/subject for the graphql request.

I'm not sure that simplifies anything actually. In reality, I need the response to include the unsubscribe method so I can remove graphql subscription connections when a subscription unmounts or changes.


> I would have some questions about this approach:
>
> Your example looks to use a new connection for each subscription query. Is that the case, and if so, is that the common practice for most graphql subscription/websocket clients?
> Are we handling connection states such as an the closing of a connection?

Actually it's doing the opposite. When you init the `subscriptionClient` that is what creates the websocket connection. Internally, every time you run `executeSubscription` it sends a frame through the ws connection for the backend to know about:

image


Then, when you run `executeUnsubscribeSubscription` it sends new frames to tell the backend they have stopped:

image

src/lib/client.ts
```diff
@@ -103,33 +118,82 @@ export const createClient = (opts: ClientOptions): Client => {
...
+
+      executeOperation(operation, subscriptionSubject);
+    };
+
+    const executeUnsubscribeSubscription = (query: Subscription) => {
+      const operation = createOperation(OperationType.Subscription, query, {
+        unsubscribe: true,
+      });
+
+      // remove cached rx subscription
```
kitten: Some indentation weirdness?

src/components/client.tsx
```diff
@@ -58,9 +76,30 @@ export class UrqlClient<MutationDeclarations> extends Component<
...
 
+  deactiveSubscriptions = (subscriptions: Subscription[] = []) => {
+    subscriptions.forEach(sub =>
+      this.state.client.executeUnsubscribeSubscription(sub)
+    );
+  };
+
+  activateSubscriptions = () => {
+    (this.props.subscriptions || []).forEach(sub =>
+      this.state.client.executeSubscription(sub)
```
kitten: I find it weird that every single subscription is handled separately here. I think the more idiomatic Rx approach would be to merge them all here in the component into a single subscription / observable

andyrichardson:
> I need the response to include the unsubscribe method

@blainekasten  the suggestion does just that - the only difference is that you're allowing the user to pass the observable to the client rather than getting them to mutate an existing subject created by the client.

It would look something like this in urql:
```
// Call client provided function
const subscription = opts.onSubscription(query).subscribe(
  next,
  error,
  complete,
);

// At unsubscribe
subscription.unsubscribe();
```

It's exactly the same as what you're proposing but without the need for mutation.

---

> Actually it's doing the opposite.

Sorry, this is my lack of understanding... If we use a single connection (which makes sense), when we receive a response, how do we determine which subscription operation the response is for?

kitten:
@blainekasten as discussed this is ready to go and only needs some minor touch ups. However, due to the observable refactor it'll need some changes. Let me know if I should quickly make those changes or not :)

PR #165, opened 2019-01-14
Enhancement/lint on commit

andyrichardson:


package.json
```diff
@@ -50,6 +50,9 @@
     "*.{js,jsx,ts,tsx,json,md}": [
       "prettier --write",
       "git add"
+    ],
+    "*.{ts, tsx}": [
+      "tslint --fix"
```
kitten: This is missing git add :) let's add prettier here as well and remove ts/tsx above so it's being kept in the right order and doesn't run in parallel

andyrichardson: Damn son... you're fast!

Good shout

PR #166, opened 2019-01-16
V1 Replace RxJS with Wonka

kitten:
This replaces RxJS in the v1 refactor with [Wonka](https://github.com/kitten/wonka). It'll require some dogfooding potentially, but works really well for our purposes and I have only encountered a single quirk, which wasn't technically a bug. _Hooray!_

From a development standpoint changes are almost not noticeable, so the following code in RxJS converts almost 1:1 to Wonka code:

```js
import { of } from 'rxjs';
import { map, filter } from 'rxjs/operators';

of('Hello').pipe(
  map(x => x + 'World'),
  filter(x => x.length > 3)
).toPromise();
```

```js
import { pipe, fromValue, map, filter, toPromise } from 'wonka';

pipe(
  fromValue('Hello'),
  map(x => x + 'World'),
  filter(x => x.length > 3),
  toPromise
)
```

The code is just as well typed but fully functional instead of Rxjs' mix. More importantly, Wonka has specific characteristics that will come in handy. Like RxJS it creates cold observables by default, since it follow the callbag spec, but runs completely and predictably synchronously _all the time_.

This can be seen in a lot of tests that are now perfectly synchronous without sacrificing readability. In fact, they're pretty predictable now.

This creates two nice features for urql:

- Iterables become a thing which are really helpful if we model other parts of the client
- **suspense** becomes super simple, since the exchange pipeline with an assumed, synchronous cache exchange would be able to resolve a result immediately


andyrichardson:
API looks similar enough, no objections here 👍

Might be worth merging the work from @blainekasten first? I'm happy to spend some time today making that mege happen so we can get this in today too.

blainekasten:
@andyrichardson I appreciate the call out. It would definitely be _nice_ if I could get my PR merged first so I don't have to learn a whole new API will still updating things.

The only down side is I'm on site at a client for the next 2 weeks and free time is going to be much harder to come by. Not sure where this leaves things. If someone wants to pick up my PR and wrap up the work I'd be happy to help explain how to move things across the finish line.

kitten:
Merging this refactor in in a broken state since #160 introduced some conflicts which are not worth fixing since the next refactor will swap out some client logic.

PR #168, opened 2019-01-22
[WIP] V1: Refactor the client's operation management

kitten:
Leftover todos:

- [ ] Update tests
- [ ] Validate new approach and correctness
- [x] Move example's dependencies into `example/package.json` (And maybe use CRA for simplicity?)
- [x] Readd TS typings export
- [x] Check biggest bundlesize contributors
- [ ] Readd `refetch` (Maybe?)
- [ ] Fix up `example/`
- [ ] Readd option to pass in `fetch` ponyfill (to prevent forced polyfilling)


src/exchanges/index.ts
```diff
@@ -2,6 +2,7 @@ export { cacheExchange } from './cache';
 export { subscriptionExchange } from './subscription';
 export { dedupeExchange } from './dedup';
 export { fetchExchange } from './fetch';
+export { composeExchanges } from './compose';
```
kitten: This should make it easier for people to compose some existing exchanges—be them internal to `urql` or third-party—outside of urql as well, in case someone writes tests or even a library that needs some of our exchanges to work. In such a case someone can use `composeExchanges` instead of exposing an exchange array, so hopefully a nice addition

PR #184, opened 2019-02-16
Support DocumentNode and refactor query keying

### LINKED ISSUE #50: Tagging the query/mutations to enable syntax highlighting possibility
Syntax highlighting libraries have no idea if a template string is GraphQL or not. In styled-components, they can detect it by `css` tag, or `styled.X`. 

For GraphQL, highlighting extensions they can detect `gql` tagged strings. So I tend to define this: 
```js
const gql = String.raw

const UserQuery = gql`query { user }`
```
And I get syntax highlighting! (I use [this](https://github.com/kumarharsh/graphql-for-vscode) in VSCode)

So as an idea we can refer to this method at the end of `readme`, as a "Syntax Highlighting" section, or we can export this from the library which might not be the best thing. I can do a PR as well.
 
Your thoughts? Thanks!

kitten:
Fix #50

- Support `DocumentNode | string` on `GraphQLRequest.query`
- Switch `Operation.query` to `DocumentNode` only
- Remove parse/print from `typenames` helper
- Add new `queryFromKey` helper

This now uses [djb2](http://www.cse.yorku.ca/~oz/hash.html) for hashing which is
more concise, but might lead to some duplicate keys. This is pretty acceptable, since we're
not keeping track of a lot of operations at any given time (Not in the magnitude of 10,000s). If we're unhappy with cache collisions we should switch to a simple counter and maps.


andyrichardson:
Looks pretty great!

Hashing wise, for uniqueness, we could use _SHA-256_ (SHA-1 isn't supported by edge for some reason...). 

We'll need to add a conditional (low-key polyfill) for IE11 support but other than that, browser support is strong and would be more performant (uses hardware accelleration).

```
const sha256 = async (str) => {
  const buf = await crypto.subtle.digest("SHA-256", new TextEncoder("utf-8").encode(str));
  return Array.prototype.map.call(new Uint8Array(buf), x=>(('00'+x.toString(16)).slice(-2))).join('');
}
```

PR #185, opened 2019-02-17
add 'creating exchange' section

andyrichardson:


docs/extending-and-experimenting.md
```diff
@@ -147,3 +147,130 @@ subscribe to the result of `executeQuery`, keep track
...
 some state. This all can be reapplied when you write your
 own APIs.
+
+## Making your own Exchange
+
+As with any library, you're going to find yourself wanting to change (or adapt) the way things work.
+Since day one, with `urql`, we have aimed to make adaptability our number one priority through the use of _Exchanges_.
+Lets run through how this works by creating our own Dedup exchange - a middleware to debounce incoming query operations.
+
+> Note: We've already provided a bunch of exchanges that we thought you would find
```
kitten: This will need to point to the basics section, not the source code. I assume people can find the source rather easy but will actually want to know which ones are standard first

docs/extending-and-experimenting.md
```diff
@@ -147,3 +147,130 @@ subscribe to the result of `executeQuery`, keep track
...
+import { Exchange } from 'urql';
+
+export const dedupExchange: Exchange = ({ forward }) => {
+  const inFlight = new Set<string>();
+
+  return ops$ => forward(ops$);
+};
+```
+
+You can see from the above example that an exchange is a [curried function](https://medium.com/javascript-scene/curry-and-function-composition-2c208d774983). It takes some setup arguments and returns a function which receives an operation stream (`ops$`) and forwards it to the next exchange (`forward`).
```
kitten: I'd prefer not to point to Medium. Docs like Redux' one used to explain this fine without and I'm pretty sure I've already got an explanation for this in the Architecture section that avoids complex terms

docs/extending-and-experimenting.md
```diff
@@ -147,3 +147,130 @@ subscribe to the result of `executeQuery`, keep track
...
+    return pipe(
+      forward(preFlight$),
+      // This `tap` is called after an operation has been completed
+      tap(response => inFlight.delete(response.operation.key))
+    );
+  };
+};
+```
+
+Our exchange now tracks operations and their completion, but for this to be useful,
```
kitten: I think at this point we'll need to explain the operators in detail and how the operations now flow through this exchange

docs/extending-and-experimenting.md
```diff
@@ -147,3 +147,130 @@ subscribe to the result of `executeQuery`, keep track
...
+} from 'urql';
+import { dedupExchange } from './myDedupExchange';
+
+export const urqlClient = createClient({
+  url: 'https://localhost:3000/graphql',
+  exchanges: [dedupExchange, debugExchange, cacheExchange, fetchExchange],
+});
+```
+
+To conclude, here's a quick recap of what we've achieved:
```
kitten: Lets ret rid of the bullet point list summarising sole things :) it seems out of place given the short length of this section

docs/extending-and-experimenting.md
andyrichardson: Trying to do this on my phone on the tube - I swear GitHub had a quick fix option added 😕

PR #197, opened 2019-02-22
fix typename recursion and add test

### LINKED ISSUE #195: Mutations not invalidating cache
I'm fairly certain that this is a bug... Or, at least a change in behavior between v0.x and v1.0.

Whenever I execute a mutation, it seems like the cache does not update unless I manually refetch. 

For example, when I run this mutation: 
```
mutation createPost($input: CreatePostInput!) {
  createPost(input: $input) {
    post {
      id
      content
      createdAt
    }
  }
}`
```
I get back a response with the new post data (as the docs say is required). However, my `posts` collection does not update unless I manually make a `refetch` call. The pre-1.0 code seemed to do this without having to manually refetch.

On a related note, are there any sorts of best practices around optimistic updates? So, if I execute a mutation, is there some way to directly manipulate the cache until it's invalidated?

andyrichardson:
Fix #195

### DIFF
--- a/src/utils/typenames.ts
+++ b/src/utils/typenames.ts
@@ -6,21 +6,34 @@ import {
   visit,
 } from 'graphql';
 
-const getTypeNameFromField = (obj: object) =>
-  Object.values(obj).reduce((all, val) => {
-    if (typeof val !== 'object' || val === null) {
-      return all;
+interface EntityLike {
+  [key: string]: EntityLike | EntityLike[] | any;
+  __typename: string | null | void;
+}
+
+const collectTypes = (obj: EntityLike | EntityLike[], types: string[]) => {
+  if (Array.isArray(obj)) {
+    obj.forEach(inner => collectTypes(inner, types));
+  } else if (typeof obj === 'object' && obj !== null) {
+    for (const key in obj) {
+      if (obj.hasOwnProperty(key)) {
+        const val = obj[key];
+        if (key === '__typename' && typeof val === 'string') {
+          types.push(val);
+        } else if (typeof val === 'object' && val !== null) {
+          collectTypes(val, types);
+        }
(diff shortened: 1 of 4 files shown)


src/utils/typenames.ts
```diff
@@ -7,15 +7,15 @@ import {
...
-  );
+  Object.values(obj).reduce((all, val) => {
+    if (typeof val !== 'object' || val === null) {
+      return all;
+    }
+
+    return val.__typename !== undefined
+      ? { ...all, [val.__typename]: true, ...getTypeNameFromField(val) }
+      : { ...all, ...getTypeNameFromField(val) };
+  }, {});
```
kitten: There are some problems here:

- Arrays are not taken into account
- Nested arrays are not taking into account
- We have an object to keep track of types but then also `filter` afterwards

I'd go for a recursive function that isn't exported, like `getTypeNameFromField` right now, but make it completely mutable and filter afterwards, i.e.:

```
interface EntityLike {
  __typename: ?string,
  [key: string]: EntityLike | EntityLike[] | string | number | null
}

const collectTypes = (obj: EntityLike, types: string[]) => {
  if (Array.isArray(obj)) {
    obj.forEach(inner => collectTypes(inner, types));
  } else {
    for (const key in obj) {
      if (object.hasOwnProperty(key)) {
        const val = obj[key];
        if (key === '__typename' && typeof val === 'string') {
          types.push(val);
        } else if (typeof val === 'object') {
          collectTypes(val, types);
        }
      }
    }
  }

  return types;
};

export const gankTypeNamesFromResponse = (response: object) => {
  const typeNames = collectTypes((response as EntityLike), []);
  return typeNames.filter((v, i, a) => a.indexOf(v) === i);
};
```

PR #200, opened 2019-02-22
Rehydrate GraphQLErrors properly

kitten:


src/utils/error.ts
```diff
@@ -21,21 +19,29 @@ const generateErrorMessage = (
...
 
-const rehydrateGraphQlError = (error: string | GraphQLError): Error => {
+const rehydrateGraphQlError = (error: any): GraphQLError => {
   if (typeof error === 'string') {
-    return new Error(error);
-  } else if (error.message) {
-    return new Error(error.message);
+    return new GraphQLError(error);
+  } else if (typeof error === 'object' && error.message) {
+    return new GraphQLError(
```
andyrichardson: This is gross - We should make a PR... probably won't though.

PR #203, opened 2019-02-25
User overloaded types for createRequest

andyrichardson:


andyrichardson:
Expected this to just work. Looks like TypeScript won't currently infer overloaded types when providing a union type [source](https://github.com/Microsoft/TypeScript/issues/14107)

PR #204, opened 2019-02-26
Add demo and update docs for @urql/exchange-graphcache

kitten:
This is still in development so this pull request is a **work in progress**. As the API for `@urql/exchange-graphcache` matures, we'll add docs for it to this PR and make sure everything's working.

See: https://github.com/kitten/urql-exchange-graphcache


andyrichardson:
Looks like the only difference between this and the other examples is this line:

`exchanges: [dedupExchange, debugExchange, cacheExchange({}), fetchExchange],`

Do we need an example for this? I think it would be much cleaner to put usage examples in the eventual graphcache repo and link to it from our docs.

kitten:
Yep, this is an example for testing, and eventually it'll probably make sense to live in here for visibility. Anyway, `cacheExchange` doesn't take any parameters yet, but it eventually will. For now though this is just a draft PR, so no action/merge necessary :)

andyrichardson:
> Yep, this is an example for testing

Sounds like we're going monorepo soon?

Looking forward to seeing this new cache in action!

kitten:
I'll close this for now since it'll be while until it's ready anyway :+1:

PR #207, opened 2019-03-06
Link to egghead screencast

ijones16:
I thought I could get the embed working but it turns out github flavored mardown doesnt support video embed.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -155,13 +155,10 @@ const TodoList = ({ limit = 10 }) => {
 Similarly to the `<Query>` component, `useQuery` will start the request
 as soon as it's mounted and will rerun it when the query or variables change.
 
+> A tutorial on the `useQuery` hook is also available as a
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -157,6 +157,12 @@ as soon as it's mounted and will rerun it when the query or variables change.
 
 [Read more about the result's API in the Architecture's Results section.](architecture.md#operation-results)
 
+#### Watch `useQuery` React hook screencast on egghead
+
+[Query Graphql Data With Urql Using React Hooks][egghead]
```
kitten: I'd opt for just having it inline and more prominently featured :) can you put it above the "Read more about..." link and make it a quote, which we use for notes? e.g.

```md
> A tutorial on the `useQuery` hook is also available as a [screencast on Egghead](https://egghead.io/lessons/graphql-query-graphql-data-with-urql-using-react-hooks).
```

(Inline links are completely fine since this is not reused :+1: )

ijones16: Sounds great!

kitten:
Awesome! It'd be great to touch base if you're planning to make more material soon 👌

PR #214, opened 2019-03-27
Escape pipe characters (|) in Markdown tables.

parkerziegler:
This PR adds a small fix to escape pipe characters in Markdown tables in `api.md`. While the pipes represent union types in the TS world, they also represent columns in Markdown land 😂. Previously, this meant that some types and descriptions of specific props were either cutoff or entirely missing.
image
 
By escaping the pipe characters, our types and descriptions rise like phoenixes from the ashes 🔥 🐦. 
image

Just noticed this while working on `reason-urql` and thought I'd fix it!

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -137,12 +137,12 @@ interface UseSubscriptionState<T> {
 
 #### Props
 
-| Prop      | Type                                    | Description                                                                                           |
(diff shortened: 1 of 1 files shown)


kitten:
Oops 😅

Issue #215, opened 2019-03-28
Cache exchange not forwarding teardown correctly


federicobadini:
Hi,

While I was working on a custom exchange I think I have found something wrong with how the default cache exchange forward requests down to the other exchanges downstream.
I have created a [codesandbox](https://codesandbox.io/s/3j32j8z2p) to show you the problem. 

Basically I have created an exchange that deduplicates queries by cancelling the old request and forwarding the new one (switchMapExchange). 
In the example attached, I issue a couple of query at the same point in time. 
You'll see that even though I forward teardown actions to cancel the first request of the two, no request is effectively cancelled and unexpectedly both are completed.

Trying to solve the problem, I'm quite convinced that it derives from how the default cacheExchange handle the forward action.
It basically creates different forwarding stream for new and skipped ops and merge those streams afterwards
```js
const newOps$ = pipe(
  sharedOps$,
  filter(op => !shouldSkip(op) && !isOperationCached(op)),
  map(mapTypeNames),
  forward,
  tap(response => {
    if (response.operation.operationName === 'mutation') {
      handleAfterMutation(response);
    } else if (response.operation.operationName === 'query') {
      handleAfterQuery(response);
    }
  })
);

const skippedOps$ = pipe(
  sharedOps$,
  filter(op => shouldSkip(op)),
  forward
);

return merge([cachedOps$, newOps$, skippedOps$]);
```
In the codesandbox linked above I have included also a slightly modified cache exchange, in which forwarded operation streams are handled differently
```js
const forwardedOps$ = pipe(
  merge([
    pipe(
      sharedOps$,
      filter(op => !shouldSkip(op) && !isOperationCached(op)),
      map(mapTypeNames)
    ),
    pipe(
      sharedOps$,
      filter(op => shouldSkip(op))
    )
  ]),
  forward,
  tap(response => {
    if (response.operation.operationName === "mutation") {
      handleAfterMutation(response);
    } else if (response.operation.operationName === "query") {
      handleAfterQuery(response);
    }
  })
);

return merge([cachedOps$, forwardedOps$]);
```
This approach solved my problem forwarding correctly teardown actions to default fetchExchange and cancelling request when needed.
Please let me know what you think

kitten:
That seems absolutely right :+1: `forward` should also not be called twice. It's possible, but obviously leads to subtle bugs, since it spreads out some of the code.

[...]

PR #222, opened 2019-04-06
Cache exchange not forwarding teardown correctly #215

federicobadini:


kitten:
Thanks for taking care of this 🙌🙌

Issue #232, opened 2019-04-26
Export UseQueryResponse et.al.


good-idea:
I'm building a Shopify Storefront API library, and it uses urql under the hood. There are a few custom hooks that are essentially simple wrappers for making graphql requests with some defaults, such as:

```ts
import { useQuery, RequestPolicy, UseQueryResponse } from 'urql'
import { defaultQuery, QueryResult } from './defaultQuery'

interface FetchConfig {
	skip?: boolean
	query?: string
	requestPolicy?: RequestPolicy
}

export function useFetchProductByHandle<ExpectedResult = QueryResult>(
	handle: string,
	config: FetchConfig = {},
): UseQueryResponse<ExpectedResult> {
	const { query, requestPolicy } = config
	const variables = {
		handle,
	}
	return useQuery({
		query: query || defaultQuery,
		variables,
		requestPolicy,
	})
}
```

It would be handy in this case to have `UseQueryResponse` exported from `hooks/useQuery.tsx`.

Any objection to this? I'm going to put together a quick PR right now.

good-idea:
Closed in #233

PR #233, opened 2019-04-26
Export hook response types

good-idea:
Re: #232


mxstbr:
This is a great start, we also need `Use<Query|Mutation|Subscription>Args` to be exported.

kitten:
@mxstbr let's merge this and get a separate PR going since this has been open for a while

Issue #228, opened 2019-04-22
Skip request


matheus-rosin:
Apollo has a feature called `skip` that, being a boolean, if `true`, will skip the query, for example. This would be awesome for Urql, specially because of the hooks approach.

If I use the component `Query`, I can check for myself if some needed variable is missing and, just return `null` in that case, for instance. But, if I were to reproduce this behavior using the `useQuery` hook, I would need to create a component (HOC) for returning null in order to skip the api request - but I would still be using the component approach, and the hook would lose sense.

So, what if we could pass an prop like `skip`, either using a component or a hook, that receives a boolean value, and that if `true` would skip of calling the api?

Thanks

matheus-rosin:
I made a PR for this. There are some checks to be made. Now it is up to you 🙂

parkerziegler:
Closed via #237.

PR #237, opened 2019-05-11
Add support for skip to Query and useQuery.

parkerziegler:
This PR picks up on the changes in #229 and would close issue #228. @kitten I read through your comments in #229 and attempted to address them in this PR. I also added a few tests to ensure things are behaving as expected, but definitely would appreciate both of your 👀 in verifying that this behaves as expected.

### DIFF
--- a/src/components/Query.tsx
+++ b/src/components/Query.tsx
@@ -11,6 +11,7 @@ interface QueryHandlerProps {
   variables?: object;
   client: Client;
   requestPolicy?: RequestPolicy;
+  skip?: boolean;
   children: (arg: QueryHandlerState) => ReactNode;
 }
 
@@ -27,6 +28,10 @@ class QueryHandler extends Component<QueryHandlerProps, QueryHandlerState> {
 
   executeQuery = (opts?: Partial<OperationContext>) => {
     this.unsubscribe();
+
+    if (this.props.skip) {
+      return;
+    }
 
     this.setState({ fetching: true });
 
@@ -58,7 +63,11 @@ class QueryHandler extends Component<QueryHandlerProps, QueryHandlerState> {
     this.executeQuery();
   }
 
-  componentDidUpdate() {
+  componentDidUpdate(prevProps: QueryHandlerProps) {
+    if (this.props.skip && prevProps.skip !== this.props.skip) {
+      return;
+    }
+
     const newRequest = createRequest(this.props.query, this.props.variables);
     if (newRequest.key !== this.request.key) {
       this.request = newRequest;
--- a/src/components/Query.tsx
+++ b/src/components/Query.tsx
@@ -65,6 +65,7 @@ class QueryHandler extends Component<QueryHandlerProps, QueryHandlerState> {
 
   componentDidUpdate(prevProps: QueryHandlerProps) {
     if (this.props.skip && prevProps.skip !== this.props.skip) {
+      this.unsubscribe();
       return;
     }
 
--- a/src/components/Query.tsx
+++ b/src/components/Query.tsx
@@ -29,25 +29,29 @@ class QueryHandler extends Component<QueryHandlerProps, QueryHandlerState> {
   executeQuery = (opts?: Partial<OperationContext>) => {
     this.unsubscribe();
 
-    if (this.props.skip) {
-      return;
-    }
-
     this.setState({ fetching: true });
 
-    const [teardown] = pipe(
-      this.props.client.executeQuery(this.request, {
-        requestPolicy: this.props.requestPolicy,
-        ...opts,
-      }),
-      subscribe(({ data, error }) => {
-        this.setState({
-          fetching: false,
-          data,
-          error,
-        });
-      })
-    );
+    let teardown = noop;
+
+    if (!this.props.skip) {
+      [teardown] = pipe(
+        this.props.client.executeQuery(this.request, {
+          requestPolicy: this.props.requestPolicy,
+          ...opts,
+        }),
+        subscribe(({ data, error }) => {
+          this.setState({
+            fetching: false,
+            data,
+            error,
+          });
+        })
+      );
+    } else {
+      this.setState({
+        fetching: false,
+      });
+    }
 
     this.unsubscribe = teardown;
   };
@@ -63,12 +67,7 @@ class QueryHandler extends Component<QueryHandlerProps, QueryHandlerState> {
     this.executeQuery();
   }
 
-  componentDidUpdate(prevProps: QueryHandlerProps) {
-    if (this.props.skip && prevProps.skip !== this.props.skip) {
-      this.unsubscribe();
-      return;
-    }
-
+  componentDidUpdate() {
     const newRequest = createRequest(this.props.query, this.props.variables);
     if (newRequest.key !== this.request.key) {
       this.request = newRequest;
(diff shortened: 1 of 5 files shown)


src/components/Query.tsx
```diff
@@ -58,7 +63,11 @@ class QueryHandler extends Component<QueryHandlerProps, QueryHandlerState> {
     this.executeQuery();
   }
 
-  componentDidUpdate() {
+  componentDidUpdate(prevProps: QueryHandlerProps) {
+    if (this.props.skip && prevProps.skip !== this.props.skip) {
```
kitten: If `props.skip` is `true` here, we'll never reach `this.unsubscribe` above 🤔

src/components/Query.tsx
```diff
@@ -28,6 +29,10 @@ class QueryHandler extends Component<QueryHandlerProps, QueryHandlerState> {
   executeQuery = (opts?: Partial<OperationContext>) => {
     this.unsubscribe();
 
+    if (this.props.skip) {
```
parkerziegler: Per your suggestions @kitten moved this after the call to `unsubscribe`. Moreover, this would return before `fetching` ever gets set to `true`, so I don't think we need to handle resetting `fetching` to `false`.

src/hooks/useQuery.test.tsx
```diff
@@ -158,7 +164,21 @@ describe('on change', () => {
 describe('execute query', () => {
   it('triggers query execution', () => {
     renderer.create(<QueryUser {...props} />);
-    act(() => execute());
+    act(() => execute!());
```
parkerziegler: Non-nullary operator for tests, since `execute` can, from a type perspective, be `undefined` in the course of running these tests.

src/components/Query.tsx
parkerziegler: Ah gotcha. So we'd hit a case where we have a current query subscription, `skip` updates to `true`, this block gets hit and `this.unsubscribe` never gets called. What if we just call `this.unsubscribe()` before the early return here?

kitten: Yea, that'd be equivalent but it should always be preferred to just call execute again since it always unsubscribes and resubscribes if necessary, so that might make things easier on terms of it always aligning the subscription with the props

parkerziegler: Agreed, good point. Let me do some refactoring and re-push.

parkerziegler:
@kitten @andyrichardson refactored this just now per @kitten's suggestions, let me know what the next steps you think we should take are 😄 Thanks for the review thus far 💯

src/components/Query.tsx
```diff
@@ -30,19 +31,27 @@ class QueryHandler extends Component<QueryHandlerProps, QueryHandlerState> {
...
-        this.setState({
-          fetching: false,
-          data,
-          error,
-        });
-      })
-    );
+    let teardown = noop;
+
+    if (!this.props.skip) {
```
parkerziegler: Moving the check here ensures that both `this.unsubscribe` will get executed and redefined in preparation for the next query. No more futzing with `componentDidUpdate` 😅

src/hooks/useQuery.ts
```diff
@@ -39,27 +40,34 @@ export const useQuery = <T = any, V = object>(
...
-        subscribe(({ data, error }) =>
-          setState({ fetching: false, data, error })
-        )
-      );
+      let teardown = noop;
+
+      if (!args.skip) {
+        [teardown] = pipe(
+          client.executeQuery(request, {
+            requestPolicy: args.requestPolicy,
```
parkerziegler: Looking at this now, shouldn't `args.requestPolicy` also be included in the dependency array for this `useCallback`? Otherwise we might not recompute it if it changes.

andyrichardson: Yup I believe so. We're currently allowing users to specify the request policy on a manual query execution but I believe this goes against the way hooks work.

If the request policy changes, that should be requested in the components state and flow all the way down to the useQuery hook.

parkerziegler: Pushed!

src/hooks/useQuery.ts
```diff
@@ -39,27 +40,34 @@ export const useQuery = <T = any, V = object>(
...
     },
-    [request.key]
+    [request.key, args.skip, args.requestPolicy]
   );
 
   useEffect(() => {
     executeQuery();
     return unsubscribe;
-  }, [request.key]);
+  }, [request.key, args.skip]);
```
parkerziegler: Last question here for both of you @andyrichardson and @kitten -> should `executeQuery` be included in the dependency array here? The Hooks FAQ make a big deal about it being unsafe to exclude functions that reference `props` / `state` from the dependency array of `useEffect` calls. So basically this dep array would become just `[executeQuery]`. I'm also starting to think we might want to `useRef` `unsubscribe`. Thoughts?

parkerziegler: React FAQ on functions as dependencies: https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies

parkerziegler:
@kitten @andyrichardson how do you feel about merging this? I have some follow up changes that 1) integrate `typescript-eslint` _just_ for the React hooks rules, 2) fix some dependency array issues flagged by the ESLint plugin, and 3) show what a test suite for `useQuery` would look like with [`react-hooks-testing-library`](https://github.com/mpeyper/react-hooks-testing-library). Would love to get those changes up for review soon!

src/hooks/useQuery.ts
kitten: Sorry, didn't get to this :scream: Including it into the array here will probably mean that `requestPolicy` can change at any time which is generally good :+1:

parkerziegler: 🆒 that'll come up in the next PR, gonna merge this as is 😄

andyrichardson: @parkerziegler - yeah I think that is necessary 👍

andyrichardson:
PR looks strong 👍👍👍

@parkerzielger minor suggestion about the wording, maybe wording could be more specific than 'skip'? I'm thinking 'autoFetch' or 'fetchOnMount'. Not a big deal either way - just my 2 cents. 

Is 'cacheOnly' the same as skip? In theory I would expect cacheOnly to not fetch on mount but still fetch when calling the refetch function. I had this problem with Tipple and settled on allowing both options despite the redundancy.

Also, +1 on migrating to eslint. The react linting rules suck at times (like if you want an effect to be called exclusively on mount - this would be a linting error) and adding a bunch of ignores to valid use cases is a pain but that's better than missing a required dependency.

kitten:
@andyrichardson I think auto fetch or fetch on mount could imply that we encourage hooks to remain passive on mount which is regarded as an antipattern by the React team.

For now this naming was suggested because it's also what Apollo uses I suppose, and it nicely matches up with the existing skip directive. But yea, not a fan.

It's not the same as cache Only as it'll stop to fetch at all. So it is useful to "freeze" the directives logic completely.

andyrichardson:
@kitten what does "passive on mount" mean? Do you have a link to the react docs about the antipattern?

If in getting this right, you're proposing that manual refetch isn't possible at all when using cache only? If we're going this route we'll need to update our types for useQuery

matheus-rosin:
About the word "skip", when I suggested it, I didn't even wonder if that is the most plausible 😄 (yeah, it was based on Apollo already available nomenclature). So, if I could suggest a better word (and more JS-ish), it would be `prevent` or even `block`. I rather the first one, but the second seems decent too.

good-idea:
Any idea on when this update will be published to NPM?

parkerziegler:
@good-idea #239 will have this, which should get merged today. I'd guess we'd be able to release it next week. I'll follow up w/ Phil there and we'll try to ship it soon!

Issue #216, opened 2019-03-28
Add promisified request methods on the Client (especially for non-React API calls)


CaitlinWeb:
Thanks for this library. I'm new to GraphQL and wanted to find a library that would work with React and had active maintainers. (For instance, both `Lokka` and `graphql-request` seem to have been abandoned 🤷‍♀️ )

My only issue is I have a use case with the need for a GraphQL request outside of React. (The use case is on the server, checking if the user exists and create one if not.) Therefore, I can't use `<Query />` or `useQuery`, etc. It took a bit, but I figured out this is possible using `urql` thanks to the "Extending & Experimenting" documentation:

```javascript
const { pipe, subscribe } = require('wonka');
const { client, createRequest } = require('urql');

// create the client
const client = createClient(/*...*/);
// create the request
const request = createRequest(`query { ... }`, { variables });
// send to api
pipe(
	client.executeQuery(request),
	subscribe(({ data, error }) => {
		console.log(data, error);
	})
);
```

While this works, it's not pretty and Wonka doesn't currently have much documentation. I'm not sure if this code works in all cases.

I would *love* it if there was a utility function exposed that does the `pipe(...)` portion in the above code sample. Something that returns a promise I can chain off of, like:
```javascript
// send to api
sendRequest(client, request)
    .then((data, error) => {/*...*/})

// could also be something like
client.sendRequest(request)
   .then((data, error) => {/*...*/})
```

Thoughts?

kitten:
I do have to write up documentation and release a nice site for Wonka! Apologies that that process isn't as easy yet as it should be 😢 

We actually have left off promisified variants of `executeQuery` after multiple refactors, since we weren't sure whether they're any use to someone, so this is definitely a signal that they indeed _are useful_.

You can also use the `toPromise` helper in Wonka to make this nicer for now, this would mean that you can at least convert the `Source<OperationResult>` to a promise, if you don't care about cancellation.

This is the operation in question: https://github.com/kitten/wonka/blob/master/src/web/wonka_sink_toPromise.d.ts

So to combine your code snippets into "one" by using `toPromise`, this would become:

```js
const { pipe, toPromise } = require('wonka');
const { client, createRequest } = require('urql');

const client = createClient(/*...*/);
const request = createRequest(`query { ... }`, { variables });

pipe(client.executeQuery(request), toPromise)
  .then(({ data, error }) => {/*...*/})
```

[...]

PR #241, opened 2019-05-25
[WIP] Add client.createQuery and client.createMutation

good-idea:
Addresses #216

This PR adds two methods to the client, `.createQuery` and `.createMutation`. Both of these create - but do not run - queries and mutations that can be used later.

**Todo**

- [ ] - review tests
- [ ] - subscription support

*(WIP documentation, will update)*

### `createQuery`

Accepts an object as its input:

```ts
interface CreateQueryInput {
  query: DocumentNode | string;
  variables?: object;
}
```

Returns a function that can be called with (optional) new variables. This function executes the query when called and returns the results.

Example:

```ts
const MY_QUERY = `
	query { ... }
`

const defaultVariables = {
  x: 'y'
}

const myQuery = client.createQuery({ query: MY_QUERY, variables: defaultVariables })

/** later */

const defaultResults = await myQuery()
const otherResults = await myQuery({ x: 'z' })
```

### `createMutation`

Similar to `createQuery`, except its only argument is a query string.

Example:

```ts
const MY_MUTATION = `
	mutation { ... }
`

const myMutation = client.createMutation(MY_MUTATION)

const result = myMutation({ a: 'b' })
```

-----

Some questions:

 - This is set up to more or less match the signatures of `useQuery` and `useMutation`. But, it might be simpler if these two methods had the same structure. (Is there any reason why `useMutation` does not use `({ query: MY_MUTATION })` instead of just `(MY_MUTATION)`?)
 - Please take a look at my tests -- I'm not too familiar with the internals and unsure if there is anything else that should be covered. The mocks for `executeQuery` and `executeMutation` were just yanked from other tests.

I've tested this in an existing project and it's working.

### DIFF
--- a/src/__snapshots__/client.test.ts.snap
+++ b/src/__snapshots__/client.test.ts.snap
@@ -3,7 +3,9 @@
 exports[`createClient passes snapshot 1`] = `
 Client {
   "activeOperations": Object {},
+  "createMutation": [Function],
   "createOperationContext": [Function],
+  "createQuery": [Function],
   "createRequestOperation": [Function],
   "dispatchOperation": [Function],
   "exchange": [Function],
--- a/src/__snapshots__/context.test.ts.snap
+++ b/src/__snapshots__/context.test.ts.snap
@@ -16,7 +16,9 @@ Object {
     "_currentRenderer2": null,
     "_currentValue": Client {
       "activeOperations": Object {},
+      "createMutation": [Function],
       "createOperationContext": [Function],
+      "createQuery": [Function],
       "createRequestOperation": [Function],
       "dispatchOperation": [Function],
       "exchange": [Function],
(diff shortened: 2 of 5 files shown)


kitten:
Hiya :wave: Sorry that we never got around to merging this! We may implement something like this in the future, sorry that it won't be merged with your name on the repo 😢 

We're thinking of likely making these methods return the raw Wonka streams and simplifying the inputs only. The reason for that is that it's not quite safe to only use the promises, since that messes with cancellability and won't make much sense for queries and subscriptions.

Either way, I hope we can solve this soon, but we first have to validate that the API makes sense and check whether it works well. We've also started on writing a Guide doc on this, so we'll hopefully be able to validate that the use-case matches the future API :pray:

good-idea:
No worries at all, looking forward to your updates!

PR #243, opened 2019-05-30
Bugfixes for useQuery

### LINKED ISSUE #240: Avoid stale state updates
I think that `useQuery` needs to track if it's been unmounted to avoid stale updates.

We're seeing:

```
Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all sub
```

And I think it's coming from `useQuery`. I think the `setState` calls need to be conditional based on some variable that we toggle from `true` to `false` when the component gets updated.

Much like this `usePromise` example: https://gist.github.com/mjackson/05c7749430d0ec87b66612733805bad6.

Happy to make the PR if you're happy with this proposal?

andyrichardson:
## Changes
- Fix #240 
- Prevent unsubscribe function being rewritten on render (potential bug)
- Prevent key being unnecessarily regenerated on re-render


src/hooks/useQuery.ts
```diff
@@ -77,8 +78,12 @@ export const useQuery = <T = any, V = object>(
 
   useEffect(() => {
     executeQuery();
-    return unsubscribe.current;
-  }, [executeQuery]);
+
+    return () => {
+      isMounted.current = false;
```
kitten: Small bug here since the effect runs more times than just for mount & unmount

src/hooks/useSubscription.ts
```diff
@@ -47,22 +48,26 @@ export const useSubscription = <T = any, R = T, V = object>(
...
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [request.key, handler, client]);
+  }, [client, handler, request]);
 
   useEffect(() => {
     executeSubscription();
-    return unsubscribe.current;
-  }, [executeSubscription]);
+    return () => {
+      unsubscribe.current();
```
kitten: Same here 🙌

src/hooks/useQuery.ts
```diff
@@ -51,23 +52,23 @@ export const useQuery = <T = any, V = object>(
     (opts?: Partial<OperationContext>) => {
       unsubscribe.current();
 
-      setState(s => ({ ...s, fetching: true }));
+      if (args.pause) {
+        unsubscribe.current = noop;
+        return;
```
parkerziegler: I think the early `return` here is the reason for the test failures. Basically, `waitForNextUpdate` expects some asynchronous thing to happen (i.e. a `Promise` to be resolved when running the `useEffect`), otherwise it'll never get a signal that the effect completed. You should be able to switch the failing tests to sync tests (remove `async` keyword and `await waitForNextUpdate`) and I'd expect them to pass.

src/hooks/useQuery.ts
andyrichardson: Haven't had time to look at this but I'm assuming we'll need a separate effect for managing mounting / unmounting.

src/hooks/useQuery.ts
andyrichardson: Thanks Parker, I'll investigate tonight 👍

src/hooks/useQuery.ts
parkerziegler: That makes sense to me, keep the effects separate (which also makes them easier to reason about).

src/hooks/useQuery.ts
parkerziegler: Looking a little more closely, it appears that the failed tests are related to the effect not re-running when new variables, query, and requestPolicy are received 🤔 Not totally sure why that would be.

src/hooks/useQuery.ts
andyrichardson: The addition of the request key to the effect was to cause a fetch not only on mount but whenever the query changes. I think this is what users would expect but will leave that for a seperate PR.

andyrichardson: @parkerziegler is the `useQuery.spec.ts` file addition intentional?

src/hooks/useQuery.ts
andyrichardson: @parkerziegler I've managed to fix the tests by removing the waitForUpdate call. Given tests themselves are checking that a re-fetch is triggered we can get away without waiting for state updates.

As for why those state updates aren't taking place in a test environment, I suspect it's related to the mock implementation of executeQuery.

parkerziegler: Gotcha, will dig into this more tomorrow to confirm. But agree, as long as we're sure that `executeQuery` is being called again when a new `query`, `variables`, or `requestPolicy` are received, then I'm not too worried about the state updates.

src/hooks/useQuery.spec.ts
```diff
@@ -130,7 +130,6 @@ describe('useQuery', () => {
     `;
```
parkerziegler: Think I saw a comment in an email from you @andyrichardson, but yep, to explain the naming behind this file, I basically wanted a way to differentiate it from `useQuery.test.tsx`. I'd be happy to use a different name. Part of the reason for bringing in `react-hooks-testing-library` was to make us less dependent on `enzyme`, to avoid re-assignment of let bindings in tests, _and_ to be able to get the direct values returned by our hooks so we can assert on them (which I think has some value). Perhaps we could specify these two files as `useQuery.enzyme.test.tsx` and `useQuery.rhtl.test.tsx`? I'm also fine with removing them if we don't think they have value or we don't like the API, I just find the it a bit easier to work w/ than `enzyme`.

andyrichardson: Cheers @parkerziegler - I initially asked but then found your PR with the explanation. No need to make changes 👍

I'm pretty sure `react-hooks-testing-library` is using react-test-renderer under the hood (we're not using enzyme). I agree the reassignment + added complexity is a pain 😑

Issue #245, opened 2019-05-31
`fetching` initially `false`?


tlrobinson:
All the documentation and code examples suggest if you check that `fetching` is `true` and `error` is falsey then you can assume `data` is not null.

e.x.

```
const YourComponent = () => {
  const [{ fetching, data }] = useQuery({ query: `{ todos { id } }` });
  return fetching ? <Loading /> : <List data={data.todos} />;
};
```

https://github.com/FormidableLabs/urql#quick-start-guide

However, that's not what I'm seeing. `fetching` is initially `false`:

<img width="373" alt="Screenshot 2019-05-31 10 43 54" src="https://user-images.githubusercontent.com/18193/58724387-4903e780-8391-11e9-9e1c-bd17d7610e3b.png">

This happens for both the `Query` component and `useQuery` hook. Unless you explicitly check that `data` is not null then these examples will crash.

Or am I missing something?

Version: 1.0.5

luisrudge:
I'm having the same issue right now 😬

d3vhound:
Yeah doesn't really make sense. It's cumbersome to have to work around `fetching` always being false at initial render. Hope this isn't intentional.

kitten:
the PRs #250 and #256 ensure that the initial state you receive from `useQuery` will always reflect what's going on internally.

They also lay the ground work necessary to implement SSR. So you can expect all of this to ship in v1.1

larsbs:
I still don't understand the reasoning behind this, because now, we can't just simply rely on `fetching` to know if we should render a loading state or the final component removing the possibility of having one single source of truth.

 Also, that always generates an extra rendering cycle for every component. First, `fetching = false` and `data = undefined`, and then `fetching = true` and `data = whatever`.

kitten:
@larsbs as posted above though, this will be fixed in the next version.

larsbs:
@kitten Oh sorry, I understood actually the opposite, that it was expected behavior and it would ship in the final version. Thanks.

kitten:
@larsbs sorry, I'll post a more extensive explanation 😅

Basically this was fixed by the two linked PRs which happened to set up some preliminary behaviour for the SSR PR as well.

During initial mount they cause effects and state changes to run synchronously due to two special hooks. Since our exchange pipeline is built to be mostly synchronous this means that the initial state will match the ongoing / completed operation more closely.

When it starts fetching the state change setting it to true will immediately be applied on mount without an additional render.

[...]

Issue #218, opened 2019-03-31
Server side rendering


shashkovdanil:
Do you plan to support server rendering?

kitten:
Yep, that's connected to this issue: #208

Since we were using hooks we needed another major building block for SSR which is now done: https://github.com/FormidableLabs/react-ssr-prepass

It's definitely one of the reasons why we haven't announced urql v1 publicly yet

ivan-kleshnin:
@kitten would be great if you guys add a NextJS integration example once SSR is finished.

BjoernRave:
any updates on this. I would love to try out urql, but this is keeping me from doing so :/

kitten:
@BjoernRave yea, This issue is at the top of my todo list so I'm expecting to get this done quite soon :)

helloguille:
Let us know if there's an alpha we could try. I'd be happy to help testing or help debugging.

kitten:
#250 and #256 introduce the groundwork for this to be completed. I'll write the API for getting the cache's state and rehydrating it soon, so this should be resolved shortly and we'll publish SSR support as part of v1.1

PR #248, opened 2019-06-01
(fix) - fetching on initial render

JoviDeCroock:
Fixes: https://github.com/FormidableLabs/urql/issues/245

After the initial render the effect gets executed that sets fetching to `true` 
So when initially rendering we see that fetching is false implying we have data, since after this we'll always see fetching -> data fetching -> data

Only on initial render we see a scenario where: fetching false but no data -> fetching -> data

Made this as a draft because I understand the why but maybe we could resolve it differently with a loaded property or something?

Snapshot:

```

exports[`on initial useEffect initialises default state 1`] = '
Object {
  "data": undefined,
  "error": undefined,
  "fetching": false,
}
';

```

Fails atm due to the initialState change

### DIFF
--- a/src/components/Query.tsx
+++ b/src/components/Query.tsx
@@ -60,7 +60,7 @@ class QueryHandler extends Component<QueryHandlerProps, QueryHandlerState> {
     executeQuery: this.executeQuery,
     data: undefined,
     error: undefined,
-    fetching: !this.props.pause,
+    fetching: false,
   };
 
   componentDidMount() {
--- a/src/components/Query.tsx
+++ b/src/components/Query.tsx
@@ -60,7 +60,7 @@ class QueryHandler extends Component<QueryHandlerProps, QueryHandlerState> {
     executeQuery: this.executeQuery,
     data: undefined,
     error: undefined,
-    fetching: false,
+    fetching: this.props.pause !== true,
   };
 
   componentDidMount() {
--- a/src/components/Query.tsx
+++ b/src/components/Query.tsx
@@ -60,7 +60,7 @@ class QueryHandler extends Component<QueryHandlerProps, QueryHandlerState> {
     executeQuery: this.executeQuery,
     data: undefined,
     error: undefined,
-    fetching: false,
+    fetching: !this.props.pause,
   };
 
   componentDidMount() {
--- a/src/components/Query.tsx
+++ b/src/components/Query.tsx
@@ -56,12 +56,15 @@ class QueryHandler extends Component<QueryHandlerProps, QueryHandlerState> {
     this.unsubscribe = teardown;
   };
 
-  state = {
-    executeQuery: this.executeQuery,
-    data: undefined,
-    error: undefined,
-    fetching: this.props.pause !== true,
-  };
+  constructor(props) {
+    super(props);
+    this.state = {
+      executeQuery: this.executeQuery,
+      data: undefined,
+      error: undefined,
+      fetching: props.pause !== true,
+    };
+  }
 
   componentDidMount() {
     this.executeQuery();
--- a/src/hooks/useQuery.ts
+++ b/src/hooks/useQuery.ts
@@ -35,9 +35,10 @@ export const useQuery = <T = any, V = object>(
 ): UseQueryResponse<T> => {
   const isMounted = useRef(false);
   const unsubscribe = useRef(noop);
-  const initialState = useRef({ fetching: true });
-  const updateState = useRef(update => {
-    initialState.current = update;
+  const initialState = useRef<UseQueryState<T>>({
+    data: undefined,
+    error: undefined,
+    fetching: true,
   });
 
   const client = useContext(Context);
@@ -53,7 +54,7 @@ export const useQuery = <T = any, V = object>(
         requestPolicy: args.requestPolicy,
       }),
       subscribe(({ data, error }) => {
-        updateState.current({ data, error, fetching: false });
+        initialState.current = { data, error, fetching: false };
       })
     );
     unsubscribe.current = teardown;
--- a/src/hooks/useQuery.ts
+++ b/src/hooks/useQuery.ts
@@ -33,19 +33,33 @@ export type UseQueryResponse<T> = [
 export const useQuery = <T = any, V = object>(
   args: UseQueryArgs<V>
 ): UseQueryResponse<T> => {
+  const isMounted = useRef(false);
   const unsubscribe = useRef(noop);
+  const initialState = useRef({ fetching: true });
+  const updateState = useRef(update => {
+    initialState.current = update;
+  });
 
   const client = useContext(Context);
+
   const request = useMemo(
     () => createRequest(args.query, args.variables as any),
     [args.query, args.variables]
   );
 
-  const [state, setState] = useState<UseQueryState<T>>({
-    fetching: !args.pause,
-    error: undefined,
-    data: undefined,
-  });
+  if (!isMounted.current) {
+    const [teardown] = pipe(
+      client.executeQuery(request, {
+        requestPolicy: args.requestPolicy,
+      }),
+      subscribe(({ data, error }) => {
+        updateState.current({ data, error, isFetching: false });
+      })
+    );
+    teardown();
+  }
+
+  const [state, setState] = useState<UseQueryState<T>>(initialState.current);
 
   const executeQuery = useCallback(
     (opts?: Partial<OperationContext>) => {
@@ -76,7 +90,16 @@ export const useQuery = <T = any, V = object>(
   );
 
   useEffect(() => {
-    executeQuery();
+    isMounted.current = true;
+    return () => {
+      isMounted.current = false;
+    };
+  }, []);
+
+  useEffect(() => {
+    if (isMounted.current) {
+      executeQuery();
+    }
     return unsubscribe.current;
   }, [executeQuery]);
 
--- a/src/hooks/useQuery.ts
+++ b/src/hooks/useQuery.ts
@@ -48,7 +48,7 @@ export const useQuery = <T = any, V = object>(
     [args.query, args.variables]
   );
 
-  if (!isMounted.current) {
+  if (!isMounted.current && !args.pause) {
     const [teardown] = pipe(
       client.executeQuery(request, {
         requestPolicy: args.requestPolicy,
--- a/src/hooks/useQuery.ts
+++ b/src/hooks/useQuery.ts
@@ -90,18 +90,18 @@ export const useQuery = <T = any, V = object>(
   );
 
   useEffect(() => {
-    isMounted.current = true;
-    return () => {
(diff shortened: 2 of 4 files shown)


kitten:
@JoviDeCroock Cheers for opening a PR! :tada: We'll have to do some other fixes though but I think that could be solved in combination with #218.

The reason is that we possibly want the initial state to match up with what the cache exchange gives us back synchronously. So depending on the cache it can also make sense to set `fetching` to `false` and have a result synchronously, so we'll have to probably solve this in combination with the small suspense rewrite. Currently since we're only triggering an exchange inside `useEffect` on the initial render we might not see the correct state reflected immediately.

JoviDeCroock:
@kitten agreed, do you want me to try and work on #218 I have time for this tomorrow

kitten:
@JoviDeCroock If you'd like to help out with that that'd be amazing! :heart: I don't have much time this weekend myself but would otherwise take a look at that next week.

Essentially we can use the fact that the exchange pipeline _should_ run synchronously. So on an initial run of the hook we could try to get the initial state with `executeQuery`. So if we get a synchronous result from it, which we should sometimes do from the cache, we can then immediately use that as the initial state. Otherwise we'd set `fetching` to `true`.

So far so good I hope 😓

This would then allow me to add suspense support to the client later on, since we're running the initial state query synchronously, which means I can easily add some more code to throw a promise in the client.

JoviDeCroock:
I find it hard to determine the fact that your client.execute would return synchronously isn't the point of that subscribe to not have to worry about that? Sorry if I'm totally missing the point.

kitten:
@JoviDeCroock It doesn't return anything synchronously, but if there isn't anything concurrent going on in the exchange chain it'll step through each operator synchronously. So the callback in `subscribe` might be called before anything else happens; at least that's what's supposed to happen when the cache already has a result

JoviDeCroock:
Yes, I can understand that and then we aren't fetching but as far as I know we can't be aware of a sync return in the first render since the setting of the state happens asynchronously anyway. I'll try and think a bit more in depth about this, don't want to bother you about these things if you don't have a lot of time

kitten:
@JoviDeCroock It should in theory work, since we're only dealing with initial mount, i.e. in some pseudo-code-ish fashion

```js
const useQuery = (/* ... */) => {
  const isInitialRender = useRef(true);
  const initialState = useRef({ fetching: true });
  const updateState = useRef(update => { initialState.current = update; });
  const teardown = useRef(null);

  if (isInitialRender.current) {
    isInitialRender.current = false;
    teardown.current = pipe(
      client.executeQuery(/* ... */),
      subscribe(result => updateState.current(result))
    );
  }

  const [state, setState] = useState(initialState.current);
  
  useEffect(() => {
    // this is where the old updating logic goes with request.key diffing etc
  }, [/* ... */]);

  // ...
};
```

Can probably be simplified but I can totally see a way to make this work so that on initial mount we get the correct initial state 🤔

src/hooks/useQuery.ts
```diff
@@ -76,7 +90,16 @@ export const useQuery = <T = any, V = object>(
   );
 
   useEffect(() => {
-    executeQuery();
+    isMounted.current = true;
```
JoviDeCroock: Thought we could also use this for future use cases instead of isInitialRender

src/hooks/useQuery.ts
```diff
@@ -76,7 +90,16 @@ export const useQuery = <T = any, V = object>(
...
+    return () => {
+      isMounted.current = false;
+    };
+  }, []);
+
+  useEffect(() => {
+    if (isMounted.current) {
+      executeQuery();
+    }
     return unsubscribe.current;
```
JoviDeCroock: This will cleanup the effect from the initial mount

JoviDeCroock:
Implemented this for useQuery, will try to find a way to test this asap and extend it to Query component as well.

src/hooks/useQuery.ts
```diff
@@ -33,19 +33,33 @@ export type UseQueryResponse<T> = [
...
     () => createRequest(args.query, args.variables as any),
     [args.query, args.variables]
   );
 
-  const [state, setState] = useState<UseQueryState<T>>({
-    fetching: false,
-    error: undefined,
-    data: undefined,
-  });
+  if (!isMounted.current) {
```
kitten: I really like the idea of reusing `isMounted` for this 👌 

I suppose we could also reuse `executeQuery = useCallback( /* ... */` now?

Maybe it could check whether `isMounted` is still `false` to update `initialState`? That would also mean that `updateState` becomes unnecessary.

I'm also not sure how `useState` works with Fiber but `setState` on `Component` was always batched, so it would only update state once. This even worked when the `Component` was still mounting. At least during researching the current SSR implementation I found out that `useState` still works similarly to that there. So if it also does in non-SSR then we could simply call `setState` and not worry about `initialState` either, since it might be batched into the first render

JoviDeCroock: I don't see how we can reuse executeQuery since to use that we already need  our setState defined, that useState inturns relies on initialState.current.

I did implement the second remove `updateState` part,  will ook more in depth later. Thanks for your excellent and through guidance. Amazing!

kitten: Yea, true, reusing `executeQuery` definitely hinges on whether `useState` can again come before it and whether it's actually going to change the `initialState`. At least in SSR this works because calling it triggers another "loop" through the component, so not technically a rerender but a rerun. Not sure if that's the case in normal, non-SSR React

src/hooks/useQuery.ts
```diff
@@ -33,19 +33,34 @@ export type UseQueryResponse<T> = [
 export const useQuery = <T = any, V = object>(
   args: UseQueryArgs<V>
 ): UseQueryResponse<T> => {
+  const isMounted = useRef(false);
```
andyrichardson: @JoviDeCroock heads up, this is going to conflict with [an existing pr](https://github.com/FormidableLabs/urql/pull/243).

src/hooks/useQuery.ts
```diff
@@ -33,19 +33,34 @@ export type UseQueryResponse<T> = [
...
 
   const client = useContext(Context);
+
   const request = useMemo(
     () => createRequest(args.query, args.variables as any),
     [args.query, args.variables]
   );
 
-  const [state, setState] = useState<UseQueryState<T>>({
-    fetching: false,
```
andyrichardson: Can we simplify this PR to only change the initial state?

This line being changed from `fetching: false` to `fetching: args.pause !== true` would set the initial state as fetching by default if pause is not specified.

kitten: `args.pause !== true` is in theory not necessary anymore if we handle this correctly via `executeQuery` being either active or not active after it's run.

I suppose this PR is still in a more "experimental" stage so maybe we can merge yours for now and if that leads to conflicts I can later resolve them 🤔

src/hooks/useQuery.ts
andyrichardson: I've added a few suggestions but if I'm not mistaken, the issue described leaves us only concerned about the initial state on mount.

It looks like we have two scenarios:
 - Pause is false and the initial state is  { fetching: true }
 - Pause is true and the initial state is { fetching: false }

If that's the case, we can make this change without the need for refs or additional logic.

```tsx
const [state, setState] = useState<UseQueryState<T>>({
  fetching: args.pause !== true,
  error: undefined,
  data: undefined,
});
```

kitten: The other scenario is some preparation for suspense SSR so we'll also would want to keep the executeQuery excursion that runs on initial mount outside of useEffect 

But depending on how React works the executeQuery useCallback could maybe be reused for that, simplifying this PR. We'll see how that pans out I guess. Might or might nor work 😅

src/hooks/useQuery.ts
andyrichardson: > args.pause !== true is in theory not necessary anymore if we handle this correctly via executeQuery being either active or not active after it's run.

Adding this logic to executeQuery would not solve the problem of the initial rendering state though?

src/hooks/useQuery.ts
andyrichardson: > SSR

I'm out of here :eyes: 

Are you saying that following SSR, the state and ref values are not synchronized from the backend?

src/hooks/useQuery.ts
JoviDeCroock: Talking about pause this should probably also be accounted in first render trying to get from cache else we’d make that option obsolete for first call.

src/hooks/useQuery.ts
JoviDeCroock: Up to you what this PR becomes, my initial solution was to make the initial state based on pause. If the current wip helps out more for future work maybe we can make a seperate simple solution PR and keep this one for the experimental work.

src/hooks/useQuery.ts
JoviDeCroock: Jup, saw that but in theory that invokes a will mount, no?

src/components/Query.tsx
```diff
@@ -56,12 +56,15 @@ class QueryHandler extends Component<QueryHandlerProps, QueryHandlerState> {
...
   };
 
-  state = {
-    executeQuery: this.executeQuery,
-    data: undefined,
-    error: undefined,
-    fetching: false,
-  };
+  constructor(props) {
+    super(props);
```
JoviDeCroock: I have further implemented the SSR approach but the sync would imply we execute our query here, this isn't possible since this.setState would result in a noop.

src/components/Query.tsx
```diff
@@ -56,12 +56,15 @@ class QueryHandler extends Component<QueryHandlerProps, QueryHandlerState> {
...
+    this.state = {
+      executeQuery: this.executeQuery,
+      data: undefined,
+      error: undefined,
+      fetching: props.pause !== true,
```
JoviDeCroock: This approach was suggested but isn't good for SSR as far as I understood

kitten:
@JoviDeCroock So I tested out my hypothesis and it works :tada: `setState` will synchronously update the initial state on initial mount before running any effects that will ensure that the user only sees the initial state that we want them to see. So just by running `executeQuery`. I definitely want to credit your work so would you like to review this and copy some of the changes over? Tests are WIP https://github.com/FormidableLabs/urql/pull/250

JoviDeCroock:
@kitten Will do, thanks for the response I'll try and check your PR tonight. What code would you need there?

Learned something new today thanks!

kitten:
@JoviDeCroock However you'd like to help 🙌 I also don't have tests for it yet... Mostly, I wouldn't have landed on that implementation this quickly without your help so at the very least I'll add you as a co-author to the merge commit. But let me know what you think about it and whether you think it'll cover everything we've talked about. That'd be super helpful. Cheers! :heart:

Issue #213, opened 2019-03-26
Generics support for Component


dotansimha:
Hi! 

I'm the maintainer of https://github.com/dotansimha/graphql-code-generator . 
[One of our community members suggested we should add a plugin](https://github.com/dotansimha/graphql-code-generator/issues/1575) that generates ready-to-use type-safe Components and hooks using your library. 
It sounds like a cool idea, but we currently not able to implement it because your implementation does not allow to override the type of `data` (https://github.com/FormidableLabs/urql/blob/master/src/components/Query.tsx#L16) or `variables`.

In the hooks implementation, we can pass the type using generics and override the `data` and `variables` types, which is great, because we can generate ready-to-use wrappers for it. 

It would be great to have support for generics, and then we'll implement the plugin for the codegen :)

kitten:
Yes, we weren't quite sure how to implement the generics on the component so that inference works correctly in all edge cases, and since put more focus on our hooks now we haven't quite been spending time on that 😅

Open to suggestions on what's the norm though 👍

dotansimha:
@kitten because React Components are just classes, you should easily add generics and then pass it to the props type, something like that:

```typescript

type QueryProps<T> = {
   variables: T;
}

type QueryState<T> = {
   data: T;
}

export class Query<TData = any, TVariables = any> extends React.Component<QueryProps<TVariables>, QueryState<TData>> {
  render() {
     // ...
  }
}
```

Then, to override it (we'll generate a similar output with the codegen):

```typescript
type MyCustomQueryData = {

};

type MyCustomQueryVariables = {

};

export class MyCustomQuery extends Query<MyCustomQueryData, MyCustomQueryVariables> { ... }
```

kitten:
@dotansimha yep, that so far is clear, but when I've implemented generics on components in the past that'd mean that when inference breaks on JSX one would have to manually assign generics to the component which is troublesome in a `tsx` file.

dotansimha:
@kitten any chance you can provide an example? I never noticed this issue before...

kitten:
@dotansimha I don't have any TS codebase readily available, but I did hit this issue a long time ago, so I'm wondering whether it's simply been fixed.

kitten:
Ok, seems like this is a non-issue now, since JSX+Generics will explicitly become easier in TS https://mariusschulz.com/blog/typescript-2-9-passing-generics-to-jsx-elements

parkerziegler:
[...]

PR #249, opened 2019-06-03
Add generics support for Query component.

parkerziegler:
This PR begins to address #213 by adding generics support to the `Query` component. I wanted to keep the diff small and reviewable so we can see how we like this pattern before adopting it in the other components and hooks.

Previously, the `data` returned by `Query`'s render prop was typed as `any`, so users got no type safety around the data returned by their GraphQL API. With these changes, `data` becomes typed by a generic `T` that's supplied to the `Query` component.

For example, let's say our `Query` component returns data of the following shape:

```json
{
   "data": {
      "todos": [
         "id": 1
         "text": "Todo"
         "complete": false
      ]
   }
}
```

We can then use generics to type `data` in our `Query` component.

```typescript
interface Todo {
   id: number,
   text: string,
   complete: boolean;
}

interface Data {
   todos: Todo[];
}

const TodoQuery = gql`
  query {
    todos {
      id
      text
      complete
    }
  }
`;

const Example: FC = () => {
  return (
    <Query<Data> query={TodoQuery}>
      {({ data }) => data.todos.map(todo => <Todo key={todo.id} {...todo} />)}
    </Query>
  );
};
```

If we attempt to access an unknown field (i.e. `data.todo`) or use a property in a non-type safe way, the compiler will alert us 🎉 The generic `T` will default to `any`, so this doesn't affect any existing code using `urql` (it's 100% opt in), so `<Query query={TodoQuery}>` in the above example is still totally valid.

### DIFF
--- a/src/components/Query.tsx
+++ b/src/components/Query.tsx
@@ -1,28 +1,31 @@
 import { DocumentNode } from 'graphql';
-import React, { Component, FC, ReactNode } from 'react';
+import React, { Component, ReactNode } from 'react';
 import { pipe, subscribe } from 'wonka';
 import { Client } from '../client';
 import { Consumer } from '../context';
 import { Omit, OperationContext, RequestPolicy } from '../types';
 import { CombinedError, createRequest, noop } from '../utils';
 
-interface QueryHandlerProps {
+interface QueryHandlerProps<T> {
   query: string | DocumentNode;
   variables?: object;
   client: Client;
   requestPolicy?: RequestPolicy;
   pause?: boolean;
-  children: (arg: QueryHandlerState) => ReactNode;
+  children: (arg: QueryHandlerState<T>) => ReactNode;
 }
 
-interface QueryHandlerState {
+interface QueryHandlerState<T> {
   fetching: boolean;
-  data?: any;
+  data?: T;
   error?: CombinedError;
   executeQuery: (opts?: Partial<OperationContext>) => void;
 }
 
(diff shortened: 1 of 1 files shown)


src/components/Query.tsx
```diff
@@ -84,8 +87,10 @@ class QueryHandler extends Component<QueryHandlerProps, QueryHandlerState> {
   }
 }
 
-export type QueryProps = Omit<QueryHandlerProps, 'client'>;
+export type QueryProps<T> = Omit<QueryHandlerProps<T>, 'client'>;
 
-export const Query: FC<QueryProps> = props => (
-  <Consumer>{client => <QueryHandler {...props} client={client} />}</Consumer>
-);
+export function Query<T = any>(props: QueryProps<T>) {
```
parkerziegler: Here's where the default `any` comes in. I elected to refactor from `const` to `function` because of the strange syntax involved in supplying generics to `const` declarations with JSX, i.e. `export const Query = <T extends {}>(props: QueryProps<T>) = {}`. The above feels more readable to me.

kitten:
@parkerziegler Is #252 good enough as well? Since removes the discrete component implementations it might supersede this PR

parkerziegler:
@kitten yep, I just took a peak at #252 and I like that a lot! Looks like it should support the same passing of generics to components, so I'm all for it. I'll close this PR cc/ @dotansimha #252 should address #213.

PR #250, opened 2019-06-03
Implement immediate effect and refactor

### LINKED ISSUE #245: `fetching` initially `false`?

kitten:
Fix #245 

Initial stuff that will become necessary for #218

This implements two hooks:

- `useRequest` to simplify the creation of requests while keeping a stable reference, so we don't have to check for `request.key`
- `useImmediateEffect` to simplify running an effect immediately outside of `useEffect` for SSR/suspense, which then otherwise behaves like a normal `useEffect`.

### DIFF
--- a/src/hooks/useImmediateEffect.ts
+++ b/src/hooks/useImmediateEffect.ts
@@ -1,31 +1,34 @@
 import { useRef, useEffect, useCallback } from 'react';
 
-enum EffectState {
-  BeforeMount = 0,
-  AfterMount = 1,
-  Render = 2,
+enum LifecycleState {
+  WillMount = 0,
+  DidMount = 1,
+  Update = 2,
 }
 
 type Effect = () => void | (() => void);
 
 /** This executes an effect immediately on initial render and then treats it as a normal effect */
-export const useImmediateEffect = (effect: Effect, changes?: any[]) => {
-  const state = useRef(EffectState.BeforeMount);
+export const useImmediateEffect = (
+  effect: Effect,
+  changes: ReadonlyArray<any>
+) => {
+  const state = useRef(LifecycleState.WillMount);
   const execute = useCallback(effect, changes);
 
   useEffect(() => {
     // Initially we skip executing the effect since we've already done so on
     // initial render, then we execute it as usual
-    if (state.current === EffectState.Render) {
+    if (state.current === LifecycleState.Update) {
       return execute();
     } else {
-      state.current = EffectState.Render;
+      state.current = LifecycleState.Update;
     }
   }, [execute]);
 
   // On initial render we just execute the effect
-  if (state.current === EffectState.BeforeMount) {
-    state.current = EffectState.AfterMount;
+  if (state.current === LifecycleState.WillMount) {
+    state.current = LifecycleState.DidMount;
     execute();
   }
 };
--- a/src/hooks/useImmediateEffect.ts
+++ b/src/hooks/useImmediateEffect.ts
@@ -1,0 +1,30 @@
+import { useRef, useEffect } from 'react';
+
+enum EffectState {
+  BeforeMount = 0,
+  AfterMount = 1,
+  Render = 2,
+}
+
+type Effect = () => void | (() => void);
+
+/** This executes an effect immediately on initial render and then treats it as a normal effect */
+export const useImmediateEffect = (effect: Effect) => {
+  const state = useRef(EffectState.BeforeMount);
+
+  useEffect(() => {
+    // Initially we skip executing the effect since we've already done so on
+    // initial render, then we execute it as usual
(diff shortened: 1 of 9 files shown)


src/hooks/useImmediateEffect.ts
```diff
@@ -0,0 +1,31 @@
...
+  BeforeMount = 0,
+  AfterMount = 1,
+  Render = 2,
+}
+
+type Effect = () => void | (() => void);
+
+/** This executes an effect immediately on initial render and then treats it as a normal effect */
+export const useImmediateEffect = (effect: Effect, changes?: any[]) => {
+  const state = useRef(EffectState.BeforeMount);
```
JoviDeCroock: Could we rename this to for example lifecycleState or something? Don't want to nit but this could be confusing for newer contributors

src/hooks/useImmediateEffect.ts
```diff
@@ -0,0 +1,31 @@
...
+  const execute = useCallback(effect, changes);
+
+  useEffect(() => {
+    // Initially we skip executing the effect since we've already done so on
+    // initial render, then we execute it as usual
+    if (state.current === EffectState.Render) {
```
JoviDeCroock: Won't this lead to double execution with the if on line 26? Shouldn't we just do `execute()`?

src/hooks/useImmediateEffect.ts
kitten: Lifecycle is a good call 👌

JoviDeCroock:
Have made a little seperate branch where I'm testing some implementations for Query and made my suggested changes. Will  PR against this branch

kitten:
@JoviDeCroock I've got another PR here that just drops compatibility with older React versions and reuses hooks, if you mean the `Query` component? Wdyt? https://github.com/FormidableLabs/urql/pull/252

JoviDeCroock:
That fully removes the need for it yes, allright. I think we're almost there then, just alter the snapshots and we're finished 🎉

kitten:
@JoviDeCroock Cheers! We're working on fleshing out the core library so we can move on to write more exchanges soon!

The unsubscription in useImmediateEffect and the main hooks should be more solid now. Just waiting for a couple of reviews from other core contributors so this can go in 🙌

After this SSR shouldn't be too hard to get done

kitten:
@andyrichardson I'll merge this for now, but let's have a retroactive review at some point :+1:

src/hooks/useRequest.test.ts
```diff
@@ -0,0 +1,31 @@
+import { renderHook } from 'react-hooks-testing-library';
+import { queryGql } from '../test-utils';
+import { useRequest } from './useRequest';
+
+it('preserves instance of request when key has not changed', () => {
+  let { query, variables } = queryGql;
+
+  const { result, rerender } = renderHook(
+    ({ query, variables }) => useRequest(query, variables),
```
parkerziegler: I think we might be unintentionally shadowing `query` and `variables` here. Typically I'll define what's in `initialProps` to be different from my test data to ensure that all calls to `rerender` are using the data passed in `initialProps` or `rerender`, i.e.:

```
const { result, rerender } = renderHook(
   ({ q, v }) => useRequest(q, v),
   { initialProps: { q: query, v: variables }
);
```

Functionally it doesn't matter b/c `query` (inside `renderHook`) -> `initialProps.query` -> `query` (from `queryGql`), and we use `let` bindings to update the references to `variables` below, so this is more of a readability suggestion to make it clear where things are coming from 🤗

src/hooks/useRequest.ts
```diff
@@ -0,0 +1,23 @@
+import { DocumentNode } from 'graphql';
+import { useRef, useMemo } from 'react';
+import { GraphQLRequest } from '../types';
+import { createRequest } from '../utils';
+
+/** Creates a request from a query and variables but preserves reference equality if the key isn't changing */
+export const useRequest = (
+  query: string | DocumentNode,
+  variables?: any
```
parkerziegler: How would you feel about using a generic here so we can pass it through from `useQuery` / `useSubscription`? Happy to make a quick PR if you think it's worthwhile.

PR #252, opened 2019-06-03
Remove create-react-context and bump React peer dependency

### LINKED ISSUE #213: Generics support for Component

kitten:
It's probably been a mistake to include `create-react-context` and support non-hook implementations in practice. This will basically get rid of most of the maintenance burden we have right now, since we won't have to maintain two APIs for each hook/component.

Fix #213


src/components/Mutation.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export interface MutationState<T, V> extends UseMutationState<T> {
+  executeMutation: (variables?: V) => Promise<OperationResult<T>>;
+}
+
+export function Mutation<T = any, V = any>({
+  children,
+  query,
+}: MutationProps<T, V>): ReactNode {
+  const [state, executeMutation] = useMutation<T, V>(query);
```
JoviDeCroock: This change requires a peerDep of React 16.8

src/components/Query.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export interface QueryState<T> extends UseQueryState<T> {
+  executeQuery: (opts?: Partial<OperationContext>) => void;
+}
+
+export function Query<T = any, V = any>({
+  children,
+  ...args
+}: QueryProps<T, V>): ReactNode {
+  const [state, executeQuery] = useQuery<T, V>(args);
```
JoviDeCroock: Same here, I don't think we can set peerdep to 16.6.1 is we're using hooks in our render methods

kitten: Whoops

src/components/Subscription.ts
```diff
@@ -0,0 +1,22 @@
...
+export interface SubscriptionProps<T, V, R = T> extends UseSubscriptionArgs<V> {
+  handler?: SubscriptionHandler<T, R>;
+  children: (arg: UseSubscriptionState<R>) => ReactNode;
+}
+
+export function Subscription<T = any, R = T, V = any>({
+  children,
+  handler,
+  ...args
+}: SubscriptionProps<T, V, R>): ReactNode {
```
parkerziegler: Super nit – it looks like `SubscriptionProps` receives its generics in the order `<T, V, R>`, but `Subscription`'s generics (and `useSubscription`) are in the order `<T, R, V>`. I'd keep the ordering consistent if possible here just to keep these signatures easy to read.

kitten: Good catch! 👍

PR #255, opened 2019-06-04
Bump dependencies

### LINKED ISSUE #251: bs-platform peer dependency warning
After installing urql 1.0.5, I get this warning from npm:
```
npm WARN bs-rebel@0.2.3 requires a peer of bs-platform@>=2.2.0 but none is installed. You must install peer dependencies yourself.
```
It seems like this dependency is brought in by wonka, used by urql:
```
$ npm list bs-rebel
<project>@0.1.0 /mnt/c/Users/mjadcza/Projects/<etc>
└─┬ urql@1.0.5
  └─┬ wonka@2.0.2
    └── bs-rebel@0.2.3
```
Do I actually need to install this peer dependency? If this is to do with the internal workings of wonka, I think urql should just explicitly require it, no? And if this is something I should install myself, it should probably be documented.

kitten:
Fix #251


src/exchanges/__snapshots__/fetch.test.ts.snap
```diff
@@ -144,7 +144,7 @@ Object {
         ],
         "results": Array [
           Object {
-            "isThrow": false,
+            "type": "return",
```
andyrichardson: @kitten is this change in return type expected?

andyrichardson:
Other than the one comment, looks good!.

Assuming everyone else is a fan, I think we should be pinning our dependencies.

src/exchanges/__snapshots__/fetch.test.ts.snap
kitten: Yep, seems to be due to the Jest upgrade

kitten:
@andyrichardson Pinning in libraries is not as advantageous as in apps. `dependencies` must not be pinned so we increase the likelihood of deduplication in a user's app; for instance `prop-types` and `graphql-tag` should never duplicate.

For `devDependencies` we could indeed pin more, but I think it makes more sense to provoke breakage locally, especially since we're technically tied to one version due to our `yarn.lock` file anyway.

andyrichardson:
TIL: pinning dependencies in libraries causes more duplication.

Cheers man - approved :+1:

PR #256, opened 2019-06-04
Ensure that state changes are applied immediately on mount

### LINKED ISSUE #209: useMutation value should be constant
I'm using a useMutation hook in a parent component, and passing the result down to a child component as the only prop. The child component rerenders whenever the parent rerenders. This means that the result value from useMutation keeps changing. It would be better if it were constant so that the child didn't rerender.

### LINKED ISSUE #245: `fetching` initially `false`?

kitten:
Fix #245
Fix #209 (See 17feeca)

It seems that in the current version React does
trigger *two* renders in our case.

We call executeQuery immediately which has a couple
of setStates. These are batched, it seems, but
do trigger a second update of the underlying
component, which is a shame.

Instead, we know that the initial executeQuery can
just safely merge its results into the initial
state, if it's during the initial mount.

This is why useImmediateState is added, which does
just that. Not the nicest way to solve this, but
the safest until React ships batched updates by
default.

This is basically a solution that will continue to
ensure that we won't have to add any state-related
stuff outside of `executeQuery`

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -25,7 +25,12 @@
     "@typescript-eslint/no-explicit-any": "off",
     "@typescript-eslint/array-type": "off",
     "react-hooks/rules-of-hooks": "error",
-    "react-hooks/exhaustive-deps": "warn",
+    "react-hooks/exhaustive-deps": [
+      "warn",
+      {
+        "additionalHooks": "useImmediateEffect"
+      }
+    ],
     "react/prop-types": "off",
     "react/no-children-prop": "off",
     "sort-keys": "off",
--- a/src/hooks/useImmediateEffect.test.ts
+++ b/src/hooks/useImmediateEffect.test.ts
@@ -1,5 +1,5 @@
 import React from 'react';
-import { renderHook } from 'react-hooks-testing-library';
+import { renderHook, act } from 'react-hooks-testing-library';
 import { useImmediateEffect } from './useImmediateEffect';
 
 it('calls effects immediately on mount', () => {
@@ -8,14 +8,37 @@ it('calls effects immediately on mount', () => {
   const effect = jest.fn();
 
   spy.mockImplementation(useEffect);
-  renderHook(() => useImmediateEffect(effect, [effect]));
+
+  renderHook(() => {
+    useImmediateEffect(effect, [effect]);
+    expect(effect).toHaveBeenCalledTimes(1);
+  });
 
   expect(effect).toHaveBeenCalledTimes(1);
-  expect(effect).toHaveBeenCalledTimes(1);
-
   expect(useEffect).toHaveBeenCalledWith(expect.any(Function), [
     expect.any(Function),
   ]);
 
-  useEffect.mockRestore();
+  spy.mockRestore();
+});
+
+it('behaves like useEffect otherwise', () => {
+  const spy = jest.spyOn(React, 'useEffect');
+  const effect = jest.fn();
+
+  const { result } = renderHook(() => {
+    const [ref, setState] = React.useState({});
+    useImmediateEffect(effect, [ref]);
+
+    expect(effect).toHaveBeenCalledTimes(1);
+    expect(spy).toHaveBeenCalled();
+
+    const forceUpdate = () => setState({});
+    return forceUpdate;
+  });
+
+  act(() => result.current()); // forceUpdate
+  expect(effect).toHaveBeenCalledTimes(2);
+
+  spy.mockRestore();
 });
--- a/src/hooks/useImmediateEffect.ts
+++ b/src/hooks/useImmediateEffect.ts
@@ -9,7 +9,7 @@ enum LifecycleState {
 
 type Effect = () => () => void;
 
-/** This executes an effect immediately on initial render and then treats it as a normal effect */
+/** This is a drop-in replacement for useEffect that will execute the first effect that happens during initial mount synchronously */
 export const useImmediateEffect = (
   effect: Effect,
   changes: ReadonlyArray<any>
--- a/src/hooks/useImmediateEffect.ts
+++ b/src/hooks/useImmediateEffect.ts
@@ -1,4 +1,6 @@
-import { useRef, useEffect, useCallback } from 'react';
+/* eslint-disable react-hooks/exhaustive-deps */
+
+import { useRef, useEffect } from 'react';
 import { noop } from '../utils';
 
 enum LifecycleState {
@@ -16,22 +18,21 @@ export const useImmediateEffect = (
 ) => {
   const teardown = useRef(noop);
   const state = useRef(LifecycleState.WillMount);
-  const execute = useCallback(effect, changes);
 
   // On initial render we just execute the effect
   if (state.current === LifecycleState.WillMount) {
     state.current = LifecycleState.DidMount;
-    teardown.current = execute();
+    teardown.current = effect();
   }
 
   useEffect(() => {
     // Initially we skip executing the effect since we've already done so on
     // initial render, then we execute it as usual
     if (state.current === LifecycleState.Update) {
-      return (teardown.current = execute());
+      return (teardown.current = effect());
     } else {
       state.current = LifecycleState.Update;
       return teardown.current;
     }
-  }, [execute]);
+  }, changes);
 };
--- a/src/hooks/useImmediateState.ts
+++ b/src/hooks/useImmediateState.ts
@@ -1,0 +1,32 @@
+import { useRef, useEffect, useState, useCallback } from 'react';
+
+type SetStateAction<S> = S | ((prevState: S) => S);
+type SetState<S> = (action: SetStateAction<S>) => void;
+
+export const useImmediateState = <S extends {}>(init: S): [S, SetState<S>] => {
+  const isMounted = useRef(false);
+  const initialState = useRef<S>({ ...init });
+  const [state, setState] = useState<S>(initialState.current);
+
+  const updateState: SetState<S> = useCallback((action: SetStateAction<S>) => {
+    // If the component is currently before its initial mount, update
+    // the state immediately
+    if (isMounted.current) {
+      setState(action);
+    } else if (typeof action === 'function') {
+      const update = (action as any)(initialState.current);
+      Object.assign(initialState.current, update);
+    } else {
+      Object.assign(initialState.current, action as any);
+    }
+  }, []);
+
+  useEffect(() => {
+    isMounted.current = true;
+    return () => {
+      isMounted.current = false;
+    };
+  }, []);
+
+  return [state, updateState];
+};
--- a/src/hooks/useImmediateState.ts
+++ b/src/hooks/useImmediateState.ts
@@ -3,14 +3,15 @@ import { useRef, useEffect, useState, useCallback } from 'react';
 type SetStateAction<S> = S | ((prevState: S) => S);
 type SetState<S> = (action: SetStateAction<S>) => void;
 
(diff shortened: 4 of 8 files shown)


kitten:
cc @JoviDeCroock fyi: We still get a double render on mount since it seems that React will schedule a batched update on setState during initial mount instead of aborting to render that component, so this makes set State mutable during initial mount

src/hooks/useImmediateState.ts
```diff
@@ -0,0 +1,32 @@
...
+type SetStateAction<S> = S | ((prevState: S) => S);
+type SetState<S> = (action: SetStateAction<S>) => void;
+
+export const useImmediateState = <S extends {}>(init: S): [S, SetState<S>] => {
+  const isMounted = useRef(false);
+  const initialState = useRef<S>({ ...init });
+  const [state, setState] = useState<S>(initialState.current);
+
+  const updateState: SetState<S> = useCallback((action: SetStateAction<S>) => {
+    // If the component is currently before its initial mount, update
```
JoviDeCroock: For this comment shouldn't it be !isMounted.current?

src/hooks/useImmediateState.ts
```diff
@@ -0,0 +1,32 @@
...
+    // the state immediately
+    if (isMounted.current) {
+      setState(action);
```
JoviDeCroock: so we set the potential function as state? Sorry if this question is weird trying to grasp it

src/hooks/useImmediateState.ts
kitten: Oh the comment goes for the entire block; I'll pull it out of useCallback to clarify

src/hooks/useImmediateState.ts
kitten: If we're in "will mount" basically we don't call setState and update the state mutably

JoviDeCroock: Oh, yes good point thanks for clarifying.

src/hooks/useImmediateState.ts
JoviDeCroock: Ah allright, sorry got confused

kitten: No worries 👌 I'll add more comments

src/hooks/useImmediateState.ts
```diff
@@ -0,0 +1,32 @@
...
+
+type SetStateAction<S> = S | ((prevState: S) => S);
+type SetState<S> = (action: SetStateAction<S>) => void;
+
+export const useImmediateState = <S extends {}>(init: S): [S, SetState<S>] => {
+  const isMounted = useRef(false);
+  const initialState = useRef<S>({ ...init });
+  const [state, setState] = useState<S>(initialState.current);
+
+  const updateState: SetState<S> = useCallback((action: SetStateAction<S>) => {
```
andyrichardson: It looks like this does the following:
 - on first render
 - hijack `setState` call
 - uses mutation to update state value

Is that the case?

kitten: Ehhh, yes if you mean wrapping it by hijacking 👍😅

andyrichardson: Updated!

I found this pretty tricky to understand but I see what you're doing now. I think the same could be achieved by currying the executeQuery function and passing in:
 - A mutation setState function on initial render
 - The actual setState function on further renders

That would get rid of the need for an abstraction.

```tsx
const useQuery = () => {
    const isMounted = useRef(false);
    const initialState = useRef({ ... });
    const [state, setState] = useState(initialState);

    const updateState = useMemo(() => isMounted.current ? setState : (arg) => Object.assign(initialState, arg), [isMounted.current]);

    const executeQuery = useMemo(() => doExecuteQuery(updateState), [updateState])
    
    // ...
}

const doExecuteQuery = (setState, otherArgs) => () => // ...
```

kitten: It's not quite the same as you'd then trigger an update since `executeQuery` changes, since `updateState` changes after `initialMount` changes... yea, long chain of stuff happening. This is also not a deterministic update since it's depending on a ref, which means we get an unexpected query on the next update, whenever it comes 😅 anyway, since this is not about the code snippet:

I wouldn't inline this as the abstraction is nice imho; I think it reduces the amount of complexity for external contributors since they're just looking at drop-in replacements for `useEffect` and `useState` that they can assume to work the same.

Given that all bugs / things we've now fixed are really subtle variations on how one would usually write these features or this behaviour, the hooks ensure that we can continue to write them normally, mostly/hopefully bug-free, without making our code take on more concerns.

kitten:
@andyrichardson Added some tests for `useImmediateState` and an additional one for `useImmediateEffect`. They should both ensure that these hooks behave as we expect, so like a normal `useState` and `useEffect` respectively apart from the special behaviour on initial mount.

src/hooks/useImmediateState.ts
```diff
@@ -0,0 +1,33 @@
+import { useRef, useEffect, useState, useCallback } from 'react';
+
+type SetStateAction<S> = S | ((prevState: S) => S);
+type SetState<S> = (action: SetStateAction<S>) => void;
+
+/** This is a drop-in replacement for useState, limited to object-based state. During initial mount it will mutably update the state, instead of scheduling a React update using setState */
+export const useImmediateState = <S extends {}>(init: S): [S, SetState<S>] => {
+  const isMounted = useRef(false);
+  const initialState = useRef<S>({ ...init });
```
andyrichardson: Just had a thought about this - we aren't using this object anywhere other than initial render so the use of useRef might be misleading/unnecessary.

kitten: We are using this as a throwaway target for unmounted setState calls but otherwise yea, not really needed on client-side.

But during React server-side rendering this can actually matter, since some calls will reset component rendering and start over, which means that since our client is external we might get subtly different results.

It probably doesn't matter but that's why I chose to use refs.

src/hooks/useImmediateState.ts
```diff
@@ -0,0 +1,33 @@
...
+  const [state, setState] = useState<S>(initialState.current);
+
+  // This wraps setState and updates the state mutably on initial mount
+  // It also prevents setting the state when the component is unmounted
+  const updateState: SetState<S> = useCallback((action: SetStateAction<S>) => {
+    if (isMounted.current) {
+      setState(action);
+    } else if (typeof action === 'function') {
```
andyrichardson: Can we return from these conditions rather than having to parse multiple conditions.

```tsx
if (isMounted.current) {
  return setState(action);
}

const actionValue = typeof action === "function" ? action(initialState.current) : action;
Object.assign(initialState.current, actionValue);
```

kitten: Doesn't really make a difference though so not to be too bikesheddy, but since we just follow some loose eslint rules,
you can always change it to what you prefer when you happen to refactor or rewrite this part of the codebase 😅

.eslintrc.json
```diff
@@ -25,7 +25,12 @@
     "@typescript-eslint/no-explicit-any": "off",
     "@typescript-eslint/array-type": "off",
     "react-hooks/rules-of-hooks": "error",
-    "react-hooks/exhaustive-deps": "warn",
+    "react-hooks/exhaustive-deps": [
+      "warn",
+      {
+        "additionalHooks": "useImmediateEffect"
```
parkerziegler: ❤️this really is such a nice surprise!

src/hooks/useImmediateEffect.test.ts
```diff
@@ -8,14 +8,37 @@ it('calls effects immediately on mount', () => {
...
-  renderHook(() => useImmediateEffect(effect, [effect]));
 
-  expect(effect).toHaveBeenCalledTimes(1);
-  expect(effect).toHaveBeenCalledTimes(1);
+  renderHook(() => {
+    useImmediateEffect(effect, [effect]);
+    expect(effect).toHaveBeenCalledTimes(1);
+  });
 
+  expect(effect).toHaveBeenCalledTimes(1);
```
parkerziegler: Not sure I totally get the logic behind the second `.toHaveBeenCalledTimes` here. As I read this, I'd expect `useImmediateEffect` to get called once on first render (which we verify with the first assertion), then I wouldn't expect us to need to assert again. So can you explain the reasoning behind having this second assertion?

kitten: react-hooks-testing-library had this weird thing where it'd execute effects after the initial mount automatically.

So if the initial effect was actually executed after manual call inside the initial mount, we'd see two calls here

src/hooks/useImmediateEffect.ts
```diff
@@ -1,4 +1,6 @@
-import { useRef, useEffect, useCallback } from 'react';
+/* eslint-disable react-hooks/exhaustive-deps */
```
parkerziegler: Is this intentional? Seems somewhat dangerous and easy to overlook. What would you say to just using `eslint-disable-next-line` where needed?

parkerziegler: What would you say to moving this above the dependency array and adding a comment on why we don't want `effect` included. I personally am interested to know why (looks like you can't do `[...changes, effect]` or `changes.concat([effect]);`). It seems like `effect` could potentially get stale if not included in the dependency array.

kitten: The effect itself can get stale in fact, but this follows React's logic. So we'll only update when the specified effects update not the passed closure. I think since this is quite small I just didn't add it inline 😅

src/hooks/useImmediateState.ts
```diff
@@ -0,0 +1,33 @@
...
+      const update = (action as any)(initialState.current);
+      Object.assign(initialState.current, update);
```
parkerziegler: Alright, totally could be missing something, but don't we need to do the assignment to `initialState.current` here to ensure the mutable ref gets updated, i.e.

```
initialState.current = Object.assign(initialState.current, update)
```

kitten: Nah, the first argument is being assigned to mutably 😀

src/hooks/useQuery.ts
```diff
@@ -28,61 +29,54 @@ export type UseQueryResponse<T> = [
...
         unsubscribe.current = noop;
         return;
+      } else {
+        setState(s => ({ ...s, fetching: true }));
       }
 
-      setState(s => ({ ...s, fetching: true }));
-
-      const [teardown] = pipe(
+      [unsubscribe.current] = pipe(
```
parkerziegler: A bit confused by how this works. Is this array destructuring also capable of reassignment, i.e. this basically turns into:

```
const [teardown] = pipe(/* stuff */)
unsubscribe.current = teardown;
```

It seems like a dope feature of TS if true.

src/hooks/useImmediateState.ts
parkerziegler: Ah right, I forgot that `Object.assign` mutates the original object 👍

src/hooks/useQuery.ts
kitten: Yep, works in pure JS as well, since the destructuring target can almost be any expression afaik

PR #258, opened 2019-06-04
Add proposed promisified query/mutation calls to Client

### LINKED ISSUE #216: Add promisified request methods on the Client (especially for non-React API calls)

kitten:
Fix #216

We can maybe continue the discussion here? This is a proposal for providing very simple, programmatic `query` and `mutation` methods. They'd simply be used in situations where a user probably isn't writing new bindings (e.g. using `urql` without the React API but with Vue) but is instead just doing a one off call.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1,7 +1,11 @@
 /* eslint-disable @typescript-eslint/no-use-before-define */
+
+import { DocumentNode } from 'graphql';
+
 import {
   filter,
   makeSubject,
+  toPromise,
   onEnd,
   onStart,
   pipe,
@@ -25,6 +29,8 @@ import {
   OperationType,
 } from './types';
 
+import { createRequest } from './utils';
+
 /** Options for configuring the URQL [client]{@link Client}. */
 export interface ClientOptions {
   /** Target endpoint URL such as `https://my-target:8080/graphql`. */
@@ -33,6 +39,11 @@ export interface ClientOptions {
   fetchOptions?: RequestInit | (() => RequestInit);
   /** An ordered array of Exchanges. */
   exchanges?: Exchange[];
+}
+
+export interface MakeOperation extends OperationContext {
+  query: string | DocumentNode;
+  variables?: object;
 }
 
 interface ActiveOperations {
@@ -158,6 +169,7 @@ export class Client {
     }
   };
 
+  /** Returns a stream executing a GraphQL query given a request */
   executeQuery = (
     query: GraphQLRequest,
     opts?: Partial<OperationContext>
@@ -166,6 +178,7 @@ export class Client {
     return this.executeRequestOperation(operation);
   };
 
+  /** Returns a stream executing a GraphQL subscription given a request */
   executeSubscription = (
     query: GraphQLRequest,
     opts?: Partial<OperationContext>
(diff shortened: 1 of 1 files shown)


CaitlinWeb:
Didn't try it out, but looks good to me! Very convenient!

Maybe it's nitpicking but should the name of `client.query` and `client.mutation` reflect the action you're taking with the query and mutation? In this case it's a "request", so it would be `client.requestQuery` and `client.requestMutation`. That way there's a clear difference between `executeQuery` (stream) and `requestQuery` (promise).

andyrichardson:
Looks really good! Removing the need to create an operation is a pretty strong shout.

I'd be interested to hear if people would prefer to receive a stream by default and have the option of casting to a promise without the need for Wonka imports.

In that case, all three of these would work

```tsx
// Original style
pipe(
  client.executeQuery(arg),
  toPromise
)

// New subscribe abstraction
client.executeQuery(arg).subscribe(...);

// New to promise abstraction
await client.executeQuery(arg).toPromise();
```

This would also address the inconsistency and redundancy in the API which would otherwise look like this:

```tsx
// For subscription
const operation = createOperation(notOp);
pipe(
  client.executeQuery(operation),
  subscribe(...)
)

// For promise
client.query(notOp)
```

BjoernRave:
I would really like to see this get merged :)

kitten:
@BjoernRave we will eventually merge some proposal that fulfils all our requirements but this one is just a really minimal layer tha my uses Wonka operators minimally. Do you can already copy the added methods to get the same result :)

That being said, why is our SSR implementation not cutting it for you?

andyrichardson:
@bjoernrave there are some challenges with implementing a promise-only API. Other than the additional complexity, adding promise support conflicts with the underlying architecture of URQL.

Queries specifically are tricky as responses are streamed to the client. We could use `take(1)` to only get the first response and then `toPromise` but there are some challenges with this:
 - what if the cache policy is `cache-then-network`
 - what if the first streamed result is from another client request (race-condition)

Exchanges can be added to solve these problems (such as the dedupe exchange preventing race conditions) but the client is not and should not be coupled/dependent on any exchanges and the requested feature is a client addition.

Hopefully what im explaining isn't too abstract - our upcoming devtools will better visualize the inner workings of Urql. Maybe we'll find a way to effectively solve this problem but right now I'm not sure a change to the client is going to be of value. 

The good news is thanks to the work of @kitten and others on the team, Wonka now has documentation which should help with anyone who isn't sure how to turn their query/mutation streams into a promise.

BjoernRave:
@andyrichardson Thanks for the explanation :) I will wait then and do what @kitten said and just copy/paste that stuff over.

@kitten I am not saying the SSR implementation is not cutting it, but in next.js we have this method: `getInitialProps` to fetch some initial data before the page has loaded and to do that with graphql a promise-based query mechanism is needed, right?

kitten:
@BjoernRave yep, just to clarify though, if you’re looking to load data for your UI that you’d usually just want to load with our hooks or components, then our SSR example with Next.js basically shows how to automate this without using getInitialProps :)

kitten:
Closing due to inactivity

avetisk:
While I somewhat understand why it's not that simple to add a promisified simple `query()` and `mutation()`, I find it extremely frustrating as I have different cases where I would need it.

There are many cases where there's no direct dependency to the UI, and you need to make calls without going through React.

JoviDeCroock:
This has been added @avetisk `client.mutation` should work

avetisk:
@JoviDeCroock Oh! I actually missed this in the docs: https://github.com/FormidableLabs/urql/blob/master/docs/api/core.md#query-and-mutation

Maybe it would be good to add it in https://formidable.com/open-source/urql/docs/urql-outside-react/ ?

JoviDeCroock:
@avetisk We're working on the new documentation as we speak at the moment. If you want to try them out you can always visit the open PR, navigate to `packages/site` install deps and yarn start. That would be of great help to us! Feedback is so much appreciated

kitten:
Also some of the new content is here and is pending finalisation :) https://github.com/FormidableLabs/urql/pull/556

avetisk:
Didn’t have time to look deeply into that but I have to thank you a lot guys. The bundle size difference is just amazing 🙏

PR #261, opened 2019-06-05
Implement server-side rendering support

### LINKED ISSUE #218: Server side rendering

kitten:
Fix #218

Server-side rendering can now be achieved; an app using `urql` can do this with just minor modifications. Instead of using `getDataFromTree` like `react-apollo` does we're using our more generic [`react-ssr-prepass` library.](https://github.com/FormidableLabs/react-ssr-prepass)

This still needs to be documented but essentially on the server the client used like so:

```js
import ssrPrepass from 'react-ssr-prepass';

import {
  createClient,
  dedupExchange,
  cacheExchange,
  ssrExchange,
  fetchExchange
} from 'urql';

// We create an ssrExchange, which is an additional cache for client-side rehydration
const ssr = ssrExchange();

// The client can also be shared, but `suspense` needs to be set to
// false on the client-side
const client = createClient({
  url,
  exchanges: [dedupExchange, cacheExchange, ssr, fetchExchange],
  suspense: true
});

// The user has their app somewhere
const App = () => {/* ... */};
// We execute the SSR prepass
await ssrPrepass(<App />);

// Now we can extract the data:
const data = ssr.extractData();

// ...
// This is where renderToString or something goes
// We then need to send down the urql ssr data in the HTML as well
const html = `<script>window.DATA = (${JSON.stringify(data)});</script>`
```

And on the client we also use the SSR exchange to rehydrate the data:

```js
import {
  createClient,
  dedupExchange,
  cacheExchange,
  ssrExchange,
  fetchExchange
} from 'urql';

// We create an ssrExchange, which is an additional cache for client-side rehydration
const ssr = ssrExchange({ initialState: window.DATA });

// The above can also be used as:
// ssr.restoreData(window.DATA);

const client = createClient({
  url,
  exchanges: [dedupExchange, cacheExchange, ssr, fetchExchange],
  suspense: false // NOTE this is false on the client
});

// ...
// ReactDOM.rehydrate or otherwise
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,6 +104,8 @@
     "react": "^16.8.6",
     "react-dom": "^16.8.6",
     "react-hooks-testing-library": "^0.5.1",
+    "react-is": "^16.8.6",
+    "react-ssr-prepass": "^1.0.5",
     "react-test-renderer": "^16.8.6",
     "rimraf": "^2.6.2",
     "terser": "^4.0.0",
--- a/src/exchanges/ssr.test.ts
+++ b/src/exchanges/ssr.test.ts
@@ -35,14 +35,20 @@ it('caches query results correctly', () => {
 
   const data = ssr.extractData();
   expect(Object.keys(data)).toEqual(['' + queryOperation.key]);
-  expect(data).toEqual({ [queryOperation.key]: queryResponse });
+
+  expect(data).toEqual({
+    [queryOperation.key]: {
+      data: queryResponse.data,
+      error: undefined,
+    },
+  });
 });
 
 it('resolves cached query results correctly', () => {
   const onPush = jest.fn();
 
   const ssr = ssrExchange({
-    initialState: { [queryOperation.key]: queryResponse },
+    initialState: { [queryOperation.key]: queryResponse as any },
   });
 
   const [ops$, next] = input;
@@ -65,7 +71,7 @@ it('deletes cached results in non-suspense environments', () => {
   const onPush = jest.fn();
   const ssr = ssrExchange();
 
-  ssr.restoreData({ [queryOperation.key]: queryResponse });
+  ssr.restoreData({ [queryOperation.key]: queryResponse as any });
   expect(Object.keys(ssr.extractData()).length).toBe(1);
 
   const [ops$, next] = input;
--- a/src/exchanges/ssr.test.ts
+++ b/src/exchanges/ssr.test.ts
@@ -85,4 +85,7 @@ it('deletes cached results in non-suspense environments', () => {
 
   expect(Object.keys(ssr.extractData()).length).toBe(0);
   expect(onPush).toHaveBeenCalledWith(queryResponse);
+
+  // NOTE: The operation should not be duplicated
+  expect(output).not.toHaveBeenCalled();
 });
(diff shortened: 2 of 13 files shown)


kitten:
I'm currently investigating a remaining issue where during client-side rehydration a race condition leads to a request being sent unnecessarily due to a race condition (potentially?)

src/exchanges/ssr.ts
```diff
@@ -0,0 +1,127 @@
...
+/** Deserialize plain JSON to an OperationResult */
+const deserializeResult = (
+  operation: Operation,
+  result: SerializedResult
+): OperationResult => {
+  const { error, data } = result;
+  const deserialized: OperationResult = { operation, data, error: undefined };
+  if (error !== undefined) {
+    deserialized.error = new CombinedError({
+      networkError: new Error(error.networkError),
```
jevakallio: Will the user find it surprising that the networkError is an `Error` instance but the graphQLErrors are strings?

Apollo seems to wrap these into `GraphQLError`s

kitten: We do have the graphQLErrors array as well below in the a CombinedError and networkError is just for miscellaneous runtime errors.

I haven't implemented extensive error rehydration yet because I'm sure it won't come up as an issue too soon 😅 but it's definitely something that needs to be added in the future

package.json
```diff
@@ -104,6 +104,8 @@
     "react": "^16.8.6",
     "react-dom": "^16.8.6",
     "react-hooks-testing-library": "^0.5.1",
+    "react-is": "^16.8.6",
```
jevakallio: I'm not seeing any usages of `react-is` but I might be dumb?

kitten: It's a peer dependency for `react-ssr-prepass` ✨

src/exchanges/ssr.ts
jevakallio: Ah yeah, this checks out 👍

I didn't expect the CombinedError constructor to rehydrate the GraphQL error types. gg

kitten:
_Will potentially be retroactively reviewed again_

src/exchanges/ssr.test.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  const ssr = ssrExchange();
+  const [ops$, next] = input;
+  const exchange = ssr(exchangeInput)(ops$);
+
+  publish(exchange);
+  next(queryOperation);
+
+  const data = ssr.extractData();
+  expect(Object.keys(data)).toEqual(['' + queryOperation.key]);
```
parkerziegler: Is this `'' + queryOperation.key` purely for string coercion? I'd probably just use a template literal or `.toString` to keep intent clear.

src/exchanges/ssr.ts
```diff
@@ -0,0 +1,127 @@
...
+
+/** Serialize an OperationResult to plain JSON */
+const serializeResult = ({
+  data,
+  error,
+}: OperationResult): SerializedResult => {
+  const result: SerializedResult = { data, error: undefined };
+  if (error !== undefined) {
+    result.error = {
+      networkError: '' + error.networkError,
```
parkerziegler: Same question here as above, is this purely for string coercion?

src/exchanges/ssr.test.ts
kitten: It's just for string coercion and I prefer `'' + x` over the template string variant at least for testing tbh 😅

helloguille:
Works nicely. I'll keep testing

PR #264, opened 2019-06-05
Add setter for fetchOptions updates

### LINKED ISSUE #234: Feature request: allow to return promise in fetchOptions function
there are often cases when in order to obtain a JWT token, we have to call an async function. For example with AWS amplify: https://aws-amplify.github.io/docs/js/authentication#retrieve-current-session

it would be nice if urql would allow us to return fetch options from the `fetch`  asynchronously.

currently fetch options type is defined here: https://github.com/FormidableLabs/urql/blob/9a49818c4a54583d1f457772323dab58c78211df/src/client.ts#L32

and as you can see it only accepts a static object or a function which returns synchronously.

andyrichardson:
Fix #234

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -94,18 +94,6 @@ export class Client {
     };
   };
 
-  createRequestOperation = (
-    type: OperationType,
-    { key, query, variables }: GraphQLRequest,
-    opts?: Partial<OperationContext>
-  ): Operation => ({
-    key,
-    query,
-    variables,
-    operationName: type,
-    context: this.createOperationContext(opts),
-  });
-
   /** Counts up the active operation key and dispatches the operation */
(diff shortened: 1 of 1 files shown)


kitten:
@andyrichardson Do you think this is safe to merge now that @JoviDeCroock has written a comprehensive auth example/guide in #361?

andyrichardson:
@kitten no objections here, I'll un-draft the PR and get CI triggered

JoviDeCroock:
@andyrichardson updating the snapshots should suffice for the CI

kitten:
Hiya, I think I actually confused this PR 💕😅 Do we still need this PR given the auth exchange example in the guide?

kitten:
Closing due to inactivity

PR #269, opened 2019-06-06
Add server-side rendering example with next.js

kitten:


yakovlevyuri:
Wow, thanks a lot! Are you planning to add next.js typescript example as well?

kitten:
@yakovlevyuri Hiya, I wasn't planning on adding that currently as this is just a minimal example for server-side rendering. But it'd be nice to contribute these examples to the `next.js` repository instead :heart:

PR #271, opened 2019-06-06
(tests) - coverage

JoviDeCroock:
Hey,

Saw some untested files so I thought to chime in there and introduce the components testing.

Made one example, if you agree with it then I'll continue else let's get it right together! 
I had to upgrade the react dependencies to 16.9.0-alpha.0 for async act.

Good coverage doesn't exclude bugs but it can make it easier for bugs to be found.


src/components/Mutation.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+    cleanup();
+  });
+
+  it('Should execute the mutation', done => {
+    let execute = () => {},
+      props = {};
+    const Test = () => <p>Hi</p>;
+    const App = () => {
+      return (
+        // @ts-ignore
```
JoviDeCroock: I really can't find why this is needed to be fair, it's complaining about Element | null while I'm returning a React element

package.json
```diff
@@ -114,12 +114,12 @@
     "microbundle": "^0.11.0",
     "npm-run-all": "^4.1.5",
     "prettier": "^1.17.1",
-    "react": "^16.8.6",
-    "react-dom": "^16.8.6",
+    "react": "^16.9.0-alpha.0",
```
kitten: I'm not quite sure whether we really want to test against `16.9.0` quite yet. It potentially will cause us to miss bugs in non-batching mode (I assume `16.9` is batching by default?) and async act sounds like it'd also batch some effects in testing, which might not be desirable 😅 

Any idea of what kind of impact this might have on us?

JoviDeCroock: I know the author of react-testing-library was pretty fond of upgrading since without it your console will be spammed with act warning, I can revert but we will have some console.errors in console (won't impact test outcome though)

kitten: We do have a workaround somewhere in the repo where we filter out these errors 👍

JoviDeCroock: Aha, then I'll find that and use it. Generally the approach is good?

kitten: It does work. It's still a little noisy but tbh I trust it more than a future React version where things possibly become more async while people are targeting stable 😅

JoviDeCroock: Allright, I agree though the less risk the better didn't think about it that way yet

JoviDeCroock:
After making the test for Subscription it seems to be so that `fetching` remains true all the time? This is normal I assume?

kitten:
_Quick note:_ This might need a rebase, since we've switched CI over to SailCI and coverage to Codecov :+1:

JoviDeCroock:
Thanks for the notice @kitten I'll try to make some time this week to finish this up

JoviDeCroock:
I don't know why but sourcetree really acts up when I try to rebase :/

PR #274, opened 2019-06-07
Fix IE11 support and duplicate helpers by adding new Rollup config

kitten:
Fix #273 

- This uses TypeScript, then Bublé, then Babel to transpile everything (heavy stuff I know)
- Ensure rest spread helper is not duplicated
- Ensure that `Object.assign` is transpiled away to guarantee IE11 compatibility
- Integrate minification into Babel config

Before we had some duplicate helpers from Bublé in our bundle. So this is replacing Microbundle with a custom config.

We do ultimately have a replacement for `Object.assign` in our bundle, but that helps with IE11 compatibility, which is :100:

Ultimately it seems that the gzipped size actually remains the same! :tada:


kitten:
Merging this since this replaces Travis CI, which is now switched off :+1: Nothing critical has changed though

PR #276, opened 2019-06-07
Fix Submit function in examples/3-ssr-with-nextjs

ijones16:


andyrichardson:
Looks good - thanks @ijones16 :+1:

PR #278, opened 2019-06-08
Modify use of pause

andyrichardson:
The proposed changes here are aimed to simplify the use-cases associated with the `pause` argument.

Rather than pause preventing requests outright - it will _pause_ the functionality of auto fetching  - whether that be on mount or when the query has changed.

The reasons for this are:
 - It allows users to pause auto fetching while having the ability to manually fetch on demand (currently not an option - see [example problem here](https://www.reddit.com/r/reactjs/comments/bxu4hc/urql_v11_now_with_serverside_rendering_support/eqb3sj0?utm_source=share&utm_medium=web2x))
 - It improves the seperation of concerns. Config options should influence how urql acts, manual operations can be prevented with a conditional inside component logic. 

There's also a small typing fix for _useImmediateEffect_.

<details>
<summary>Example usage</summary>

```tsx
const MySearch = () => {
  const [variables, setVariables] = useState({ searchString: "" });
  const [searchResults, fetchSearchResults] = useQuery({ query, variables, skip: true });

  const throttledFetch = throttle(fetchSearchResults, 200); 

  const handleChange = (e) => {
    setVariables({ searchString: e.target.value });
    throttledFetch();
  };

  return (
    <input value={vars.searchString} onChange={handleChange} />
    <Results values={searchResults} />
  )
}
```

</details>

### Additional notes
The advantages of this approach are pretty clear but it does highlight a bug that we haven't yet addressed - updating variables/query values and immediately executing them in the same render doesnt act as expected.

I've created an [issue about this](https://github.com/FormidableLabs/urql/issues/279) and I believe the best way to go about this is to allow arguments to be overridden when manually fetching.

### DIFF
--- a/src/hooks/useImmediateEffect.ts
+++ b/src/hooks/useImmediateEffect.ts
@@ -1,7 +1,6 @@
 /* eslint-disable react-hooks/exhaustive-deps */
 
-import { useRef, useEffect } from 'react';
-import { noop } from '../utils';
+import { useRef, useEffect, EffectCallback } from 'react';
 
 enum LifecycleState {
   WillMount = 0,
@@ -9,14 +8,12 @@ enum LifecycleState {
   Update = 2,
 }
 
-type Effect = () => () => void;
-
 /** This is a drop-in replacement for useEffect that will execute the first effect that happens during initial mount synchronously */
 export const useImmediateEffect = (
-  effect: Effect,
+  effect: EffectCallback,
   changes: ReadonlyArray<any>
 ) => {
-  const teardown = useRef(noop);
+  const teardown = useRef<ReturnType<EffectCallback>>(undefined);
   const state = useRef(LifecycleState.WillMount);
 
   // On initial render we just execute the effect
--- a/src/hooks/useQuery.ts
+++ b/src/hooks/useQuery.ts
@@ -48,15 +48,7 @@ export const useQuery = <T = any, V = object>(
     (opts?: Partial<OperationContext>) => {
       unsubscribe.current();
 
-      // If useQuery is currently paused, set fetching to
-      // false and abort; otherwise start the query
-      if (args.pause) {
-        setState(s => ({ ...s, fetching: false }));
-        unsubscribe.current = noop;
-        return;
-      } else {
-        setState(s => ({ ...s, fetching: true }));
-      }
+      setState(s => ({ ...s, fetching: true }));
 
       [unsubscribe.current] = pipe(
         client.executeQuery(request, {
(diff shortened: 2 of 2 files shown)


src/hooks/useQuery.ts
```diff
@@ -68,16 +60,18 @@ export const useQuery = <T = any, V = object>(
         })
       );
     },
-    [args.pause, args.requestPolicy, client, request, setState]
+    [args.requestPolicy, client, request, setState]
   );
 
-  // This calls executeQuery immediately during the initial mount and
```
kitten: Seems good but it'd be great to readd the comments and slightly rewrite them. Especially the one here that explains with this is using useImmediateEffect 😊

andyrichardson: Yup agreed - I'll add something back in 👍

src/hooks/useImmediateEffect.ts
```diff
@@ -1,22 +1,19 @@
...
-type Effect = () => () => void;
-
 /** This is a drop-in replacement for useEffect that will execute the first effect that happens during initial mount synchronously */
 export const useImmediateEffect = (
-  effect: Effect,
+  effect: EffectCallback,
   changes: ReadonlyArray<any>
 ) => {
-  const teardown = useRef(noop);
+  const teardown = useRef<ReturnType<EffectCallback>>(undefined);
```
parkerziegler: Are we initializing this to `undefined` with the expectation that `teardown.current` will never get called until it gets reassigned to the return type of `effect`? Seems like a safe assumption based on how this effect runs (`teardown` synchronously getting re-assigned in the first `if` below) but just wanted to get the reasoning on this change.

kitten: Yea, should be safe since teardown functions can be undefined in the worst case 👍

src/hooks/useQuery.ts
```diff
@@ -68,16 +60,18 @@ export const useQuery = <T = any, V = object>(
...
-  // otherwise behaves like a normal useEffect; We call executeQuery
-  // everytime it, i.e. its input like request, changes
   useImmediateEffect(() => {
+    if (args.pause) {
```
parkerziegler: Nice I like this. So the flow here is:

- We check `args.pause` on initial mount to handle the case where a user just wants the `executeQuery` fn.
- Also check it on all subsequent changes and prevent queries when things update.

❤️Love it. My one question is should L69, if it's being treated as the cleanup function, be written as `() =>  setState(s => ({ ...s, fetching: false }));`? Seems we're returning `void` currently rather than `() => void`.

PR #281, opened 2019-06-08
Fix dedup ignoring torn down query operations

### LINKED ISSUE #280: Previously canceled queries are not refetched, but fetching is true
You can try it here https://codesandbox.io/s/tender-hodgkin-5xiqh by changing the input before the query finishes (for example 1 -> 2 -> 3 -> 2)

Version 1.1.0

kitten:
Fix #280 

When a query was torn down, the dedup exchange wasn't
removing its key from the inFlight keys.

This was causing us to ignore torn down queries forever.

### DIFF
--- a/src/exchanges/dedup.ts
+++ b/src/exchanges/dedup.ts
@@ -1,32 +1,36 @@
 import { filter, pipe, tap } from 'wonka';
-import { Exchange } from '../types';
+import { Exchange, Operation, OperationResult } from '../types';
 
 /** A default exchange for debouncing GraphQL requests. */
 export const dedupExchange: Exchange = ({ forward }) => {
-  const inFlight = new Set<number>();
+  const inFlightKeys = new Set<number>();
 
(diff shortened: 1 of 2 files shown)


src/exchanges/dedup.ts
```diff
@@ -1,32 +1,36 @@
...
+    const isInFlight = inFlightKeys.has(key);
+    inFlightKeys.add(key);
+    return !isInFlight;
+  };
 
-            if (!hasInFlightOp) {
-              inFlight.add(key);
-            }
+  const afterOperationResult = ({ operation }: OperationResult) => {
+    inFlightKeys.delete(operation.key);
```
andyrichardson: Is there a reason for not having this inline?

kitten: Just to clean it up; I thought having names like `afterOperationResult` could help when someone's trying to figure out how those work

PR #288, opened 2019-06-11
Add fetch exchange response preprocessor

daniel-edwards-iress:
This is a simple factory to create an exchange which allows user code to modify the response before allowing the fetch exchange to process this.

An example usage of this is invisibly handling authentication.
e.g. Send a request to the api server. It returns a 401 unauthorised error. The preprocessor handles this by doing the necessary steps to authenticate and calls fetch again and passes the new response to the rest of the exchange's promise chain

### DIFF
--- a/src/exchanges/fetch.ts
+++ b/src/exchanges/fetch.ts
@@ -4,133 +4,148 @@ import { filter, make, merge, mergeMap, pipe, share, takeUntil } from 'wonka';
 import { Exchange, Operation, OperationResult } from '../types';
 import { CombinedError } from '../utils/error';
 
-/** A default exchange for fetching GraphQL requests. */
-export const fetchExchange: Exchange = ({ forward }) => {
-  const isOperationFetchable = (operation: Operation) => {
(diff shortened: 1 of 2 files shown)


kitten:
Hiya, first of all cheers for the contribution 👌

I'm sorry, but I think it'll be unlikely that we'll merge this. This functionality in itself can be implemented already with a custom exchange.

Now we unfortunately don't have enough docs yet on writing such an exchange but a "post processor" or even "pre processor" would just be an exchange in front of the fetch exchange 🙂

I'll leave this open for further discussion 👍 thanks again for opening a PR. Sorry that the first reaction is not a complete positive 😅

daniel-edwards-iress:
given the default exchange list `[dedupExchange, cacheExchange, fetchExchange]` if i was to insert a custom exchange into the list e.g. `[dedupExchange, cacheExchange, myCustomExchange, fetchExchange]`.
Is it possible for myCustomExchange to redo fetchExchange for the same operation?

PR #304, opened 2019-06-16
(chore) - move wrongly placed import

JoviDeCroock:
In my PR the testing dep snuck into dependencies instead of devDependencies, apologies!


JoviDeCroock:
sailci not running for forks or?

PR #305, opened 2019-06-16
Add component source info to hooks

### LINKED ISSUE #291: Add metadata capabilities to urql
In order to expose enough information to create valuable devtools I'm proposing we add a _meta_ tag to our operations.

This tag can be set/updated at the following points:
 - on calling hooks / client execution
 - inside hooks (for sourcing the calling component)
 - inside any exchange

We could use context but I think keeping it separate (such as a property within context as @kitten has suggested) would be best as any information added to the meta property of an operation should not be mission critical (i.e. no worry of breaking anything at runtime).

**tl;dr:** add meta property to operations

Edit: clear up wording on using context

andyrichardson:
Hooks now add metadata information for devtools including the source component funciton name.

> Note: Because we're using the stack trace to determine component names, the sourced component name is considered to be it's function name - **not** the function's displayName property.

Fix #291


src/hooks/useMutation.ts
```diff
@@ -20,6 +20,7 @@ export type UseMutationResponse<T, V> = [
 export const useMutation = <T = any, V = object>(
   query: DocumentNode | string
 ): UseMutationResponse<T, V> => {
+  const devtoolsContext = useRef({ source: getHookParent() });
```
kitten: Can we maybe make this a hook that just returns a ref; or just returns the value itself if that's possible and stable?

src/utils/debug.ts
```diff
@@ -0,0 +1,14 @@
+/** Find the name of the calling component. */
+export const getHookParent = () => {
```
kitten: Nice! 👏 naming of the file might not be optimal but that's alright. Can we make sure that NODE_ENV is used here so it becomes a noop in prod?

src/utils/index.ts
```diff
@@ -1,8 +1,9 @@
-export { CombinedError } from './error';
-export { getKeyForRequest } from './keyForQuery';
-export { createRequest } from './request';
-export { formatDocument, collectTypesFromResponse } from './typenames';
-export { toSuspenseSource } from './toSuspenseSource';
+export * from './debug';
```
kitten: It'd be nice to revert this tbh 😅

It's less convenient but this way we just have a list here of what's supposed to be exposed. Fine to merge though

src/utils/debug.ts
andyrichardson: So I was using the NODE_ENV checks when using the react stack trace but I don't see any advantage to hiding these details in prod as they'll just be ignored.

kitten: It's mainly so that this compiles away in a user's production bundle since it'd otherwise be dead code 👍

src/utils/index.ts
andyrichardson: I wasn't expecting the utils index file to look like it did as the explicit re-exporting feels pretty confusing. If we're adopting the standard of importing utilities from the 'utils' folder, what advantage do we get by re-stating the exported values? Feels like more work! 😕

I've added explicit exports at the root level index file so we're not exposing internal details to end users 🔊

src/utils/debug.ts
andyrichardson: Fair enough - I'll pull this into a hook also then :+1:

src/utils/index.ts
andyrichardson: The main reason for the change here was because we're filtering internal usage of exports in our utils folder but then not filtering when exporting to the end user - I've flipped that the other way round which makes more sense imho

andyrichardson:
I'm not having much luck with getting jest to not crash but I suspect it's relating to [this issue](https://github.com/facebook/jest/issues/8493). I'm yet to find a way around this problem but will try to find a fix asap.

JoviDeCroock:
Would it be an idea to make use of `useDebugValue` in the hooks since this helps for React devtools

andyrichardson:
@kitten thanks for some of the fixes yesterday. 

I've had to add the ref back in as the stack trace is modified if we call `getHookParent` inside useMemo. Also, we can't guarantee the stack order as the function is called by the react internals.

Looks ready for approval / merge!

andyrichardson:
> Would it be an idea to make use of `useDebugValue` in the hooks since this helps for React devtools

@JoviDeCroock  - this isn't for React's devtools but maybe better integrating into that is something we should look into :+1:

src/hooks/useDevtoolsContext.ts
```diff
@@ -0,0 +1,28 @@
...
+  const stackLine = (new Error().stack as string).split('\n')[4];
+  const parsedLine = /.*at (exports\.)?(\w+)/.exec(stackLine);
+
+  return parsedLine === null || typeof parsedLine[2] !== 'string'
+    ? 'Unknown'
+    : parsedLine[2];
+};
+
+const useDevtoolsContextHook = () => {
+  const source = useRef(getHookParent());
```
kitten: As discussed yesterday, the ref here isn't necessary is it 😅

PR #306, opened 2019-06-16
Move all build tasks to Rollup

kitten:
- Declarations are now generated as part of the Rollup build
- The Rollup build overrides the tsconfig correctly now and no changes to the output were made
- The clean task now runs only before publishing as that's the only time we care about it
- Prune was removed; instead we just exclude test files inside Rollup

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "test": "jest",
     "coverage": "jest --coverage",
     "lint": "eslint . --ext .ts,.tsx",
-    "prepublishOnly": "run-s clean build",
+    "prepublishOnly": "run-s clean test build",
     "codecov": "codecov"
   },
   "author": "Formidable",
--- a/package.json
+++ b/package.json
@@ -21,17 +21,13 @@
     "exchanges"
   ],
   "scripts": {
-    "prebuild": "rimraf dist",
-    "build": "run-p build:types build:bundle",
-    "build:clean": "rimraf dist",
-    "build:types": "tsc -d --emitDeclarationOnly --outDir dist/types",
-    "build:bundle": "rollup -c rollup.config.js",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -21,17 +21,13 @@
...
-    "build:types": "tsc -d --emitDeclarationOnly --outDir dist/types",
-    "build:bundle": "rollup -c rollup.config.js",
-    "build:prune": "rimraf dist/types/**/*.test.d.ts dist/types/test-utils",
+    "clean": "rimraf ./dist ./node_modules/.cache",
+    "build": "rollup -c rollup.config.js",
+    "watch": "rollup -w -c rollup.config.js",
     "test": "jest",
     "coverage": "jest --coverage",
     "lint": "eslint . --ext .ts,.tsx",
-    "check-formatting": "prettier --write src/**/*.{ts,tsx}",
```
andyrichardson: Removed?

rollup.config.js
```diff
@@ -84,14 +84,21 @@ const plugins = [
...
     tsconfigDefaults: {
       compilerOptions: {
-        sourceMap: true,
-        declaration: true
+        sourceMap: true
       },
     },
     tsconfigOverride: {
-      compilerOptions: {
+     exclude: [
```
andyrichardson: Not crazy on having typescript settings in multiple places. How about a tsconfig.json and tsconfig.build.json? It supports extending configs!

kitten: These are literally just overrides. I'm not a big fan of adding multiple tsconfigs as that just adds another set of configs that has to be read when looking at the rollup config.

In this case this is just forcing declarations and excluding tests, which won't make much sense outside of rollup 🥳

package.json
kitten: Yes, we should always use Prettier via ESLint so this could format some stuff differently / incorrectly

PR #296, opened 2019-06-12
(fix) - fast render should update state

JoviDeCroock:
Fixes: https://github.com/FormidableLabs/urql/issues/287
Reproduced at: https://codesandbox.io/s/nostalgic-sound-psbzf?fontsize=14

I think this is the most consistent fix for this problem.

PR #310, opened 2019-06-18
Fix broken SSR on master

andyrichardson:
It looks like the SSR reproduction on the current master is broken (looks to be related to #296).

```
UnhandledPromiseRejectionWarning: Error: [react-ssr-prepass] Hooks can only be called inside the body of a function component. (https://fb.me/react-invalid-hook-call)
```

To get around the late call of useEffect (causing problems in both SSR and browser environments) we now use useLayoutEffect.

This also addresses the issue I've been seeing locally where super fast network responses cause the state not to be updated.


JoviDeCroock:
If you want I can test it against the reproduction case provided in: https://codesandbox.io/s/nostalgic-sound-psbzf?fontsize=14

andyrichardson:
@JoviDeCroock that's awesome! I'll give it a shot

PR #314, opened 2019-06-18
(docs) - propose way for third-party exchanges

JoviDeCroock:
This is a proposition as to how we could acknowledge exchanges made by others.

This is meant for people who want to add their own package, we could expand this with a guide on how to write your own exchanges etc, I think people who write exchanges that help others should be able to advertise it somehow on the main repository. Correct me if my thought process is wrong somewhere.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -35,10 +35,8 @@ the behavior of your urql-client.
 
 ### [Extending & Experimenting](extending-and-experimenting.md)
 
-Here's everything you need to know to extend,
-customise and experiment with `urql`. This section
-shows you how to use the `Client` to write new
-APIs (Be it for React or not) and how to write
-new "Exchanges" to customise `urql`'s core behaviour!
+Here's everything you need to know to customize and augment
+every aspect of the GraphQL client from how data is cached
+to how components receive their data.
 
 ### [API](api.md)
--- a/docs/README.md
+++ b/docs/README.md
@@ -30,13 +30,16 @@ the `cacheExchange`, and how to use _Subscriptions_.
 
 ### [Exchanges](exchanges.md)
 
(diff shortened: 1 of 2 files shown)


docs/README.md
```diff
@@ -28,6 +28,11 @@ Everything about the basic & default `urql` behaviour
 that comes with explanations of the `fetchExchange`,
 the `cacheExchange`, and how to use _Subscriptions_.
 
+### [Exchanges](exchanges.md)
+
+Exchanges written by the community, these enhance
+the behavior of your urql-client.
```
kitten: We should probably stuck to writing urql here as just urql in an inline code annotation and without client?

And maybe we can reuse this quote from the announcement blog post:

> customize and augment every aspect of the GraphQL client from how data is cached to how components receive their data

Also I think this list won't be just consistent of community exchanges but also "official" ones

docs/exchanges.md
```diff
@@ -0,0 +1,22 @@
+# Exchanges
+
+## Persisted Queries
```
kitten: Can we maybe structure this as an "awesome" list? Short link and description on a single line?

I know we don't have much content yet, but it'll probably scale better 👍

docs/exchanges.md
```diff
@@ -0,0 +1,16 @@
...
+
+These exchanges can be imported from the `urql` package.
+
+- cacheExchange - caches all incoming data
+- debugExchange - gives information about outgoing graphql requests in log format
+- dedupExchange - deduplicates outgoing queries
+- fetchExchange - used to manage the outgoing graphql requests
+- subscriptionExchange - used to support subscriptions
+
+## Thirth-party
```
kitten: Small typo here: ~thirth~third-party exchanges

PR #324, opened 2019-06-27
Flag responses with cache hit status

andyrichardson:
When using the default cache exchange, responses will have an additional property `cacheHit` added to their context.

This property will not be added to the global context type definition as it is specific to this cache. Future cache implementations may use this `cacheHit` property in a different way (i.e. cache hits could be flagged as partial).

### DIFF
--- a/src/exchanges/cache.ts
+++ b/src/exchanges/cache.ts
@@ -72,9 +72,9 @@ export const cacheExchange: Exchange = ({ forward, client }) => {
         }
 
         return {
-          operation,
           data: undefined,
           error: undefined,
+          operation: addMetadata(operation, { cacheOutcome: 'miss' }),
         };
       })
     );
@@ -158,10 +158,7 @@ const afterQuery = (
     return;
   }
 
-  resultCache.set(operation.key, {
-    ...response,
-    operation,
-  });
+  resultCache.set(operation.key, response);
 
   collectTypesFromResponse(response.data).forEach(typeName => {
     const operations =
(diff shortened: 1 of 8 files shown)


src/exchanges/cache.ts
```diff
@@ -145,19 +146,23 @@ const afterQuery = (
...
   if (data === undefined) {
     return;
   }
 
-  resultCache.set(key, response);
+  resultCache.set(operation.key, {
+    ...response,
+    operation: {
+      ...operation,
+      context: { ...operation.context, cacheHit: true },
```
kitten: So for the normalised cache can we introduce more of an enum, which will cover all cases:

```ts
type CacheOutcome = 'MISS' | 'PARTIAL' | 'HIT';
```

src/exchanges/cache.ts
```diff
@@ -145,19 +146,23 @@ const afterQuery = (
...
-    data,
-  } = response;
+  const { operation, data } = response;
+
   if (data === undefined) {
     return;
   }
 
-  resultCache.set(key, response);
+  resultCache.set(operation.key, {
```
kitten: Let's modify the operation only when it goes back out, so that we don't get this confused later on: https://github.com/FormidableLabs/urql/blob/113b732bd1a3dd32e5b1a9df542a7a641e82acde/src/exchanges/cache.ts#L65

Which would also allow us to mark this line as a cache miss: https://github.com/FormidableLabs/urql/blob/113b732bd1a3dd32e5b1a9df542a7a641e82acde/src/exchanges/cache.ts#L69

PR #325, opened 2019-06-27
Add latency info to context

andyrichardson:
Adds a _latency_ property to a returning operation's context. This is the duration in milliseconds that the request took to complete.

### DIFF
--- a/src/exchanges/fetch.ts
+++ b/src/exchanges/fetch.ts
@@ -2,7 +2,7 @@
 import { print } from 'graphql';
 import { filter, make, merge, mergeMap, pipe, share, takeUntil } from 'wonka';
 import { Exchange, Operation, OperationResult } from '../types';
-import { CombinedError } from '../utils/error';
+import { addMetadata, CombinedError } from '../utils';
 
 /** A default exchange for fetching GraphQL requests. */
 export const fetchExchange: Exchange = ({ forward }) => {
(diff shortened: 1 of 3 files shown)


src/exchanges/fetch.ts
```diff
@@ -75,9 +75,19 @@ const createFetchSource = (operation: Operation) => {
...
     executeFetch(operation, fetchOptions).then(result => {
       if (result !== undefined) {
-        next(result);
+        next({
+          ...result,
+          operation: {
+            ...result.operation,
+            context: {
+              ...result.operation.context,
+              latency: Date.now() - startTime,
```
kitten: This is something that I'd rather not have in `production` all the time, but I suppose it's ok to use this anyway :+1:

can we make sure that there's a default value of `0` in the `client`. Might not be nice, but we don't want people to actually  use this for something else

JoviDeCroock: Shouldn't this be renamed to __latency to avoid naming conflicts. an application at work uses `latency` for an IOT application

src/exchanges/fetch.ts
```diff
@@ -75,9 +75,19 @@ const createFetchSource = (operation: Operation) => {
         abortController !== undefined ? abortController.signal : undefined,
     };
 
+    const startTime = Date.now();
```
JoviDeCroock: Any reason as to why you're using Date.now over performance.now?

andyrichardson: No particular reason. Is there an advantage to using `performance.now` other than sub-millisecond precision?

JoviDeCroock: It’s relative to page load in comparison to unix epoch. I think this is a minimal difference, I did read that it’s not available in opera mini and IE9 so I don’t know if that’s a blocker. It’s up to you, what you prefer was just asking if there was a reason. I think either is good for the job but perf.now will probably be a micro optimisation since it revolves around a smaller number.

PR #328, opened 2019-07-01
Update type signature for SubscriptionHandler. Update associated docs.

### LINKED ISSUE #319: SubscriptionHandler type issue
The `SubscriptionHandler` used for generating a composite response from subscription events is currently typed as
```typescript
export type SubscriptionHandler<T, R> = (prev: R | void, data: T) => R;
```
but the `void` type should not be used in argument types unless it is to disallow them from being used. For instance, the trivial handler
```typescript
(updates: number = 0, response) => {
      return 42
    }
```
errors out with
```
error TS2345: Argument of type '(updates: number | undefined, response: ServerImageEventsSubscription) => number' is not assignable to parameter of type 'SubscriptionHandler<ServerImageEventsSubscription, ServerImageEventsSubscription>'.
  Types of parameters 'updates' and 'prev' are incompatible.
    Type 'void | ServerImageEventsSubscription' is not assignable to type 'number | undefined'.
      Type 'void' is not assignable to type 'number | undefined'.

151     (updates: number = 0, response) => {
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

I think that the intention with this handler is to pass in `undefined` on the first call, when there is no previous state, and the logic of `useSubscription` indeed does this. Therefore, I think that this type should be
```typescript
export type SubscriptionHandler<T, R> = (prev: R | undefined, data: T) => R;
```

EDIT:
It seems that there is another subtle issue here: because `useSubscription` sets `R=T` by default in its type parameters, the `SubscriptionHandler` expects the "state" to be of the same type as the subscription data, unless explicitly overriden. However, using
```typescript
useSubscription<
    ServerImageEventsSubscription,
    number
  >(
    {
      query: ServerImageEventsDocument,
    },
    (updates: number = 0, response) => {
      return 4
    }
  )
```

still gives a similar (but more sane) error
```
 error TS2345: Argument of type '(updates: number | undefined, response: ServerImageEventsSubscription) => number' is not assignable to parameter of type 'SubscriptionHandler<ServerImageEventsSubscription, number>'.
  Types of parameters 'updates' and 'prev' are incompatible.
    Type 'number | void' is not assignable to type 'number | undefined'.
      Type 'void' is not assignable to type 'number | undefined'.

152     (updates: number = 0, response) => {
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

parkerziegler:
This PR would fix #319. As @mjadczak points out, the `void` type is a mistake there - `handler` is called in `useSubscription` with `(s.data, data)`, where `s.data` is set initially by `useImmediateState` to `undefined`: https://github.com/FormidableLabs/urql/blob/master/src/hooks/useSubscription.ts#L36. It also fixes the signature of the mutable ref inside of `useRequest`, which is initialized to `undefined` .

I'm also not sure if we were attempting to take advantage of `undefined` and `null` being sub-types of `void`, so feel free to push back on this change (but it seems to make sense to me).

### DIFF
--- a/examples/2-using-subscriptions/README.md
+++ b/examples/2-using-subscriptions/README.md
@@ -1,21 +1,35 @@
 ## About
 
-This is a basic example of subscriptions being used with URQL.
+This is a basic example of subscriptions being used with `urql`. It demonstrates some of the core principles around wiring up subscriptions, including:
+
+- Setting up the `subscriptionExchange` with a `subscriptionClient`.
+- Using the `useSubscription` hook.
+- Using a `handler` to accumulate the results of subscriptions.
 
 ## Usage
 
-#### 1. Install dependencies
+#### 1. Install dependencies in repo root.
+
+To get started with the example, make sure to install `urql`'s dependencies in the root directory of your clone. We `link` to the `urql` source from the `examples` directory, so its dependencies need to be installed.
 
 ```bash
+# In root directory
 yarn
 ```
 
-#### 2. Start server
+#### 2. Navigate to this directory and install dependencies.
+
+```bash
+cd examples/2-using-subscriptions
+yarn
+```
+
+#### 3. Start the example app and server (this is done in a single command).
 
 ```bash
 yarn start
 ```
 
-#### 3. Open browser
+#### 4. Open your browser
 
 Navigate to example at [http://localhost:3000/](http://localhost:3000/).
--- a/examples/2-using-subscriptions/yarn.lock
+++ b/examples/2-using-subscriptions/yarn.lock
@@ -602,11 +602,6 @@ array-unique@^0.3.2:
   resolved "https://registry.yarnpkg.com/array-unique/-/array-unique-0.3.2.tgz#a894b75d4bc4f6cd679ef3244a9fd8f46ae2d428"
   integrity sha1-qJS3XUvE9s1nnvMkSp/Y9Gri1Cg=
 
-asap@~2.0.3:
-  version "2.0.6"
-  resolved "https://registry.yarnpkg.com/asap/-/asap-2.0.6.tgz#e50347611d7e690943208bbdafebcbc2fb866d46"
-  integrity sha1-5QNHYR1+aQlDIIu9r+vLwvuGbUY=
-
 asn1.js@^4.0.0:
   version "4.10.1"
   resolved "https://registry.yarnpkg.com/asn1.js/-/asn1.js-4.10.1.tgz#b9c2bf5805f1e64aadeed6df3a2bfafb5a73f5a0"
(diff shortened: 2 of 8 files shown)


andyrichardson:
Makes sense 👍

Void isn't actually a type so I think it should be avoided where possible. The only place I've found it valuable is when stating "I don't care about this value" which is almost always exclusive to function arguments and their return types.

```tsx
interface Props {
  onClick: (e: MouseEvent) => void;
}
```

examples/2-using-subscriptions/README.md
```diff
@@ -1,21 +1,35 @@
 ## About
 
-This is a basic example of subscriptions being used with URQL.
+This is a basic example of subscriptions being used with `urql`. It demonstrates some of the core principles around wiring up subscriptions, including:
```
andyrichardson: We need to standardize the casing - I vote "Urql". Capitalized (proper noun) and not as spammy/agressive as URQL.

Not something for this pr, but worth us thinking about.

examples/2-using-subscriptions/yarn.lock
```diff
@@ -4869,12 +4816,10 @@ wide-align@^1.1.0:
   dependencies:
     string-width "^1.0.2 || 2"
 
-wonka@^2.0.1:
-  version "2.0.1"
-  resolved "https://registry.yarnpkg.com/wonka/-/wonka-2.0.1.tgz#3769f9b050683edf2280e04d2b01afbdfe5fd9d4"
-  integrity sha512-s0nGPVbbA9a0FhUarNJino4i1J1UyQ5rGA1MmA65U/+xUXmBQEw4cCKceRXkDhgL8ZJ2HVhvxthHFZZTf02gBA==
-  dependencies:
-    bs-rebel "^0.2.3"
+wonka@^3.0.0:
```
andyrichardson: Looks like our lock file was out of sync with our package.json

examples/2-using-subscriptions/README.md
kitten: In all of the markdown of our docs we've now gone with "urql" (also when possible stylised as monospaced) which was discussed on Slack.

We can reopen that discussion, but frankly I think optically "urql" is fine and works well, "Urql" is OK but would now be an inconsistency given the docs, and "URQL" would be highly inconsistent and also doesn't read rather nicely (definitely shouty)

andyrichardson: My issue with all lowercase is that there's not much emphasis on the name which leaves it feeling a little deflated. We can highlight the name in markdown which is awesome but in other places such as a tweet, the name urql just gets lost amongst the words.

Sounds like all lowercase is the standard for now so let's go with that. Maybe there are other examples of all lowercase being used for branding? If you find any forward them to me!

examples/2-using-subscriptions/yarn.lock
parkerziegler: This was just for the example project, so not too concerned about it for the time being. I verified the example starts up as expected 👍

parkerziegler:
@JoviDeCroock yep, the change for `useRequest` was included as well, it's at the very bottom of the diff 👍

JoviDeCroock:
> @JoviDeCroock yep, the change for `useRequest` was included as well, it's at the very bottom of the diff 👍

I looked over it, my bad

parkerziegler:
Looks like Sail CI may be down?

kitten:
@parkerziegler seems to be working now. Just a failing snapshot; not sure what happened there but the stack might be inaccurate? I've got an alternative implementation for that since we're launching devtools soon and want that to be super nice and accurate :+1:

parkerziegler:
@kitten do you want me to merge this and we'll assume that #329 will fix the build once it gets merged? Or wait until #329 is merged, rebase off of it, and then merge?

parkerziegler:
cc/ @kitten @andyrichardson @JoviDeCroock merging this with the snapshots updated to pass the build. When #329 gets merged we can update those snapshots in tests.

PR #329, opened 2019-07-02
Replace useDevtoolsContext Error.stack with Fiber approach

kitten:
(I added some tests for `useDevtoolsContext` first, before I started, just to check that it's still doing the right thing)

This replaces the `Error` stack to determine the current component name with a React Fiber approach, by looking at `ReactCurrentOwner`. This is generally safe and the data structure hasn't changed much since `react-reconciler` was implemented. There's also some guards in the implementation though just in case. But this is just for development after all.

The motivation behind this was that the stack approach can lead to incorrect names at times depending on the browser, minification in the app bundle, testing approach, and other factors, where the stack is unreliable.

So this looks at `ReactCurrentOwner`, which is a ref, checks whether we have a fiber (which we should) and whether it's a component (which it must be since we're checking from a hook). Then it gets the component and will use `displayName` (hooray, display names!) or `name` to determine the current component's name.

If the name is found to be an internal one (`Query`, `Mutation`, `Subscription`) then it'll walk up by one fiber and use the parent instead to get a more useful name. This is currently done using a heuristic as it should be sufficient for a development mode implementation.

I also added some support for class components, since walking one component up via the error stack is pretty tough (if not impossible?)

@andyrichardson: Hope you like it :tada: tl;dr I think it's potentially more stable and allows us to get the `displayName` and find the parent component for our own components :+1:

### DIFF
--- a/src/hooks/useDevtoolsContext.ts
+++ b/src/hooks/useDevtoolsContext.ts
@@ -43,7 +43,7 @@ const useDevtoolsContextImpl = (): Partial<OperationContext> => {
 };
 
 /** Creates additional context values for serving metadata to devtools. */
-export const useDevtoolsContext: () => Partial<OperationContext> | void =
+export const useDevtoolsContext: () => Partial<OperationContext> | undefined =
   // NOTE: We check for CurrentOwner in case it'll be unexpectedly changed in React's source
   process.env.NODE_ENV !== 'production' && !!CurrentOwner
     ? useDevtoolsContextImpl
--- a/src/hooks/useDevtoolsContext.ts
+++ b/src/hooks/useDevtoolsContext.ts
@@ -1,28 +1,47 @@
-import { useMemo, useRef } from 'react';
+import * as React from 'react';
+import { OperationContext } from '../types';
 
-/** Find the name of the calling component. */
-const getHookParent = () => {
-  /**
-   * Line values could be:
-   * - named export             " at exports.Home (Home.tsx?9b1b:17)"
-   * - default export / inline  " at Home (Home.tsx?9b1b:17)"
-   */
-  const stackLine = (new Error().stack as string).split('\n')[3];
-  const parsedLine = /.*at (exports\.)?(\w+)/.exec(stackLine);
+const {
+  ReactCurrentOwner: CurrentOwner,
+} = (React as any).__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED;
 
-  return parsedLine === null || typeof parsedLine[2] !== 'string'
-    ? 'Unknown'
-    : parsedLine[2];
-};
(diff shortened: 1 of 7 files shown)


src/hooks/useDevtoolsContext.ts
```diff
@@ -1,28 +1,47 @@
...
 
-  return useMemo(() => [{ meta: { source: source.current } }], []);
+    return { meta: { source } };
+  }, []);
 };
 
 /** Creates additional context values for serving metadata to devtools. */
-export const useDevtoolsContext =
-  process.env.NODE_ENV === 'production'
-    ? () => [undefined]
```
kitten: I've changed the output not to be an array, since it doesn't need to be one :) this is an internal hook so we can always change it without breaking anyone's apps luckily :+1:

parkerziegler:
Wowza, this is pretty dang 🆒! I like the implementation a lot @kitten 🎉

src/hooks/useDevtoolsContext.ts
andyrichardson: A bit of a deviation from the standard`[value, mutation/s]` which we've also been following throughout the project so far (useQuery, useMutation, etc).

Not sure this change is necessary/beneficial.

andyrichardson:
@kitten what's the inventive behind the `void` additions?

kitten:
@andyrichardson it’s definitely still a little dodgy but stable. I already had this in the back of my mind for the chance that the error stacktraces become unreliable. It is the internal Fiber architecture, but since this is one of the most important React internals it won’t just disappear, and if it does there are some guards.

The void additions are because TS didn’t expect undefined arguments anymore for the options or operation spreads once the destructure was gone for some reason.

Regarding removing the array destructure itself, it’s an internal hook and there’s no convention around making hooks return arrays/tuples, that’s just a convention for our public hooks.

Since the typings document this return value internally and externally this hooks isn’t in use, I think it’s unnecessary to keep it :)

andyrichardson:
> The void additions are because TS didn’t expect undefined arguments anymore for the options or operation spreads once the destructure was gone for some reason.

I'm not sure adding void is a good way to solve this problem given we're only expecting our argument properties to be of type `T | undefined`. Can you drop the error message you're getting here?

parkerziegler:
Would love to see this formally get merged if possible. @kitten @andyrichardson what can I help with to move this along? It seems like we want to investigate the usage of `void` and why TS is complaining. Anything else?

andyrichardson:
> Would love to see this formally get merged if possible. @kitten @andyrichardson what can I help with to move this along? It seems like we want to investigate the usage of `void` and why TS is complaining. Anything else?

@kitten @parkerziegler I'm not aware of any reasons why this can't get merged. I'll give it a test later today on the latest devtools and then merge if all looks swell 🚀

Issue #14, opened 2018-01-29
Support file uploads


jaydenseric:
If there is interest I can raise a PR to add support for file uploads via GraphQL mutation variables according to the [GraphQL multipart request spec](https://github.com/jaydenseric/graphql-multipart-request-spec) pretty easily.

It would add [less than 1kb to the client bundle](https://bundlephobia.com/result?p=extract-files) and the API will be unchanged so it won't get in the way for users who don't need uploads.

The logic to add to the fetcher is only a few lines; you can see [how it works in `apollo-upload-client`](https://github.com/jaydenseric/apollo-upload-client/blob/v7.0.0-alpha.4/src/index.mjs#L49).

I looked to see if there is some sort of middleware system that could be used as an alternative to baking it in, but there isn't one?

kenwheeler:
That would be dope! Also, what would your ideal middleware setup look like?

jaydenseric:
I suggest per-request config, overridable by the user:

```js
import { Client } from 'urql'
import getViewerToken from './helpers'

const client = new Client({
  request: request => {
    // Can also be an async function if you want to await a few things to build the config.
    // `request` properties match https://developer.mozilla.org/en-US/docs/Web/API/Request#Properties
    // Populated with defaults the user can overide. For example, a request may
    // have a JSON or FormData body, depending if there are uploads.

    request.url = 'http://localhost:3000/graphql'

    // Token is retrieved at the time of each request.
    const token = getViewerToken()
    if (token)
      request.headers = {
        authorization: `Bearer ${token}`
      }

    // Will be used immediately by fetch.
    return request
  }
})
```

This allows a user to do almost anything you can imagine, without having to learn a system of links or middleware.

By design it also prevents a lot of common noob mistakes such as setting credentials at client initialization instead of per-request.

It is crazy easy to implement for a really lean bundle; There is no need for deep merging of request config set at client initialization and in middleware.

jaydenseric:
I edited the example above a few times for polish.

savovs:
That would be awesome!

kenwheeler:
can this be accomplished with the fetchOptions/functional fetchOptions settings on the Client instance currently?

pyros2097:
Yes most probably. 
If you see apollo-upload client source,
https://github.com/jaydenseric/apollo-upload-client/blob/master/src/index.mjs#L35
It just modifies the fetchOptions if it is a file upload and changes the variables a bit.

[...]

PR #332, opened 2019-07-03
(POC) - upload exchange

JoviDeCroock:
I have set up a really small exchange that would handle file uploads, during the making of this I realised this should maybe be added to our current `fetchExchange` since these are conflicting (or we build a check in it whatever you prefer). Since post-exchanges won't get called when returning false or do they? I might be making some thinking errors, this being the main reason as to why I put this up for draft.

Combining these would come at a disadvantage for `urql` consumers since this would impose an additional 500 bytes dependency that wouldn't be treeshaken when using the fetch exchange.

A second thing bothering me is that this is very hard to test, I'm going to dive deeper into how to test file uploads.

Thanks for reading and hope to hear your initial thoughts on whether or not to combine these.

PS: sorry for the integrity being removed from the lockfiles, updated my `yarn` on this pc now.

Based on:
- https://github.com/jaydenseric/graphql-multipart-request-spec
- https://github.com/FormidableLabs/urql/issues/14

### DIFF
--- a/src/exchanges/files.ts
+++ b/src/exchanges/files.ts
@@ -1,0 +1,61 @@
+import { print } from 'graphql';
+import { extractFiles } from 'extract-files';
+import { filter, pipe, map } from 'wonka';
+import { Operation } from 'src/types';
+
+const fileExchange = ({ forward }) => {
+  const isMutation = (operation: Operation) =>
+    operation.operationName === 'mutation';
+
+  return ops$ => {
+    const preFlight$ = pipe(
+      ops$,
+      filter(isMutation),
+      map(operation => {
+        const { url } = operation.context;
+        let { files } = extractFiles(operation);
+        if (!files.length) return true;
+
+        files = files.values();
+        const extraOptions =
+          typeof operation.context.fetchOptions === 'function'
+            ? operation.context.fetchOptions()
+            : operation.context.fetchOptions || {};
+
+        const fetchOptions = {
+          method: 'POST',
+          headers: {
+            ...extraOptions.headers,
+          },
+        };
+
+        const form = new FormData();
+        form.append(
+          'operations',
+          JSON.stringify({
+            query: print(operation.query),
+            variables: Object.assign({}, operation.variables, { file: null }),
+          })
+        );
+
+        const map = files.reduce((acc, path, i) => ({ ...acc, [i]: path }), {});
+        form.append('map', JSON.stringify(map));
+        files.forEach((file, i) => {
+          form.append(`${i}`, file, file.name);
+        });
+
+        (fetchOptions as any).body = form;
+
+        fetch(url, fetchOptions)
+          .then(res => res.json())
+          .then(json => json);
+
+        return false;
+      })
+    );
+
+    return forward(preFlight$);
+  };
+};
+
+export default fileExchange;
--- a/package.json
+++ b/package.json
@@ -134,6 +134,7 @@
     "react-dom": ">= 16.8.0"
   },
   "dependencies": {
+    "extract-files": "^5.0.1",
     "fast-json-stable-stringify": "^2.0.0",
     "graphql-tag": "^2.10.1",
     "prop-types": "^15.6.0",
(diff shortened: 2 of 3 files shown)


src/exchanges/files.ts
```diff
@@ -0,0 +1,61 @@
...
+
+const fileExchange = ({ forward }) => {
+  const isMutation = (operation: Operation) =>
+    operation.operationName === 'mutation';
+
+  return ops$ => {
+    const preFlight$ = pipe(
+      ops$,
+      filter(isMutation),
+      map(operation => {
```
JoviDeCroock: Not quite sure about `map` here.

src/exchanges/files.ts
```diff
@@ -0,0 +1,61 @@
...
+        const { url } = operation.context;
+        let { files } = extractFiles(operation);
+        if (!files.length) return true;
```
JoviDeCroock: Since it's a map I check for the length

src/exchanges/files.ts
```diff
@@ -0,0 +1,61 @@
...
+
+        files = files.values();
```
JoviDeCroock: We don't really need the keys as far as I know so I convert it to this instead

src/exchanges/files.ts
```diff
@@ -0,0 +1,61 @@
...
+
+        const fetchOptions = {
+          method: 'POST',
+          headers: {
+            ...extraOptions.headers,
+          },
+        };
+
+        const form = new FormData();
+        form.append(
```
JoviDeCroock: Add the operation to our formData

src/exchanges/files.ts
```diff
@@ -0,0 +1,61 @@
...
+          'operations',
+          JSON.stringify({
+            query: print(operation.query),
+            variables: Object.assign({}, operation.variables, { file: null }),
+          })
+        );
+
+        const map = files.reduce((acc, path, i) => ({ ...acc, [i]: path }), {});
```
JoviDeCroock: Build the filesmap

src/exchanges/files.ts
```diff
@@ -0,0 +1,61 @@
...
+        form.append('map', JSON.stringify(map));
+        files.forEach((file, i) => {
+          form.append(`${i}`, file, file.name);
+        });
+
+        (fetchOptions as any).body = form;
+
+        fetch(url, fetchOptions)
```
JoviDeCroock: Fetch and return false, here I'm not sure what the right course of action is. This feels wrong to me

kitten:
So, this is very likely an exchange that we wouldn’t want in the core library, i.e. this repository. But it could totally be a separate exchange published to npm. It’s good to have this solution in place in general, but I don’t think it’s one of the “core basic features” that need to be maintained for urql.

JoviDeCroock:
All right, thanks I'll move it to a separate repository. Thanks for the info

src/exchanges/files.ts
Narigo: When you map the operation, shouldn't you return the operation here instead of returning a boolean?

src/exchanges/files.ts
Narigo: I suppose you based this exchange on the example in issue #14 ? Over there, everything was done in the filter operation, working only with side effects instead of creating real operations.

I think it's better to base this on the official fetch-exchange, which is a bit more sophisticated in this regard.

src/exchanges/files.ts
Narigo: I suppose this should stream the result and merge the stream like fetch-exchange does...

kitten:
This should likely become a separate exchange package, I suppose?

This repository could be used as a template for publishing this: https://github.com/FormidableLabs/urql-exchange-suspense

Narigo:
I don't know if @JoviDeCroock is still working on this, I just wanted to say that I am not working with a GraphQL server that is supporting this and hence I won't be using or actively maintain something like the above code. If someone wants to use it if it helps, please go ahead :)

Issue #335, opened 2019-07-08
<Mutation> docs example fails typecheck


reverie:
Looking at https://formidable.com/open-source/urql/docs/getting-started/#writing-mutations, the provided example does not typecheck.

The `Mutation` element fails with error: "JSX element type 'ReactNode' is not a constructor function for JSX elements. Type 'undefined' is not assignable to type 'Element | null'.ts(2605)"

![Screen Shot 2019-07-08 at 1 51 29 PM](https://user-images.githubusercontent.com/73446/60831471-a9075e00-a187-11e9-81ab-f32de21b556f.png)
![Screen Shot 2019-07-08 at 1 51 38 PM](https://user-images.githubusercontent.com/73446/60831472-a99ff480-a187-11e9-8ca7-c022f59a71b5.png)

parkerziegler:
Released in [v1.2.0](https://github.com/FormidableLabs/urql/releases).

PR #337, opened 2019-07-08
(fix) - return value of functional components

JoviDeCroock:
fixes: https://github.com/FormidableLabs/urql/issues/335

Normally one would use the exported type React.FC for these but since this isn't possible with generics I followed this line from the React types to come to this conclusion.

```
type SFC<P> = StatelessComponent<P>;
interface StatelessComponent<P> {
 (props: P & { children?: ReactNode }, context?: any): ReactElement<any>;
}
```

### DIFF
--- a/src/components/Mutation.ts
+++ b/src/components/Mutation.ts
@@ -1,11 +1,11 @@
-import { ReactNode } from 'react';
+import { ReactElement } from 'react';
 import { DocumentNode } from 'graphql';
 import { OperationResult } from '../types';
 import { useMutation, UseMutationState } from '../hooks';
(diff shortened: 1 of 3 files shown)


JoviDeCroock:
Sorry about the notification spam, I missclicked on assignees there :/

andyrichardson:
Looks good to me!

I'd go as far as to say we shouldn't be explicitly stating function return types unless absolutely necessary. This can sometimes lead to upcasting and in the case of these components, an error will be thrown through implicit inference when used as a component and the return type is not `ReactElement | null`.

PR #343, opened 2019-07-16
Fix error in collectTypes method

mxstbr:
`obj.hasOwnProperty` can sometimes be undefined even though `obj` is an actual object, see this conversation over in another repo for reference: https://github.com/domchristie/humps/issues/35

I am (somehow through a million hoops) running into this in my project, and their suggested fix of using `Object.prototype.hasOwnProperty.call(obj, key)` works beautifully and fixes my issue.

### DIFF
--- a/src/utils/typenames.ts
+++ b/src/utils/typenames.ts
@@ -16,7 +16,7 @@ const collectTypes = (obj: EntityLike | EntityLike[], types: string[] = []) => {
     obj.forEach(inner => collectTypes(inner, types));
   } else if (typeof obj === 'object' && obj !== null) {
     for (const key in obj) {
-      if (obj.hasOwnProperty(key)) {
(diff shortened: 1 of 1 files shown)


mxstbr:
Any chance somebody could cut a release? :pray:

kitten:
cc @andyrichardson ❓

mxstbr:
Friendly ping we are blocked by this right now :blush:

andyrichardson:
Makes sense and looks good to me!

I suspect we had this conditional for TS reasons rather than runtime execution? Can't reproduce type errors on my end without the conditional so maybe this has been addressed in later updates. Either way, let's get this merged and look into that later :+1:

mxstbr:
Thank you @andyrichardson! A release would be great :pray:

andyrichardson:
@mxstbr I'll aim to get a release out tonight!

Issue #345, opened 2019-07-18
cache-and-network seems to not fire the network


dusty:
I am brand new to urql and while testing it out in my app, it appears to me that cache-and-network is not working as I had expected.

I created a different issue than #299 as that issue was primarily concerned about the executeQuery part of useQuery. 

My expectations from reading the docs is that it would return what is in cache, then it would also fire a network request in the background, update the cache, and cause a re-render if the data has changed.

Let me know if those expectations are incorrect.

I created this codesandbox to demonstrate.  https://codesandbox.io/s/youthful-fast-7g373

I created two components that each utilize the same query and have the requestPolicy set to cache-and-network.  We are grabbing the first new story from hackernews using graphqlhub api. 

When I load a component for the first time, it shows me the network request.  However, when moving between the components, it is only pulling from cache and not firing the network request.

I added debugExchange after cacheExchange.  It only logs on the first network request, which I believe is demonstrating that the cacheExchange is not forwarding on.

Below is a video:

![Fgi0NS0wxe](https://user-images.githubusercontent.com/792/61467886-43f2fb80-a931-11e9-8b2b-030909985e06.gif)

parkerziegler:
Thanks for the report @dusty. Agreed that I would expect `cache-and-network` to do what you expect – pull from the cache and then _always_ trigger a new network request. I believe this line in the `cache` exchange is the one we're looking to hit: https://github.com/FormidableLabs/urql/blob/d0bfd4928763029d155427554fe3022d0cee62e8/src/exchanges/cache.ts#L62.

Happy to look into this more and see what may be going awry. I'm also curious to investigate what using the same query for two components does for how we manage the cache.

One last thing – I popped open the CodeSandbox you provided (❤️thank you so much for that) and I ran into this error. Looks like the access on `data` may just be a bit off. I can work with this tho!

![image](https://user-images.githubusercontent.com/19421190/61475070-de0d7080-a93e-11e9-96ee-a5620fd27584.png)

dusty:
Thanks for the response and the pointer to the location.  

I've got it running locally and with some old fashioned console.log debugging, I traced it through and it is passing those checks and calling ```reexecuteOperation```.  

It makes its way over to and calling ```this.dispatchOperation(operation)``` in client.ts.  I see that it has the context.requestPolicy set to 'network-only' at this point as well.  However, its still not initiating the network request.

[...]

PR #349, opened 2019-07-21
(fix) - should reexecute operation on cache-and-network

JoviDeCroock:
This doesn't feel entirely right yet since cache-and-network should return AND dispatch a new request. This isn't the case yet was just making this PR to see if I'm on the correct path for this.

### DIFF
--- a/src/exchanges/cache.test.ts
+++ b/src/exchanges/cache.test.ts
@@ -18,7 +18,7 @@ import {
   subscriptionResult,
   undefinedQueryResponse,
 } from '../test-utils';
-import { Operation, OperationResult } from '../types';
+import { Operation, OperationResult, RequestPolicy } from '../types';
 import { afterMutation, cacheExchange } from './cache';
 
 let response;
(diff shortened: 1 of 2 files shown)


kitten:
This isn't quite correct. We have to let the operstions there through so that they're reexecuted later although they are indeed cached.

I believe the problem is that they're being deduped, which is totally technically correct behaviour, but i forgot that this means that this special policy won't work by default.

I'll have a think about how we can work around this nicely, but we'll probably just have to ensure that the operation is dispatched synchronously somehow after the dedup had a chance to receive the result.

JoviDeCroock:
Will also try give it another look, thanks for the explanation.

parkerziegler:
Ah makes sense, thanks for that explanation @kitten! I was trying to trace it through the exchanges (`dedup`, `cache`, `fetch`) and noticed that it hits the `dedupExchange` every time after `this.dispatchOperation` is called in the client, but it doesn't make it through to the `cacheExchange` or `fetchExchange`, so it'd make sense that it's getting `filter`ed out there before getting forwarded. Linking this to #345 so that subscribers there can track progress here 😄

JoviDeCroock:
@kitten 

When I check it with the following codesandbox it doesn't feel like it gets deduped but rather stopped in the cacheExchange as I did assume with making this PR

https://codesandbox.io/s/cool-ishizaka-9k4pf

kitten:
@JoviDeCroock I'm actually not seeing the same thing 😅  I'm seeing the operation in question disappear when it enters the `dedupExchange`, which does make sense to me. Currently the `reexecuteOperation` in `cacheExchange` is part of the exchange itself, so it will be stopped by the dedup. I think I'll attempt to alter `reexecuteOperation` to wait until the current operation has passed through the exchange pipeline, but I'll have to check whether a `dedupExchange` change may be sufficient

JoviDeCroock:
@kitten Hmm, I might have misunderstood my test. I'll see if I can reproduce it in a test.

kitten:
@JoviDeCroock I haven't tested this yet at all, but this should work 🤔  https://github.com/FormidableLabs/urql/pull/356

Issue #346, opened 2019-07-18
Add OperationContext to useQuery arguments


isker:
Hi!  I am trying to wrap `useQuery` in a custom hook so that I can provide new per-query functionality that interacts with a custom exchange I've written.  This is done by attaching per-query data in my hook via the `OperationContext`'s arbitrary-string properties.  But this data can only be provided by calling the `executeQuery` callback from `useQuery` ourselves; `useQuery` takes no `OperationContext` while `executeQuery` does:
https://github.com/FormidableLabs/urql/blob/fe99b20850ae18b9c9cb4b8d7aa0937e041a5ad1/src/hooks/useQuery.ts#L49-L50

It's not clear why this is the case.  To work around it in my hook, I have to pause the query on the call to `useQuery` then immediately invoke `executeQuery`, taking care not to create an infinite loop on render.  Are you open to adding this argument to `useQuery` directly?

kitten:
That's true. I think we've removed the rest spread some time before we were done but it should definitely be readded so that this can be provided cleanly

PR #351, opened 2019-07-23
(feat) - allow passing operationContext to useQuery

JoviDeCroock:
Fixes: https://github.com/FormidableLabs/urql/issues/346


src/hooks/useQuery.ts
```diff
@@ -13,6 +13,7 @@ export interface UseQueryArgs<V> {
   query: string | DocumentNode;
   variables?: V;
   requestPolicy?: RequestPolicy;
+  operationContext?: Partial<OperationContext>;
```
kitten: I was actually thinking we could just make this a rest spread? Or name it `context` at least, I suppose? I'm not quite sure.

What do you think, @andyrichardson?

Also, you're just too fast, Jovi 😆

src/hooks/useQuery.ts
```diff
@@ -63,7 +65,14 @@ export const useQuery = <T = any, V = object>(
         })
       );
     },
-    [args.requestPolicy, client, devtoolsContext, request, setState]
+    [
+      args.operationContext,
```
kitten: I think this is potentially a huge problem. I think people may accidentally misunderstand the context and cause too many refetches. This does mean that the rest spread may be the worse option and `args.context` the better one, but it's still something we should keep in mind

JoviDeCroock: So rename `operationContext` --> context and spread it out in the args array? Since that would enable our users to not having to memoize creating an object in useMemo argument?

kitten: I think spreading isn't a good idea, yea, as said it'd make it even harder to make sure the reference doesn't change. It was something we had at some point, I believe, but removed it. Maybe it was even due to this problem

JoviDeCroock: I remember us having a similar problem at one of my former projects, we ended up making a hook `usePreviousValue` and deeply comparing them. I would suggest leaving args.context out of the array to be honest.... This is against the rules of hooks but in theory it doesn't impact this concept of ours

kitten: Yea, I don't think we should go down that road; maybe let's just add "args.context" for now and not worry about this too much?

JoviDeCroock: That would make me having to adjust the tests though, I'm not too fond about that

JoviDeCroock: I pushed a fix for the failing test, should something like this fix it:

```
const context = React.useMemo(() => args.context, [...args.context]);
```

This however imposes an issue with `fetchOptions`

src/hooks/useQuery.ts
andyrichardson: The word _context_ in our client is just confusing when working with React 😄 

Prefixing it with operation makes sense but I'm happy with either.

JoviDeCroock: I understand what @andyrichardson is saying, context could be confusing another option would be to use the word `defaultOperationContext` but that could be quite long ^^

src/hooks/useQuery.ts
isker: FWIW I had to deal with this same problem in my hook that I described in #346.  Deciding that users of my hook (who are supplying values that go into the context) were very unlikely to provide referentially equal values across renders, I excluded those values from the react hook dependencies and added a lint to force my hook to be used with constant values 🙃.

Obviously you are dealing with a much more general problem than I am so this may still be the right solution, but for my use case I will have to keep pausing and invoking executeQuery myself to get around this dependencies check.

kitten:
If there are no objections, I'd just like to merge this as is for now 😅 Potentially this could obviously be nicer, but since for most users `context` won't be passed, the simple solution may be best here. In other cases for "advanced" use-cases with `context`, I'd assume that people would spot the required memoisation rather quickly.

On this note though, we may want to add the same modification to `useMutation` and `useSubscription` as well:

- `useMutation`: Potentially an optional second argument on `executeMutation`? https://github.com/FormidableLabs/urql/blob/master/src/hooks/useMutation.ts#L35
- `useSubscription`: Possibly the exact same https://github.com/FormidableLabs/urql/blob/master/src/hooks/useSubscription.ts#L12

JoviDeCroock:
Hmm, weird issue in the `useSubscription` test. When logging it calls with url but when checking the mock it doesn't seem to be present. Will look further after a break

src/hooks/useSubscription.test.tsx
```diff
@@ -70,6 +72,24 @@ describe('on initial useEffect', () => {
   });
 });
 
+it('should support setting context in useSubscription params', () => {
+  const context = { url: 'test' };
+  act(() => {
```
JoviDeCroock: This only works when act is present

JoviDeCroock:
@kitten just updating docs atm and then I think it's ready to merge, is there a spot that should be safe to note about the context and memoization of it?

kitten:
@JoviDeCroock I suppose we should add an entire section about the context somewhere, where we're also talking about customising urql. So maybe in "Extending & Experimenting". I'm thinking that doc can be extended with "FAQ-style" guides.

E.g.: "How to write custom exchanges", "How to extend the context", etc

JoviDeCroock:
All right, added it thanks for the heads up. It's rather small now but we can iterate on this.

kitten:
@jovidecroock let's discuss this tomorrow! There's a lot of todos for the docs anyway, so we can put down some specific todos for what sections need to be written 🙌

PR #355, opened 2019-07-26
(feat)  Add support for the top-level extensions key in the GraphQL specification

### LINKED ISSUE #353: Top level extensions are not passed through
GraphQL supports an `extensions` key at the top level, as siblings of the `data` and `error` keys. This is often used for cross cutting concerns or performance metrics etc. Note this is not the error extensions.

> The response map may also contain an entry with key extensions. This entry, if set, must have a map as its value. This entry is reserved for implementors to extend the protocol however they see fit, and hence there are no additional restrictions on its contents. 
https://graphql.github.io/graphql-spec/June2018/#sec-Response-Format

Currently, this is not parsed by the fetch exchange and I think is not considered at multiple levels below that too meaning it's ultimately not accessible.

Would a PR be accepted?

adamscybot:
Resolves #353.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -174,7 +174,7 @@ Every operation that enters the exchange pipeline will receive a result, either
 immediately (read: synchronously) or eventually as the result of a network
 request comes in.
 
-The raw GraphQL result from an API is typically: `{ data?: T, errors?: GraphQLError[], extensions?: Record<string, any> }`.
+The raw GraphQL result from an API is: `{ data?: T, errors?: GraphQLError[], extensions?: Record<string, any> }`.
 And `urql`'s operation results are very similar: `{ data?: T, error?: CombinedError, extensions?: Record<string, any>}`.
 
 The [`CombinedError` is a very simple wrapper](https://github.com/FormidableLabs/urql/blob/master/src/utils/error.ts)
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -174,8 +174,8 @@ Every operation that enters the exchange pipeline will receive a result, either
 immediately (read: synchronously) or eventually as the result of a network
 request comes in.
 
-The raw GraphQL result from an API is typically: `{ data?: T, errors?: GraphQLError[] }`.
-And `urql`'s operation results are very similar: `{ data?: T, error?: CombinedError }`.
+The raw GraphQL result from an API is typically: `{ data?: T, errors?: GraphQLError[], extensions?: Record<string, any> }`.
+And `urql`'s operation results are very similar: `{ data?: T, error?: CombinedError, extensions?: Record<string, any>}`.
 
 The [`CombinedError` is a very simple wrapper](https://github.com/FormidableLabs/urql/blob/master/src/utils/error.ts)
 that has either a `networkError` property with any unexpected errors that might occur,
--- a/src/exchanges/cache.ts
+++ b/src/exchanges/cache.ts
@@ -74,6 +74,7 @@ export const cacheExchange: Exchange = ({ forward, client }) => {
         return {
           data: undefined,
           error: undefined,
+          extensions: undefined,
           operation: addMetadata(operation, { cacheOutcome: 'miss' }),
         };
       })
--- a/src/exchanges/cache.ts
+++ b/src/exchanges/cache.ts
@@ -153,13 +153,13 @@ const afterQuery = (
   resultCache: ResultCache,
   operationCache: OperationCache
 ) => (response: OperationResult) => {
-  const { operation, data } = response;
+  const { operation, data, error } = response;
 
   if (data === undefined) {
     return;
   }
 
-  resultCache.set(operation.key, response);
+  resultCache.set(operation.key, { operation, data, error });
 
   collectTypesFromResponse(response.data).forEach(typeName => {
     const operations =
--- a/src/hooks/useMutation.test.tsx
+++ b/src/hooks/useMutation.test.tsx
@@ -6,7 +6,7 @@ jest.mock('../client', () => {
   const mock = {
     executeMutation: jest.fn(() =>
       pipe(
-        fromValue({ data: 1, error: 2, extensions: 3 }),
+        fromValue({ data: 1, error: 2, extensions: { i: 1 } }),
         delay(200)
       )
     ),
@@ -140,7 +140,7 @@ describe('on subscription update', () => {
     await execute();
     wrapper.update(<MutationUser {...props} />);
 
-    expect(state).toHaveProperty('extensions', 3);
+    expect(state).toHaveProperty('extensions', { i: 1 });
   });
 
   it('sets fetching to false', async () => {
--- a/src/hooks/useMutation.test.tsx
+++ b/src/hooks/useMutation.test.tsx
@@ -6,7 +6,7 @@ jest.mock('../client', () => {
   const mock = {
     executeMutation: jest.fn(() =>
       pipe(
-        fromValue({ data: 1, error: 2 }),
+        fromValue({ data: 1, error: 2, extensions: 3 }),
         delay(200)
       )
     ),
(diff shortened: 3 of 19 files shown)


src/hooks/useMutation.test.tsx
```diff
@@ -6,7 +6,7 @@ jest.mock('../client', () => {
   const mock = {
     executeMutation: jest.fn(() =>
       pipe(
-        fromValue({ data: 1, error: 2 }),
+        fromValue({ data: 1, error: 2, extensions: 3 }),
```
JoviDeCroock: I'd make this an object to conform to what we expect, this can just be { i: x }

docs/architecture.md
```diff
@@ -174,8 +174,8 @@ Every operation that enters the exchange pipeline will receive a result, either
 immediately (read: synchronously) or eventually as the result of a network
 request comes in.
 
-The raw GraphQL result from an API is typically: `{ data?: T, errors?: GraphQLError[] }`.
-And `urql`'s operation results are very similar: `{ data?: T, error?: CombinedError }`.
+The raw GraphQL result from an API is typically: `{ data?: T, errors?: GraphQLError[], extensions?: Record<string, any> }`.
```
kitten: Small nit, we can remove "typically" from this sentence because I was referring to extensions as an exception from the typical results

src/hooks/useMutation.test.tsx
adamscybot: Yeh, let me fix this one sec

src/hooks/useSubscription.ts
```diff
@@ -45,11 +47,12 @@ export const useSubscription = <T = any, R = T, V = object>(
 
     [unsubscribe.current] = pipe(
       client.executeSubscription(request, devtoolsContext),
-      subscribe(({ data, error }) => {
+      subscribe(({ data, error, extensions }) => {
         setState(s => ({
           fetching: true,
           data: handler !== undefined ? handler(s.data, data) : data,
           error,
+          extensions,
```
kitten: I don't think we should pass this on to the "user-level" as exchanges would be the only consumers of these special fields.

Or is there any use case you can think of? I'm not sure whether Apollo passes them on?

This can especially confusing because cached results wouldn't have any extensions, which makes this an easy pitfall as well

adamscybot: I don't think its the case that exchanges are the only consumers -- what's in there is entirely arbitrary and actually could have functional reasons. For us, we are passing down informations regarding authorisations and cross-cutting UI concerns that arent actually in the "graph". This is quite a purist view of graphQL -- but there are definitely use cases.

Another use case might be to surface some performance metrics in the UI.

> This can especially confusing because cached results wouldn't have any extensions, which makes this an easy pitfall as well

Ah, I was hoping they *would* come in. I think my changes to the `cacheExchange` sees to that?

docs/architecture.md
adamscybot: Sorted :)

src/hooks/useMutation.test.tsx
adamscybot: Fixed :)

src/hooks/useSubscription.ts
adamscybot: Heres an example we actually have in production: Sometimes a user can see part of an entity but they aren't allowed to see the wider detail of it -- hence, they cant "click through" on these entities. Data about this is sent in the extensions and the UI reacts to it. We dont send this data in the data itself as its very much meta and a cross cutting thing that can affect any entity.

Apollo doesn't pass them on. But it's because its a bug: https://github.com/apollographql/react-apollo/issues/2059

kitten: I think this very much could also be part of the data bit fair enough. The problem here though is that we have to assume that the extensions are bound to the specific request.

This means that we shouldn't cache it as it's stale when it's read back from the cache, especially if it contains tracing data and performance data. This probably should also not be included in the SSR result as the frontend shouldn't be concerned about request information from the backend.

I know this is controversial, but I have the feeling that this is very much "non best-practice territory".

If we're thinking of authorisation in general, if it's meta data then a custom exchange would possibly handle that. Otherwise it may be better to include it in data.

A normalised cache would never be able to cache this data, as queries can be fulfilled entirely offline.

So there's definitely a point I believe to surfacing this to the user for "advanced use cases" but the extensions probably shouldn't be cached, shouldn't be part of the SSR serialised data, and shouldn't be guaranteed.

I'm unsure however about some details around this 😅

adamscybot: Sure, I can see where you are coming from. It doesn't help the spec for extensions is essentially "object for stuff". We ourselves have wondered hard if we are overloading this thing. We could stretch the definition of an error for some of these use cases and use that. Those have a `path` in the spec back to the data and so are inherently linked.

If the extensions relate to the data, they could be cached. If the extensions don't relate directly to the data (perf metrics), they should not be cached. 

Back to the drawing board :(. I guess my next questions is should I just limit this whole PR to the fetchExchange? People could then do what they want with them in their own exchanges and potentially opt to put them in the data where they will be cached.

adamscybot: One potential pitfall is how one would get the data out of the hooks if they chose to make a custom exchange. You could overload the data, but that violates the types -- so you'd have to make a new custom hook to surface it.

As well as the necessary change to `fetchExchange` what about spreading any unknown keys to the state. Perhaps boxing them in a "meta" key or something.

That way, people can surface anything they want in the hooks via a custom exchange.

kitten: I mean, I'm definitely all for passing extensions on. Yea, that's a nice way of getting some extra data out of exchanges.

But none of the default behaviour should cache or guarantee the extensions. That's definitely custom exchange territory.

So the ssrExchange should probably be modified to remove extensions from the serialised result.

adamscybot: Right, so just so Im 100% clear before I sort this out: We want to surface extensions in the hooks, we just want to remove my bad assumption re: caching the extensions?

I think as well as ssrExchange i need to modify `cache.ts` to filter extensions before it does a cache set operation.

kitten: > We want to surface extensions in the hooks, we just want to remove my bad assumption re: caching the extensions?

yea, let's surface it. We probably won't put too much focus on it in the docs, but it's particularly useful for also adding (client-only) information/metadata/data from the exchanges, without modifying `data`, like you said.

> I think as well as ssrExchange i need to modify cache.ts to filter extensions before it does a cache set operation.

I think you might need to keep it when you set, because otherwise it may be missing during SSR after suspense is done 🤔  But we definitely want to filter it out from the data that is returned by `serializeData`, so that it's not sent to the client from the server.

Cheers for taking the time to do this btw 🙌

adamscybot: I've committed a fix to ssrExchange to not add extensions during serialisation.

I think I'm getting slightly confused over the cache in ssrExchange and cacheExchange.  In cacheExchange I've now removed the setting of extensions in there by filtering it -- I think we might accidentally be referring to different caches in above conversations.

Might be best for you to sanity check what I've just done.

kitten: Yea, I'll give it a quick review again 👌

src/exchanges/cache.ts
```diff
@@ -152,13 +153,13 @@ const afterQuery = (
...
 ) => (response: OperationResult) => {
-  const { operation, data } = response;
+  const { operation, data, error } = response;
 
   if (data === undefined) {
     return;
   }
 
-  resultCache.set(operation.key, response);
+  resultCache.set(operation.key, { operation, data, error });
```
kitten: Hm, this is probably a change we need for another PR, but it isn't safe to store the `error` I think. On the other hand it's obviously also not safe to store the data at all when there's an `error`. I suppose we have to revisit this

kitten:
so, on `exchanges/ssr` we need to filter out the `extensions` on `ssr.extractData`, since that's the ssr cache data that will be serialized and sent to the client. Unfortunately it won't be enough to remove it when we deserialize 😅

src/exchanges/cache.ts
adamscybot: So I think beforehand it was still operation, data and error that are set in the cache (it caches the whole response object containing those keys). This change just prevents caching the extensions.

kitten: yes, sorry, you're right. I'm just noticing somthing completely unrelated 😅  we may not want to cache the result when there's been a more severe error. For instance when the `CombinedError` has a `networkError`, but that's definitely sth for another PR. This change LGTM :+1:

kitten:
Yep, that's right! I've reviewed the full thing now and it looks perfect! I'll merge it and see if we need more tests and start seeing whether it all behaves as expected :+1:

Thanks again for taking the time to submit this!

PR #357, opened 2019-07-30
Add fetch option to Client args

### LINKED ISSUE #352: Provide custom fetch implementation
We have our custom fetch implementation (that mirrors the API for the default implementation) that does things like:

- Authentication.
- Handling custom response headers.
- Configuring and appending the API root.
- Handling the server-specific nuances.

This is quite a common pattern I believe. With `apollo-link-http` you can pass a fetch function down and it uses that.

Is it possible to do something similar? This is more generic and flexible than the current limitation of only configuring headers -- as you can't add any global/cross-cutting custom response handling using the options.

I could make my own exchange, but that feels like a large overhead when I just want the default fetch exchange but using my fetch function.

This might help mocking fetch in test env's too.

kitten:
Fix #352

Allow a custom fetch implementation to be passed to the client, which will be used in the fetchExchange.

We previously removed this option, but it may make sense to reintroduce it for ease of use.


kitten:
@JoviDeCroock We used to have this at some point I'm sure and then removed it. So let's wait for @andyrichardson's review as he might remember more details around why we don't have it anymore.

andyrichardson:
@kitten I don't recall any particular reason other than us already having a fetchExchange. Given how much of an edge case this is, I would assume anyone looking to change the fetch implementation would also be looking to implement their own fetchExchange. Is there a demand/use case for this?

kitten:
@andyrichardson as discussed, the issue has some comments on that and if the goal for someone is to only prevent a global fetch polyfill or reuse an exisiting fetch drop-in replacement, copying the exchange is a lot to ask

PR #361, opened 2019-08-01
(docs) - custom exchanges

JoviDeCroock:
In this first commit I've made an exchange that looks at the token and refreshes when needed, we build iteratively up until this point.

This covers the concepts:
- HOC exchange
- Basic exchange
- operations + forward
- link to wonka docs
- usage of mergeMap to decide a direction

Sandboxes:

Server: https://codesandbox.io/s/urql-issue-template-server-0ufyz
Client: https://codesandbox.io/s/recursing-shadow-t8b6g


docs/extending-and-experimenting.md
```diff
@@ -149,6 +149,171 @@ of the unsubscription (`teardown`) and update
...
+to forward being invoked. So this is basically an empty
+exchange.
+
+Next up is writing some code that refreshes our token, so imagine
+the following method:
+
+> note that these methods should be replaced with your own logic.
+
+```js
+const query = gql`
```
kitten: I think we should show this example with just REST, or even with imaginary `fetchToken` functions. It's a minor detail, since people know how to write those and if we want to forward a "refresh token" query there's also an option of forwarding it to the next exchange, which is a little trickier, so not worth doing here imo

andyrichardson:
This looks good! I think we've been needing something like this for a while and the use case is a valid one (although I'm not sure if blocking operations from being forwarded while waiting for async actions is something we want to encourage?).

There's some work here that might be salvageable - https://github.com/FormidableLabs/urql/pull/186/files

Would love to see the docs that you've done get merged 🙌

JoviDeCroock:
Thanks for the feedback @andyrichardson I do think it's needed since ideally when using a 
`refreshToken` mechanism you don't want to request unless your token is refreshed. When it's still valid it will just complete with the `fromValue` anyway.

I'll add one of your examples in there as well thanks for the link!

kitten:
@andyrichardson @JoviDeCroock I don't think we need to add a comment on this. In a lot of cases there'll be an existing access token mechanism that people will have to work with and refreshing on the fly isn't always an option but trivial to add to this example.

So the main thing I'd like to highlight is that these guides aren't about best practices but just examples so people can write any exchange they need and not just the ones we're writing about :)

It's just not realistic to add a super complex auth example on a guide here

JoviDeCroock:
I made some small adjustements to the language, removed a link since it was redundant and formatted the long lines.

JoviDeCroock:
Moved the sandbox with the completed refreshToken exchange to the `urql` codesandbox team

docs/guides.md
```diff
@@ -0,0 +1,413 @@
...
+```js
+import { pipe, fromPromise, map } from 'wonka';
+
+export const refreshTokenExchange = ({ forward }) => {
+  return operations$ => {
+    return pipe(
+      operations$,
+      pipe(
+        fromPromise(refreshToken()),
+        map(newToken => ({ ...op, context: { ...op.context, token: newToken } }))
```
tbrannam: what's `op` here?

JoviDeCroock: From the mergemap see line 351 or the codesandbox

PR #367, opened 2019-08-04
Add a higher-order function a fetch exchange

rzane:
Some features, such as file uploads, will require defining a new exchange to send a request. The `createFetchExchange` function will allow developers to customize requests at a low level, without
reimplementing the entire fetch exchange.

With this HOC, it should be relatively simple to build an exchange that handles file uploads.

### DIFF
--- a/src/exchanges/fetch.test.ts
+++ b/src/exchanges/fetch.test.ts
@@ -2,7 +2,7 @@ import { empty, fromValue, pipe, Source, subscribe, toPromise } from 'wonka';
 import { Client } from '../client';
 import { queryOperation } from '../test-utils';
 import { OperationResult, OperationType } from '../types';
-import { fetchExchange } from './fetch';
+import { fetchExchange, createFetchExchange } from './fetch';
 
 const fetch = (global as any).fetch as jest.Mock;
 const abort = jest.fn();
@@ -167,5 +167,27 @@ describe('on teardown', () => {
 
     expect(fetch).toHaveBeenCalledTimes(0);
     expect(abort).toHaveBeenCalledTimes(0);
+  });
+});
+
+describe('createFetchExchange', () => {
+  it('creates a new exchange', async () => {
+    const headers = { foo: 'bar' };
+    const buildRequest = jest.fn(() => ({ headers }));
+    const customExchange = createFetchExchange(buildRequest);
+
+    fetch.mockResolvedValue({});
+
+    await pipe(
+      fromValue(queryOperation),
+      customExchange(exchangeArgs),
+      toPromise
+    );
+
+    expect(buildRequest).toHaveBeenCalledWith(queryOperation);
+    expect(fetch).toHaveBeenCalledWith(
+      expect.any(String),
+      expect.objectContaining({ headers })
+    );
   });
 });
--- a/src/exchanges/fetch.ts
+++ b/src/exchanges/fetch.ts
@@ -1,11 +1,13 @@
 /* eslint-disable @typescript-eslint/no-use-before-define */
 import { print } from 'graphql';
 import { filter, make, merge, mergeMap, pipe, share, takeUntil } from 'wonka';
-import { Exchange, Operation, OperationResult } from '../types';
+import { Exchange, Operation, OperationResult, RequestBuilder } from '../types';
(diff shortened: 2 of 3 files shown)


kitten:
Hiya, sorry for the late reply 👋

This is basically also a response to: https://github.com/FormidableLabs/urql/issues/14#issuecomment-518352418

We're not completely done discussing this and wanted to be careful how to phrase this internally.

Basically, the background is that we understand all exchanges in the urql repository to be the absolute minimum. They all have minimal API surfaces and provide the basic functionality of a GraphQL client.

We're not entirely sure how we'll approach the issue of exchanges that blur the lines of customisability and solid standard behaviour. So these are all things that we're still actively trying to figure out.

Our default API for customising urql's behaviour will always be custom exchanges, via third-party packages and via one-off exchanges.

This is also why the auth exchange is just a guide and not a built-in "context exchange".

Things get more complicated when we approach the fetch exchange. It already has three options on the client (url, fetch, fetchOptions) itself to customise it's behaviour. Optimally these options would be on the fetchExchange itself, which would complicate its API however and would mean that every user would have to pass the exchanges array to the client, even if they're just getting started.

The problem here is that we can't anticipate every modification that people would like to make to our default exchanges, especially the fetch exchange. So we'd prefer not to throw more API surface area at it and instead encourage custom exchanges.

This does mean that copy pasting is sometimes one of our proposed solutions, which you've of course addressed in your comment. This is an obvious drawback, but on the other hand, once you've written a custom exchange it shouldn't have to ever be changed unless you personally have new requirements.

We're very dedicated to keeping the exchanges' rules and API contracts extremely stable.

Specifically for the fetchExchange you don't have to even copy all of it. You may not need the redirect options or other parts of it. I'd be happy to also make smaller templates for exchanges like this.

We'll be thinking about the fetchExchange specifically a little more in the coming days, but we have to keep customisability via exchanges (as a default recommendation) and small API surface area in our mind, as our main goals.

I hope that makes sense 🙌 I'm sure we can find a compromise on achieving these goals without adding more options.

I think your solution is extremely elegant and I actually think we can probably merge it with some minor changes, but we'll have to be absolutely sure that this complies with the goals of the library, since we can't revert any of these major changes once they're merged 😅

rzane:
Yeah, I totally understand and I appreciate your dedication to API stability.

I think another way to solve my *very specific* use case (which is file uploads) would be to just detect if the `body` is `FormData`. In which case, just remove the `application/json` header. With that change, I can have a custom exchange that runs before the `fetchExchange` that sets the `body` on the `fetchOptions`. I'm more than happy to make a PR to implement it this way.

Another thing to consider about the specific problem of file uploads, is that while there is a proposed standard (https://github.com/jaydenseric/graphql-multipart-request-spec), it's not an accepted standard yet. As such, there are other ways of doing it.

For example, I'm using Absinthe (an Elixir library for building GraphQL APIs) on a project. They have a built-in `Upload` type, and it expects the request to come in a little bit differently.

That means, if we were to take the copy-pasta approach, there will be at least three copies of the fetch exchange living out in the wild.

kitten:
Hiya, given the copy&paste guide we'll likely make some modifications that make it less painful to create a `fetchExchange` from scratch, including:

- `makeResult` and `makeErrorResult` from https://github.com/FormidableLabs/urql/blob/master/src/utils/result.ts which are new
- Removing devtools' metadata from `urql` (#388), which simplifies the code a lot: https://github.com/FormidableLabs/urql/blob/b3a8a22928e16e4d22561bd87b566315da61a2e4/src/exchanges/fetch.ts

We simply don't think that a higher-order function is the right approach. Conceptually it _does_ fit our API patterns really well and we'd typically merge this! However, it does encourage more logic to be moved _inside_ the fetch exchange, which makes it less reusable unfortunately and we don't want users to be encouraged to go down that path.

Apart from that use-cases like persisted queries wouldn't be covered by this modification as well (see https://github.com/Daniel15/urql-persisted-queries), so it doesn't unfortunately even cover all use-cases. That doesn't mean that we should go after making the `fetchExchange` way more reusable, but we believe that exposing `result.ts` and making `fetchExchange` slimmer is the right thing to do.

Hope that makes sense! Sorry for having this PR stick around for this long :pray:

PR #369, opened 2019-08-05
Fix ssrExchange's client-mode defaulting based on suspense-mode

kitten:
Previously it was impossible to use the `ssrExchange` together with `suspense` mode on the client-side, since the `ssrExchange` assumes that it's on the server, when the `suspense` mode is turned on.

This PR adds a separate `isClient` option to `ssrExchange`, so it can be told whether it's running on the client-side manually.

### DIFF
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -230,6 +230,15 @@ The exchange returned by `ssrExchange()` should be added after the `cacheExchang
 (or any other custom cache exchange you've defined), and before any
 asynchronous exchanges like the `fetchExchange`.
 
+If you're also using suspense mode on the client, you can
+additionally set the `isClient` option, which tells the `ssrExchange` manually
+whether it's on the server or client, so that you can enable the `suspense`
+mode on the client-side as well.
+
+```js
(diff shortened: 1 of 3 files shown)


iamstarkov:
nice

docs/basics.md
```diff
@@ -230,6 +230,15 @@ The exchange returned by `ssrExchange()` should be added after the `cacheExchang
 (or any other custom cache exchange you've defined), and before any
 asynchronous exchanges like the `fetchExchange`.
 
+If you're also using suspense mode on the client, you can
+additionally set the `isClient` option, which tells the `ssrExchange` manually
+whether it's on the server or client, so that you can enable the `suspense`
+mode on the client-side as well.
+
+```js
+const ssrCache = ssrExchange({ isClient: !!process.browser });
```
iamstarkov: afaik, `typeof window !== 'undefined'` is better

kitten: Yea, totally, process.browser is a "Webpack-ism" which is noted here but that's fine too. I'm just assuming that people will be able to decide themselves :)

iamstarkov: true

PR #370, opened 2019-08-05
(docs) - guided exchange

JoviDeCroock:
Making a file upload exchange guide


docs/guides.md
```diff
@@ -411,3 +411,100 @@ by adding it into your exchanges array as `refreshTokenExchange()`.
...
+`File` or `FileList` and send it to the server in the right format.
+
+The only thing left to do is to include this new exchange by doing:
+
+```js
+import { createClient, dedupExchange, cacheExchange } from 'urql';
+import { myFetchExchange } from './fetchExchange';
+
+const client = createClient({
+  exchanges: [dedupExchange, cacheExchange, myFetchExchange],
```
rzane: This is hard for me to follow, because `myFetchExchange` is never defined anywhere in this document.

kitten: That's a good point! Maybe we should be a bit clearer of the exact changes we're making to the fetchExchange.

That being said; this guide will also have a CodeSandbox link at the end soon, so that you can just read the full source code 👍

JoviDeCroock: yes still have to finish the sandbox for server and client will do so tomorrow

rzane: While I think this documentation is really valuable, I think it could also probably be summed up with a link to the `src/exchanges` directory and saying "hey, you can copy and paste these things and modify them as you see fit".

I'm sort of waiting to see if #367 gets merged, at which point, I'm hoping to make a package to support uploads. If a package exists that implements file uploads, is this worth documenting?

JoviDeCroock:
@kitten the sandbox etc is uploaded I think we are ready to merge this

PR #375, opened 2019-08-06
Make further bundle size optimisations

kitten:
This accomplishes a couple of things.

Firstly it sets up Rollup to compile `process.env.NODE_ENV` to be `"production"` for our minified bundles so those accurately reflect what `urql` would compile to in a production environment.

Then it adds some initial bundle size optimisations:

- Remove rest spread from Query/Subscription components as it's not needed
- Replace `debugExchange` with a noop in production
- Remove usage of `addMetadata` with latency times from `fetchExchange` in production
- Remove fallback and fetch error for unknown operations in production
- Extract response/error parsing from `fetchExchange` and `subscriptionExchange` into a common file

This appears to currently save us about `0.3kB (4.62 kB => 4.32kB) min+gzip`

### DIFF
--- a/package.json
+++ b/package.json
@@ -122,6 +122,7 @@
     "rollup-plugin-buble": "^0.19.8",
     "rollup-plugin-commonjs": "^10.0.1",
     "rollup-plugin-node-resolve": "^5.2.0",
+    "rollup-plugin-replace": "^2.2.0",
     "rollup-plugin-terser": "^5.1.1",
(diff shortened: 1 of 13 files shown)


JoviDeCroock:
I think this in combination with the removal of graphql-tag will drop the bundle size on phobia down by quite a good margin. I think it's ready personally don't see that much aspects to make it smaller atm (apart from things that reduce readability). We could also mangle really internal properties

PR #379, opened 2019-08-07
(chore) - continue without graphql-tag

JoviDeCroock:
This PR forces consumers to pass in a `DocumentNode` instead of us parsing this double, since this is already a common practice in the wild.

### DIFF
--- a/src/exchanges/subscription.ts
+++ b/src/exchanges/subscription.ts
@@ -37,6 +37,8 @@ export interface ObservableLike<T> {
 }
 
 export interface SubscriptionOperation {
+  // TODO: why does this work?
+  // And why doesn't it work with DocumentNode
   query: string;
   variables?: object;
   key: string;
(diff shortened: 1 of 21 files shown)


src/exchanges/subscription.ts
```diff
@@ -37,6 +37,8 @@ export interface ObservableLike<T> {
 }
 
 export interface SubscriptionOperation {
+  // TODO: why does this work?
```
JoviDeCroock: Note this here, somehow changing query to DocumentNode doesn't work since `print` expects an ASTNode

kitten:
So if we only allow a DocumentNode that's a breaking change though, so maybe we should just preserve the parsing w/o graphql-tag. If we don't accept strings we won't be able to have some special cases in the future where we just accept strings and prevent bundling of the graphql package entirely 😅

Issue #377, opened 2019-08-07
Remove dependency on graphql-tag


kitten:
There are several improvements we can make to ensure that we don't need to depend on graphql-tag.

Previously though we ran into issues where using graphql-tag and urql in one project led to import issues with the graphql package itself for some reason, which is why we just used its utility for parsing GraphQL DSL queries.

If we switch back to parsing ourselves we have to probably make some improvements to the hashing. Specifically graphql-tag guarantees that the same two queries always yield the same instance of a DocumentNode on which we can store a hash.

Issue #206, opened 2019-03-01
Uncaught TypeError: Object(...) is not a function


sun2rise:
Hi,
as soon as I try to make a query I have this error:

```
Uncaught TypeError: Object(...) is not a function
    at Client.createRequestOperation (urql.es.js:506)
    at Client.executeQuery (urql.es.js:522)
    at urql.es.js:951
    at urql.es.js:966
    at commitHookEffectList (react-dom.development.js:17860)
    at commitPassiveHookEffects (react-dom.development.js:17887)
    at HTMLUnknownElement.callCallback (react-dom.development.js:147)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:196)
    at invokeGuardedCallback (react-dom.development.js:250)
    at commitPassiveEffects (react-dom.development.js:19513)
    at wrapped (scheduler-tracing.development.js:177)
    at flushFirstCallback (scheduler.development.js:99)
    at flushWork (scheduler.development.js:216)
    at MessagePort.push../node_modules/scheduler/cjs/scheduler.development.js.channel.port1.onmessage (scheduler.development.js:619)
```

It's strange because the same exact project inside codesandbox.io was working perfectly but now I have the same error in local and on netlify.
I cannot share this project, I'm sorry

I am using this (from `package.json`):
```json
"dependencies": {
    "@material-ui/core": "3.9.2",
    "@material-ui/icons": "3.0.2",
    "axios": "0.18.0",
    "date-fns": "1.30.1",
    "graphql": "14.1.1",
    "lodash": "4.17.11",
    "react": "16.8.3",
    "react-dom": "16.8.3",
    "react-router-dom": "4.3.1",
    "react-scripts": "2.0.3",
    "recharts": "1.5.0",
    "urql": "1.0.4"
  },
```

kitten:
Hiya, that's certainly odd! I apologise, we haven't announced v1 publicly for this very reason. We're still searching for good error messages that we can add here and there for more clarity.

Do you have a small snippet at least of your usage of useQuery? A completely anonymised snippet would be completely fine! Feel free to comment out the actual query, variables, and component around it 👌

sun2rise:
Sure!, I hope this is enough

I start like this:
```jsx
import React from 'react';
import ReactDOM from 'react-dom';
import { BrowserRouter as Router } from 'react-router-dom';
import Main from './pages/Main';
import * as serviceWorker from './serviceWorker';
import { Provider, createClient } from 'urql';

[...]

PR #380, opened 2019-08-07
(refactor) - remove gql-tag dependency

JoviDeCroock:
Well this makes it a lot easier on us :D

This should reduce the footprint of urql to 6.5KB

Resolves: https://github.com/FormidableLabs/urql/issues/377

### DIFF
--- a/package.json
+++ b/package.json
@@ -135,7 +135,6 @@
   },
   "dependencies": {
     "fast-json-stable-stringify": "^2.0.0",
-    "graphql-tag": "^2.10.1",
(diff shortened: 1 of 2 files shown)


kitten:
Found the old issue related to this. We have to ensure that this bug doesn't occur again: https://github.com/FormidableLabs/urql/issues/206

Let's also check whether we need to change the hashing to be more efficient before we merge this?

JoviDeCroock:
@kitten do you mean we have to strip whitespaces and stuff like that?

Issue #253, opened 2019-06-03
Better packaging for non React consumers


tchak:
I am working on `Ember.js` addon for `urql`. It would make my life easier (and I suspect of any implementor not using `React` stuff) to have a separate package with core API.

tchak:
Or at least provide a standalone build with no references to `React`.

kitten:
So, this is a tricky one. While we've built urql so that no parts of React or parts of our logic blend into one another we would like to keep the amount of libraries we publish as low as possible; at least for the time being and probably not forever.

So this is definitely something we'll consider if other libraries for different UI frameworks become popular.

For now though tree shaking should be able to remove all parts of the urql code that you don't use. So kind of controversial, but you could bundle urql into your own library without actually depending on it 🤔

kitten:
I'll close this as a duplicate of https://github.com/FormidableLabs/urql/issues/423 which is a concrete RFC on how we're going to move forward with this without introducing another package :+1:

PR #385, opened 2019-08-10
Add guide on using urql outside of React.

parkerziegler:
This PR is currently WiP, but I figured I'd get feedback early from folks. The purpose of this PR is to create a guide for using `urql` outside of React. Specifically, I'm focusing on highlighting how to use `urql`'s `client` directly, since that's the easiest way for non-React consumers to get access to the core functionality.

Currently, I have two examples for using `urql` with NodeJS. I'd like to add a basic example with Vue as well before merging.

The biggest issue I see with using `urql` in non-React codebases is the peer dependency on React, and the way our internal references to `react` require it being installed even in a Node or Vue environment where it isn't used. I'd be curious to see if our tree-shaking works out in a Vue environment. Regardless, this makes me think of two things:

1. We should probably have a Caveats section in this doc that makes this plain. It's not a big deal if you're just using `urql` to write quick little Node scripts / CLIs for personal use, but the peerDep on `react` makes me question how useful this is.
2. As @kitten suggested in #253, there may be a time in the future where we consider splitting the `client` into a separate package to support non-React consumers who just want access to the client. But I don't think we're there yet 😅 We should continue to focus on making this the best it can be for React.

### DIFF
--- a/docs/extending-and-experimenting.md
+++ b/docs/extending-and-experimenting.md
@@ -20,6 +20,7 @@ and hooks or even React at all.
 The Client is structured so that it can easily be reused to create different
 APIs or integrate `urql` with other libraries and frameworks than React.
 [Have a look at the API docs to see a full list of the client's methods.](api.md#client-class)
+Also take a look at [how to use `urql` outside of a React context.](url-outside-react.md)
 
 While it's possible to write new APIs outside of React or new components
 quite easily, this section will only focus on hooks, for illustrative
--- a/docs/urql-outside-react.md
+++ b/docs/urql-outside-react.md
@@ -1,17 +1,18 @@
 # `urql` Outside of React
 
-While you'll get the best experience with `urql` in a React codebase, it's still possible to use parts of `urql`'s functionality outside of React. In this guide, we'll take a look at using `urql` in environments other than React, including NodeJS and Vue.
+While `urql` is a GraphQL client that's mainly targeting React, apart from our React-specific APIs (i.e. the hooks and components), all other elements of the library function the same outside of a React context. In this guide, we'll take a look at using `urql` in environments other than React, including NodeJS and Vue.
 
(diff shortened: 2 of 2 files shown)


parkerziegler:
> Do we want this to even include the HTTP server bits? This doesn't showcase a spec- compliant GraphQL server, so it may be a confusing example

I'll remove these. They're included purely because `http` is used in the CodeSandbox examples to write a response to the client (so that the user sees something), but I agree that they're just implementation details that are confusing in this context 👍

dusty:
I've just been experimenting with urql.  We have an app that is angular 1 and react.  All new things are in react, but we have a ton of old angular components that we don't have time to rewrite.  Here is what I ended up using in my experimentation.  So far I just played with queries.

I export my urql client that I created and then pull that in to an angular factory.

```
import angular from 'angular'
import { pipe, subscribe } from 'wonka'
import { createRequest } from 'urql'
import { client } from '@/graphql'  // this is the client I created

angular.module('common').factory('GraphQL', GraphQL)

GraphQL.$inject = ['$q']
function GraphQL($q) {
  return { query }

  function query({ query, variables }) {
    const request = createRequest(query, variables)
    return $q(resolve =>
      pipe(
        client.executeQuery(request),
        subscribe(resolve)
      )
    ).then(({ data, error }) => (error ? $q.reject(error) : data))
  }
}
```

Somewhere else

```
return GraphQL.query({
      query,
      variables: { id }
    }).then(data => data.theThing)
```

If we ever decide to move forward with urql (we are currently playing with apollo) I'd be happy to share whatever we end up doing.

But, it sounds like you were asking for feedback, so I thought I'd pass this along.

parkerziegler:
Thanks for the example @dusty. I'm glad to see that folks already seem to be figuring a lot of what this guide will try to formalize. I think we're currently planning to keep the examples to Node and Vue, but if there's more interest I'd definitely be happy to have an Angular or Ember example in there as well; I'd probably just want to ensure we were showing examples in the latest versions of those frameworks.

parkerziegler:
@JoviDeCroock definitely. I never did see an invite come through, would you mind resending it?

PR #393, opened 2019-08-17
(fix) - typos

JoviDeCroock:
I should also fix these in the PR to `urql-docs`, the others mentioned are already fixed in `urql` master but aren't reflected to the docs publish yet


JoviDeCroock:
Forgot my HEAD on the windows pc is totally broken sorry about that....

PR #396, opened 2019-08-20
Add pollInterval argument to queries

mxstbr:
This PR adds `pollInterval` option to queries, which causes them to be refetched every `pollInterval` ms. Example:

 ```JS
 useQuery({
   query,
   pollInterval: 1000
 })
 ```

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -12,6 +12,7 @@ interface UseQueryArgs {
   variables?: any;
   requestPolicy?: RequestPolicy;
   pause?: boolean;
+  pollInterval?: number;
   context?: Partial<OperationContext>;
 }
 ```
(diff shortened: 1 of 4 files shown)


kitten:
This is a pretty awesome and useful addition! :clap: Definitely going to merge this ⭐️ 

I was just wondering—and it's just a small detail—should this be part of the client logic? That'd move it away from the hooks and make it accessible in Node/other environments:

```js
  executeQuery = (
    query: GraphQLRequest,
    opts?: Partial<OperationContext>
    const operation = this.createRequestOperation('query', query, opts);
    const response$ = this.executeRequestOperation(operation);

    if (opts.pollInterval) {
      return pipe(
        merge([
          fromValue(0),
          interval(opts.pollInterval),
        ]),
        mergeMap(() => response$),
      );
    } else {
      return response$;
    }
  };
```

This is a change to the client `executeQuery` method and it only applies to queries. The nice thing here is, we're automatically taking care of the cancellation/switching logic internally with stream operators

mxstbr:
Closing in favor of #397!

Issue #403, opened 2019-08-22
Fragments are no longer deduplicated as of 1.4.0


good-idea:
**Describe the bug**
When a query string has multiple references to the same fragment, GraphQL servers will return an error like "There can be only one fragment named \"LinkFragment\"."

When using fragments to compose queries, it's common to use fragments-within-fragments, like so:

```js
/**
 * Fragments
 */
export const linkFragment = /* GraphQL */ `
	fragment LinkFragment on PageLink {
		linkData
	}
`

export const textBlockFragment = /* GraphQL */ `
	fragment TextBlockFragment on TextBlock {
		title
		body
		link {
			...LinkFragment
		}
	}
	${linkFragment}
`

export const imageBlockFragment = /* GraphQL */ `
	fragment ImageBlockFragment on ImageBlock {
		title
		caption
		link {
			...LinkFragment
		}
	}
	${linkFragment}
`

export const contentSectionFragment = /* GraphQL */ `
	fragment ContentSectionFragment on ContentSection {
		items {
			... on ImageBlock {
				...ImageBlockFragment
			}
			... on TextBlock {
				...TextBlockFragment
			}
		}
		backgroundImage {
			...SanityImageFragment
		}
	}
	${imageBlockFragment}
	${textBlockFragment}
`
/**
 * Homepage Query
 */
export const HomepageQuery = /* GraphQL */`
	query HomepageQuery {
		Homepage(id: "homepage") {
			_id
			contentSections {
				...ContentSectionFragment
			}
		}
	}
	${contentSectionFragment}
`
```

With urql's removal of graphql-tag, these duplicated fragments are not removed from the final query. This can be fixed by using graphql-tag in your project, i.e.:

```js
export const HomepageQuery = gql`
	query HomepageQuery {
		Homepage(id: "homepage") {
			_id
			contentSections {
				...ContentSectionFragment
			}
		}
	}
	${contentSectionFragment}
`
```

**Steps to reproduce**

Compose a query like the above that contains multiple inline references to the same fragment

**Expected behavior**
The query should work as it did before, with urql de-duplicating the fragments. Perhaps an exchange that does something similar to [this](https://github.com/apollographql/apollo-client/issues/897#issuecomment-260087424).
**Or**, an error message that detects the problem ahead of time and suggests to import and use `gql` manually.

**Actual behavior**
Duped fragments are no longer removed, Graphql API complains

kitten:
[...]

PR #404, opened 2019-08-22
fix: (WIP) deduplicate fragment definitions

good-idea:
Addresses #403 

This includes a simple exchange that filters out duplicated fragment definitions.

- [ ] There's a typescript error on line 13 - it's complaining that "name" doesn't exist on type "DefinitionNode" - while `DefinitionNode` is an enum with a handful of types that *do* have a `name`. I wasn't sure how to fix this.
- [ ] The test isn't working. I'm not familiar enough with wonka, etc, to make it all happen - although, I've tested this in a project of my own and it works.

### DIFF
--- a/src/exchanges/dedupFragments.test.ts
+++ b/src/exchanges/dedupFragments.test.ts
@@ -1,0 +1,64 @@
+import { empty, pipe, fromValue, toPromise, Source } from 'wonka';
+import { dedupFragmentsExchange } from './dedupFragments';
+import { queryOperation } from '../test-utils';
+import { createRequest } from '../utils';
+import { OperationResult } from '../types';
+import { Client } from '../client';
+
+const linkFragment = `
+fragment LinkFragment on Link {
+  url
+}
+`;
+
+const imageFragment = `
+fragment ImageFragment on Image {
+  src 
+  link {
+    ...LinkFragment
+  }
+}
+${linkFragment}
+`;
+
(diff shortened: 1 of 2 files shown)


JoviDeCroock:
Hey @good-idea,

I don't think an exchange is the best way to put this logic, this happens at query parse time, most people bypass this issue by using `graphql-tag` and other relevant libraries. What we do now is when an unparsed string is passed we assume you are passing a simple query and we use the native graphql parser.

kitten:
@good-idea 	Hiya :wave: As Jovi said, sorry, we're not going to merge this into the core library. We deem `graphql-tag` as the default solution for this as it also has some nice runtime warnings. Essentially, we don't need to take care of this and I think we're "too late" in the chain and `graphql-tag` is a nice abstraction already.

That being said you could consider publishing this as a separate exchange package on npm and we'd be happy to also give you a package name on the `@urql` scope :+1:

good-idea:
@kitten totally understandable! I had the same question when I was putting the PR together - if this should be happening in an exchange or not, or if it made sense to include in the package at all. But, I wanted to mess around with exchanges as I hadn't done that before
@JoviDeCroock, I missed your comment, sorry for the late response.

Back to the original issue --- I still think it would be good to have a more helpful error message, as the removal of `graphql-tag` is a breaking change in some scenarios.

PR #405, opened 2019-08-23
(feat) - promisified requests

### LINKED ISSUE #216: Add promisified request methods on the Client (especially for non-React API calls)

JoviDeCroock:
Resolve #216

This is a POC with a minimal amount of added code to support this concept.

Wanted to give this a try, some concerns I'm having is what if we are polling? Multiple results will come in.
I would opt for no polling but yeah...

Been trying out some stuff but this does not seem to result in any responses during our tests

I've tried this in the examples and it works

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -204,16 +204,13 @@ export class Client {
     }
   };
 
-  query({
+  async query({
     query,
     variables,
     context,
   }: PromisfiedOperation): Promise<OperationResult> {
     const request = createRequest(query, variables);
-    return pipe(
-      this.executeQuery(request, context),
-      toPromise
-    );
+    return toPromise(this.executeQuery(request, context));
   }
 
   executeQuery = (
@@ -248,10 +245,7 @@ export class Client {
     context,
   }: PromisfiedOperation): Promise<OperationResult> {
     const request = createRequest(query, variables);
-    return pipe(
-      this.executeMutation(request, context),
-      toPromise
-    );
+    return toPromise(this.executeMutation(request, context));
   }
 
   executeMutation = (
--- a/src/client.ts
+++ b/src/client.ts
@@ -57,7 +57,7 @@ interface ActiveOperations {
 export const createClient = (opts: ClientOptions) => new Client(opts);
 
 type PromisfiedOperationResult = Source<OperationResult> & {
-  toPromise: () => Promise<any>;
+  toPromise: () => Promise<OperationResult>;
 };
 
 const withPromise = (
--- a/src/client.ts
+++ b/src/client.ts
@@ -55,7 +55,7 @@ interface ActiveOperations {
 }
 
 interface PromisfiedOperation {
-  context: Partial<OperationContext>;
+  context?: Partial<OperationContext>;
   query: string | DocumentNode;
   variables?;
 }
@@ -204,13 +204,17 @@ export class Client {
     }
   };
 
-  async query({
+  query({
     query,
     variables,
     context,
   }: PromisfiedOperation): Promise<OperationResult> {
     const request = createRequest(query, variables);
-    return toPromise(this.executeQuery(request, context));
+    return pipe(
+      this.executeQuery(request, context),
+      take(1),
+      toPromise
+    );
   }
 
   executeQuery = (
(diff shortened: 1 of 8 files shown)


src/client.ts
```diff
@@ -222,6 +243,19 @@ export class Client {
...
+  mutation({
+    query,
+    variables,
+    context,
+  }: PromisfiedOperation): Promise<OperationResult> {
+    const request = createRequest(query, variables);
+    return pipe(
+      this.executeMutation(request, context),
+      take(1),
+      toPromise
```
kitten: So for the reason of multiple results I'd still opt for no promisification by default. It doesn't make sense to create a shortcut API that then compromises on only using promises which also hides some of what you could otherwise do without problems 😅

src/client.ts
```diff
@@ -222,6 +243,19 @@ export class Client {
     return this.executeRequestOperation(operation);
   };
 
+  mutation({
+    query,
```
kitten: I'd just expand these into three arguments, the reason being that they're likely to be three separate variables in user code anyway.

src/client.ts
JoviDeCroock: So you would remove take(1) and toPromise and just give the user the option to subscribe?

kitten: Yep, to subscribe, call toPromise, themselves, etc. Otherwise I wouldn't even be sure what we'd be adding these methods for 😅

Admittedly promises are widespread, but are we really just adding these shortcuts for server code?

Because in any other case the cancellation and stream will likely be important to people

JoviDeCroock: What do you think about this: we offer the possibility for a fourth argument that is called `promisify` this will return a `take(1), toPromise`. As a comfort API or are we just settling on `return this.executeMutation(request, context)`

src/client.ts
```diff
@@ -54,6 +56,22 @@ interface ActiveOperations {
 
 export const createClient = (opts: ClientOptions) => new Client(opts);
 
+type PromisfiedOperationResult = Source<OperationResult> & {
+  toPromise: () => Promise<OperationResult>;
```
kitten: Lets extract this type and make it a generic?

src/client.ts
```diff
@@ -54,6 +56,22 @@ interface ActiveOperations {
...
+};
+
+const withPromise = (
```
kitten: We can make this a helper potentially

src/client.ts
```diff
@@ -196,6 +214,16 @@ export class Client {
     }
   };
 
+  query(
+    query: DocumentNode | string,
+    variables?: object,
+    context?: Partial<OperationContext>
+  ): PromisfiedOperationResult {
+    return withPromise(
+      this.executeQuery(createRequest(query, variables), context)
```
kitten: Beautiful 👏💛

src/utils/withPromise.ts
```diff
@@ -0,0 +1,12 @@
...
+import { PromisifiedSource } from '../types';
+
+export function withPromise<T = any>(source$: Source<T>): PromisifiedSource<T> {
+  (source$ as PromisifiedSource<T>).toPromise = () =>
+    pipe(
+      source$,
+      take(1),
+      toPromise
+    );
+  return source$ as PromisifiedSource<T>;
```
kitten: This cast shouldn't be necessary and it may be better to remove the any fallback

PR #410, opened 2019-08-26
Implement active teardowns and add useQuery features to useSubscription

### LINKED ISSUE #407: Completed subscription not reflected in hook data
There are two cases where a subscription query completed. Either it unmounts or the server completes the source. i was expecting that on source completion the "fetching" value of useSubscription would turn to false but it still stays the same.

I went through the code of the subscription exchange and found out that the complete event from forwardSubscription is sent to the wonka stream via the complete function but for an unknown reason it does not propagate up to the parent.

kitten:
Resolve #407 

I think the idea of adding "active teardowns" is pretty valuable. It may be required in the future for someone to actively cancel ongoing queries from exchanges. This is already necessary to correctly implement ending subscriptions.

This hence also adds missing features to the `useSubscription` hook, like `executeSubscription` and `args.pause`. It also ensures that the `handler` arg is a constant ref, since reexecuting subscriptions when it changes wouldn't make much sense.

Wonka had a slight bug in `onEnd`. The operator wasn't propagating `Close` signals correctly, so the use of `onEnd` would prevent unsubscriptions from propagating upwards.

### DIFF
--- a/package.json
+++ b/package.json
@@ -136,6 +136,6 @@
   },
   "dependencies": {
     "fast-json-stable-stringify": "^2.0.0",
-    "wonka": "^3.2.0"
+    "wonka": "^3.2.1"
   }
 }
--- a/src/client.ts
+++ b/src/client.ts
@@ -8,6 +8,7 @@ import {
   share,
   Source,
   take,
+  takeUntil,
   merge,
   interval,
   fromValue,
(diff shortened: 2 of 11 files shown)


kitten:
Turns out I just forgot to use `useImmediateEffect`, but it should be fine now :+1:

kitten:
Verified this to be working on CodeSandbox, but more testing needed. There's a prerelease on npm under the `next` tag with version `1.4.1-next.1`

PR #424, opened 2019-09-10
(feat) - implement separate client build

### LINKED ISSUE #423: RFC: Client specific packaging
In light of https://github.com/FormidableLabs/urql/issues/253 I would actually opt to take a similar approach as we’ve taken in [Preact](https://runpkg.com/?preact@10.0.0-rc.1/dist/preact.js) and that is to publish one urql package, but we just make a separate entry-point with non-react-specifics. In our case this would be the client and exchanges. In the normal urql package we can just use relative imports to the src of urql/client. This would imply no breaking changes at all

for example `import  Client from 'urql/client'`. This effectively solves our issues with non-React consumers needing a dependency on React or the package crashes, I don’t think a lot of vue users would like to add react to their dependencies just to make the client import work. This breaks because the entry-point imports the hooks and those require React. This will be tree-shaken out in the production build but yes...

The only concern I'm having is where to put the /utils, obviously `toSuspenseSource` is part of React but it's used in `client`. `createRequest` is a neutral one but used in `useRequest` and `useMutation` as well as in the `client` package. `withPromise` is only used in the `client` so that can easily be moved. The utils used in exchanges seem to be limited to only there.

Specific steps to produce this would be:

- make a sub-folder /client/src and move our client and exchanges there.
- update imports in /src to /client/src ones
- introduce build for /client

This encourages others to make thirth party packages like `urql-vue`. I think this is the best effort to having a non-react entry point while not resorting to a monorepo.

- #311
- #263

JoviDeCroock:
Fix #423

Will test this out with one of our examples.

Rearranged some imports to make it easier on our consumer


client/example/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "urql-client-example",
```
kitten: I'd rather put this example in the main examples folder. 😅just to have them all in one place. It may also be better to generate this client folder only on demand so it doesn't confuse anyone

rollup.config.js
```diff
@@ -140,17 +140,17 @@ const makePlugins = (isProduction = false) => [
...
   treeshake: {
     propertyReadSideEffects: false
   }
-};
+});
 
 export default [
   {
-    ...config,
+    ...makeConfig('./src/index.ts', './dist'),
```
kitten: Rollup has a feature to alias separate inputs to names that generate separate bundles that import one another as necessary.

This is good because it means that people can use both the core entry and the main-entry without duplicating code or running into global state issues

JoviDeCroock: But can you generate them at separate places then?

rollup.config.js
```diff
@@ -140,55 +140,99 @@ const makePlugins = (isProduction = false) => [
...
     propertyReadSideEffects: false
   }
-};
+});
 
 export default [
   {
-    ...config,
-    plugins: makePlugins(false),
+    ...makeConfig('./src/index.ts'),
```
kitten: We'll need to be able to use Rollup's code splitting feature. We'd use it by aliasing the core and main entries. It can then generate separate bundles that depend on one another.

This way people won't run into global state issues or duplicated code when they're using both, which is highly likely as we add more third party libraries

client/example/package.json
JoviDeCroock: How would you generate that on demand?

kitten: A prepublish script that creates the JSON and the folder which can probably just point to the dist folder (../dist/x.js)

package.json
```diff
@@ -21,14 +21,15 @@
...
+    "prepare-core": "node ./scripts/prepare-core.js",
     "clean": "rimraf ./dist ./node_modules/.cache",
     "build": "rollup -c rollup.config.js",
     "watch": "rollup -w -c rollup.config.js",
     "check": "tsc --noEmit",
     "test": "jest",
     "coverage": "jest --coverage",
     "lint": "eslint . --ext .ts,.tsx",
-    "prepublishOnly": "run-s clean test build",
+    "prepublishOnly": "run-s pepare-core clean test build",
```
kitten: Sorry another not 😅 let's add the core folder to the rimraf clean task and Run prepare after?

There's also a small typo in the tasks name here

scripts/prepare-core.js
```diff
@@ -0,0 +1,34 @@
...
+
+const pathToPkgJson = path.resolve(dir, 'package.json');
+if (fs.existsSync(pathToPkgJson)) {
+  fs.unlinkSync(pathToPkgJson);
+}
+
+const contents = JSON.stringify({
+  name: 'urql-core',
+  version: '1.5.0',
+  description:
```
kitten: May make sense to set the private field to true here so we don't actually publish this accidentally. Does it make sense to use the name "@urql/core" anyway and publish a stub package to block the name?

kitten: Lets maybe also copy some of the fields from the main package json?

JoviDeCroock:
I think I found a bug in rollup with making this, I have made an issue report for the time being https://github.com/rollup/rollup/issues/3109

PR #426, opened 2019-09-16
Replace fast-json-stable-stringify

kitten:
The problem with having `fast-json-stable-stringify` is that it's the only package that doesn't support ESM. This means that by using it we completely forgo the possibility of supporting ESM browser imports even for `urql/core` in the future.

This replaces the package with an embedded implementation that works the exact same. It's exported so that it can be reused in `@urql/exchange-graphcache`, however it works the same, so it's completely backwards compatible.

This also shaves off some bytes, since we're getting rid of some options and configuration in `fast-json-stable-stringify` that we don't need.

### DIFF
--- a/src/utils/stringifyVariables.ts
+++ b/src/utils/stringifyVariables.ts
@@ -1,0 +1,46 @@
+const seen = new Set();
+
+const stringify = (x: any): string => {
+  if (x === undefined) {
+    return '';
+  } else if (typeof x == 'number') {
+    return isFinite(x) ? '' + x : 'null';
+  } else if (typeof x !== 'object') {
+    return JSON.stringify(x);
+  } else if (x === null) {
+    return 'null';
+  }
+
+  let out = '[';
+  if (Array.isArray(x)) {
+    for (let i = 0, l = x.length; i < l; i++) {
+      if (i > 0) out += ',';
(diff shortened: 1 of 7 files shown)


src/utils/stringifyVariables.ts
```diff
@@ -0,0 +1,46 @@
...
+    return out + ']';
+  } else if (seen.has(x)) {
+    throw new TypeError('Converting circular structure to JSON');
+  }
+
+  const keys = Object.keys(x).sort();
+
+  seen.add(x);
+  for (let i = 0, l = keys.length; i < l; i++) {
+    const key = '' + keys[i];
```
kitten: divergence from `fast-json-stable-stringify`: We don't expect anything but strings and numbers here, so we can totally just cast to a string using `'' + x`.

src/utils/stringifyVariables.ts
```diff
@@ -0,0 +1,46 @@
+const seen = new Set();
+
+const stringify = (x: any): string => {
+  if (x === undefined) {
+    return '';
```
kitten: divergence from `fast-json-stable-stringify`: We don't expect to stringify `undefined`, so we can just return an empty string here, which makes the return type consistently a `string` and simplifies lines 19 and 34

PR #435, opened 2019-09-24
(chore) - don't destructure getInitialProps

kysley:
While spinning up my first nextjs project (experienced with react/urql), I ran into a eslint error [where `ctx` was not defined for `getInitialProps`](https://github.com/FormidableLabs/urql/compare/master...kysley:fix/example-nextjs-ssr#diff-d531117c918a8b6531456cf0f269e13bL17). 

**The HOC works as expected without this change, but was confusing for somebody who is new to nextjs.** 

The other changes are prettier formatting.


JoviDeCroock:
I've placed a similar PR since in this state getInitialProps does not work when used in a component

PR #436, opened 2019-09-25
(chore) - make CI run for external pull requests

JoviDeCroock:
Changed the trigger to pull_request this enables the CI for forks. push will only trigger for pushes/pull requests on the exact repository.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,5 @@
 name: CI
-on: [push]
+on: [pull_request]
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,5 +1,5 @@
 name: CI
-on: [push]
+on: [pull_request]
```
kitten: Didn't consider that 🤔 
But this means we'll have to narrow down this event to ignore stuff we don't care about, right? https://help.github.com/en/articles/events-that-trigger-workflows#pull-request-event-pull_request

JoviDeCroock: I assume this has to be opened or ready_for_review

PR #438, opened 2019-09-27
(feat) - add suspense flag to context

JoviDeCroock:


src/client.ts
```diff
@@ -194,7 +194,10 @@ export class Client {
       onEnd<OperationResult>(() => this.onOperationEnd(operation))
     );
 
-    return this.suspense && operationName !== 'subscription'
+    if (operation.context.suspense === undefined)
+      operation.context.suspense = this.suspense;
+
+    return operation.context.suspense && operationName !== 'subscription'
```
JoviDeCroock: to be honest, we could save some bytes by changing this to `operationName === 'query'`

src/client.ts
```diff
@@ -210,8 +213,9 @@ export class Client {
   query(
     query: DocumentNode | string,
     variables?: object,
-    context?: Partial<OperationContext>
+    context: Partial<OperationContext> = {}
   ): PromisifiedSource<OperationResult> {
+    context.suspense = false;
```
kitten: Maybe:

```
if (typeof context.suspense !== 'boolean') {
  context.suspense = false;
}
```

src/client.ts
kitten: Or:

```
operationName === 'query' &&
(operation.context.suspense === true || this.suspense)

PR #441, opened 2019-10-01
(docs) - fix file upload exchange example

lukasluecke:
add missing "variables." prefix to path map entries

### DIFF
--- a/docs/guides.md
+++ b/docs/guides.md
@@ -480,7 +480,7 @@ if (isFileUpload) {
   const map = {}
   let i = 0
   files.forEach(paths => {
-    map[++i] = paths
+    map[++i] = paths.map(path => `variables.${path}`)
   });
 
   fetchOptions.body.append('map', JSON.stringify(map));


lukasluecke:
I am also not sure about the 1-based index for the path map - in the spec they use 0-based in their examples. Not that it changes anything, was just wondering .

lukasluecke:
Also the example seems to have diverted from the "real" fetchExchange quite a bit already, would you want me to update the code as well, so it more closely resembles the current structure (and you can just copy / paste the "file upload" part into it)?

JoviDeCroock:
The variables should not be needed, I've implemented a codesandbox where you can see it at work.

lukasluecke:
@JoviDeCroock Really weird.. I can see your example working, but when I tried it in my project, the `paths` were missing the `variables.` part, and this change fixed it.. 🤔 I'll try to find out what's happening there.

JoviDeCroock:
Maybe your server implementation is using another spec?

lukasluecke:
Doesn't all this happen before anything is sent to the server though?

```
files.forEach(paths => {
    console.log(paths);
    map[++i] = paths;
});
```

In your codesandbox, this logs `["variables.file"]` as expected, but using the code from the guide (which seems to be pretty identical to your example) locally logs `["file"]`, which then doesn't work.

JoviDeCroock:
Maybe something changed in the extract-files implementation? It's correct for the current state of that plugin. I'm not at my pc but I'll test it tomorrow

lukasluecke:
@JoviDeCroock I found the difference - in your example you use `mutate({ variables: { file } })` to call the mutation - but the signature of the mutator returned by `useMutation` expects a call like `mutate({file})` - which is also what I'm using, and therefore adding the prefix is necessary.

```js
export declare const useMutation: <T = any, V = object>(query: string | DocumentNode) => [UseMutationState<T>, (variables?: V | undefined, context?: Partial<OperationContext> | undefined) => Promise<OperationResult<T>>];
```

Do you agree, or am I missing something here?

lukasluecke:
No worries, my implementation is working for me now, so there's no rush. Just wanted to let you know 🙂

JoviDeCroock:
Oh, then yours is good to merge fi we updater the sandbox

JoviDeCroock:
Thank you a lot for this PR @lukasluecke I'm sorry it took so long for me to get to it, updated the sandboxes and stuff! Thanks a lot!

PR #443, opened 2019-10-05
add delete and add todo to example 1

kandros:
This PR adds the feature to delete and add new todos to the example, it helps to better understand how urql caching works when changing adding/removing things


examples/1-getting-started/src/app/Home.tsx
```diff
@@ -32,28 +66,28 @@ export const Home: FC = () => {
...
+            {...todo}
+            loading={toggleTodoMutation.fetching}
+            disabled={
+              toggleTodoMutation.fetching || deleteTodoMutation.fetching
+            }
+          />
         ))}
       </ul>
     );
   }, [res]);
```
JoviDeCroock: isn't the linter complaining here because toggleTodoMutation is missing and so on?

PR #444, opened 2019-10-08
RFC: Allow fetching to be forced to true by exchanges

kitten:
This allows `OperationResult` to contain `fetching: true` which would force the `useQuery` hook to keep `fetching: true` although data has already been served by an exchange.

This is useful to indicate that more data is being fetched during `cache-and-network`. Hence the legacy `fetching` flag could still be reproduced by doing `oldFetching = fetching && !data`.

### DIFF
--- a/src/exchanges/cache.ts
+++ b/src/exchanges/cache.ts
@@ -57,16 +57,19 @@ export const cacheExchange: Exchange = ({ forward, client }) => {
       filter(op => !shouldSkip(op) && isOperationCached(op)),
       map(operation => {
         const cachedResult = resultCache.get(operation.key);
-        if (operation.context.requestPolicy === 'cache-and-network') {
-          reexecuteOperation(client, operation);
-        }
-
-        return {
+        const result: OperationResult = {
           ...cachedResult,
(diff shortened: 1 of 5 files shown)


kitten:
Yes, that's why this is still an RFC. I think we mainly have to validate:

- is this breaking? (If we use `fetching: true`, then yes)
- should this be a separate state, i.e. `stale`?
- how granular are we considering `cache-and-network` and partial background fetches to be?
- can we expect this state to expand even more?
- should there be a proactive flag as discussed previously where we observe using a subscription when the `Client` is fetching? (potentially expensive and complex)

lukasluecke:
In my opinion, having separate `stale` and `partial` flags would be great (as @JoviDeCroock suggested). If someone wanted to "merge" them back into `fetching`, they could just wrap that in a custom hook or something.

JoviDeCroock:
The nice thing about stale is that we could consolidate stale and partial to be the same thing since essentially they are both stale. That would actually be something I can live with since I don't want to overcomplicate our result a lot.

kitten:
Will be replaced by a new PR with `stale` based on `master`

PR #445, opened 2019-10-08
RFC: Configurable debugExchange

StevenLangbroek:
Hey folks!

We're running URQL on both Server & Client, and we'd like to be able to use `pino` for server logging. I haven't really spent much time on this yet, as I'd like to understand whether you'd be interested in this, and if this is the right approach. Would love to hear your thoughts.

This is a breaking change, as it changes the signature of `debugExchange` to be an exchange factory, rather than an exchange.

New API:

```ts
interface DebugExchangeOptions {
  // default: console.log
  logFn?: (...args) => void;
  logFormat?: (message, data) => void;
}

type CreateDebugExchange = (options: DebugExchangeOptions) => Exchange;
```

Thanks! ❤️

edit: just to clarify, this is absolutely not mergeable in current shape, this is just to have a discussion around it.

## Todo
- [ ] Try out in current project
- [ ] Rename to `logExchange`
- [ ] Configure `logExchange` to `debugExchange`

### DIFF
--- a/src/exchanges/debug.test.ts
+++ b/src/exchanges/debug.test.ts
@@ -29,7 +29,7 @@ beforeEach(() => {
 it('forwards query operations correctly', async () => {
   jest.spyOn(global.console, 'log').mockImplementation();
   const [ops$, next, complete] = input;
-  const exchange = debugExchange(exchangeArgs)(ops$);
+  const exchange = debugExchange({})(exchangeArgs)(ops$);
 
   publish(exchange);
   next(queryOperation);
@@ -52,7 +52,7 @@ describe('production', () => {
   it('is a noop in production', () => {
     const [ops$] = input;
 
-    debugExchange({
+    debugExchange({})({
       forward: ops => {
         expect(ops).toBe(ops$);
       },
--- a/src/exchanges/debug.ts
+++ b/src/exchanges/debug.ts
@@ -1,19 +1,26 @@
 import { pipe, tap } from 'wonka';
 import { Exchange } from '../types';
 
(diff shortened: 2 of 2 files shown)


StevenLangbroek:
Yeah I think it might make sense to rename this to `logExchange`, as it changes its use-case to logging on production, where `debug` kind of suggests it's for local development (for me anyway). Thoughts?

StevenLangbroek:
If we do that, `debugExchange` could just preconfigure `logExchange` 🤔

kitten:
I think that’s a good shout!

StevenLangbroek:
So, I can either implement changes as is, or I can develop it in our application and submit it back upstream when we've figured out whether this _definitely_ fits all our needs. If you're not in a rush I'd maybe suggest the latter.

kitten:
We’re definitely not in a rush 🙌 so do feel free to take your time ✨

StevenLangbroek:
🙇

ThisIsMissEm:
What if we do this:

1. Create a new logExchange package with the above API
2. Migrate debugExchange to it using defaults
3. In the next major version do a breaking change to debugExchange to accept options

StevenLangbroek:
@thisismissem im just gonna define it in line in our project for now rather than going through the whole package infra hoopla right now.

ThisIsMissEm:
@StevenLangbroek I guess we should close this then & create a ticket in our backlog to get back to contributing this unless someone beats us to it?

kitten:
Closing for now due to inactivity :)

PR #446, opened 2019-10-09
Allow passing a custom client instance.

### LINKED ISSUE #429: Multiple clients / instances?
In my application I have to talk to several (two) GraphQL endpoints, but I think URQL's Context is a singleton, so I can't create two Providers.

I need to use both endpoints in the same component (one to fetch external data, the other to apply mutations to another data source), so I can't isolate Providers.

How can I solve this?
Thank you!

fubhy:
Resolve #429

Hi there. Apparently this came up with another user too: https://github.com/FormidableLabs/urql/issues/429

I had the same need so instead of continuing to roll a custom useQuery hook I thought I'd write a quick PR.

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -11,16 +11,16 @@ export const Consumer = Context.Consumer;
 
 let hasWarnedAboutDefault = false;
 
-export const useClient = (client?: Client): Client => {
-  const context = useContext(Context);
+export const useClient = (overrideClient?: Client): Client => {
+  const client = useContext(Context);
 
-  if (!!client) {
-    return client;
+  if (overrideClient) {
+    return overrideClient;
   }
 
   if (
     process.env.NODE_ENV !== 'production' &&
-    context === defaultClient &&
+    client === defaultClient &&
     !hasWarnedAboutDefault
   ) {
     hasWarnedAboutDefault = true;
@@ -33,5 +33,5 @@ export const useClient = (client?: Client): Client => {
     );
   }
 
-  return context;
+  return client;
 };
--- a/src/context.ts
+++ b/src/context.ts
@@ -11,12 +11,16 @@ export const Consumer = Context.Consumer;
 
 let hasWarnedAboutDefault = false;
 
-export const useClient = (): Client => {
-  const client = useContext(Context);
+export const useClient = (client?: Client): Client => {
+  const context = useContext(Context);
+
+  if (!!client) {
+    return client;
+  }
 
   if (
     process.env.NODE_ENV !== 'production' &&
-    client === defaultClient &&
+    context === defaultClient &&
     !hasWarnedAboutDefault
   ) {
     hasWarnedAboutDefault = true;
@@ -29,5 +33,5 @@ export const useClient = (): Client => {
     );
   }
 
-  return client;
+  return context;
 };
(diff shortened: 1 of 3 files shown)


src/hooks/useQuery.ts
```diff
@@ -7,13 +7,15 @@ import { CombinedError, noop } from '../utils';
 import { useRequest } from './useRequest';
 import { useImmediateEffect } from './useImmediateEffect';
 import { useImmediateState } from './useImmediateState';
+import { Client } from 'src/client';
```
JoviDeCroock: This import won't work in build

src/context.ts
```diff
@@ -11,12 +11,16 @@ export const Consumer = Context.Consumer;
 
 let hasWarnedAboutDefault = false;
 
-export const useClient = (): Client => {
-  const client = useContext(Context);
+export const useClient = (client?: Client): Client => {
+  const context = useContext(Context);
```
JoviDeCroock: I would personally prefer keeping this named client

fubhy: I'd vote against that as the internal variable name inside this function is less important than the name of the parameter.

fubhy: If you feel strongly about this though then I can rename it to e.g. "customClient".

kitten: I’d say, either are not that important, but either aren’t really clear. I’d call the inner variable “client” and the argument “override” or “overrideClient” as it overrides the context’s one.

The usage is already clear since the hook itself is called “useClient”

fubhy: I still believe that "client" is the best name for the parameter but overrideClient works for me too.

src/context.ts
```diff
@@ -11,12 +11,16 @@ export const Consumer = Context.Consumer;
...
+
+  if (!!client) {
```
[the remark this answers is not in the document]
fubhy: Yeah, that's just my personal style. I like to coerce variables in conditions to booleans explicitly. Will remove if you prefer that.

src/hooks/useQuery.ts
```diff
@@ -33,7 +35,7 @@ export const useQuery = <T = any, V = object>(
   args: UseQueryArgs<V>
 ): UseQueryResponse<T> => {
   const unsubscribe = useRef(noop);
-  const client = useClient();
+  const client = useClient(args.client);
```
kitten: Oh before I forget 😅 let’s get this added to the other hooks as well.

`useSubscription` will just need another arg, just like `useQuery`. For `useMutation` we can probably use an optional second argument.

kitten:
Apologies for keeping this around for so long! I think for now we won't end up merging this, sorry! 😞

Mainly we're concerned that this will A) make not a lot of sense as a natural evolution of the hook, and B) be an irreversible API change that is not inline with React.

I'll post another comment in the issue.

ptbrowne:
I also stumbled upon this problem and ended up having to wrap my component into another urql.Provider to support this usecase. This is not ideal as it means the children hierarchy of the wrapped component has a different provider, and this might be unexpected for the children (in my case it's fine, the component is very small and its children do not use useQuery).

It would be much simpler if useQuery and consorts could accept a client parameter.

@kitten Can you expand on why this would not be natural for this hook and why it would not be inline with React ?

PR #447, opened 2019-10-10
Replace useImmediate* hooks with react-wonka

kitten:
This is intended to improve the stability of our hooks and ensure that we won't run into problems with concurrent mode in the future. At its core is a tiny library I've written, `react-wonka`. https://github.com/kitten/react-wonka

It takes care of transforming changing inputs to streams and using the given output stream to then return output values from its hooks, synchronously if possible, and with updates if not.

It is aware of concurrent mode and uses an internal counter to trigger updates, which seems to be popular these days 😓 

It's tests are more thorough than what we could write realistically for each of our hooks: https://github.com/kitten/react-wonka/blob/master/src/index.test.ts

In practice this may even make `useQuery` and `useSubscription` look a little nicer, if someone is open to help me tweak the implementation a little. Doesn't look 100% pretty yet 😅 

In practice, our bundle-size only changes by `+0.02kB min+gzip` (this branch compared to the last release) when I'm including `react-wonka` in the bundle, which seems better for a fair comparison.

This should also improve our handling of suspense. Our `useImmediateEffect` hook runs normal effects after the initial mount. When we update a hook we'd right now  throw a suspense-promise in a normal effect, which is not allowed.


src/hooks/useMutation.ts
```diff
@@ -25,7 +24,8 @@ export const useMutation = <T = any, V = object>(
   query: DocumentNode | string
 ): UseMutationResponse<T, V> => {
   const client = useClient();
-  const [state, setState] = useImmediateState<UseMutationState<T>>({
+
+  const [state, setState] = useState<UseMutationState<T>>({
```
kitten: Small note on this change; This is not a breaking change. Since mutations are realistically run only after a component has mounted, `useImmediateState` was never doing anything here

PR #451, opened 2019-10-12
Fix react-wonka update bug that occurs with Context Providers

### LINKED ISSUE #450: Updating `res.fetching` doesn't trigger Context consumer updates
**Describe the bug**

Hi folks! Big fan of Urql here. In the update to `1.6.0`, interaction between `useQuery` and the React Context API stopped working in the same way. Here's the relevant code (pulled out from my own project, codesandbox link below):

```ts
// viewer-context.tsx

export const ViewerContext = createContext<ViewerContextValue>({
  fetching: true,
  viewer: null
})

export function ViewerProvider({ children }: { children: ReactNode }) {
  const [res] = useQuery<{ viewer: User | null }>({
    query: viewerQuery
  })

  const value = {
    fetching: res.fetching,
    viewer: res.data && res.data.viewer ? res.data.viewer : null
  }

  return (
    <ViewerContext.Provider value={value}>{children}</ViewerContext.Provider>
  )
}
```

```tsx
// Main.tsx, which consumes that context using `useContext`

export default function Main() {
  const { viewer, fetching } = useContext(ViewerContext)
  console.log(fetching)

  // Still loading
  if (fetching) return null

  return <h1>Hi!</h1>
}
```

I'm able to reliably reproduce differing behavior between `1.5.1` and `1.6.0`. In the former, when `res.fetching` changes, consumers of the context are updated (the `console.log` in `Main.tsx` follows the updates). That is not the behavior I'm seeing in `1.6.0` - the Context API [checks for object equality using `Object.is`](https://reactjs.org/docs/context.html#contextprovider), so I'm guessing something has changed there (presumably in #447)? That doesn't really make sense, but updating `urql` is the only change I've made while testing this out.

In setting up the below codesandbox, I've realized that maybe using a Context for not repeating the same query is unnecessary thanks to `urql`'s caching - I'd love your opinion there, but either way I'm bringing this up as a likely unintentional regression in `1.6.0`!

**Steps to reproduce**

A codesandbox link (thanks for the template below!): https://codesandbox.io/embed/urql-issue-template-client-vj73c

**Expected behavior**

I'd expect consumers of a Context that sets a value using `res.fetching` to correctly update consumers of the context.

**Actual behavior**

The value is updated in the provider itself, but not in the consumers. You can toggle between `1.5.1` and `1.6.0` in the Codesandbox link to see it in action!

Thanks! Let me know how I can help ❤️

kitten:
Fix #450 
See: https://github.com/kitten/react-wonka/pull/1

This upgrade `react-wonka` to fix an issue where updates weren't propagated to context providers. This seems to be closely related to some kind of React internals that cause these kinds of updates to be run twice (presumably in two separate Fiber phases) to update the context consumers.

This also now externalises `react-wonka`, in case we need to push out another patch.

### DIFF
--- a/package.json
+++ b/package.json
@@ -133,7 +133,7 @@
     "react-dom": ">= 16.8.0"
   },
   "dependencies": {
-    "react-wonka": "^1.0.1",
+    "react-wonka": "^1.0.2",
     "wonka": "^3.2.1"
   }
 }
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -16,8 +16,6 @@ if (pkgInfo.peerDependencies)
 if (pkgInfo.dependencies)
   external.push(...Object.keys(pkgInfo.dependencies));
 
-external = external.filter(x => x !== 'react-wonka');
-
 const externalPredicate = new RegExp(`^(${external.join('|')})($|/)`);
 const externalTest = id => {
   if (id === 'babel-plugin-transform-async-to-promises/helpers') {
--- a/yarn.lock
+++ b/yarn.lock
@@ -4588,10 +4588,10 @@ react-test-renderer@^16.0.0-0, react-test-renderer@^16.10.2:
     react-is "^16.8.6"
     scheduler "^0.16.2"
 
-react-wonka@^1.0.1:
-  version "1.0.1"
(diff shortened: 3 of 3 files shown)


kitten:
Retroactive review should be required later

PR #459, opened 2019-10-29
Fix SSR entering infinite loop when multiple query hooks update

### LINKED ISSUE #458: Imperative Calls are throwing error
I wanted to implement the imperative way to make queries, but I keep on getting the following error:
```
index.js:1 Uncaught (in promise) Error: The error you provided does not contain a stack trace.
```

So I thought maybe I am doing something wrong, tested the example from the repo and the same happens. I also tested it in various browsers, the same behavior everywhere.

kitten:
Fix #458 

This bug can be replicated by:

- Having a `useQuery` hook in a component
- Either: having a second with the same operation or another hook in the
  component
- Triggering an update causing the component to rerender

This will cause the component to rerender in SSR.
The `useCallback` is a noop so some dependencies will trigger,
and this will subsequently trigger another update, which then
enters an infinite loop.

This is because `useCallback` is a noop in SSR: https://github.com/facebook/react/blob/0f64703edf5970b12a878ea3b5e1e30ef1d71c74/packages/react-dom/src/server/ReactPartialRendererHooks.js#L448-L454


parkerziegler:
Wow that's _wild_. I was actually hitting this yesterday working on a small NextJS app and found that changing the dependency array arbitrarily on a `useCallback` didn't appear to be doing anything. Great find!

kitten:
@parkerziegler it really is and really was hard to pin down. I'm totally aware that `useMemo` and `useCallback` make no strong guarantees, but I was totally expecting `useCallback` to still work in most simple cases.

parkerziegler:
:thonking-hard: 😅

PR #461, opened 2019-11-05
Add more information to main README

kitten:
Our docs right now aren't as good and navigable as they could be, but at the same time it's very easy to get started with `urql`! A lot of people probably dive in just by reading the basics and diving straight in, but that means that they may miss the subtle API additions that we have to make their lives easier (like `requestPolicy`, `pause`, normalized caching, etc)

This aims to add all relevant information to the main README. Each section has been written to not go too much into detail, but provide a thorough but brief overview of all of our features (minus subscriptions).

It highlights how the document cache works, what our hooks look like, how to set up normalized caching, and how to set up SSR.

I've also added diagrams that we've created for our React Advanced workshop, since I already have them and they may add some more clarity but also eye-candy 🍬


README.md
```diff
@@ -39,6 +39,7 @@
 - ⚙️ Fully **customisable** behaviour [via "exchanges"](#-add-on-exchanges)
 - 🗂 Logical but simple default behaviour and document caching
 - ⚛️ Minimal React components and hooks
+- 🌱 Normalized caching via [`@urql/exchange-graphcache`](https://github.com/FormidableLabs/urql-exchange-graphcache)
```
parkerziegler: I've been wondering whether or not we want to have an "Ecosystem" section where we mention the normalized cache, the dev tools, `next-urql`, and `reason-urql`. I could also see those just becoming list items here, i.e.

```md
- ⚡️ SSR support via [`next-urql`](https://github.com/FormidableLabs/next-urql).
- DevTools via [`urql-devtools-exchange`](https://github.com/FormidableLabs/urql-devtools-exchange).
- Reason support via [`reason-urql`]https://github.com/FormidableLabs/reason-urql).
```

kitten: Yea, we can probably change up this section: https://github.com/FormidableLabs/urql/blob/chore/improved-readme/README.md#-add-on-exchanges

Not sure what that'd look like, but shouldn't be too hard to figure out.

kitten:
I just realised I'm missing two sections still: Suspense and Server-side Rendering

README.md
```diff
@@ -59,35 +60,26 @@ Some of the available exchanges that extend `urql` are listed below in the ["Add
...
-  </Provider>,
-  document.body
-);
-```
+The [`useQuery` hook](https://formidable.com/open-source/urql/docs/api/#usequery-hook) is
+used to send GraphQL queries and will provide GraphQL results from your API.
 
-This allows you to use the `useQuery` hook in your function component to
-fetch data from your server:
+When you're using `useQuery` it'll accept a configuration object that may contain a `query` and `variables`.
```
parkerziegler:
```suggestion
When you're using `useQuery` it'll accept a configuration object that may contain keys for `query` and `variables`.
```

README.md
```diff
@@ -97,24 +89,174 @@ const YourComponent = () => {
     query: `{ todos { id } }`,
   });
 
-  const { fetching, data } = result;
-  return fetching ? <Loading /> : <List data={data.todos} />;
+  if (result.error) return 'Oh no!';
+  if (result.fetching) return 'Loading...';
```
parkerziegler: I think I tend to prefer when docs show some actual components rendered in situations like this than just strings, i.e.:

```typescript
if (result.error) {
   return <Error message={result.error.message} />;
}

if (result.fetching || !result.data) {
    return <Loading />;
}
```

The example above also handles the `fetching: false` but `data` false-y case – is that still an issue we're hitting?

kitten: Good point! False-y data shouldn’t be an issue as long as the API can be trusted to not have nullable data, so that’s purely schema-dependent.

In all usual cases it’s either: fetching, error, or data, where data and error may co-exist depending on whether it’s partial GraphQL errors that don’t affect the entire query

PR #465, opened 2019-11-06
remove confusing example

JoviDeCroock:
We don't allow for passing variables to `executeQuery` as far as I know, these are only made through the `useQuery` and `useRequest`

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,9 +143,7 @@ Mutations will only execute once the `executeMutation` method is called with som
 The `useQuery` hook can actually be used similarly. The hook also provides an `executeQuery`
 function that can be called imperatively to change what query the hook is running.
 
-Unlike the `useMutation` hook, the `useQuery`'s `executeQuery` function actually accepts more than just variables.
-Instead it accepts a full object of all of `useQuery`'s normal options, so it can be used to change the entire query, variables, or
-other options that the hook uses to run the query.
+Unlike the `useMutation` hook, the `useQuery`'s `executeQuery` function accepts an `OperationContext` as the argument, this allows you to for example override the `requestPolicy` or even the `fetchOptions`.
 
 ```js
 const [result, executeQuery] = useQuery({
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -154,7 +152,7 @@ const [result, executeQuery] = useQuery({
 });
 
 // executeQuery can trigger queries and override options
-const update = () => executeQuery({ variables: { sort: 'by-text' } });
+const update = () => executeQuery({ requestPolicy: 'network-only' });
```
kitten: This example is already used at the end of this section :)

README.md
```diff
@@ -143,9 +143,7 @@ Mutations will only execute once the `executeMutation` method is called with som
 The `useQuery` hook can actually be used similarly. The hook also provides an `executeQuery`
 function that can be called imperatively to change what query the hook is running.
 
-Unlike the `useMutation` hook, the `useQuery`'s `executeQuery` function actually accepts more than just variables.
-Instead it accepts a full object of all of `useQuery`'s normal options, so it can be used to change the entire query, variables, or
-other options that the hook uses to run the query.
+Unlike the `useMutation` hook, the `useQuery`'s `executeQuery` function accepts an `OperationContext` as the argument, this allows you to for example override the `requestPolicy` or even the `fetchOptions`.
```
kitten: This is a bit technical due to the mention of “OperationContext”. We also mention “requestPolicy” somewhere else. We also can’t use “fetchOptions” because those also don’t have been mentioned and aren’t self-explanatory 😅

Let’s maybe just keep it more as it was before, but remove the example, since other examples do follow here

JoviDeCroock: what do you suggest using instead? Maybe a link to the docs or?

kitten: I think it’s good as it was before, but as there’s a link to the docs already, maybe we can write some words about the context then and inline link it to the API docs?

JoviDeCroock: So just remove the example and elaborate on context?

PR #467, opened 2019-11-13
exchange authoring guide clarifications for Urql noobs

a-laughlin:
These updates would have helped me when learning to write exchanges.  They add context for the noop exchange, comment the examples' anonymous ExchangeIO function, explain what happens with forwarded + returned streams, where operationResults get created, and how request/response relate to forward/return.

### DIFF
--- a/docs/guides.md
+++ b/docs/guides.md
@@ -82,23 +82,58 @@ For reference, this is a basic template for an exchange:
 
 ```js
 const noopExchange = ({ client, forward }) => {
-  return operations$ => {
-    return forward(operations$);
+  return operation$ => {
+    // <-- The ExchangeIO function
+    const operationResult$ = forward(operations$);
+    return operationResult$;
   };
 };
 ```
 
-In this form the exchange doesn't do anything yet.
-When you create a client and pass it an array of exchanges, all exchanges
-will be composed together into a single one. They will each be called with
-an object that contains the `client` itself, and a `forward` function which
-is the next `ExchangeIO` function.
+By convention, variables ending with `$` are streams. Each exchange's `ExchangeIO` function receives an [`Operation`](https://formidable.com/open-source/urql/docs/api/#operation-type) stream (`operation$`) and must return an [`OperationResult`](https://formidable.com/open-source/urql/docs/api/#operationresult-type) stream (`operationResult$`).
 
-The `ExchangeIO` function is what each exchange returns. This is a function
-that receives the stream of operations and must return a stream of results.
+### Exchange Composition
 
-So our `noopExchange` is the minimal template that fulfils this requirement.
-It just receives the operations stream and passes it on to `forward`.
+When you create a client and pass it an array of exchanges, Urql composes them left-to-right into a single exchange using [`composeExchanges`](https://github.com/FormidableLabs/urql/blob/master/src/exchanges/compose.ts). Here's the noopExchange in context:
+
+```js
+import { Client, dedupeExchange, fetchExchange } from 'urql';
+
+const noopExchange = ({ client, forward }) => {
+  // ExchangeIO function receives Operation object stream from dedupeExchange
+  return operation$ => {
+    // noopExchange forwards Operation object stream to fetchExchange,
+    // and receives a stream with an OperationResult object created by fetchExchange
+    const operationResult$ = forward(operations$);
+    // then returns the OperationResult stream to dedupeExchange
+    return operationResult$;
+  };
+};
+
+const client = new Client({
+  exchanges: [dedupeExchange, noopExchange, fetchExchange],
+});
+```
+
+### Request and Response Handling
+
+Interacting with a request occurs before the forward() call. Interacting with a response occurs after the forward call. Here's what the noopExchange looks like with explicit request and reponse handlers.
+
+```js
+const handleRequest = operation$ => {
+  /*do stuff*/ return operation$;
+};
+const handleResponse = operationResult$ => {
+  /*do stuff*/ return operationResult$;
+};
+
+const noopExchange = ({ client, forward }) => {
+  return operations$ => {
+    // <-- The ExchangeIO function
+    return handleResponse(forward(handleRequest(operations$)));
+  };
+};
+```
 
 ### One operations stream only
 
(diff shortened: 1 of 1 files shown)


docs/guides.md
```diff
@@ -78,27 +78,58 @@ Before we jump into writing some exchanges, there are a couple of
...
+  exchanges: [dedupeExchange, noopExchange, fetchExchange],
+});
 ```
 
-In this form the exchange doesn't do anything yet.
-When you create a client and pass it an array of exchanges, all exchanges
-will be composed together into a single one. They will each be called with
-an object that contains the `client` itself, and a `forward` function which
-is the next `ExchangeIO` function.
+#### Forward + Return = Request + Response
```
kitten: I think this section is overall more confusing than helpful. It takes about handlers and ordering, but that’s potentially confusing and factually a little off since we’re dealing with streams.

The example isn’t doing anything so this example isn’t too helpful.
Let’s maybe remove this section since it doesn’t add more explanation, even compared with the deleted section 🤔

docs/guides.md
```diff
@@ -78,27 +78,58 @@ Before we jump into writing some exchanges, there are a couple of
...
 
+### Forward Operation Streams. Return OperationResult Streams.
+
 For reference, this is a basic template for an exchange:
 
 ```js
 const noopExchange = ({ client, forward }) => {
-  return operations$ => {
-    return forward(operations$);
+  return operation$ => { // <-- The ExchangeIO function
```
kitten: I think these comments are actually not indicating a lot of information if we’re repeating them too often, so let’s maybe annotate “Exchange” and “ExchangeIO” separately

docs/guides.md
```diff
@@ -78,27 +78,58 @@ Before we jump into writing some exchanges, there are a couple of
...
+    return operationResult$;
+  };
+};
+```
+
+By convention, variables ending with `$` are streams. Each exchange's `ExchangeIO` function receives a stream containing an [`Operation`](https://formidable.com/open-source/urql/docs/api/#operation-type) object (`operation$`) and must return a stream of [`OperationResult`](https://formidable.com/open-source/urql/docs/api/#operationresult-type) object(s)(`operationResult$`).
+
+#### Forward and Return Composition
+
+When you create a client and pass it an array of exchanges, Urql composes them left-to-right into a single exchange using [`composeExchanges`](https://github.com/FormidableLabs/urql/blob/master/src/exchanges/compose.ts). `composeExchanges` orchestrates stream forward and return between exchanges. Here's the noopExchange in context:
```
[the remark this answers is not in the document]
a-laughlin: I wondered about `urql` formatting.  Great to know!

> We also always code highlight variables.
Missed that one.  Thanks for catching it!

docs/guides.md
a-laughlin: I agree that this section is "factually off".  It especially wouldn't make sense without fetchExchange on the end making a network request.  I also haven't thought of a better way to address the underlying reason for adding this section yet;  Forward's mismatching semantics and behavior were confusing at the beginning.  The word forward doesn't imply returning a different shape, or returning at all. Calling `forward` is (loosely) like passing a chicken  to `promise.then`, and receving a [turducken](https://en.wikipedia.org/wiki/Turducken) in the next call. That unexpected behavior was confusing when learning.  `getResults`, or `pullResults` semantics would more closely match forward's actual behavior, making it easier to learn.

That said, I didn't want to suggest a name change to a broadly used core function while still learning, hence adding this section.

a-laughlin: Maybe the example above, showing forward in context, with comments, is sufficient for now, so I'll delete this section.  

If you ever want to deprecate the name `forward`, you'll have my full support. 😁

docs/guides.md
a-laughlin: Agreed that they aren't adding information.  I think annotating "Exchange" is unnecessary since exchanges are called fooExchange by convention.

"ExchangeIO" annotations repeat a few times (not the whole doc) for three reasons:
- it is always anonymous by convention
- learners are learning enough new concepts in this section for their working memory to drop the first instances before retaining it
- Repetition facilitates memory retention

docs/guides.md
kitten: When explaining this, what’s missing here is the hint that `forward` is just the next exchange’s `ExchangeIO` function hence the name.

You pass it a stream of operations you want to _forward_ and you receive the next exchange’s results stream in return.

Maybe a simple explanation like this can go a long way at explaining this.

We also have diagrams from a workshop that I could modify to be included here 🤔

- https://advanced-graphql-workshop.netlify.com/55 (This is very different from urql but can easily be modified to convey a similar amount of information)
- https://advanced-graphql-workshop.netlify.com/65
- https://advanced-graphql-workshop.netlify.com/64

docs/guides.md
kitten: Fair, that’s a great point 👍 let’s keep it!

docs/guides.md
a-laughlin: I think saying "`forward` calls the next exchange's `ExchangeIO` function" is accurate enough early in the exchange learning process.

"Accurate enough" because my brain choked briefly on the sentence "forward is just the next exchange’s ExchangeIO function".  There are many lower level concepts to parse when reading it as a learner, like an exchange being a function, and each argument in that function, and argument type, and which ones are relevant to the sentence, and operation order, and stream order, and return values.  That's just unpacking the sentence itself.  Saying `forward` _is_ `ExchangeIO` adds the cognitive load of dual names for the same concept.  That's unnecessarily confusing when already remembering the other stuff.  Saying "`forward` calls the next exchange's `ExchangeIO` function" is cognitively easier to keep straight, hence accurate enough.

I can add "`forward` calls the next exchange's `ExchangeIO` function" to the contextualized `noopExchange` example if you're cool with it.

Or... you could always rename the argument `nextExchangeIO`. 😁

Like the video!

a-laughlin: Contextual noopExchange example expanded.

Some gaps in my own knowledge stood out in the process of explaining pipeline's ExchangIO function role, so I added each step in the process to hopefully prevent others' gaps in understanding.

PR #471, opened 2019-11-18
Auto-population of mutation bodies

### LINKED ISSUE #472: RFC: Auto populate mutation body
It would be handy if mutation bodies could be auto populated based on the set of currently watched queries. As inspired by [Relay](https://relay.dev/docs/en/classic-guides-mutations#a-complete-example).

## How do updates currently work?

```gql
type Query {
  todos: [Todo]
}
	
type Mutation {
  toggleTodo(id: ID!): Todo
  addTodo(text: String!): Todo
  deleteTodo(id: ID!): Todo
}

type Todo {
  id: ID
  text: String
  complete: Boolean
}
```
Given the schema above. We could request the `todos` using the following query.

```gql
query {
  todos {
	__typename
    id
    text
    complete
  }
}
```

Whilst a component is mounted with this associated query, it is identified as watched. This means any changes to fields or selections within it will trigger reactive UI updates whenever its data changes.

---

Mutations that affect it will also trigger updates so long as the entity can matched to one in the cache.

```gql
mutation($id: ID!) {
  toggleTodo(id: $id) {
	__typename
    id
    complete
  }
}
```

A mutation body should include any fields that may change as a result of the mutation. Ideally these fields at least match those that have already been requested. We can enforce this by spreading common fragments, however, this can become chaotic as the schema and application grow.

The following works quite well but can result in overfetching since not all requested fragments may be mounted but they are now deeply coupled to component fragments.

```gql
fragment Todo on Todo {
	__typename
    id
	text
    complete
}

query {
  todos {
    ...Todo
  }
}

mutation($id: ID!) {
  toggleTodo(id: $id) {
	...Todo
  }
}
```

# Proposal

Instead of declaring the mutation body explicitly, we can infer the body from the currently watched queries. Using schema introspection we can infer that for the following watched query a `Todo` fragment exists.

```diff
# This query belongs to a currently mounted component
query {
  todos {
-	__typename
-    id
-	text
-   complete
}

fragment Todo on Todo {
+	__typename
+    id
+	text
+   complete
}

```

We can also infer that a mutation may require a Todo fragment, however we cannot infer where in its selection that may be. Using a directive we can be explicit about where we'd like to auto populate.

```diff
mutation($id: ID!) {
-  toggleTodo(id: $id) @populate
}

# Replaces the scalar with a selection set populated by the watched fragment
mutation($id: ID!) {
+  toggleTodo(id: $id) {
+	...Todo
+  }
}
```

As queries are mounted and unmounted the mutation body dynamically updates to accommodate the immediate requirements of the UI.

# What do we need to do this?

Since this behaviour is additive to urql it would be best served as an opt-in exchange. We'll need access to the following in order to achieve the above.

- Knowledge of currently watched queries via mount/teardown events.
- Handling for the custom `@populate` directive. These need to be replace by selections at runtime.
- Type info for the annotated selection set or field ideally via introspection since we'll reliably need the typename in all cases

andyrichardson:
Fix #472

## Todo
 - [x] Create fragment definitions instead of inline fragments
 - [x] Handle/consider issue where mutation is called but no "prior knowledge" is available
 - [x] Allow populating fields with existing selection sets
 - [x] Add support for interfaces
 - [x] Only add fragments to the query when they are required (currently all user-provided fragments are appended)
 - [x] Add support for recurqling (pronounced _re-cur-cul_)

## Example
Kapture 2019-11-18 at 17 32 07


src/exchanges/mutateBody.ts
```diff
@@ -0,0 +1,159 @@
...
+  DocumentNode,
+  print,
+  ASTNode,
+  buildClientSchema,
+  visitWithTypeInfo,
+  TypeInfo,
+  FragmentDefinitionNode,
+} from 'graphql';
+import { parse, visit, NonNullTypeNode } from 'graphql';
+import { type } from 'os';
```
JoviDeCroock: I think this is not intended

src/exchanges/mutateBody.ts
```diff
@@ -0,0 +1,159 @@
...
+import { object } from 'prop-types';
```
JoviDeCroock: This is unintended I assume

andyrichardson: Yeah it's in a messy state - not ready for review/merge yet :)

kitten: Prefixed it with `WIP` so that noone finds this and assumes it's about to be imminently merged in :)

imranolas:
#472 The rationale behind this for anyone that is interested

andyrichardson:
> I like the notion this exchange introduces, I'm just a bit on the ropes if it belongs inside the repository, the thing that makes me reluctant is quite superficial since if unused this will be tree-shaken off _but_ in most bundle-analyzers like bundlePhobia this will add a lot of size with the 2 imported methods.

I hear you - but I don't think we should allow stats from services like bundlephobia to influence architectural decisions 😞 

I'm not sure where else it would make sense to serve this exchange and this is an exchange which I can see most Urql users wanting to use! Did you have any alternatives in mind?

kitten:
@andyrichardson From an architectural standpoint; this is super useful in combination with `@urql/exchange-graphcache`, so if that's separate, then having this separate makes sense, but there's some flexibility in that:

- We can add it to `@urql/exchange-graphcache`. That's good since they'd be used together but bad because it's technically a separate thing and shouldn't be "hidden"
- We can make `@urql/exchange-graphcache` into a monorepo that has some kind of name that implies that it makes `urql` into a full "fat" client. I do like that, because then we'd have a repo that basically says: "Hey, this is what you need if you're replacing Relay/Apollo"
- We can make it a separate `@urql/exchange-populate` repo (don't mind the name)
- We could move to a "fat" monorepo that includes the above and graphcache

andyrichardson:
I think grouping it with graphcache isn't a bad shout if the bundle size impact is too large. 

While it can be used (and we should encourage use) with any cache, it doesn't make much sense to use this with the default cache.

kitten:
Here's a discussion thread on Spectrum, so we can discuss this publicly but with a lot more messages. I suspect this is going to be a longer discussion 😆  https://spectrum.chat/urql/general/core-discussion-monorepo-more-repos~51619a7d-dedd-4b35-99ab-0085cf15e233

src/exchanges/populate.ts
```diff
@@ -0,0 +1,331 @@
...
+export const populateExchange = ({
+  schema: ogSchema,
+}: ExchangeArgs): Exchange => ({ forward }) => {
+  const schema = buildClientSchema(ogSchema);
+  /** List of operation keys that have already been parsed. */
+  const parsedOperations = new Set<number>();
+  /** List of operation keys that have not been torn down. */
+  const activeOperations = new Set<number>();
+  /** Collection of fragments used by the user. */
+  let userFragments: UserFragmentMap = {};
```
kitten: `userFragments` and `typeFragments` should likely use `Object.create(null)` so we can ensure that weird fragment names don't mess things up (like `fragment prototype` or `fragment hasOwnProperty` 🙈 )

src/exchanges/populate.ts
```diff
@@ -0,0 +1,331 @@
...
+  /** Collection of type fragments. */
+  let typeFragments: TypeFragmentMap = {};
+
+  /** Handle mutation and inject selections + fragments. */
+  const handleIncomingMutation = (op: Operation) => {
+    if (op.operationName !== 'mutation') {
+      return op;
+    }
+
+    const activeSelections = Object.entries(typeFragments).reduce(
```
kitten: If we use `Object.entries` it'll need to be polyfilled, but that is probably not really desirable since we don't need to take on a polyfill for this. (This should be auto-poyfilled, but either way we can make due without)

So let's either maybe just use good ol' `Object.keys`. Also nit: The spreading isn't necessary in a reduce. Of course it reads more nicely but we're cleaning up Graphcache to avoid unnecessary allocations to alleviate GC pressure, so it'd be awesome to carry that over to new code in `urql` as well :)

src/exchanges/populate.ts
```diff
@@ -0,0 +1,331 @@
...
+        typeFragments: activeSelections,
+        userFragments: userFragments,
+        query: op.query,
+      }),
+    };
+  };
+
+  /** Handle query and extract fragments. */
+  const handleIncomingQuery = ({ key, operationName, query }: Operation) => {
+    activeOperations.add(key);
```
kitten: Is this supposed to be below the escape hatch below? seems a little extreme to track all operations here if don't need all of them (in the case of unknown operations that may become an issue)

Mainly I'm concerned that the `activeOperations.delete(key)` for teardowns is in a different place, which makes it hard to track the above and the add/delete for teardowns

src/exchanges/populate.ts
```diff
@@ -0,0 +1,331 @@
...
+      (state, fragment) => ({
+        ...state,
+        [fragment.name.value]: fragment,
+      }),
+      userFragments
+    );
+
+    typeFragments = newSelections.reduce((state, { selections, type }) => {
+      const current = state[type] || [];
+      const entry: TypeFragment = {
```
kitten: Code Golfing question cc @JoviDeCroock 

We may be able to reduce the size here by using the GraphQL constructors (i.e. `new NameNode`, etc). They're probably already included so we may as well try to use them and see whether it reduces the size here a little

src/exchanges/populate.ts
```diff
@@ -0,0 +1,331 @@
...
+}
+
+/** Gets typed selection sets and fragments from query */
+export const extractSelectionsFromQuery = ({
+  schema,
+  query,
+}: MakeFragmentsFromQueryArg) => {
+  let selections: { selections: SelectionSetNode; type: string }[] = [];
+  let fragments: FragmentDefinitionNode[] = [];
+  const typeInfo = new TypeInfo(schema);
```
kitten: Seems like this instance should maybe be preserved and only be instantiated once?

src/exchanges/populate.ts
```diff
@@ -0,0 +1,331 @@
...
+      Field: {
+        enter: node => {
+          if (
+            !node.directives ||
+            !node.directives.find(d => d.name.value === 'populate')
+          ) {
+            return;
+          }
+
+          const directives = node.directives.filter(
```
kitten: nit: you could avoid the double iteration (find + filter) by checking `!node.directives` once and checking `directives.length < node.directives.length` below instead 👀

src/exchanges/populate.ts
```diff
@@ -0,0 +1,331 @@
...
+            return [
+              ...p,
+              ...typeFragments[t.name].map(({ fragment }) => {
+                // Add used fragment for insertion at Document node
+                getUsedFragments(fragment).forEach(f =>
+                  requiredUserFragments.add(userFragments[f])
+                );
+
+                // Add fragment for insertion at Document node
+                additionalFragments = {
```
kitten: There's a lot of local spreading/copying in this file (inside reduces, here as well, etc). Generally I don't mind that at all, but it all adds up in terms of extra-allocations which we can avoid to make this run nice and fast :)

src/exchanges/populate.ts
```diff
@@ -0,0 +1,331 @@
...
+          };
+        },
+      },
+      Document: {
+        leave: node => {
+          return {
+            ...node,
+            definitions: [
+              ...node.definitions,
+              ...Object.values(additionalFragments),
```
kitten: similarly to `Object.entries` we'll want to avoid `Object.values`

src/exchanges/populate.ts
```diff
@@ -0,0 +1,331 @@
...
+    })
+  );
+};
+
+/** Get all possible types for node with TypeInfo. */
+const getTypes = (schema: GraphQLSchema, typeInfo: TypeInfo) => {
+  const type = typeInfo.getType();
+
+  if (!type || !('ofType' in type)) {
+    console.warn('PopulateExchange: Unsupported type at populate decorator.');
```
kitten: I think this is a little cryptic 🤔  can we make this message clearer somehow? I'm not quite sure what it's trying to tell the user about. Also, do we want to log this out identically multiple times?

src/exchanges/populate.ts
```diff
@@ -0,0 +1,331 @@
...
+      names = [...names, f.name.value];
+    },
+  });
+
+  return names;
+};
+
+const getType = (t: TypeInfo) => {
+  const type = t.getType() as any;
+  return type.ofType || type.toString();
```
kitten: This seems interesting; I get that this could be almost anything (https://github.com/graphql/graphql-js/blob/c7f11f3b95367247cc4ce3adb4c24313d2943c5b/src/type/definition.d.ts#L100-L114) but what are we trying to get here with `toString()`?

src/exchanges/populate.ts
andyrichardson: Just checked this out. If we're going to go this route for security reasons we would need to handle this on a project wide scope.

This will be an issue on every use of spreads. I'll see if there are any babel/rollup plugins about which cover this.

Screenshot 2019-11-20 at 15 30 32

kitten: We would have to take care of this every time we use an object as a record, yea, but (as far as I'm aware) we've taken care of that everywhere (?)
We rarely use objects as records anyway in `urql` (there are a lot of cases in Graphcache that do use `Object.create(null)`) but often we've also just gone for `Map` wherever possible

src/exchanges/populate.ts
JoviDeCroock: This should be the case yes, should also enforce some more streamlined objects shapes which v8 loves

andyrichardson: Ooft this would be nice. I tried this at first but got errors that XyzNode is a type.

Will play around.

src/exchanges/populate.ts
andyrichardson: Good catch - fixed

src/exchanges/populate.ts
andyrichardson: TypeInfo is stateful and is used while traversing the schema alongside a given query/mutation. Not too sure how else we would go about this - any examples?

src/exchanges/populate.ts
andyrichardson: For real, we need to be prepared for nodes which have 1000s+ directives! 😉 

But more seriously, this is probably a good time to mention that there are some potential performance improvements that can be done (e.g. `getUsedFragments` which calls `visit` inside of an existing `visit` (multiple traversals of the same subtree).

If my thinking is correct, this would be a time complexity of `O(2n)` which should be negligible but I'd like to see this in practice on larger structures before resting on that assumption.

src/exchanges/populate.ts
andyrichardson: Sounds like Map is what we want to be using

src/exchanges/populate.ts
andyrichardson: This is something I wanted to talk to you and @imranolas about. I'm assuming we're going to see users adding this decorator on a field with the following return types:
 - GraphQLObjectType
 - GraphQLInterfaceType

Are there any sub/parent/sibling types you can see being used?

src/exchanges/populate.ts
andyrichardson: This should really be

```suggestion
  return type.ofType.toString() || type.toString();
```

FYI:
 - ofType = List/NonNull
 - toString = Object/Interface/string

andyrichardson: Edit

```suggestion
  return (type.ofType && type.ofType.toString()) || type.toString();
```

src/exchanges/populate.ts
kitten: Ok, didn’t realise this was stateful 👍 the name didn’t really imply it 😂

src/exchanges/populate.ts
kitten: We could always optimise this in a separate PR though, so let’s maybe not worry about that yet? All the visiting logic in Graphcache is custom. The AST is pretty compact and easy to traverse so we could also make some specialised traversal logic if that makes things easier

src/exchanges/populate.ts
kitten: Oh, I see what it’s used for now then 😅 wasn’t too clear on that.

I’d suspect we’d be able to use it on any field that can have a selection set, so basically: returning objects or unions or interfaces

I feel like that may be it

src/exchanges/populate.ts
andyrichardson: `Object.create(null)` will do the trick - transforms prevent that spreading issue :+1:

src/exchanges/populate.ts
andyrichardson: Yeah lets visit that in a later PR.

Would be great to try this exchange out on some larger ASTs

src/exchanges/populate.ts
andyrichardson: Overlooked unions - adding a test for that now.

andyrichardson:
Merged to graphcache

PR #474, opened 2019-11-21
Add executeExchange for using an executable schema

andyrichardson:
## About
Exchange for executing queries locally using an executable schema.

This may be particularly useful for us when making example projects as we could mock responses here rather than implementing a server.

### DIFF
--- a/exchanges/execute/README.md
+++ b/exchanges/execute/README.md
@@ -1,20 +1,6 @@
 <h2 align="center">@urql/exchange-execute</h2>
-<p align="center">
-<strong>An exchange for executing queries against a local schema in <code>urql</code></strong>
-<br /><br />
-<a href="https://npmjs.com/package/@urql/exchange-graphcache">
-  <img alt="NPM Version" src="https://img.shields.io/npm/v/@urql/exchange-graphcache.svg" />
-</a>
-<a href="https://bundlephobia.com/result?p=@urql/exchange-graphcache">
-  <img alt="Minified gzip size" src="https://img.shields.io/bundlephobia/minzip/@urql/exchange-graphcache.svg?label=gzip%20size" />
-</a>
-<a href="https://github.com/FormidableLabs/urql-exchange-graphcache#maintenance-status">
-  <img alt="Maintenance Status" src="https://img.shields.io/badge/maintenance-active-green.svg" />
-</a>
-<a href="https://spectrum.chat/urql">
-  <img alt="Spectrum badge" src="https://withspectrum.github.io/badge/badge.svg" />
-</a>
-</p>
+
+<p align="center"><strong>An exchange for executing queries against a local schema in <code>urql</code></strong></p>
 
 `@urql/exchange-execute` is an exchange for the [`urql`](https://github.com/FormidableLabs/urql) GraphQL client which executes queries against a local schema.
 This is a replacement for the default _fetchExchange_ which sends queries over HTTP/S to be executed remotely.
--- a/exchanges/execute/README.md
+++ b/exchanges/execute/README.md
@@ -33,19 +33,17 @@ You'll then need to add the `executeExchange`, that this package exposes, to you
 by replacing the default cache exchange with it:
 
 ```js
-import { createClient, dedupExchange, fetchExchange } from 'urql';
-
-import { cacheExchange } from '@urql/exchange-execute';
+import { createClient, dedupExchange, cacheExchange } from 'urql';
+import { executeExchange } from '@urql/exchange-execute';
 
 const client = createClient({
   url: 'http://localhost:1234/graphql',
   exchanges: [
     dedupExchange,
-    // Replace the default cacheExchange with the new one
+    cacheExchange,
     executeExchange({
       /* config */
     }),
-    fetchExchange,
   ],
 });
 ```
--- a/exchanges/execute/README.md
+++ b/exchanges/execute/README.md
@@ -1,0 +1,103 @@
+<h2 align="center">@urql/exchange-execute</h2>
+<p align="center">
+<strong>An exchange for executing queries against a local schema in <code>urql</code></strong>
+<br /><br />
+<a href="https://npmjs.com/package/@urql/exchange-graphcache">
+  <img alt="NPM Version" src="https://img.shields.io/npm/v/@urql/exchange-graphcache.svg" />
+</a>
+<a href="https://bundlephobia.com/result?p=@urql/exchange-graphcache">
+  <img alt="Minified gzip size" src="https://img.shields.io/bundlephobia/minzip/@urql/exchange-graphcache.svg?label=gzip%20size" />
+</a>
+<a href="https://github.com/FormidableLabs/urql-exchange-graphcache#maintenance-status">
+  <img alt="Maintenance Status" src="https://img.shields.io/badge/maintenance-active-green.svg" />
+</a>
+<a href="https://spectrum.chat/urql">
+  <img alt="Spectrum badge" src="https://withspectrum.github.io/badge/badge.svg" />
+</a>
+</p>
+
+`@urql/exchange-execute` is an exchange for the [`urql`](https://github.com/FormidableLabs/urql) GraphQL client which executes queries against a local schema.
+This is a replacement for the default _fetchExchange_ which sends queries over HTTP/S to be executed remotely.
+
+## Quick Start Guide
+
+First install `@urql/exchange-execute` alongside `urql`:
+
+```sh
+yarn add @urql/exchange-execute
+# or
+npm install --save @urql/exchange-execute
+```
+
+You'll then need to add the `executeExchange`, that this package exposes, to your `urql` Client,
+by replacing the default cache exchange with it:
+
+```js
+import { createClient, dedupExchange, fetchExchange } from 'urql';
+
+import { cacheExchange } from '@urql/exchange-execute';
+
+const client = createClient({
+  url: 'http://localhost:1234/graphql',
+  exchanges: [
+    dedupExchange,
+    // Replace the default cacheExchange with the new one
+    executeExchange({
+      /* config */
+    }),
+    fetchExchange,
+  ],
+});
+```
+
+## Usage
+
+The exchange takes the same arguments as the [_execute_ function](https://graphql.org/graphql-js/execution/#execute) provided by graphql-js.
+
+Here's a brief example of how it might be used:
+
+```js
+import { buildSchema } from 'graphql';
+
+// Create local schema
+const schema = buildSchema(`
+  type Todo {
+    id: ID!
+    text: String!
+  }
+
+  type Query {
+    todos: [Todo]!
+  }
+
+  type Mutation {
+    addTodo(text: String!): Todo!
+  }
+`);
+
+// Create local state
+let todos = [];
+
+// Create root value with resolvers
+const rootValue = {
+  todos: () => todos,
+  addTodo: (_, args) => {
+    const todo = { id: todos.length.toString(), ...args };
+    todos = [...todos, todo];
+    return todo;
+  }
+}
+
+// ...
+
+// Pass schema and root value to executeExchange
+executeExchange({
+  schema,
+  rootValue,
+}),
+// ...
+```
+
+## Maintenance Status
+
+**Active:** Formidable is actively working on this project, and we expect to continue for work for the foreseeable future. Bug reports, feature requests and pull requests are welcome.
--- a/exchanges/execute/README.md
+++ b/exchanges/execute/README.md
@@ -30,7 +30,7 @@ npm install --save @urql/exchange-execute
 ```
 
 You'll then need to add the `executeExchange`, that this package exposes, to your `urql` Client,
-by replacing the default cache exchange with it:
+by replacing the default fetch exchange with it:
 
 ```js
 import { createClient, dedupExchange, cacheExchange } from 'urql';
@@ -41,6 +41,7 @@ const client = createClient({
   exchanges: [
     dedupExchange,
     cacheExchange,
+    // Replace the default fetchExchange with the new one.
     executeExchange({
       /* config */
     }),
--- a/exchanges/execute/package.json
+++ b/exchanges/execute/package.json
@@ -29,12 +29,6 @@
       "require": "dist/urql-exchange-execute.cjs.js",
       "types": "dist/types/index.d.ts",
       "source": "src/index.ts"
-    },
-    "./extras": {
-      "import": "dist/urql-exchange-execute-extras.esm.js",
-      "require": "dist/urql-exchange-execute-extras.cjs.js",
-      "types": "dist/types/extras/index.d.ts",
-      "source": "src/extras/index.ts"
     }
   },
   "files": [
--- a/exchanges/execute/package.json
+++ b/exchanges/execute/package.json
@@ -19,17 +19,18 @@
     "formidablelabs",
     "exchanges"
   ],
-  "main": "dist/urql-exchange-execute.cjs.js",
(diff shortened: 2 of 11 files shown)


kitten:
I think @mxstbr is already maintaining a package for that. We may just want to send a PR to update it a little https://github.com/mxstbr/urql-exchange-schema

mxstbr:
Happy to merge that into core!

andyrichardson:
Let me know what you want to do :+1:

src/exchanges/execute.ts
```diff
@@ -0,0 +1,71 @@
...
+      filter(f => targetOperationTypes.includes(f.operationName)),
+      map(async o => {
+        try {
+          const r = await execute(
+            schema,
+            o.query,
+            rootValue,
+            contextValue,
+            o.variables,
+            o.operationName,
```
andyrichardson: Breaking change added to graphql-js caused this to fail.

We don't need to pass operation name so this should be null.

```suggestion
            null,
```

kitten:
@JoviDeCroock @andyrichardson assuming we'll want to get this merged soon, but we'll only need some minor documentation for this after a rebase, and let's stick it into a new `exchanges/execute` package?

andyrichardson:
@kitten Yeah I'm happy to sort that out but you're going to have to explain the monorepo to me.

The exchange is 61 lines, no new dependencies, and will be tree shaken if not in use. Do we want this exchange to be a standalone package in `/exchanges`?

Also, where do the docs for standalone packages go?

kitten:
I think bundling an exchange would only make sense if a majority of users use it? Although, I can totally see an exchange being "moved" into core, when it crosses that threshold. For now, I'd play it safe. It looks like when `execute` is bundled (be that because it's bundled accidentally, or tree-shaking fails, or not an option, etc) it adds a lot of heft: https://github.com/apollographql/apollo-link/pull/968

I've got some docs in the `CONTRIBUTING.md` guide that should help :+1: but happy to also pair on that or sth :+1: https://github.com/FormidableLabs/urql/blob/master/CONTRIBUTING.md#how-do-i-add-a-new-package

andyrichardson:
@kitten thanks for the explanation - I've moved it to _/exchanges_

exchanges/execute/package.json
```diff
@@ -0,0 +1,68 @@
...
+  "types": "dist/types/index.d.ts",
+  "source": "src/index.ts",
+  "exports": {
+    ".": {
+      "import": "dist/urql-exchange-execute.esm.js",
+      "require": "dist/urql-exchange-execute.cjs.js",
+      "types": "dist/types/index.d.ts",
+      "source": "src/index.ts"
+    },
+    "./extras": {
```
JoviDeCroock: Why does the execute exchange have extra's?

exchanges/execute/package.json
```diff
@@ -0,0 +1,68 @@
...
+  },
+  "keywords": [
+    "urql",
+    "exchange",
+    "execute",
+    "executable schema",
+    "formidablelabs",
+    "exchanges"
+  ],
+  "main": "dist/urql-exchange-execute.cjs.js",
```
JoviDeCroock: Rebase your branch, we altered the conventions we use to be

```
main: dist/urql-exchange-execute
module: dist/urql-exchange-execute.mjs
```

This has to be done for the exports map as well but these need the path "./" included for node 13

exchanges/execute/src/execute.test.ts
```diff
@@ -0,0 +1,120 @@
...
+});
+
+describe('on unsupporte doperation', () => {
+  const operation = {
+    ...queryOperation,
+    operationName: 'teardown',
+  } as const;
+
+  it('returns operation result', async () => {
+    const [stream, next] = makeSubject<any>();
```
JoviDeCroock: The new wonka returns an object of subject$ and next

exchanges/execute/src/execute.ts
```diff
@@ -0,0 +1,60 @@
...
+  GraphQLFieldResolver,
+  GraphQLTypeResolver,
+  execute,
+} from 'graphql';
+import { Exchange, makeResult, makeErrorResult } from '@urql/core';
+
+interface ExecuteExchangeArgs {
+  schema: GraphQLSchema;
+  rootValue?: any;
+  contextValue?: any;
```
kitten: It’d be great to just call this just `context` and make it accept a function returning an object as well, inline with Apollo Server and other servers 🙌

Same for `rootValue`; it’d be great to have that aligned with Apollo Server as well

exchanges/execute/src/execute.ts
```diff
@@ -0,0 +1,60 @@
...
+      map(async o => {
+        try {
+          const r = await execute(
+            schema,
+            o.query,
+            rootValue,
+            contextValue,
+            o.variables,
+            null,
+            fieldResolver,
```
kitten: These two resolvers fields shouldn’t be a thing for  `execute` and executable schemas. Can they go?

exchanges/execute/src/execute.ts
```diff
@@ -0,0 +1,60 @@
...
+export const executeExchange = ({
+  schema,
+  rootValue,
+  contextValue,
+  fieldResolver,
+  typeResolver,
+}: ExecuteExchangeArgs): Exchange => ({ forward }) => {
+  return ops$ => {
+    const sharedOps$ = share(ops$);
+    const targetOperationTypes = ['query', 'mutation', 'subscription'];
```
kitten: I don’t think you’d want subscriptions here because this is missing the necessary support stuff from `graphql-subscriptions`. Let’s maybe keep it to queries and mutations for now? 🤔

exchanges/execute/package.json
andyrichardson: @JoviDeCroock  Like this? 

```suggestion
  "main": "./dist/urql-exchange-execute",
  "module": "./dist/urql-exchange-execute.mjs",
```

exchanges/execute/src/execute.test.ts
andyrichardson: How does this look?

```suggestion
    const { stream, next } = makeSubject<any>();
```

exchanges/execute/src/execute.ts
andyrichardson: The naming is identical to the [graphqljs execute](https://github.com/graphql/graphql-js/blob/master/src/execution/execute.js#L146) function.

exchanges/execute/src/execute.ts
andyrichardson: This just proxies arguments to `graphql.js` [(source)](https://github.com/graphql/graphql-js/blob/master/src/graphql.js#L71)

exchanges/execute/src/execute.ts
andyrichardson: Looks like there's [no combined function yet](https://github.com/graphql/graphql-js/issues/2144) either - maybe something worth looking at in the future.

```suggestion
    const targetOperationTypes = ['query', 'mutation'];
```

leebenson:
FWIW, this exchange fixed an issue I was experiencing with https://github.com/mxstbr/urql-exchange-schema, in that queries via SSR were being executed _once for each exchange_.

So a typical set-up of:

```ts
exchanges: [
  dedupExchange,
  cacheExchange,
  ssr,
  schemaExchange(schema, { context }),
]
```

... would actually run the query **four** times.

This is a subtle bug that's easy to overlook, since the first query that's run is the one that winds up being rendered back to the client.

Just a heads-up @mxstbr in case it made sense to update your own package while waiting for this to merge.

exchanges/execute/src/execute.ts
```diff
@@ -0,0 +1,60 @@
...
+  rootValue?: any;
+  contextValue?: any;
+  fieldResolver?: GraphQLFieldResolver<any, any>;
+  typeResolver?: GraphQLTypeResolver<any, any>;
+}
+/** Exchange for executing queries locally on a schema using graphql-js. */
+export const executeExchange = ({
+  schema,
+  rootValue,
+  contextValue,
```
kitten, line 21: One small thing we could do is to allow this to be a function so new contexts could be created per operation, which would allow the user to derive the context from `operation.context` (like extract an API token or sth)

exchanges/execute/src/execute.ts
```diff
@@ -0,0 +1,60 @@
...
+  fieldResolver,
+  typeResolver,
+}: ExecuteExchangeArgs): Exchange => ({ forward }) => {
+  return ops$ => {
+    const sharedOps$ = share(ops$);
+    const targetOperationTypes = ['query', 'mutation'];
+
+    const executedOps$ = pipe(
+      sharedOps$,
+      filter(f => targetOperationTypes.includes(f.operationName)),
```
kitten, line 31: nit: It'd be great to just match the other exchanges in just comparing it directly, i.e. `operation => operation.operationName === 'query' || operation.operationName === 'subscription'`. Not that big of a deal though which one we go for.

exchanges/execute/src/execute.ts
```diff
@@ -0,0 +1,60 @@
...
+      map(async o => {
```
kitten, line 32: Nit: I think this exchange uses short variable names a lot `o` for `operation`, `f` for "to be filtered operation", `r` for result, etc, which makes it a little hard to follow

exchanges/execute/src/execute.ts
```diff
@@ -0,0 +1,60 @@
+import { pipe, share, filter, map, fromPromise, mergeMap, merge } from 'wonka';
+import {
+  GraphQLSchema,
+  GraphQLFieldResolver,
+  GraphQLTypeResolver,
+  execute,
+} from 'graphql';
+import { Exchange, makeResult, makeErrorResult } from '@urql/core';
+
+interface ExecuteExchangeArgs {
```
kitten, line 10: Will this need to be exported?

kitten:
To get this up-to-date we just need to make sure that the config files match (`package.json` fields, some are linted by the `prepare` script), `tsconfig.json` changes, creating an empty `CHANGELOG`, etc

amyboyd:
> To get this up-to-date we just need to make sure that the config files match (`package.json` fields, some are linted by the `prepare` script), `tsconfig.json` changes, creating an empty `CHANGELOG`, etc

The branch has been rebased on master and these files are matching the other extensions.

A few other misc changes -- see commit log.

exchanges/execute/package.json
amyboyd: Done.

exchanges/execute/src/execute.ts
amyboyd: Accepts a function now.

exchanges/execute/src/execute.ts
amyboyd: Accepts a function now.

exchanges/execute/src/execute.ts
amyboyd: I've gone for:

```
const SUPPORTED_OPERATION_TYPES = ['query', 'mutation'];

......

.filter(operation =>
     SUPPORTED_OPERATION_TYPES.includes(operation.operationName)
),
```

Happy to change to your suggestion if you prefer.

exchanges/execute/src/execute.ts
amyboyd: Agreed, short variable names are hard to read. Have renamed:

`f` to `operation`

`o` to `operation`

`r`  to `result`

`p` to `pipe`

exchanges/execute/src/execute.ts
amyboyd: I don't see a need for it to be exported right now. If left private, it can be exported later. If exported now, it can't be undone in the future (without a breaking change).

exchanges/execute/src/execute.ts
```diff
@@ -0,0 +1,67 @@
...
+      ),
+      map(async operation => {
+        try {
+          const result = await execute(
+            schema,
+            operation.query,
+            rootValue,
+            typeof contextValue === 'function' ? contextValue() : contextValue,
+            operation.variables,
+            operation.operationName,
```
kitten, line 44: I believe this needs to be filled in by this function, which we can just copy over: https://github.com/FormidableLabs/urql/blob/bde241f911a0109abd86b3ecb9e717dfb8243890/packages/core/src/internal/fetchOptions.ts#L13

Due to historical reasons `operationName` is used for two things in `GraphQL`: The operation type and the operation name;

For us the type is `query | subscription | mutation | teardown`, but for GraphQL it sometimes refers to the name that a user has given a query document, e.g. `query MyName`

exchanges/execute/README.md
```diff
@@ -0,0 +1,102 @@
...
+</a>
+<a href="https://bundlephobia.com/result?p=@urql/exchange-graphcache">
+  <img alt="Minified gzip size" src="https://img.shields.io/bundlephobia/minzip/@urql/exchange-graphcache.svg?label=gzip%20size" />
+</a>
+<a href="https://github.com/FormidableLabs/urql-exchange-graphcache#maintenance-status">
+  <img alt="Maintenance Status" src="https://img.shields.io/badge/maintenance-active-green.svg" />
+</a>
+<a href="https://spectrum.chat/urql">
+  <img alt="Spectrum badge" src="https://withspectrum.github.io/badge/badge.svg" />
+</a>
```
kitten, line 16: These seem out-of-date and copied from `@urql/exchange-graphcache`. We can also just remove them, since most exchanges readmes don't need this information as long as they'll be documented in the future in the docs, e.g. https://github.com/FormidableLabs/urql/commit/b8a7548e9b66f5028d30137e872088b8911bd6aa#diff-9190c36c53f94a10dac5a70ca4510a03R1

exchanges/execute/src/execute.ts
```diff
@@ -0,0 +1,67 @@
...
+      filter(operation =>
+        SUPPORTED_OPERATION_TYPES.includes(operation.operationName)
+      ),
+      map(async operation => {
+        try {
+          const result = await execute(
+            schema,
+            operation.query,
+            rootValue,
+            typeof contextValue === 'function' ? contextValue() : contextValue,
```
kitten, line 42: I think the only thing that's left here is provide some useful data to `contextValue()`; I think we could just pass `operation` or `operation.context`. Mainly this will need to give the user access to `operation.context` so that options like `operation.context.fetchOptions` are accessible.

Also, I think, I mentioned this in a nother comment, but some implementations, like `express-graphql` choose to name this option `context` instead of `contextValue`. That could potentially be confusing, since we already have `operation.context`, but technically it's not only a value anymore given that it can be computed: https://graphql.org/graphql-js/express-graphql/

exchanges/execute/src/execute.ts
```diff
@@ -0,0 +1,67 @@
...
+    const sharedOps$ = share(ops$);
+
+    const executedOps$ = pipe(
+      sharedOps$,
+      filter(operation =>
+        SUPPORTED_OPERATION_TYPES.includes(operation.operationName)
+      ),
+      map(async operation => {
+        try {
+          const result = await execute(
```
kitten, line 38: The last thing that's kind of missing is cancellation via the `teardown` operations; we can pair on this later, but generally we need to ensure that the result isn't issued when it's been cancelled, specifically for React Suspense 🙌

exchanges/execute/src/execute.ts
amyboyd: Now called as `context(operation)`

exchanges/execute/README.md
amyboyd: Removed.

exchanges/execute/src/execute.ts
amyboyd: Have wrapped with `getOperationName()`

PR #482, opened 2019-12-10
try the closure compiler in our build process

JoviDeCroock:
Sizes:

|  Name | Past | PR  |
|---|---|---|
|  urql.js | 7.25  |  6.22 |
|  urql.es.js | 7.25 | 6.06  |
|  urql.min.js | 4.83  | 4.79  |
|  urql.es.min.js | 4.86  | 4.5  |

In theory parse time should speed up as well due to the closure compiler, something that might be worth looking into:  https://github.com/angular/tsickle

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -6,6 +6,7 @@ import buble from 'rollup-plugin-buble';
 import babel from 'rollup-plugin-babel';
 import replace from 'rollup-plugin-replace';
 import { terser } from 'rollup-plugin-terser';
+import compiler from '@ampproject/rollup-plugin-closure-compiler';
 import transformPipe from './scripts/transform-pipe';
 
 const pkgInfo = require('./package.json');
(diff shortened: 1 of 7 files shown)


rollup.config.js
```diff
@@ -169,23 +174,28 @@ export default [
     ]
   }, {
     ...config,
-    plugins: makePlugins(true),
+    plugins: makePlugins(true, 'ADVANCED_OPTIMIZATIONS'),
```
JoviDeCroock: `esm` is the only place where we can do advanced optimizations since `closure` is made for `import/export` rather than `require`.

rollup.config.js
```diff
@@ -169,23 +174,28 @@ export default [
...
-        file: './dist/urql.min.js',
-        format: 'cjs'
-      },
+        file: './dist/urql.es.min.js',
+        format: 'esm'
+      }
+    ]
+  }, {
+    ...config,
+    plugins: makePlugins(true, 'SIMPLE_OPTIMIZATIONS'),
```
kitten: I suppose the problem is that people won’t use any of our heavily optimised mini field bundles by default, which negates some of the benefits 😔

JoviDeCroock: I can split it up in 4 so that we can do advanced optimizations on the normal es bundle as well

PR #485, opened 2019-12-10
fix problem with stringifying date objects

### LINKED ISSUE #478: Error in stringifyVariables during mutation
We recently updated our version of urql from 1.3.1 to 1.6.3 and our mutations started breaking.

The error logged in the console is
```
Uncaught TypeError: Cannot read property 'length' of undefined
    at stringify (stringifyVariables.ts:36)
    at stringify (stringifyVariables.ts:35)
    at stringify (stringifyVariables.ts:35)
    at stringify (stringifyVariables.ts:35)
    at stringify (stringifyVariables.ts:35)
    at stringifyVariables (stringifyVariables.ts:49)
    at createRequest (request.ts:36)
    at useMutation.ts:47
```
We have Momentjs fields in the object that is being passed into the mutation variables and it appears this is where the embedded stringify is breaking.  Version 1.3.1 was able to stringify these just fine with the default string version of a Moment being an ISO date.

Any guidance would be appreciated.

### LINKED ISSUE #479: refetch not triggered when variable changed
I have the following code to fetch data:
```
  const [{ data, fetching, error }, fetchMovements] = useQuery<
    Get_MovementsQuery
  >({
    query: GET_MOVEMENTS,
    variables: { skip: startRow, first: rowNumber, orderBy, where: filter },
    requestPolicy: 'cache-and-network'
  })
```

Where filter is an object with all the different filter options. 

I have a datetime-range filter. The structure of the filter object for that case looks like this:
```
{
    date: {
      gte: DateTime,
      lte: DateTime
    }
  }
```

When changing one of the values for the first time, it correctly detects a change in the variables and triggers a refetch, but when changing the same datetime another time, it doesn't cause a refetch.

If necessary I can provide a reproduction.

Maybe it's also just my own stupidity, but I am pretty sure it's related to urql.

I am using the latest version with the urql-graphexchange.

BjoernRave:
For some reason when I tried to include the library locally I got an error about Invalid hook calls:

`Error: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:`

So I was not able to test if it fixes those for sure

This possibly: 
fixes #479 and fixes #478

### DIFF
--- a/src/utils/stringifyVariables.ts
+++ b/src/utils/stringifyVariables.ts
@@ -1,9 +1,10 @@
 const seen = new Set();
 
 const stringify = (x: any): string => {
-  if (x && x.toJson) x = x.toJSON();
   if (x === undefined) {
     return '';
+  } else if (x && x.toJSON) {
+    return x.toJSON();
   } else if (typeof x == 'number') {
     return isFinite(x) ? '' + x : 'null';
   } else if (typeof x !== 'object') {
--- a/src/utils/stringifyVariables.ts
+++ b/src/utils/stringifyVariables.ts
@@ -1,6 +1,7 @@
 const seen = new Set();
 
 const stringify = (x: any): string => {
+  if (x && x.toJson) x = x.toJSON();
   if (x === undefined) {
     return '';
   } else if (typeof x == 'number') {
--- a/src/utils/stringifyVariables.ts
+++ b/src/utils/stringifyVariables.ts
@@ -3,14 +3,14 @@ const seen = new Set();
 const stringify = (x: any): string => {
   if (x === undefined) {
     return '';
-  } else if (x && x.toJSON) {
-    return x.toJSON();
   } else if (typeof x == 'number') {
     return isFinite(x) ? '' + x : 'null';
   } else if (typeof x !== 'object') {
     return JSON.stringify(x);
   } else if (x === null) {
     return 'null';
+  } else if (x.toJSON) {
+    return x.toJSON();
   }
 
   let out = '';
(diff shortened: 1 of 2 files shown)


JoviDeCroock:
Can we add a test for this?

src/utils/stringifyVariables.ts
```diff
@@ -1,6 +1,7 @@
 const seen = new Set();
 
 const stringify = (x: any): string => {
+  if (x && x.toJson) x = x.toJSON();
```
kitten: Seems like this isn’t quite the right check. We could also move it down to avoid the double null check and do a proper `typeof` being “function” check?

I’m also wondering, what are we really accomplishing for moment here? Don’t we want to convert it and dates to a UTC string?

JoviDeCroock: If I recall correctly the normal way for serializing a complex object is through the `.toJSON` function which is present on most libraries like `moment`, `day-js`,...

BjoernRave: When writing the test I found out that it doesn't work this way, because for some reason `x.toJson` is undefined. At least for javascript dates `JSON.stringify(x)` works though

BjoernRave:
Not sure if this test is enough or if we should import sth like moment to test it with their dates as well

src/utils/stringifyVariables.ts
```diff
@@ -3,6 +3,8 @@ const seen = new Set();
 const stringify = (x: any): string => {
   if (x === undefined) {
     return '';
+  } else if (x && x.toJSON) {
```
JoviDeCroock: can we just move this after `x !== 'object` for then we can probably remove `x &&`

BjoernRave: I moved it on the last position, otherwise it would be called when `x === null`

PR #496, opened 2020-01-09
Support concurrent mode

JoviDeCroock:


src/exchanges/dedup.ts
```diff
@@ -10,7 +10,7 @@ export const dedupExchange: Exchange = ({ forward }) => {
     if (operationName === 'teardown') {
       inFlightKeys.delete(key);
       return true;
-    } else if (operationName !== 'query') {
+    } else if (operationName !== 'query' && operationName !== 'subscription') {
```
andyrichardson: Good shout :+1:

src/hooks/useSource.ts
```diff
@@ -0,0 +1,98 @@
...
+  output: R;
+  input?: T;
+  task?: any;
+  [key: string]: any;
+}
+
+const useIsoEffect =
+  typeof window !== 'undefined' ? useLayoutEffect : useEffect;
+
+export const useSubjectValue = <T, R>(
```
andyrichardson: File name is `useSource`

andyrichardson:
Any chance of a breif explaination of `useSubjectValue` for those out of the loop (...me)

andyrichardson:
Also possibly off topic but for support with concurrent mode, we probably want to implement a some preloader functionality.

Here's what I [documented on tipple for concurrent support](https://github.com/andyrichardson/tipple/issues/55) and here's an [example preload util and hook](https://github.com/andyrichardson/tipple/blob/suspense/src/preloadRequest.ts#L19)

I think this would be a good way to go as I wouldn't expect useQuery to work and the [React docs encourage against this pattern](https://reactjs.org/docs/concurrent-mode-suspense.html#for-library-authors) for use with concurrent mode.

JoviDeCroock:
> Also possibly off topic but for support with concurrent mode, we probably want to implement a some preloader functionality.

Prelaoder is mainly meant for `Suspense` and is not really in line with our current Suspense approach (which we have to reevaluate)

> I think this would be a good way to go as I wouldn't expect existing useQuery and useSubscription hooks to work and the React docs encourage against this pattern for use with concurrent mode.

Yes this will work for both Concurrent and non-concurrent which should be the best way forward

I'll leave some comments explaining everything in a bit

kitten:
So mainly the preloader approach, as far as Relay and the recommended use of Suspense goes is that `useQuery` is used as usual, but is used as the preloader itself. Below that `useQuery` you should then be able to (somehow) have a suspense-based loader.

That doesn't currently align with our approach, but is worth thinking about for sure :+1:

rollup.config.js
```diff
@@ -7,6 +7,7 @@ import babel from 'rollup-plugin-babel';
 import replace from 'rollup-plugin-replace';
 import { terser } from 'rollup-plugin-terser';
 import transformPipe from './scripts/transform-pipe';
+console.log(Object.keys(require('scheduler')))
```
JoviDeCroock: TODO: remove this log

src/hooks/testie.ts
```diff
@@ -0,0 +1,66 @@
+import { useReducer, useRef, useMemo, useEffect, useLayoutEffect } from 'react';
```
JoviDeCroock: TODO: remove this file (no clue why git didn't pick this up)

src/hooks/useOperator.ts
```diff
@@ -0,0 +1,134 @@
...
+
+interface State<R, T = R> {
+  subject: Subject<T>;
+  onValue: Dispatch<R>;
+  teardown: null | (() => void);
+  task: null | CallbackNode;
+  value: R;
+}
+
+const useIsomorphicEffect =
```
JoviDeCroock: Support SSR (useLayoutEffect not working there)

andyrichardson:
Yeah okay so it sounds like we haven't decided on how preloading is going to be handled. I think it should be pretty straightforward (it looks to be with tipple) and just requires us to add a new hook for handling preloaded queries [(tipple version)](https://github.com/andyrichardson/tipple/blob/suspense/example/src/components/Posts.tsx#L19)

JoviDeCroock:
@andyrichardson that approach can't preload url parameters and stuff right? What if we want to react to props?

andyrichardson:
> @andyrichardson that approach can't preload url parameters and stuff right? What if we want to react to props? It feels like that approach would only work if you don't need variables which is rare, no?

I think the idea is that these are two separate use cases (hence the seperate APIs) 

> Although it’s technically doable, Suspense is not currently intended as a way to start fetching data when a component renders. Rather, it lets components express that they’re “waiting” for data that is already being fetched. Building Great User Experiences with Concurrent Mode and Suspense describes why this matters and how to implement this pattern in practice.

[source](https://reactjs.org/docs/concurrent-mode-suspense.html#for-library-authors)

PR #509, opened 2020-01-15
Update file exchange example to work with wonka v4

lukasluecke:
Not sure if the other guides need to be updated as well, but this is the change I needed to make in my project for the file exchange.

### DIFF
--- a/docs/guides.md
+++ b/docs/guides.md
@@ -465,7 +465,7 @@ import { extractFiles } from 'extract-files';
 
 ...
 const createFetchSource = (operation) => {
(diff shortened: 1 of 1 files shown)


docs/guides.md
```diff
@@ -465,7 +465,7 @@ import { extractFiles } from 'extract-files';
 
 ...
 const createFetchSource = (operation) => {
-  return make(([next, complete]) => {
+  return make(({next, complete}) => {
```
JoviDeCroock: Let's add some spaces

```suggestion
  return make(({ next, complete }) => {
```

PR #514, opened 2020-01-17
Fix concurrent-mode edge cases using useSubscription (adds useSo…

kitten:
This pull request aims to fix even more concurrent-mode edge cases by changing our approach. Previously we disqualified the use of `useSubscription` because we didn't have a way to construct a side-effect-less `getCurrentValue` function.

As it turns out we can indeed create a `getCurrentValue` that creates short, synchronous subscriptions to get the current value of a source and avoid issues where `react-wonka` doesn't start/stop subscriptions correctly (see [`react-wonka/#11`](https://github.com/kitten/react-wonka/pull/11))

This is more of a temporary issue as `use-subscription` will often lead to a bailout of many concurrent mode features (see [`use-subscription` readme](https://github.com/facebook/react/tree/master/packages/use-subscription#limitations-in-concurrent-mode))

Realistically, we want to wait for this feature and then replace the two utilities that this PR creates (see [`react/#15317`](https://github.com/facebook/react/issues/15317#issuecomment-575761699))

> We're working on a new hook for mutable external values that deopts in scenarios where it would produce an inconsistent tree. By doing a full batched render in synchronous mode. It's "safe" and it does allow a tree to commit as one.

This PR **changes** the behaviour of `teardown` operations. They will be seen much more frequently (twice on initial mount) and will hence become a little more noisy. The `fetchExchange` has been adapted however so that it avoids calling `fetch` for immediately cancelled operations.

We can reduce this churn a little to _one teardown per initial mount_ by replacing `useSubscription` in this PR's new `useSource` hook, **in a follow up PR**. This would check passively in `useEffect` whether a new value has come in. Currently `useEffect` actively calls `getCurrentValue` synchronously in `useEffect` to check whether it needs to update and it calls it on every subscription event. Due to our strong guarantees we can avoid this.

We'll have to monitor changes in React to be able to improve this behaviour but for now it seems to be the safest option available to us.


src/hooks/useSource.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export const useSource = <T>(source: Source<T>, init: T): T =>
+  useSubscription(
+    useMemo((): Subscription<T> => {
+      let hasUpdate = false;
+      let currentValue: T = init;
+
+      const updateValue = pipe(
+        source,
+        onPush(value => (currentValue = value))
```
JoviDeCroock: NIT: does this need to return a value?

src/hooks/useSource.ts
```diff
@@ -0,0 +1,74 @@
...
+    const prevValue$ = pipe(
+      fromValue(value),
+      map(() => prevValue)
+    );
+
+    const source = concat([prevValue$, subject.source]);
+
+    const next = (value: T) => {
+      prevValue = value;
+      subject.next(value);
```
JoviDeCroock:
```suggestion
      subject.next(prevValue = value);
```

PR #519, opened 2020-02-03
implement conversion for useHTTPGetForQueries

JoviDeCroock:
The only uncertainty (no documentation anywhere) is how to use fragments for this.


src/exchanges/fetch.ts
```diff
@@ -146,3 +151,24 @@ const executeFetch = (
       }
     });
 };
+
+export const convertToGet = (uri: string, body: Body): string => {
```
andyrichardson: This looks good!

src/exchanges/fetch.ts
```diff
@@ -99,6 +97,13 @@ const createFetchSource = (operation: Operation) => {
         abortController !== undefined ? abortController.signal : undefined,
     };
 
+    if (fetchOptions.method === 'GET' && operation.operationName === 'query') {
```
andyrichardson: I'd go the ternary route here or pull fetchOptions parsing into a function with a return. That way fetchOptions doesn't change after assignment

JoviDeCroock: This is kind off implicit logic so I'll probably move this to `client.useGetForQueries` instead and attach it to the `Operation` this will grant the benefit of us also being able to clearly document this

src/exchanges/fetch.ts
```diff
@@ -146,3 +151,24 @@ const executeFetch = (
...
 };
+
+export const convertToGet = (uri: string, body: Body): string => {
+  const queryParams: string[] = [];
+  if (body.query) {
+    queryParams.push(`query=${encodeURIComponent(body.query)}`);
+  }
+
+  if (body.operationName) {
+    queryParams.push(`operationName=${encodeURIComponent(body.query)}`);
```
kitten: This isn’t correct afaik. This is usually the query’s name that is to be executed, but should be optional when only one `DocumentNode` is present (so for us all the time)

Since we’re not adding this for `POST` we shouldn’t add it to `GET` or we should add it for both.

src/exchanges/fetch.ts
```diff
@@ -146,3 +151,24 @@ const executeFetch = (
       }
     });
 };
+
+export const convertToGet = (uri: string, body: Body): string => {
+  const queryParams: string[] = [];
+  if (body.query) {
```
kitten: We should probably assume that `query` will be present.

src/exchanges/fetch.ts
```diff
@@ -146,3 +151,24 @@ const executeFetch = (
...
+    queryParams.push(`operationName=${encodeURIComponent(body.query)}`);
+  }
+
+  if (body.variables) {
+    queryParams.push(
+      `variables=${encodeURIComponent(JSON.stringify(body.variables))}`
+    );
+  }
+
+  // TODO: fragments
```
kitten: Not sure what this means. Fragments are part of the document and shouldn’t need any special logic

JoviDeCroock: I thought there was a complex issue with this on a previous thing I made. Let's see when I test it

src/exchanges/fetch.ts
kitten: I think the naming here as a little shaky, but I’d expect that we’re going a little off-spec and standards here we could also make some assumptions and call the option `preferGetMethod`

src/exchanges/fetch.ts
```diff
@@ -29,7 +29,7 @@ export const fetchExchange: Exchange = ({ forward }) => {
           filter(op => op.operationName === 'teardown' && op.key === key)
         );
 
-        return pipe(createFetchSource(operation), takeUntil(teardown$));
+        return pipe(createFetchSource(operation, operation.operationName === 'query' && client.preferGetMethod), takeUntil(teardown$));
```
kitten: We shouldn’t ever have to read from the client directly. Let’s move this to the operation context.

The reason for that has generally always been that the client’s option are opaque and separate from the operations and how they’re created and are dealt with in the exchange pipeline. So any option on the context can be seen, set, and manipulated by other exchanges. The same doesn’t go for the client.

To be fair, we should probably type the `Client` on the exchange using an interface that only contains the method we’d like to be exposed on the exchanges.

PR #532, opened 2020-02-10
Fixes links to graphcache documation

than:
Since the package moved, the README links were no longer working.


than:
Looks like #530 beat me to it.

PR #533, opened 2020-02-11
Provide possibility to mask the __typename in client results

JoviDeCroock:
## Summary

It solves problems where you don't want to manually strip in variables/useQuery result before sending it off somewhere

## Set of changes

- add export to urql/core (maskTypename)
- adds the option to client to enable this on all results


JoviDeCroock:
> Doesn't it conceptually make more sense to be able to use `stripTypename` on `variables`?

So to me it makes sense being able to use it on the response for your query since we implicitly add `__typename`. This allows us to reverse that (optionally). `variables` with `__typename` seems like a bad practice for a request so I enforced it there? What would you change?

kitten:
@JoviDeCroock That's true. What I'm unsure of is that we basically have an asymmetry already:

- We add `__typename` but that cannot be disabled or reversed on query documents
- We may remove `__typename` but that's on a hook level in this PR
- Removing `__typename` is often relevant when passing results as inputs but we'd be applying it to results instead of to variables

So I'm not sure whether the place where we're applying this makes sense?

packages/core/src/utils/stripTypename.ts
```diff
@@ -0,0 +1,19 @@
+export const stripTypename = (data: any): any => {
+  if (!data) return data;
+
+  const result = {};
+
+  Object.keys(data).forEach((key: string) => {
```
kitten: I'd suggest to either use a `for in` loop here or switch to `reduce`

JoviDeCroock:
The main point you are hinting at is to keep it exclusively to `useRequest` right? I just think it's very ergonomic to provide this to `useQuery` in case of for instance `initialValues` of a form, when using object schema validation additional properties often result in errors.

packages/core/src/utils/stripTypename.ts
```diff
@@ -0,0 +1,22 @@
+export const stripTypename = (data: any): any => {
+  if (!data) return data;
+
+  return Object.keys(data).reduce((acc, key: string) => {
+    const value = data[key];
+    if (key === '__typename') {
+      return acc;
+    } else if (Array.isArray(value)) {
+      return {
+        ...acc,
```
kitten: We don't need to spread acc here to avoid excessive object allocations and can just mutate (as before).

Does it make sense to add `__typename` as `Object.defineProperty({ value, enumerable: false, writable: true, configurable: true })`?

packages/core/src/utils/stripTypename.ts
```diff
@@ -0,0 +1,21 @@
+export const stripTypename = (data: any): any => {
+  if (!data || typeof data !== 'object') return data;
+
+  return Object.keys(data).reduce((acc, key: string) => {
```
JoviDeCroock: Wondering if there's a scenario where data is an array from the start

kitten: No, since we'd only apply this to things that are objects in GraphQL (results / variables / etc) we'd never start with an array

packages/core/src/utils/stripTypename.ts
```diff
@@ -0,0 +1,21 @@
...
+    const value = data[key];
+    if (key === '__typename') {
+      Object.defineProperty(acc, '__typename', {
+        enumerable: false,
+        value,
+      });
+    } else if (Array.isArray(value)) {
+      acc[key] = value.map(stripTypename);
+    } else if (typeof value === 'object') {
```
kitten: This needs an escape hatch:

```suggestion
    } else if (typeof value === 'object' && '__typename' in value) {
```

That's because we don't know what other values may be used here. It's imaginable that a user has deserialised a result at this point where a value for instance is a `Date` object.

.changeset/modern-queens-run.md
```diff
@@ -0,0 +1,8 @@
+---
+'@urql/core': minor
+'@urql/preact': minor
+'urql': minor
+---
+
+Adds the `stripTypename` export to urql-core, this deeply makes `__typename` non-enumerable.
```
kitten:
```suggestion
Add the `stripTypename` export to `@urql/core` to make all `__typename` fields non-enumerable.
```

.changeset/modern-queens-run.md
```diff
@@ -0,0 +1,8 @@
...
+In (P)React we'll apply the above function on result payloads of mutations, subscriptions and mutations.
```
kitten:
```suggestion
In `@urql/preact` and `urql` this is applied to all `data` results.
```

marcusdarmstrong:
The implication of these changes would be that it's impossible to useQuery to fetch __typename as a part of the original query? That seems harmful.

JoviDeCroock:
> The implication of these changes would be that it's impossible to useQuery to fetch __typename as a part of the original query? That seems harmful.

No, we've always implicitly added `__typename` and result.data.field.__typename will still be defined it just won't be enumerable anymore

marcusdarmstrong:
> > The implication of these changes would be that it's impossible to useQuery to fetch __typename as a part of the original query? That seems harmful.
> 
> No, we've always implicitly added `__typename` and result.data.field.__typename will still be defined it just won't be enumerable anymore

1. `__typename` is only automatically added via cache exchanges presently, not urql itself—our usage doesn't involve that automatic addition, as we don't use the default cache or graphcache—and that's actually important for us, as we'd like to move toward not actually having query ASTs on the client/in our bundles.
2. Enumerability is the expectation for fields I request in my query. That makes this a breaking change.

kitten:
What this PR doesn't include yet is that this should and will only be enabled when an option on the client is enabled, e.g. `new Client({ maskTypenames: true })`

.changeset/modern-queens-run.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/core': minor
+---
+
+Adds the `stripTypename` export to urql-core, this deeply strips typenames from the given payload. In (P)React we add an additional parameter to useQuery to strip typenames on result and strip typenames from variables in useRequest
```
kitten: Let’s get this updated :)

.changeset/modern-queens-run.md
```diff
@@ -0,0 +1,7 @@
+---
+'@urql/core': minor
+---
+
+Adds the `maskTypename` export to urql-core, this deeply masks typenames from the given payload.
+Masking typenames has also become an option on the `client` under the name `maskTypename` setting this to `true` will automatically
```
kitten:
```suggestion
Masking `__typename` properties is also available as a `maskTypename` option on the `Client`. Setting this to true will
```

kitten:
`maskTypename` has been removed.
For a full write-up on why and more see: https://github.com/urql-graphql/urql/pull/3299#issuecomment-2194996566

Issue #537, opened 2020-02-12
[exchange-graphcache] TypeScript errors in updateQuery


iamkevinwolf:
I am getting some TypeScript errors when trying to use `cache.updateQuery` within the`updates` object of `@urql/exchange-graphcache`.

Given the following query:

```graphql
query teams {
  __typename
  teams {
    __typename
    id
    name
    slug
  }
}
```

And the following mutation:

```graphql
mutation createTeam($team: CreateTeamInput!) {
  __typename
  createTeam(team: $team) {
    __typename
    id
    name
    slug
  }
}
```

I am not being able to compile the following cache

```typescript
const cache = cacheExchange({
  updates: {
    Mutation: {
      createTeam: (result, _args, cache) => {
        cache.updateQuery({ query: graphql.TeamsDocument }, (data) => {
          return {
            ...data,
            teams: [...data.teams, result.createTeam],
          }
        })
      },
    },
  },
})
```

Getting the following errors:

```
Argument of type '(data: Data | null) => { teams: any[]; } | { teams: any[]; __typename: string; _id?: string | number | null | undefined; id?: string | number | null | undefined; }' is not assignable to parameter of type '(data: Data | null) => Data | null'.
  Type '{ teams: any[]; } | { teams: any[]; __typename: string; _id?: string | number | null | undefined; id?: string | number | null | undefined; }' is not assignable to type 'Data | null'.
    Type '{ teams: any[]; }' is not assignable to type 'Data'.
      Property '__typename' is missing in type '{ teams: any[]; }' but required in type 'SystemFields'.
```

kitten:
The important part here is:

> Property '__typename' is missing in type '{ teams: any[]; }'

I'm assuming TypeScript gets a little confused here. It may help to handle the case where `data` is `null`. This may happen when `TeamsDocument` has never been queried before.

I also just spotted that we may have a small mistake in our typings here. We ask for a return value of `Data | null` which must contain `__typename` but internally we are and can assume that `__typename` at that location will be `"Query"`, so it's actually not necessary. I'll see whether I can fix that.

Once you handle `null` this means that you may still have to add a default of `__typename: 'Query'` for now.

JoviDeCroock:
Hey, this is part of the latest graphcache release! Thanks for reporting this issue

bkiac:
I still get this error with `urql@1.10.0` and `@urql/exchange-graphcache@3.0.2`.
This is my `Cache.updateQuery` type definition:
```ts
[...]

PR #538, opened 2020-02-13
fix updateQuery updater function type

JoviDeCroock:
## Summary

This should make life easier on people using custom updaters with mutations/subscriptions

Resolves: https://github.com/FormidableLabs/urql/issues/537

## Set of changes

- Moves away from Data return value for the `updater` function for `updateQuery` and goes to `DataFields` this type doesn't need `__typename` to be present
- Adds a fallback for `__typename` in the updater function.

### DIFF
--- a/.changeset/famous-ducks-smash.md
+++ b/.changeset/famous-ducks-smash.md
@@ -2,4 +2,4 @@
 '@urql/exchange-graphcache': patch
 ---
 
-Changes the type for the `updater` function of `cache.updateQuery` to `DataFields` this doesn't require `__typename` to be defined.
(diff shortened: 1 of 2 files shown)


.changeset/famous-ducks-smash.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/exchange-graphcache': patch
+---
+
+Changes the type for the `updater` function of `cache.updateQuery` to `DataFields` this doesn't require `__typename` to be defined.
```
kitten:
```suggestion
Update the `updater` function type of `cache.updateQuery` to have a return type of `DataFields` so that `__typename` does not need to be defined.
```

Issue #541, opened 2020-02-14
Weird subscription behaviour after upgrading to latest urql


zenios:
Hi All

I have noticed an odd behaviour with subscriptions and the latest urql. I assume it has something to do with teardown called multiple times.

So the issue is that for each subscription query, the client makes a subscriptions and then immediately teardown 

After that the normal subscription execution is followed

This is worst when in StrictMode. 
For each subscription two subscriptions are created and immediately teardown and finally continues onto executing the subscription normally

See screenshots attached.

*without strict mode*
![without_strict_mode](https://user-images.githubusercontent.com/89569/74526766-bdd17e80-4f2c-11ea-99fe-3b70f42f0bb5.png)

*with strict mode*
![with_strict_mode](https://user-images.githubusercontent.com/89569/74526768-bf02ab80-4f2c-11ea-9f32-ed2ac26408ce.png)

kitten:
Yes, your observation is 100% and we intentionally let the tear down through to restart the operation later on on purpose.

Is this causing any issues? If not we’ll keep things as is. Alternatively, I can try to add a delay to the start of the subscription (which is what we’ve done in the `fetchExchange`)

zenios:
Right now it does not cause any issues assuming that you don't take into consideration the multiple stop operations coming to the server.

My only concern is what happens when the tear down does not arrive so fast and the start that is going to be "tear down" later actually reaches the server

We will have multiple subscriptions opening and closing without cause.

kitten:
The first `teardown` after mount arrives synchronously so we’re actively avoiding timing issues 🙌

I’ll look into whether we can add a delay to avoid quirks in the future for unstable websocket connections / subscriptions implementations 

This is the change we’ve made to the `fetchExchange`, so if possible it’ll likely look similar: https://github.com/FormidableLabs/urql/blob/a1935c985529fd8b549b5d5683e7efd919ab016a/packages/core/src/exchanges/fetch.ts#L115

zenios:
I just did something similar on my own subscription exchange and it works
fine now.

You might ask why i have my own subscription exchange.

The only reason is because the official urql exchange does not support
using subscription exchange also for queries and mutations. ( Not sure if
that would be something interesting to be merged upstream)

Other than that my subscription exchange and the official are 100%
identical.

kitten:
We can totally look into implementing this in the official subscription exchange 👍 I’m figuring that  this would be a rather minor addition in terms of amount of code

[...]

PR #542, opened 2020-02-14
Fix/dupe teardown subscription

JoviDeCroock:
## Set of changes

- adds a microtick delay (and a guard) for subscribing

Resolves https://github.com/FormidableLabs/urql/issues/541

### DIFF
--- a/.changeset/selfish-scissors-float.md
+++ b/.changeset/selfish-scissors-float.md
@@ -2,4 +2,4 @@
 '@urql/core': patch
 ---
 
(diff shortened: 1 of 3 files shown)


.changeset/selfish-scissors-float.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/core': patch
+---
+
+Ensure we don't dispatch network-requests on getCurrentValue or subscriptions that will tear down early
```
kitten: Let’s make sure to keep this very specific, like;

> Add one-tick delay to the `subscriptionExchange` to prevent unnecessary early tear downs.

PR #544, opened 2020-02-14
Add option to let subscriptionExchange handle all operations

### LINKED ISSUE #543: Enhanced subscription exchange
Hi All

As discussed on #541  please find below an enhanced implementation of subscription exchange that:

1. Allow using subscription exchange for queries and mutations
2. Allow passing fetchOptions just like fetch exchange
3. Fixes the problem with teardown on uninitialised subscription (#541)

Check it out and let me know. Thanks

```javascript
import { print } from "graphql";

import {
  filter,
  make,
  merge,
  mergeMap,
  pipe,
  share,
  Source,
  takeUntil
} from "wonka";

import {
  Exchange,
  ExecutionResult,
  makeErrorResult,
  makeResult,
  Operation,
  OperationContext,
  OperationResult
} from "urql";

export interface ObserverLike<T> {
  next: (value: T) => void;
  error: (err: any) => void;
  complete: () => void;
}

/** An abstract observable interface conforming to: https://github.com/tc39/proposal-observable */
export interface ObservableLike<T> {
  subscribe(
    observer: ObserverLike<T>
  ): {
    unsubscribe: () => void;
  };
}

export interface SubscriptionOperation {
  query: string;
  variables?: object;
  key: string;
  context: OperationContext;
}

export type SubscriptionForwarder = (
  operation: SubscriptionOperation
) => ObservableLike<ExecutionResult & { extensions?: Record<string, any> }>;

/** This is called to create a subscription and needs to be hooked up to a transport client. */
export interface SubscriptionExchangeOpts {
  // This has been modelled to work with subscription-transport-ws
  // See: https://github.com/apollographql/subscriptions-transport-ws#requestoptions--observableexecutionresult-returns-observable-to-execute-the-operation
  forwardSubscription: SubscriptionForwarder;

  // Allow selecting which operations should be handled by this exchange.
  // Anything that is not handled should be forwarded to the next exchange (most probably fetch)
  shouldHandleOperation: (operation: Operation) => boolean;
}
export const subscriptionExchange = ({
  forwardSubscription,
  shouldHandleOperation
}: SubscriptionExchangeOpts): Exchange => ({ client, forward }) => {
  const createSubscriptionSource = (
    operation: Operation
    const {
      context: { fetchOptions, ...restContext }
    } = operation;

    const extraOptions =
      typeof fetchOptions === "function" ? fetchOptions() : fetchOptions || {};

    // This excludes the query's name as a field although subscription-transport-ws does accept it since it's optional
    const observableish = forwardSubscription({
      key: operation.key.toString(36),
      query: print(operation.query),
      variables: operation.variables,
      context: restContext,
      ...extraOptions
    });

    return make<OperationResult>(({ next, complete }) => {
      let isComplete = false;
      let sub:
        | {
            unsubscribe: () => void;
          }
        | undefined = undefined;

      // Fixes https://github.com/FormidableLabs/urql/issues/541
      Promise.resolve().then(() => {
        if (isComplete) return;
        sub = observableish.subscribe({
          next: result => next(makeResult(operation, result)),
          error: err => next(makeErrorResult(operation, err)),
          complete: () => {
            if (!isComplete) {
              isComplete = true;
              client.reexecuteOperation({
                ...operation,
                operationName: "teardown"
              });
              complete();
            }
          }
        });
      });

      return () => {
        isComplete = true;
        if (sub !== undefined) sub.unsubscribe();
      };
    });
  };

  return ops$ => {
    const subscriptionResults$ = pipe(
      sharedOps$,
      filter(shouldHandleOperation),
      mergeMap(operation => {
        const { key } = operation;
        const teardown$ = pipe(
          sharedOps$,
          filter(op => op.operationName === "teardown" && op.key === key)
        );

        return pipe(createSubscriptionSource(operation), takeUntil(teardown$));
      })
    );

    const forward$ = pipe(
      sharedOps$,
      filter(op => !shouldHandleOperation(op)),
      forward
    );

    return merge([subscriptionResults$, forward$]);
  };
};

```

kitten:
This adds an option to the `subscriptionExchange` to let it handle `query` and `mutation` operations as well.

Resolve #543

### DIFF
--- a/packages/core/src/exchanges/subscription.ts
+++ b/packages/core/src/exchanges/subscription.ts
@@ -106,7 +106,7 @@ export const subscriptionExchange = ({
   const isSubscriptionOperation = (operation: Operation): boolean => {
     const { operationName } = operation;
     return operationName === 'subscription' ||
-      (enableAllOperations &&
+      (!!enableAllOperations &&
         (operationName === 'query' || operationName === 'mutation'));
   };
 
--- a/packages/core/src/exchanges/subscription.ts
+++ b/packages/core/src/exchanges/subscription.ts
@@ -52,13 +52,14 @@ export interface SubscriptionExchangeOpts {
   // This has been modelled to work with subscription-transport-ws
   // See: https://github.com/apollographql/subscriptions-transport-ws#requestoptions--observableexecutionresult-returns-observable-to-execute-the-operation
   forwardSubscription: SubscriptionForwarder;
-}
 
-const isSubscriptionOperation = (operation: Operation) =>
-  operation.operationName === 'subscription';
+  /** This flag may be turned on to allow your subscriptions-transport to handle all operation types */
+  enableAllOperations?: boolean;
+}
 
 export const subscriptionExchange = ({
   forwardSubscription,
+  enableAllOperations,
 }: SubscriptionExchangeOpts): Exchange => ({ client, forward }) => {
   const createSubscriptionSource = (
     operation: Operation
@@ -100,6 +101,13 @@ export const subscriptionExchange = ({
         if (sub) sub.unsubscribe();
       };
     });
+  };
+
+  const isSubscriptionOperation = (operation: Operation): boolean => {
+    const { operationName } = operation;
+    return operationName === 'subscription' ||
+      (enableAllOperations &&
+        (operationName === 'query' || operationName === 'mutation'));
   };
 
   return ops$ => {
--- a/.changeset/tiny-goats-type.md
+++ b/.changeset/tiny-goats-type.md
@@ -1,0 +1,5 @@
+---
+'@urql/core': patch
+---
+
+Add enableAllOperations option to subscriptionExchange to let it handle queries and mutations as well.


packages/core/src/exchanges/subscription.ts
```diff
@@ -52,13 +52,14 @@ export interface SubscriptionExchangeOpts {
   // This has been modelled to work with subscription-transport-ws
   // See: https://github.com/apollographql/subscriptions-transport-ws#requestoptions--observableexecutionresult-returns-observable-to-execute-the-operation
   forwardSubscription: SubscriptionForwarder;
-}
 
-const isSubscriptionOperation = (operation: Operation) =>
-  operation.operationName === 'subscription';
+  /** This flag may be turned on to allow your subscriptions-transport to handle all operation types */
+  enableAllOperations?: boolean;
```
zenios: Why don't we use filter function instead? There might be cases (I cannot think of any) where someone wants to pass only subscriptions and mutation via subscription exchange.

kitten: I don't think it makes sense to anticipate use-cases that we can't imagine. To a certain extent we tell people that "forking" our exchanges or making more specific ones is fine.

So overall we only implement what we can see as a use-case right now, instead of bloating our API surface 👍  Hope that makes sense to you though

zenios:
Also it would be nice if we could pass fetchOptions as well. Right now i am using it to pass an authentication header

kitten:
@zenios Kind of unrelated, but I think that's already possible.

https://github.com/FormidableLabs/urql/blob/73bdd698f72a17840abcc4f38708fda23d1aaee2/packages/core/src/exchanges/subscription.ts#L46-L48

You receive the entire operation in `forwardSubscription`, so you can just use `operation.context.fetchOptions`

zenios:
> @zenios Kind of unrelated, but I think that's already possible.
> 
> https://github.com/FormidableLabs/urql/blob/73bdd698f72a17840abcc4f38708fda23d1aaee2/packages/core/src/exchanges/subscription.ts#L46-L48
> 
> You receive ~the entire~ the masked operation with `context` in `forwardSubscription`, so you can just use `operation.context.fetchOptions`

Not if fetchOptions is a function

kitten:
@zenios Like in `fetchExchange` you can just run the function though

https://github.com/FormidableLabs/urql/blob/8877e0fbc1169405e385579934fc68e1daa49942/packages/core/src/exchanges/fetch.ts#L81-L84

zenios:
> @zenios Like in `fetchExchange` you can just run the function though
> 
> https://github.com/FormidableLabs/urql/blob/8877e0fbc1169405e385579934fc68e1daa49942/packages/core/src/exchanges/fetch.ts#L81-L84

If i use this subsciption exchange and fetchOptions is a function on my server fetchOptions will arrive as [object]

kitten:
@zenios That sounds like a different problem. You just need to call the function. If you're talking about serialisation, we don't plan or account for `context` ever to be sent to the server, so this shouldn't happen 😅 

Either way, if you need an answer to this feel free to open an issue, where we can discuss this separately. :+1:

PR #551, opened 2020-02-20
Fix cache-only for the caches

JoviDeCroock:
## Summary

`cache-only` currently dispatches a network-call when the result isn't in cache.

## Set of changes

- Add a `cache-only` filter so the operation does not get forwarded in doc-cache.
- Moves the `graphcache` operations outside of `forward` so we can add the same filter as introduced in doc-cache

Issue: https://github.com/FormidableLabs/next-urql/issues/39#issuecomment-588901815

### DIFF
--- a/.changeset/fluffy-guests-begin.md
+++ b/.changeset/fluffy-guests-begin.md
@@ -3,4 +3,4 @@
 '@urql/core': patch
 ---
 
-Prevents `cache-only` operations from getting forwarded outside of the cache when no result was found
+Fix `cache-only` operations being forwarded and triggering fetch requests
--- a/.changeset/fluffy-guests-begin.md
+++ b/.changeset/fluffy-guests-begin.md
@@ -1,0 +1,6 @@
+---
+'@urql/exchange-graphcache': patch
+'@urql/core': patch
+---
+
+Prevents `cache-only` operations from getting forwarded outside of the cache when no result was found
--- a/exchanges/graphcache/src/cacheExchange.ts
+++ b/exchanges/graphcache/src/cacheExchange.ts
@@ -305,7 +305,7 @@ export const cacheExchange = (opts?: CacheExchangeOpts): Exchange => ({
     const cacheOps$ = pipe(
       cache$,
       filter(res => res.outcome === 'miss'),
-      map(res => addCacheOutcome(res.operation, res.outcome))
+      map(res => addCacheOutcome(res.operation, res.outcome)),
     );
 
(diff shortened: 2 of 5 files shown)


exchanges/graphcache/src/cacheExchange.ts
```diff
@@ -342,15 +342,15 @@ export const cacheExchange = (opts?: CacheExchangeOpts): Exchange => ({
...
-        ])
-      ),
+      merge([
+        pipe(
+          inputOps$,
+          filter(op => !isCacheableQuery(op))
+        ),
+        cacheOps$,
+      ]),
+      filter(op => !(op.operationName === 'query' && op.context.requestPolicy === 'cache-only')),
```
andyrichardson: Is it fair to assume we'll never get a non-query operation with a cachePolicy?

```suggestion
      filter(op => op.context.requestPolicy !== 'cache-only'),
```

packages/core/src/exchanges/cache.ts
```diff
@@ -86,6 +86,7 @@ export const cacheExchange: Exchange = ({ forward, client }) => {
         ),
       ]),
       map(op => addMetadata(op, { cacheOutcome: 'miss' })),
+      filter(op => !(op.operationName === 'query' && op.context.requestPolicy === 'cache-only')),
```
andyrichardson: Woah, this was a bug in the default cache too? 😱

kitten:
```suggestion
      filter(op => op.operationName !== 'query' || op.context.requestPolicy !== 'cache-only'),
```

.changeset/fluffy-guests-begin.md
```diff
@@ -0,0 +1,6 @@
+---
+'@urql/exchange-graphcache': patch
+'@urql/core': patch
+---
+
+Prevents `cache-only` operations from getting forwarded outside of the cache when no result was found
```
kitten: Let's make sure this is user-friendly without implementation details?

```suggestion
Fix `cache-only` operations being forwarded and triggering fetch requests
```

subpopular:
I am still seeing a network request with `{requestPolicy 'cache-only'}` on my query on latest version.

kitten:
@subpopular I'm not quite sure what you're looking for but it sounds like maybe you'll want to open a new issue with a question? :) https://github.com/FormidableLabs/urql/issues/new?labels=question+%F0%9F%99%8B&template=Question.md&title=

PR #552, opened 2020-02-20
Add "basics" section for docs

JoviDeCroock:
## Summary

Adds the `basics` section to our docs with sections for:
- setting up
- querying
- mutating

I've purposely opened with a framework title so we can be future-proof in that regards, something I was wondering is whether or not `###` will be shown in the `Legend`.

This is mainly a first draft and helps dictate the way we want to convey information, I've found that the loose approach works well (read: sparks joy)

Related to: #548


docs/basics/mutating-data.md
```diff
@@ -0,0 +1,53 @@
...
+    mutation ($id: ID!, $title: String!) {
+      updateTodo (id: $id, title: $title) {
+        id
+      }
+    }
+  `);
+}
+```
+
+Similar to the `useQuery` output we see a tuple, the first being our result
```
wgolledge: I've tried rewording this a tiny bit, what do you think? 🙂 

Similar to the `useQuery` output, `useMutation` returns a tuple. The first item in the tuple being our result
containing: `fetching`, `error`, and `data`. At this point in time, no mutation has been performed. To mutate the data we first have to invoke the second item in the tuple - the function here named `updateTodo`.

docs/basics/setting-up-the-client.md
```diff
@@ -0,0 +1,79 @@
+---
+title: setting-up-the-client
+order: 0
+---
+
+## Installation
+
+Installing `urql` is as quick as you'd expect. Firstly, install it
+with your package manager of choice, this installation is specific for React:
```
wgolledge: Nit:

Firstly, install it with your package manager of choice. Note: this installation is specific to React.

docs/basics/mutating-data.md
```diff
@@ -0,0 +1,53 @@
+---
+title: mutating-data
+order: 2
+---
+
```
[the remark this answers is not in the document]
wgolledge: Actually I guess this is up for debate - we could just have the title on the sidebar and not in the content

PR #553, opened 2020-02-20
Part 1: New documentation

JoviDeCroock:
## Summary

Adds the `basics` section to our docs with sections for:
- setting up
- querying
- mutating

I've purposely opened with a framework title so we can be future-proof in that regards, something I was wondering is whether or not `###` will be shown in the `Legend`.

This is mainly a first draft and helps dictate the way we want to convey information, I've found that the loose approach works well (read: sparks joy)

Related to: #548


docs/basics/mutating-data.md
```diff
@@ -0,0 +1,54 @@
+---
+title: Mutations
+order: 2
+---
+
+## Mutate your data
+
+Now that we know how to query our data we'll also need to know
```
andyrichardson: Might be worth keeping docs isolated from other sections given how pages are separate. Straight to the point might be more useful.

> # Mutations
>Mutations can be triggered in Urql in one of two ways
>
> - directly via the client using `executeMutation`
> - indirectly using an adapter - (e.g. `useQuery`)
> 
> ## Client (executeMutation)
> ...
>
> ## React/Preact (useMutation)
>
> ...
>
> ## JQuery ($mutation)
>
> ...
>

docs/basics/mutating-data.md
```diff
@@ -0,0 +1,54 @@
...
+```
+
+Similar to the `useQuery` output, `useMutation` returns a tuple. The first item in the tuple being our `result`
+containing: `fetching`, `error`, and `data`. At this point in time, no mutation has been performed.
+To mutate the data we first have to invoke the second item in the tuple - the function here named `updateTodo`.
+
+### Using the result
+
+When calling this function we have two ways of getting the response from the server,
+we can get it from `updateTodoResult` or we chain a `.then` on our `updateTodo` call.
```
andyrichardson: Minor wording change!


```suggestion
we can get it from `updateTodoResult` or we can await the promise returned from our mutation trigger function.
```

Also, is it fair to assume most will be using async/await?

kitten: Agreed, let’s not make any assumptions and in general just state that it returns a result in the expected, identical shape 👍

It’s a minor detail but whenever possible it’s good to state all options and details of an API of ours whenever possible (we can add cross-references / links to API / other pages later) but stop whenever an explanation dives outside of urql / GraphQL

PR #554, opened 2020-02-20
Add testing/fixture docs

### LINKED ISSUE #550: How to write tests with Provider
Is there any documentation for how to write tests for components that make calls to `useQuery` and `useMutation`? 

What should I use in tests since the entire app is wrapped in urql's `Provider`? In tests, I need access to that context, but I need the ability to provide it mocked data. `apollo-client` offers a `MockedProvider` which takes a `mocks` prop that I can provide mocked data, but I'm not quite sure what to do in the case of urql.

andyrichardson:
## About

Docs on how to test + create fixtures with Urql!

It covers
 - Testing calls to the client
 - Forcing states (fetching, response, error, etc)
 - Handling multiple hooks
 - Simulating multiple responses

**[View markdown](https://github.com/FormidableLabs/urql/blob/add-docs/docs/core/testing.md)**

## Notes

 - This is intended to cover all the info required to get testing (not a deep dive on how Wonka streams work)
 - The examples use Enzyme mounts but principles apply to any library (including those with generic names)
 - @bstro read my mind - Fix #550


docs/core/testing.md
```diff
@@ -0,0 +1,221 @@
...
+order: 6
+---
+
+<a name="testing"></a>
+
+# Testing
+
+When testing your components, you're likely going to want to check arguments and force different states from the Urql hooks, this is the place to get started.
+
+## Communicating with hooks
```
kitten: I think this section we can remove. That’d mean that we can explicitly leave a note in the above introduction saying: “This page will show some examples with `urql`’s React hook. But the same testing approach applies. Mocking the client is a solid way to test the behaviour of the rest of your codebase.”

Then this page is pretty future proof

docs/core/testing.md
```diff
@@ -0,0 +1,221 @@
+---
+title: Testing
+order: 6
+---
+
+<a name="testing"></a>
```
kitten: This `<a>` tag isn't needed anymore btw. Also either once `basics-for-docs` is merged (or we can rebase) this is the file where this page should go: https://github.com/FormidableLabs/urql/blob/basics-for-docs/docs/advanced/testing.md

bstro:
I'm atempting to follow these instructions right now and am finding that `<Provider>` does not typecheck when I give it the following client object: 

```
const mockClient = {
  executeQuery: () => {},
  executeMutation: () => {},
  executeSubscription: () => {},
}
```

I am calling Provider like so: ```
    <Provider value={mockClient}>
      …
    </Provider>
```

and the typescript error is:
```
Type '{ executeQuery: () => void; executeMutation: () => void; executeSubscription: () => void; }' is missing the following properties from type 'Client': url, exchange, suspense, preferGetMethod, and 15 more.  TS2740
```

**Comment by andyrichardson** [comment]
@bstro this is because the mocks in the example don't match the actual client.

The way to go is to cast this as an any. I'll add a note about this to the docs.

```tsx
const mockClient: any = {
  executeQuery: () => {},
  executeMutation: () => {},
  executeSubscription: () => {},
}
```

docs/core/testing.md
```diff
@@ -0,0 +1,221 @@
...
+});
+```
+
+## Forcing states
+
+For testing render output, or creating fixtures, you may want to force the state of your components.
+
+### Fetching
+
+Fetching states can be simulated by returning a stream which never returns. Wonka provides a utility for this aptly called `never`.
```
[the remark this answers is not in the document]
andyrichardson:
```suggestion
Fetching states can be simulated by returning a stream which never returns. Wonka provides a utility for this, aptly called `never`.
```

docs/core/testing.md
andyrichardson:
```suggestion
```

PR #556, opened 2020-02-25
Part 2: New Documentation

kitten:
Related to #548


docs/basics/queries.md
```diff
@@ -0,0 +1,194 @@
...
+const Todos = () => {
+  const [result, reexecuteQuery] = useQuery({
+    query: TodosQuery,
+  });
+
+  const { data, fetching, error } = result;
+
+  if (fetching) {
+    return <p>Loading...</p>;
+  } else if (error) {
```
andyrichardson: Need to get rid of this unnecessary else - there's quite a few in the codebase also

```suggestion
  }
  
  if (error) {
```

kitten: I’m curious, do you think it’s an issue with readability? I assumed since it’s equivalent both is fine 🤗

andyrichardson: Yes I think adding unnecessary conditional logic an issue with readability

kitten:
The staging deployment seems to be working now. Here’s the link to this PR’s deployed docs 🌟 https://formidable-com-urql-staging-556.surge.sh/docs

PR #557, opened 2020-02-28
Move next-urql to monorepo

parkerziegler:
**NOTE:** Please don't squash & merge, but do a merge commit.


parkerziegler:
@JoviDeCroock @kitten and I cranked this out on Friday afternoon. I think we should be good to merge, but let me know if you'd like any additional bits brought in here. We used `git subtree` for this (thanks for showing me that Phil!) so it should preserve the whole commit history. A few snags getting the build working with the shared Rollup setup but otherwise looking good. Once we merge I'll archive the `next-urql` repo and transfer open issues over here.

packages/next-urql/README.md
```diff
@@ -0,0 +1,226 @@
...
+
+### Installation
+
+Install `next-urql` along with its `peerDependencies`.
+
+```sh
+npm install --save next-urql urql react-is isomorphic-unfetch
+```
+
+`react-is`, `styled-components`, and `isomorphic-unfetch` help to support server-side `Suspense` with `react-ssr-prepass`. This assumes you have followed the basic installation steps for `urql` [here](https://github.com/FormidableLabs/urql#installation).
```
JoviDeCroock: is `styled-components`  needed since it's not in the above install

packages/next-urql/examples/1-with-urql-client/.gitignore
```diff
@@ -0,0 +1,25 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
```
JoviDeCroock: We might want to add these to the global gitignore instead to reduce files (same for the logo maybe use them from `/docs`)

packages/next-urql/README.md
kitten: Good point! That’s not required anymore

packages/next-urql/examples/1-with-urql-client/.gitignore
kitten: I think it’s fine to make an exception for examples until those are cleaned up maybe? 🤔

packages/next-urql/README.md
parkerziegler: Good catch, will remove it. It used to be needed in previous versions of `react-ssr-prepass`, but no longer!

packages/next-urql/examples/1-with-urql-client/.gitignore
parkerziegler: ✅ Agreed, will do.

PR #558, opened 2020-02-29
Commutative updates

JoviDeCroock:
## Summary

This PR aims at making applying updates predictable, let's explain this with a scenario rather than abstract terms.

Imagine we are 3 queries that are related, 1 - 2 - 3 in that order. These can be list, details, ... If these arrive as a server response in the same order 1 - 2 - 3 this will be the outcome we desire but what happens if this happens out-of-sync? We could start overriding properties/methods we don't want.

In this PR we'll apply responses that have queries queued up before them as optimistic responses. This way we can when all preceding queries come in merge them in order.

## Set of changes

- queries are applied in order
- when an out of order query returns it's written as an optimistic layer

### DIFF
--- a/exchanges/graphcache/src/cacheExchange.ts
+++ b/exchanges/graphcache/src/cacheExchange.ts
@@ -260,16 +260,20 @@ export const cacheExchange = (opts?: CacheExchangeOpts): Exchange => ({
         if (index === 0) {
           writeDependencies = write(store, operation, data).dependencies;
           inFlightKeys.delete(operation.key);
-          // TODO: can this have an optimistic entry, to verify
-          clearOptimistic(data, operation.key);
+          if (hasOptimisticKey(store.data.records, operation.key)) clearOptimistic(store.data, operation.key);
+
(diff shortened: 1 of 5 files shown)


exchanges/graphcache/src/cacheExchange.ts
```diff
@@ -244,7 +245,59 @@ export const cacheExchange = (opts?: CacheExchangeOpts): Exchange => ({
...
+              index = inFlight.length;
+            }
+          }
+        } else {
+          /**
+           * When we're not a tthe front of the queue we need to treat this update as an
+           * optimistic layer since this update could always be subject to overriding because
+           * of things earlier in the queue.
+           */
+          writeDependencies = write(store, operation, data, operation.key).dependencies;
```
JoviDeCroock: A potential problem here is the following 1 - 2 - 3 where 3 comes in first, applies layer, 2 comes in, applies layer. This will make it so that the data of 2 has prio over 3. This fixes itself when 1 resolves

PR #563, opened 2020-03-03
Fix examples in next-urql.

parkerziegler:
Summary

This PR fixes the `next-urql` examples, which were broken in the course of the migration. The root cause appears to have to do with the working directory of the `build` step that the examples perform on `prestart`.

When running `yarn start` in, say, `packages/next-urql/examples/1-with-urql-client`, we used to run `npm run build --prefix ../../` to build `next-urql` before starting the example. However, this process doesn't change the `cwd`, so when we attempt to run `rollup` we can't actually find it in the path 😱 I believe this same problem is present in the `react-urql` examples currently as well.

image

I also fixed a pretty big bug I introduced in `next-urql` 0.3.0 where `withUrql.getInitialProps` returns `undefined` for client-side navigation if no `getInitialProps` is implemented on the wrapped page 😱 This should go out ASAP!

## Set of changes

1. Move to using `cd` in our `package.json` scripts for `prestart`, i.e.

```sh
cd ../../ && yarn build && cd examples/1-with-urql-client
```
2. Update installation instructions in examples to note needing to build `core` and `react-urql`.
3. Update `next.config.js` for all examples to point to root `node_modules`.
4. Ensure `pageProps` is initialized to an empty object in `withUrqlClient`'s `getInitialProps`.


packages/next-urql/src/with-urql-client.tsx
```diff
@@ -67,7 +67,7 @@ export function withUrqlClient(
       }
 
       // Run the wrapped component's getInitialProps function.
-      let pageProps;
+      let pageProps = {} as any;
```
parkerziegler: Here's the bugger that broke everything 😓

ryan-gilb: nice catch. i think this used to be typed as the generic `IP` over `withUrqlClient`, could we use that?

parkerziegler: Yeah I relaxed the types a bit in 0.3.0 since we now properly support wrapping `_app.js`, which made dealing with that generic a _nightmare_. If anyone _really_ wants to fix those types I'd be happy to review a PR, but creating a nice abstraction around both `_app.js` and traditional page components is quite tough. Just the reality of Next's funky API.

PR #481, opened 2019-12-10
Add retryExchange to core

mxstbr:

PR #564, opened 2020-03-03
Add retryExchange package

wgolledge:
## Summary

Add a new retryExchange based on the previous [implementation](https://github.com/FormidableLabs/urql/pull/481). This version adds some additional options.

It's an exchange factory of type `Options => Exchange`.

Accepted options:
- `initialDelayMs` - the minimum delay between retries, defaults to 1000
- `maxDelayMs` - the maximum delay that can be applied to an operation
- `randomDelay` - whether to apply a random delay to protect against thundering herd, defaults to true
- `maxNumberAttempts` - the maximum number of attempts to retry any given operation, defaults to Infinity
- `retryIf` - optional function to apply to errors to determine whether they should be retried

## Points of note

- I've created this as an external package rather than extending core
- Can we think of any other options that could be handy?


exchanges/retry/src/retryExchange.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  return ({ forward }) => ops$ => {
+    const sharedOps$ = pipe(ops$, share);
+    const { source: retry$, next: nextRetryOperation } = makeSubject<
+      OperationWithRetry
+    >();
+
+    const retryWithBackoff$ = pipe(
+      retry$,
+      mergeMap((op: OperationWithRetry) => {
```
RodolfoSilva: I believe is better avoid name abbreviations.

kitten: We use the `op` for "operation" abbreviation in multiple exchanges in the codebase already. The type annotation makes the intent clear in most cases anyway :)

exchanges/retry/package.json
```diff
@@ -0,0 +1,58 @@
+{
+  "name": "@urql/exchange-retry",
+  "version": "1.8.4",
```
kitten: You may want to start at either `0.1.0` or `1.0.0` and adjust the changelog to remove that `1.8.4` entry :)

exchanges/retry/package.json
```diff
@@ -0,0 +1,58 @@
...
+  },
+  "keywords": [
+    "urql",
+    "graphql client",
+    "formidablelabs",
+    "exchanges",
+    "react",
+    "retry"
+  ],
+  "main": "dist/urql-exchange-retry.cjs.js",
```
JoviDeCroock: Should we add an exports field similar to urql-react?

exchanges/retry/src/retryExchange.ts
```diff
@@ -0,0 +1,118 @@
...
+      result$,
+      // We let through all non-network-failed results
+      filter(res => !res.error || !res.error.networkError)
+    );
+
+    const failedResult$ = pipe(
+      result$,
+      // Only retry if there was a network error and the error passes the
+      // conditional retryIf function (if passed)
+      filter(res => !!(res.error && (!retryIf || retryIf(res.error)))),
```
JoviDeCroock: Let’s limit to networkError when there’s no retryIf, wdyt @kitten?

exchanges/retry/package.json
kitten: That isn’t strictly speaking required unless we do that for all packages. It’s only required for `/extras` right now to make sure that Node does the right thing.

So we should probably file a ticket to update `prepare` and all packages

exchanges/retry/src/retryExchange.ts
kitten: Yep, makes sense 👍

exchanges/retry/src/retryExchange.ts
```diff
@@ -0,0 +1,118 @@
...
+      // Send failed responses to be retried by calling next on the retry$ subject
+      tap(op => nextRetryOperation(op.operation)),
+      // Only let through the first failed response
+      filter(res => !res.operation.context.retryDelay)
```
kitten: I’m not 100% sure this is right; so we’re letting through the first error is what the comment says rather than the last retried error in case we run out of attempts?

JoviDeCroock:
I just realised, we should probably check for `pollInterval` being present so we don't retry those

exchanges/retry/src/retryExchange.ts
```diff
@@ -0,0 +1,118 @@
...
+  takeUntil,
+} from 'wonka';
+import { Exchange, Operation, CombinedError } from '@urql/core';
+
+interface RetryExchangeOptions {
+  initialDelayMs?: number;
+  maxDelayMs?: number;
+  randomDelay?: boolean;
+  maxNumberAttempts?: number;
+  // Optional function to determine whether an error should be retried
```
andyrichardson:
```suggestion
  /** Conditionally determine whether an error should be retried */
```

^ This will make the prompt show up on your editor when you hover over it.

exchanges/retry/src/retryExchange.ts
andyrichardson:
```suggestion
      filter(res => res.error && (!retryIf || retryIf(res.error))),
```

I'm guessing that filter will automatically evaluate the truthiness for us?

exchanges/retry/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "../../tsconfig.json",
+  "include": ["src"],
+  "compilerOptions": {
+    "baseUrl": "./",
+    "paths": {
```
andyrichardson: This is cool, but does our bundling process currently support aliasing?

I stopped using aliases after getting tired of writing alias maps in multiple places (tsconfig and webpack config)

andyrichardson:
Looks siccc 🔥

exchanges/retry/src/retryExchange.ts
andyrichardson: I agree with this, I can't see having the latest error being a problem.

wgolledge: I _think_ this should now be fixed

exchanges/retry/tsconfig.json
wgolledge: Everything seems fine as far as I can see - happy to swap them to be relative paths if preferred though

exchanges/retry/README.md
```diff
@@ -0,0 +1,67 @@
...
+import { createClient, dedupExchange, cacheExchange, fetchExchange } from 'urql';
+import { retryExchange } from '@urql/exchange-retry';
+
+const client = createClient({
+  url: 'http://localhost:1234/graphql',
+  exchanges: [
+    dedupExchange,
+    cacheExchange,
+    fetchExchange,
+    retryExchange(options), // Use the retryExchange factory to add a new exchange
```
JoviDeCroock: I'd define some basic options object since there's non visible in this context, imo a getting started can just be `copy-paste`.

exchanges/retry/src/retryExchange.ts
```diff
@@ -0,0 +1,136 @@
...
+}
+
+export interface OperationResultWithRetry extends OperationResult {
+  operation: OperationWithRetry;
+}
+
+export const retryExchange = ({
+  initialDelayMs,
+  maxDelayMs,
+  randomDelay = true,
```
JoviDeCroock: can we initialise this the same was as the others? Defaulting in function arguments is also very expensive. [Proof](https://babeljs.io/en/repl#?browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYewdgzgLgBAHjAvDAFAbwFAxgJwIZgAmIAtkjAKwA0WMANgKYQQBKBxJNAvgJRIB8MTNkbM2RUuVGt2kgD5yYANgDctHAygBXHGFyyyAKnpMZEkmq5A&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Cenv&prettier=false&targets=&version=7.8.7&externalPlugins=)

wgolledge: Wow. It seems like default params often causes more trouble than they're worth IMO

PR #565, opened 2020-03-03
(graphcache) - Implement Commutative Query Results

kitten:


exchanges/graphcache/src/cacheExchange.test.ts
```diff
@@ -841,3 +851,60 @@ it('reexecutes query and returns data on partial result', () => {
     'partial'
   );
 });
+
+it.only('applies results that come in out-of-order commutatively and consistently', () => {
```
JoviDeCroock: You left a .only in here

Issue #559, opened 2020-02-29
Best practice to update cache on mutation insert/delete for large applications?


marbemac:
This is more of a question related to best practice for create/delete mutations. I noticed that the urql graph cache does not update automatically when things are added / removed via a mutation. After exploring the documentation and examples, I was able to figure out a solution that works, but I'm not sure if it's the best way. So before doing this for all of the mutations across our application, I figured I'd check in here :).

Some background - we're using https://hasura.io to drive much of our graphql API, and their insert/delete mutations work with multiple objects (similar to batching). Here are some example mutations / queries (note they're all returning `__typename: star`).

```graphql
mutation StarNode($projectId: Int!, $nodeId: Int!) {
  createStars(objects: [{ projectId: $projectId, nodeId: $nodeId }]) {
    returning {
      id
      __typename # of type star
    }
  }
}

mutation UnstarNode($projectId: Int!, $nodeId: Int!) {
  deleteStars(where: { nodeId: { _eq: $nodeId }, projectId: { _eq: $projectId } }) {
    returning {
      id
      __typename # of type star
    }
  }
}

query GetNodeStar($projectId: Int!, $nodeId: Int!) {
  stars(where: { nodeId: { _eq: $nodeId }, projectId: { _eq: $projectId } }) {
    id
    __typename # of type star
  }
}
```

After some fiddling, this is what I came up with. Is this recommended way to make sure the UI / cache updates in response to insert/delete mutations? It works, but it feels like it won't scale well in a large application that has lots of different queries, with different variables, etc - particularly because we have to make sure to pass the originally used `variables`. The creation of this cache is quite far away from the queries happening in our application, so it feels like it will be easy to make a mistake and not cover all the variable combinations, etc.

```ts
import { cacheExchange } from '@urql/exchange-graphcache';

// @ts-ignore
clientCache = cacheExchange({
  updates: {
    Mutation: {
      createStars: (result: StarNodeMutation, args, cache, info) => {
        cache.updateQuery(
          { query: GetNodeStarDocument, variables: info.variables },
          // @ts-ignore
          (data: GetNodeStarQuery | null) => {
            if (data) data.stars.push(...(result.createStars?.returning || []));

            return data;
          },
        );
      },
      deleteStars: (result: UnstarNodeMutation, args, cache, info) => {
        cache.updateQuery(
          { query: GetNodeStarDocument, variables: info.variables },
          // @ts-ignore
          (data: GetNodeStarQuery | null) => {
[...]

PR #566, opened 2020-03-03
(graphcache) - Invalidating entities

JoviDeCroock:
## Summary

For use cases described in [this issue](https://github.com/FormidableLabs/urql/issues/559) an `invalidate` method seemed very suitable. Now we can remove an entity (with its properties) from the cache.

This has an added benefit that it removes links, if the links has no more ties with other entities it will get garbage collected.

> Note that schema awareness can cause a `null` in your data if this is marked as optional (partial result)

## Set of changes

- Adds `cache.invalidate`

### DIFF
--- a/exchanges/graphcache/src/store/store.ts
+++ b/exchanges/graphcache/src/store/store.ts
@@ -156,14 +156,15 @@ export class Store implements Cache {
   }
 
   invalidateEntity(entity: Data | string) {
-    const entityKey = typeof entity === 'string' ? entity : this.keyOfEntity(entity);
+    const entityKey =
+      typeof entity === 'string' ? entity : this.keyOfEntity(entity);
 
     if (!entityKey) {
       invariant(
         entityKey,
         `Can't generate a key for invalidateEntity(${entity}),
         You need to pass in a valid key (__typename:id) or an object with the "__typename" property and an "id" or "_id" property.`,
-        19,
+        19
       );
     }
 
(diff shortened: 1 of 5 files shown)


exchanges/graphcache/src/store/store.ts
```diff
@@ -154,11 +155,39 @@ export class Store implements Cache {
...
   }
 
+  invalidateEntity(entity: Data | string) {
+    const entityKey =
+      typeof entity === 'string' ? entity : this.keyOfEntity(entity);
+
+    if (!entityKey) {
+      invariant(
+        entityKey,
+        `Can't generate a key for invalidateEntity(${entity}),
```
kitten: This adds an empty indentation on the next line. Let’s maybe keep this in line with the other messages and just manually concatenation?

Also let’s maybe make sure that this message is as clear as this one? https://github.com/FormidableLabs/urql/blob/master/docs/graphcache/help.md#7-cant-generate-a-key-for-readfragment-

The text can be copied almost 1:1, but that one accounts for edge cases, like talking about “keys” in general instead of the default “id” fields

exchanges/graphcache/src/store/store.ts
```diff
@@ -154,11 +155,39 @@ export class Store implements Cache {
     invalidate(this, createRequest(query, variables));
   }
 
+  invalidateEntity(entity: Data | string) {
```
kitten: Is it worth calling this just “invalidate” or is that a little misleading? 🤠

JoviDeCroock: I don't mind calling it invalidate if we're planning to deprecate `invalidateQuery`

exchanges/graphcache/src/store/store.ts
```diff
@@ -154,11 +155,39 @@ export class Store implements Cache {
...
 
+  invalidate(entity: Data | string) {
+    const entityKey =
+      typeof entity === 'string' ? entity : this.keyOfEntity(entity);
+
+    invariant(
+      entityKey,
+        "Can't generate a key for invalidate(...).\n" +
+        "You have to pass an id or _id field or create a custom `keys` field for `" +
+        (entity as Data).__typename +
```
kitten: Sorry 😅  This will throw if `entity` was a string

PR #567, opened 2020-03-03
Move populateExchange to a package

JoviDeCroock:
## Summary

Let's use our new monorepo optimally and start moving some packages, this allows us to install less if it's not needed


exchanges/graphcache/src/ast/schemaPredicates.ts
```diff
@@ -9,7 +9,7 @@ import {
   GraphQLUnionType,
 } from 'graphql';
 
-import { invariant, warn } from '../helpers/help';
+import { warn, invariant } from '../../../shared/src/helpers/help';
```
kitten: Seems like this path is incorrect :)

exchanges/populate/CHANGELOG.md
```diff
@@ -0,0 +1,202 @@
+# @urql/exchange-graphcache
+
+## 2.1.1
```
kitten: Let’s trim this down to “Initial Release: Move `populateExchange` to a separate package.” and publish it as `0.1.0`? Seems like the right version to start at imho

PR #570, opened 2020-03-04
make the build-command work on windows

JoviDeCroock:
Windows doesn't recognise the `shebang`


kitten:
Glad this is the only thing I missed for Windows 😅

PR #575, opened 2020-03-06
Add bundling for d.ts typings output

kitten:


kitten:
@jovidecroock this is closer but unfortunately doesn’t work 😢 the plugin is a good alternative to what we had before but it either needs typings for all other externals, or paths have to be readded which will bundle all types which is also not quite what we want

JoviDeCroock:
@kitten let's maybe just duplicate the code for now, we'll have to duplicate `help.ts` for now anyway. I'll take care of it if you want

Issue #560, opened 2020-03-01
File-upload exchange


JoviDeCroock:
Now  that we have a monorepo and more room to add custom-exchanges, this [spectrum post](https://spectrum.chat/urql/help/upload-files~cf16dd5e-2d75-4211-b717-e5a0ae23ea10) made me think about publishing an exchange that builds upon the current `fetchExchange` but additionally allows file-uploads.

[Example file-upload exchange](https://formidable.com/open-source/urql/docs/guides/#adjusting-an-exchange)

Let's get some opinions on this and then mark it as a `Good First Issue` or refer to the conclusion here when people ask for this.

BjoernRave:
@BjoernRave I would welcome such an exchange :)

roshangm1:
Yes. Yes.

kayluhb:
Just as a headsup, I'm getting an error using this example with @urql/core@1.9.0

```
TypeError: undefined is not a function
(anonymous function)
./utils/exchanges/uploadExchange.js:42
  39 |     });
  40 | };
  41 | 
> 42 | const createFetchSource = operation =>```

JoviDeCroock:
Hey @kayluhb ,

Thank you for reporting this!
The example has probably been outdated (we're  actively working on new documentation), I'll most likely take some time to construct the exchange in the following days.

kayluhb:
for anyone that may need it, here's an updated version of the exchange that was previously in the docs (sans typescript and without support for gets).

```js
import { extractFiles } from 'extract-files';
import { filter, make, merge, mergeMap, pipe, share, takeUntil } from 'wonka';
import { Kind, print } from 'graphql';
import { makeResult, makeErrorResult } from 'urql';

const getOperationName = query => {
  const node = query.definitions.find(
    node$ => node$.kind === Kind.OPERATION_DEFINITION && node$.name,
  );

  return node !== undefined && node.name ? node.name.value : null;
};

const executeFetch = (operation, opts) => {
  const { url, fetch: fetcher } = operation.context;
  let response;

  return (fetcher || fetch)(url, opts)
    .then(res => {
      const { status } = res;
      const statusRangeEnd = opts.redirect === 'manual' ? 400 : 300;
      response = res;

      if (status < 200 || status >= statusRangeEnd) {
        throw new Error(res.statusText);
      } else {
        return res.json();
      }
    })
    .then(result => makeResult(operation, result, response))
    .catch(err => {
      if (err.name !== 'AbortError') {
[...]

PR #585, opened 2020-03-08
Add multipartFetchExchange package

JoviDeCroock:
## Summary

This adds a new exchange meant to extend the feature-set of our default `fetchExchange` to include file-uploads!

Working playground https://codesandbox.io/s/urql-upload-client-05pcn

Implements https://github.com/FormidableLabs/urql/issues/560

### DIFF
--- a/exchanges/multipart-fetch/src/multipartFetchExchange.test.ts
+++ b/exchanges/multipart-fetch/src/multipartFetchExchange.test.ts
@@ -67,9 +67,7 @@ describe('on success', () => {
 
     expect(data).toMatchSnapshot();
     expect(fetchOptions).toHaveBeenCalled();
-    expect(fetch.mock.calls[0][1].headers).toEqual({
-      'content-type': 'multipart/form-data',
-    });
+    expect(fetch.mock.calls[0][1].headers).toMatchSnapshot();
     expect(fetch.mock.calls[0][1].body).toMatchSnapshot();
   });
 
--- a/exchanges/multipart-fetch/src/multipartFetchExchange.test.ts
+++ b/exchanges/multipart-fetch/src/multipartFetchExchange.test.ts
@@ -1,0 +1,191 @@
+import { Client, OperationResult, OperationType } from '@urql/core';
+import { queryOperation } from '@urql/core/src/test-utils';
+import { empty, fromValue, pipe, Source, subscribe, toPromise } from 'wonka';
+import gql from 'graphql-tag';
+import { print } from 'graphql';
+
+import { multipartFetchExchange, convertToGet } from './multipartFetchExchange';
+
+const fetch = (global as any).fetch as jest.Mock;
+const abort = jest.fn();
+
+const abortError = new Error();
+abortError.name = 'AbortError';
+
+beforeAll(() => {
+  (global as any).AbortController = function AbortController() {
+    this.signal = undefined;
+    this.abort = abort;
+  };
+});
+
+afterEach(() => {
+  fetch.mockClear();
+  abort.mockClear();
+});
+
+afterAll(() => {
+  (global as any).AbortController = undefined;
+});
+
+const response = {
+  status: 200,
+  data: {
+    data: {
+      user: 1200,
+    },
+  },
+};
+
+const exchangeArgs = {
+  forward: () => empty as Source<OperationResult>,
+  client: {} as Client,
+};
+
+describe('on success', () => {
+  beforeEach(() => {
+    fetch.mockResolvedValue({
+      status: 200,
+      json: jest.fn().mockResolvedValue(response),
+    });
+  });
+
+  it('returns response data', async () => {
+    const fetchOptions = jest.fn().mockReturnValue({});
+
+    const data = await pipe(
+      fromValue({
+        ...queryOperation,
+        context: {
+          ...queryOperation.context,
+          fetchOptions,
+        },
+      }),
+      multipartFetchExchange(exchangeArgs),
+      toPromise
+    );
+
+    expect(data).toMatchSnapshot();
+    expect(fetchOptions).toHaveBeenCalled();
+    expect(fetch.mock.calls[0][1].body).toMatchSnapshot();
+  });
+});
+
+describe('on error', () => {
+  beforeEach(() => {
+    fetch.mockResolvedValue({
+      status: 400,
+      json: jest.fn().mockResolvedValue(response),
+    });
+  });
+
+  it('returns error data', async () => {
+    const data = await pipe(
+      fromValue(queryOperation),
+      multipartFetchExchange(exchangeArgs),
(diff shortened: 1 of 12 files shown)


zenios:
When a subscription exchange with enableAllOperations = true is added before  this exchange then file upload requests will not be forwarded to this exchange.

JoviDeCroock:
I mean, that's an issue with the subscriptionExchange rather than the `fetchExchange` and should be the case in current implementation as well

zenios:
You dont have the option in subscription exchange to opt out and let a
request pass through to the next exchange

Its either handle all or only subscriptions and i am not sure if you can
handle mulipart requests through a websocket

JoviDeCroock:
Yes but that's not the point, we establish the premise of exchanges being order-based in the documentation. We add in the `subscriptions` that this should be the final exchange due to this reason, am I missing something? If so could you please elaborate?

zenios:
Yes i want subscription exchange to handle all types of queries (queries /
mutations / subscriptions) but since it cant handle file uploads i want
mulipart-fetch-exchange to do it.

How can i achieve that with the current subscription exchange and
multipart-fetch-exchange?

JoviDeCroock:
I'd say that's a very very custom use case, I'd suggest forking the subscriptionExchange and this one and combine them on your end

zenios:
Or change subscription exchange from 

enableAllOperations?: boolean to 
enableAllOperations?: boolean | (operation: Operation) => boolean;

This will cater for simple and advanced usages

What do you think?

JoviDeCroock:
The name doesn't really serve the purpose of becoming a function, I see where you're trying to go but that would need another variable (or a breaking change).

JoviDeCroock:
Also I don't really think this is the best place to discuss a change to the `subscriptionExchange`

zenios:
Its not a change to the subscriptionExchange out of the blue.

Its a change related to handling multipart-requests which is what this PR is for

Nevertheless

In my opinion having to fork a whole exchange just to add a check
if(operation === multipart) then forward

is more complex than having such kind of functionality provided from the core.

JoviDeCroock:
It would be mixing concerns between exchanges at this point so the change would have to be in `subscriptionExchange` and not generally related to this, it would include a check whether or not this is a file-upload. The check would indeed be moved to user-land, I'd personally just prefer that in a seperate PR for readability (git history wise).

I'm not saying that it isn't needed it's more that we'd have to settle on an approach, infinitely increasing API surface isn't really what we want to do in core.

exchanges/multipart-fetch/src/multipartFetchExchange.test.ts
```diff
@@ -0,0 +1,212 @@
...
+
+describe('on success', () => {
+  beforeEach(() => {
+    fetch.mockResolvedValue({
+      status: 200,
+      json: jest.fn().mockResolvedValue(response),
+    });
+  });
+
+  it('uses a file when given', async () => {
```
wgolledge: Should we add a test with multiple files?

exchanges/multipart-fetch/src/multipartFetchExchange.ts
```diff
@@ -0,0 +1,196 @@
...
+      headers: {
+        'content-type': 'application/json',
+        ...extraOptions.headers,
+      },
+      signal:
+        abortController !== undefined ? abortController.signal : undefined,
+    };
+
+    if (!!files.size) {
+      fetchOptions.body = new FormData();
```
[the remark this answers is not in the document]
JoviDeCroock: Then we would have to clear `.body` in `shouldUseGet` which would just move that line there to be `fetchOptions.body = undefined`. Personally prefer this approach

exchanges/multipart-fetch/src/multipartFetchExchange.ts
```diff
@@ -0,0 +1,196 @@
...
+    return () => {
+      ended = true;
+      if (abortController !== undefined) {
+        abortController.abort();
+      }
+    };
+  });
+};
+
+const executeFetch = (
```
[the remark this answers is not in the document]
JoviDeCroock: My thinking was, in case we need to seperate these two more we're already prepared, else we would also introduce another export from `@urql/core` which increases the size

exchanges/multipart-fetch/src/multipartFetchExchange.test.ts
JoviDeCroock: Good shout

exchanges/multipart-fetch/src/multipartFetchExchange.ts
```diff
@@ -0,0 +1,194 @@
...
+      method: shouldUseGet ? 'GET' : 'POST',
+      headers: {
+        'content-type': 'application/json',
+        ...extraOptions.headers,
+      },
+      signal:
+        abortController !== undefined ? abortController.signal : undefined,
+    };
+
+    if (!!files.size) {
```
kitten: I assume this is part of Apollo's extended file-upload spec? Does their server support always sending multipart requests? it may be worth sending those consistently instead of JSON POST requests, to avoid duplicate code and confusion when using this.

exchanges/multipart-fetch/src/multipartFetchExchange.ts
```diff
@@ -0,0 +1,194 @@
...
+      fetchOptions.body = new FormData();
```
kitten: Is there anything specific we need to do to support this in Node.js? Can we check what `isomorphic-unfetch` / `node-fetch` requires for this to work correctly?

exchanges/multipart-fetch/src/multipartFetchExchange.ts
JoviDeCroock: You would prefer always doing FormData? This would mean  deeply converting every `variables` structure?

exchanges/multipart-fetch/src/multipartFetchExchange.ts
JoviDeCroock: Both show this working with `.body = FormData`

exchanges/multipart-fetch/src/multipartFetchExchange.ts
JoviDeCroock: The acceptance of `FormData` seems relevant to the upload `graphql-scalar` https://github.com/jaydenseric/graphql-upload/blob/master/lib/processRequest.js

kitten: Oh I see! Well as long as we follow their spec we’re good then 👍

exchanges/multipart-fetch/src/multipartFetchExchange.ts
kitten: Awesome!

PR #586, opened 2020-03-08
Skip specific query/mutation operations when enableAllOperations is on

JoviDeCroock:
## Summary

Allows for selective skipping of  (query/mutation) operations when `enableAllOperations` is turned on for subscriptions.

## Set of changes

- add `skipOperation` option to the `subscriptionExchange`


zenios:
Having a way to skip subscription exchange is something i mentioned in two prs but giving it a aecond thought, higher order exchanges is a much better approach.

PR #587, opened 2020-03-08
Fix commutative squashing failing when lowest layer resolved earlier than others

### LINKED ISSUE #574: cache.writeFragment does not rerender component
When updating cached data with cache write fragment

I dont see any component using that data beeing refreshed.
I can see the records beeing updated using console.log(cache)

Is this intended?

kitten:
The commutative squashing would previously only start when
the lowest commutative layer (the first one) was written.
This ignored the case however of when this layer has already
been written. When this was the case the stopping condition
was unreachable and commutative layers were never squashed.

The change is to continuously squash the commutative layers
with the least priorities as they're written.

This change also gets rid of the temporary `squash` queue,
which ends up making it a rather nice implementation.

Resolve #574

### DIFF
--- a/exchanges/graphcache/src/store/data.ts
+++ b/exchanges/graphcache/src/store/data.ts
@@ -100,16 +100,15 @@ export const clearDataState = () => {
     const commutativeIndex =
       data.optimisticOrder.length - data.commutativeKeys.size;
 
-    // Find `i` to be the lowest commutative layer that can be squashed
+    // Squash all layers in reverse order (low priority upwards) that have
+    // been written already
     let i = data.optimisticOrder.length;
     while (--i >= commutativeIndex) {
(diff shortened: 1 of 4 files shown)


exchanges/graphcache/src/store/data.ts
```diff
@@ -494,6 +489,10 @@ export const clearLayer = (data: InMemoryData, layerKey: number) => {
 
 /** Merges an optimistic layer of links and records into the base data */
 const squashLayer = (layerKey: number) => {
+  // Hide current dependencies from squashing operations
+  // const prevDependencies = currentDependencies;
+  // currentDependencies = new Set();
```
JoviDeCroock: Is this purposely left in?

kitten: ah, I wanted to commit that, just a moment

Issue #582, opened 2020-01-04
initUrqlClient called twice browser side


addstar34:
I noticed that `initUrqlClient` is getting called twice browser side. It first gets called in `withUrql.getInitialProps` and then again within `withUrql`.

Maybe `urqlClient` needs to get passed down here:

```
// withUrqlClient.tsx

const isBrowser = typeof window !== 'undefined'
if (isBrowser) {
  return { ...pageProps, urqlClient }
}
```

which would then be set in the `useMemo`

kitten:
Fixed by #589

PR #589, opened 2020-03-08
(next-urql) - call init once browser side

JoviDeCroock:
## Summary

Passes down `urqlClient` in `withUrqlClient.getInitialProps`, this prevents the HOC from having to call `initUrqlClient` again.

If anyone has suggestions for a test feel free to throw them out there since `typeof window === undefined` in tests

Implements https://github.com/FormidableLabs/urql/issues/582

## Set of changes

Passes down `urqlClient` in `withUrqlClient.getInitialProps`

### DIFF
--- a/packages/next-urql/src/with-urql-client.tsx
+++ b/packages/next-urql/src/with-urql-client.tsx
@@ -76,7 +76,7 @@ export function withUrqlClient(
       // getInitialProps runs on the server for initial render, and on the client for navigation.
       // We only want to run the prepass step on the server.
       if (typeof window !== 'undefined') {
-        return pageProps;
+        return { ...pageProps, urqlClient };
       }
 
       const props = { ...pageProps, urqlClient };
--- a/.changeset/dirty-tips-sin.md
+++ b/.changeset/dirty-tips-sin.md
@@ -1,0 +1,5 @@
+---
+'next-urql': patch
+---
+
+Pass the urqlClient down in withUrqlClient.getInitialProps to prevent it from being called twice.


packages/next-urql/src/with-urql-client.tsx
```diff
@@ -76,7 +76,7 @@ export function withUrqlClient(
       // getInitialProps runs on the server for initial render, and on the client for navigation.
       // We only want to run the prepass step on the server.
       if (typeof window !== 'undefined') {
-        return pageProps;
+        return { ...pageProps, urqlClient };
```
kitten: How do we prevent `next` from trying to serialise it?

JoviDeCroock: Is that a problem on client-only?

kitten: I think it may be a problem on rehydration? @parkie-doo may know more

JoviDeCroock: https://github.com/FormidableLabs/next-urql/commit/14d4bef6cf5754d11235461b1c686c1c4793d382#diff-605252a5383070a6def592e274ab50b4L124

Thought adding it there was the only thing affecting that, will investigate

parkerziegler: I think this change will likely be ok. As I reason it:

- Initial SSR occurs. We skip this block altogether on the server. On the client, the call to `initUrqlClient` occurs and we get our client-side instance.
- Client-side navigation occurs and `getInitialProps` is run. `urqlClient` gets created and returned, and used in lieu of `initUrqlClient` getting called again.

I'll apply this patch locally and test it out on `examples/4-with-navigation` and ensure we still get the expected client configuration in all environments.

parkerziegler:
Tested this out locally and everything looks good to me :shipit: 

Also, lol at this changesets name 😂 

image

PR #591, opened 2020-03-09
Enable AWS deploy script

kitten:


ryan-roemer:
Have we tested a `--dryrun` yet in localdev with full credentials?

The preferred way to "test the hook without deploying" is like this:

```
      - deploy:
          name: Deploy Production Site
          command: |
            cd packages/site
            node scripts/deploy/aws.js --dryrun
```

Then examine logs after merge to verify it **only** ran on a merge to master (not during PR) and the logs look good for what it would have done.

then in next PR remove the `--dryrun` flag

PR #594, opened 2020-03-09
Fix relative links and adapt to base path in production

kitten:


ryan-roemer:
Capturing here so it's on issue. Still seeing this:

Screen Shot 2020-03-09 at 11 02 07 AM

Screen Shot 2020-03-09 at 11 03 06 AM

PR #598, opened 2020-03-10
Codesandboxify examples

andyrichardson:
## About
Set up SW GraphQL server so examples can run in codesandbox

**[CodeSandbox of this branch](https://codesandbox.io/s/github/FormidableLabs/urql/tree/codesandbox-for-examples/packages/react-urql/examples/1-getting-started)**

## Changes
 - Add service worker for GraphQL HTTP responses
 - Add use of local IndexedDB for persistent state
 - Remove server section
 - Remove Webpack build configs
 - Add parcel for build process


packages/react-urql/examples/1-getting-started/src/pages/Home.tsx
```diff
@@ -13,16 +13,16 @@ interface QueryResponse {
   todos: ITodo[];
 }
 
-const ToggleTodo = `
-  mutation($id: ID!) {
+const ToggleTodo = gql`
```
JoviDeCroock: is this gql call needed, I removed them because it shows we don't force you to use it unless you want fragments (said in docs)

packages/react-urql/examples/1-getting-started/src/service-worker.ts
```diff
@@ -0,0 +1,34 @@
+import { execute } from 'graphql';
+import { rootValue, schema } from './graphql';
+import gql from 'graphql-tag';
+
+self.addEventListener('install', (event: any) => {
+  console.log('GraphQL service worker installed.');
+  event.waitUntil((self as any).skipWaiting());
```
kitten: You should be able to call `skipWaiting` without `waitUntil` afaik

packages/react-urql/examples/1-getting-started/src/index.tsx
```diff
@@ -2,11 +2,13 @@ import React, { FC, StrictMode } from 'react';
 import * as ReactDOM from 'react-dom';
 import { createClient, Provider, defaultExchanges } from 'urql';
 import { devtoolsExchange } from '@urql/devtools';
-import { Home } from './Home';
+import { Home } from './pages';
 import './index.css';
 
+navigator.serviceWorker.register('./service-worker.ts', { scope: '/sw' });
```
kitten: The browser will likely not update properly unless you have `.then(reg => reg.update())` here... at least that's how I remember it?

andyrichardson: Nope, this is working fine

packages/react-urql/examples/1-getting-started/src/service-worker.ts
andyrichardson: Was trying to get immediate access to network interceptions [using this example](https://serviceworke.rs/immediate-claim_service-worker_doc.html)

packages/react-urql/examples/1-getting-started/src/pages/Home.tsx
andyrichardson: Oh I see, it's not needed no. But the types also make it clear that strings can be provided no?

andyrichardson:
Merging for now as this fixes the broken example.

Codesandbox integration needs some support from the codesandbox folks by either fixing [multiple parcel issues](https://github.com/codesandbox/codesandbox-client/issues/2087) or allowing [container environments on repo import](https://github.com/codesandbox/codesandbox-client/issues/2111).

packages/react-urql/examples/1-getting-started/package.json
```diff
@@ -1,37 +1,24 @@
 {
-  "name": "@urql/example-getting-started",
+  "name": "1-getting-started",
   "version": "1.0.0",
```
kitten: Let's make sure in a follow-up PR to set `"private"` back to `true` please 🙏 Shouldn't affect CodeSandbox / the example itself, but since this isn't published it's safer to set this

Issue #596, opened 2020-03-10
cache invalidation problem with empty lists and adding the first item


gabor:
short version: when adding the first item into an empty list, the cache does not get invalidated.

long version:
imagine a have a schema where i have a `Group` type, and groups contain lists-of-users.
something like:
```graphql
type User {
  id: ID!
}

type Group {
  id: ID!
  users: [User!]!
}

type Query {
  getGroup(id:ID!):Group
}

type Mutation {
  addUserToGroup(userID:ID!, groupID:ID!):User!
}
```

now, on a page i show the details of a given group, a list of it's users, with a query like:
```graphql
getGroup(id:"42") {
  id
  users {
    id
  }
}
```

i can add a new user on the page, which executes a mutation like:
```graphql
addUserToGroup(userId:$userId, groupId: $groupId) {
  id
}
```
(i did not write out the `__typename` attributes, but they are there).

usually this works fine, i mean, if i create an user, the mutation returns the `User` typename, and the cache gets invalidated, and the list of users gets updated.

but, if i display a group, where there are no users, and i create the first user, the page does not refresh, i assume, because originally the `User` typename did not arrive, because there were no users.( i am not 100% sure about this, i am just guessing here).

is there a way to work around this problem?

JoviDeCroock:
Hey,

You can find more information about this limitation here: https://github.com/FormidableLabs/urql/issues/378

This can be solved by using `graphCache`, essentially the issue comes down to us not having the possibility to know what type that list of users is unless you are using something like `relayPagination`.

```
getGroup(id:"42") {
  id
  users {
    id
  }
}
```

If there are no users, we don't know that `__typename` users should invalidate this.

More resources on this:
- https://spectrum.chat/urql/help/query-not-refetched-after-a-mutation-using-exchange-graphcache~615d2331-2899-4648-ae26-88e6a50d2223

gabor:
hi,

thanks for the answers. some more questions:

1. in case of relay-pagination, how would this be solved? for me it seems that if there are no items, there will be no items with with relay, too.

2. you mentioned this can be solved by "graphCache". would graphCache solve this transparently for me, or would i have to write custom code to manipulate the local-cache after the mutation? the reason i'm worried about graphCache is that i have used the apollo client-side graphql libraries before, where you have a normalizing cache, and i had to write manual code to update related things after mutations, and it became  so complex, it was unmanageable. my hope was that with the urql-document-cache, i trade fetching "efficiency" for easier correctness.

[...]

PR #601, opened 2020-03-10
Add additionalTypes option to context for default cache invalidation

JoviDeCroock:
## Summary

This PR introduces the possibility of adding `additionalTypenames` to the `query` methods, this `OperationContext` variable will be used by the document-based cache to circumvent issues where for instance a list would not contain any items. In those scenario's we have no way of reacting to a mutation.

If we go with this proposal I still have to:
- write a test
- write docs

## Set of changes

- add recognition for `additionalTypenames` to the `cacheExchange` of `@urql/core`

Fixes: https://github.com/FormidableLabs/urql/issues/596

### DIFF
--- a/docs/api/core.md
+++ b/docs/api/core.md
@@ -194,16 +194,17 @@ data that can be passed from almost all API methods in `urql` that deal with
 Some of these options are set when the `Client` is initialised, so in the following list of
 properties you'll likely see some options that exist on the `Client` as well.
 
-| Prop            | Type                                  | Description                                                                                                           |
-| --------------- | ------------------------------------- | --------------------------------------------------------------------------------------------------------------------- |
-| fetchOptions    | `?RequestInit \| (() => RequestInit)` | An optional [request policy](/basics/querying-data#request-policy) that should be used specifying the cache strategy. |
-| fetch           | `typeof fetch`                        | An alternative implementation of `fetch` that will be used by the `fetchExchange` instead of `window.fetch`           |
-| requestPolicy   | `RequestPolicy`                       | An optional [request policy](/basics/querying-data#request-policy) that should be used specifying the cache strategy. |
-| url             | `string`                              | The GraphQL endpoint                                                                                                  |
-| pollInterval    | `?number`                             | Every `pollInterval` milliseconds the query will be refetched.                                                        |
-| meta            | `?OperationDebugMeta`                 | Metadata that is only available in development for devtools.                                                          |
-| suspense        | `?boolean`                            | Whether suspense is enabled.                                                                                          |
-| preferGetMethod | `?number`                             | Instructs the `fetchExchange` to use HTTP GET for queries.                                                            |
+| Prop                | Type                                  | Description                                                                                                           |
+| ------------------- | ------------------------------------- | --------------------------------------------------------------------------------------------------------------------- |
+| fetchOptions        | `?RequestInit \| (() => RequestInit)` | An optional [request policy](/basics/querying-data#request-policy) that should be used specifying the cache strategy. |
+| fetch               | `typeof fetch`                        | An alternative implementation of `fetch` that will be used by the `fetchExchange` instead of `window.fetch`           |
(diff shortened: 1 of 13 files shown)


wgolledge:
Looks like a great solution to me!

docs/api/core.md
```diff
@@ -194,16 +194,17 @@ data that can be passed from almost all API methods in `urql` that deal with
...
+| url                 | `string`                              | The GraphQL endpoint                                                                                                  |
+| pollInterval        | `?number`                             | Every `pollInterval` milliseconds the query will be refetched.                                                        |
+| meta                | `?OperationDebugMeta`                 | Metadata that is only available in development for devtools.                                                          |
+| suspense            | `?boolean`                            | Whether suspense is enabled.                                                                                          |
+| preferGetMethod     | `?number`                             | Instructs the `fetchExchange` to use HTTP GET for queries.                                                            |
+| additionalTypenames | `?number`                             | Allows you to tell the operation that it has dependencies on certain typenames (used in document-cache.)              |
```
JoviDeCroock: Sorry I had to add spaces to get everything to line up nicely

wgolledge:
```suggestion
| additionalTypenames | `?number`                             | Allows you to tell the operation that it depends on certain typenames (used in document-cache.)              |
```

docs/basics/document-caching.md
```diff
@@ -50,6 +50,5 @@ docs.](../api/core.md#requestpolicy-type)
 This cache has a small trade-off! If we request a list of data and the API returns an empty list,
 the cache won't be able to see the `__typename` of said list and won't invalidate.
 
-Once you've encountered this problem you've likely hit the limits of the _Document Caching_
-approach, and you may want to [switch to "Normalized Caching"
-instead.](../graphcache/normalized-caching.md)
+There are two ways to fix this issue, supplying `additionalTypenames` to the context of your query or [switch to "Normalized Caching"
+instead.](../graphcache/normalized-caching.md).
```
wgolledge:
```suggestion
instead](../graphcache/normalized-caching.md).
```

docs/basics/mutations.md
```diff
@@ -92,6 +92,23 @@ const Todo = ({ id, title }) => {
 };
 ```
 
+### Adding typenames
```
wgolledge: I think this is a great section to have. I wonder if it fits in `basics` though. Not really sure if there's anywhere else it could go?

packages/core/src/exchanges/cache.ts
```diff
@@ -154,14 +158,18 @@ const afterQuery = (
...
+  const { additionalTypenames } = operation.context;
 
   if (data === undefined || data === null) {
     return;
   }
 
   resultCache.set(operation.key, { operation, data, error });
 
-  collectTypesFromResponse(response.data).forEach(typeName => {
+  [
```
wgolledge: Nit: Could these go into a `generateTypenames` func? 

``` js
const generateTypenames = (responseData, additionalTypenames) => ...
```

docs/basics/mutations.md
kitten: I think it should go into "Document Caching". We already have a gotchas section there. Or maybe we want to have the actual API here in `mutations`, note that it's for "Document Caching" only, then repeat the generic API in "Document Caching"?

JoviDeCroock: I mean this can be used for multiple things, think about localState, ... I wanted to move it here so we could give a real-world example.

packages/core/src/exchanges/cache.ts
JoviDeCroock: My concern with it is the unneeded size overhead this will introduce

kitten:
Let’s not forget to rename the PR title, e.g. “Add additionalTypes option to context for default cache invalidation”

packages/core/src/exchanges/cache.ts
wgolledge: I need to become more size-concious 🙂

PR #603, opened 2020-03-11
(next-urql) - Allow user to configure `suspense` client config.

### LINKED ISSUE #583: Allow user to configure `suspense` client config.
While suspense is currently experimental, Next does offer experimental support for using it (albeit without streaming SSR). However, it’s not possible to enable it with this library. At present, adding the `suspenseExchange` results in a [warning](https://github.com/FormidableLabs/urql/blob/b8d4423e103ab94df93b675be84885a1b4930dce/exchanges/suspense/src/suspenseExchange.ts#L14):
> [@urql/exchange-suspense]: suspenseExchange is currently disabled.

It should be possible to override the [default configuration](https://github.com/FormidableLabs/next-urql/blob/321328ce26fd20edb34ddd47b0576d731be96be7/src/init-urql-client.ts#L37) which is to only enable suspense on the server.

ryan-gilb:
Resolves #583 

## Summary

We want to allow users to configure `suspense` on the client when it's running in the browser!

## Set of changes

* suspenseExchange - made some small fixes to the "missing suspense config" warning message
* next-urql - allows `suspense` to be passed into `clientOptions` and passes it in call to `createClient`
* next-urql - added example showing how to use the `suspenseExchage` with `next-urql`


.changeset/smooth-lions-brush.md
```diff
@@ -0,0 +1,6 @@
+---
+'next-urql': minor
```
kitten: Lets maybe keep this to patch? I wouldn’t consider this a ‘minor’ bump as it’s neither a new feature, nor a significant change, nor a potentially relevant change of behaviour? It seems to me that this PR restores/establishes expected behaviour 🙏

ryan-gilb: Gotcha, makes sense 😄

packages/next-urql/examples/5-with-suspense-exchange/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "5-with-suspense-exchange",
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "start": "next dev",
+    "prestart": "cd ../../ && yarn build && cd examples/3-with-custom-exchange"
```
parkerziegler: I think we need to update this to:

```suggestion
    "prestart": "cd ../../ && yarn build && cd examples/5-with-suspense-exchange"
```

Let's verify that these changes still work with the correct example getting spun up.

packages/next-urql/examples/5-with-suspense-exchange/package.json
```diff
@@ -0,0 +1,17 @@
...
+  },
+  "dependencies": {
+    "@urql/exchange-suspense": "^1.8.3",
+    "graphql": "^14.5.8",
+    "graphql-tag": "^2.10.1",
+    "isomorphic-unfetch": "^3.0.0",
+    "next": "^9.2.0",
+    "wonka": "^4.0.7"
```
parkerziegler: Starting in v0.3.1 `next-urql` now ships with a direct `urql` dependency: https://github.com/FormidableLabs/urql/blob/master/packages/next-urql/package.json#L54 This should bring `wonka` along with it, so I'd expect this example _not_ to need to list `wonka` as a dep (although I'm certain `yarn` will deduplicate it). I think we ought to remove for clarity's sake.

packages/next-urql/examples/5-with-suspense-exchange/package.json
ryan-gilb: lol, when i was initially working on this, i added suspense to example 3. that must have been why running the start command from the copy-pasta'd example 5 worked as expected 😅

PR #605, opened 2020-03-12
(docs) - Post-release update

wgolledge:
PR to amend some of the initial bugs, as well as implement some of the "nice to haves" as discussed in #548.


packages/site/src/styles/global.js
```diff
@@ -74,4 +74,12 @@ export const GlobalStyle = createGlobalStyle`
   img {
     max-width: 100%;
   }
+
+  /* fix markdown link anchors due to fixed header */
```
kitten: Let’s maybe add this to mdx.js instead? We’ll need to specifically add anchor links to headings anyway :)

wgolledge: Good shout!

PR #606, opened 2020-03-12
Proposal: Add primitives for variables and responses

### LINKED ISSUE #604: Increase type specificity for 'data' and 'variables'
## About

Make _data_ and _variable_ types more specific.

> Shoutout to @sofiapoh who started this out in urql devtools

## Types

The logic here is that all variables and data responses must be serialisable and therefore one of the following types:

```ts
/** Serialisable GraphQL primitives (excluding PrimitiveMap) */
type Primitive = number | boolean | string | null;

/** A map of primitives and primitive maps */
interface PrimitiveMap {
  [key: string]: Primitive | Primitive[] | PrimitiveMap;
}

// ...
/** A Graphql query, mutation, or subscription. */
export interface GraphQLRequest {
  /** Unique identifier of the request. */
  key: number;
  query: DocumentNode;
  variables?: PrimitiveMap; // <- no longer 'object'
}


/** Resulting data from an [operation]{@link Operation}. */
export interface OperationResult<Data extends PrimitiveMap = PrimitiveMap> { // <- no longer 'Data = any'
  /** The [operation]{@link Operation} which has been executed. */
  operation: Operation;
  /** The data returned from the Graphql server. */
  data?: Data;
  /** Any errors resulting from the operation. */
  error?: CombinedError;
  /** Optional extensions return by the Graphql server. */
  extensions?: Record<string, any>;
  /** Optional stale flag added by exchanges that return stale results. */
  stale?: boolean;
}
```

andyrichardson:
Fix #604

[Example of how vars upcast](http://www.typescriptlang.org/play/#code/PQKhAIGUFMCcEsCGAbeBnRAjZ1wHFZEAHACwEUAZcIhAW3gBd4A3aNcEYAWACgGBPIrgAKdRi1wBecKPj0mrALLFwAH3AA7AK61McNeEwB7IzkQaDaBgg0BzA9uTIDWjQBNoAM3gbobgNy8vKAQAILgtCpGntRiCmzg5m6xcuKsEcTsnLw+DHCeiADGInESykTgAN684OAA2gDW0PwAXOBWNrYAum2y8hIGfWnQdV2DpUrEgTwAvkE8hUYaVuBuiAyIAGIW0gA8ACrg0AAeee7sQ-Hl4NKXZcQAfAAUiLC2APxt+wCUNw9VNVq4Fg0AYWlgFmsWmg0zmPGCwHAADkjKt1ohEm9eGsNtsnt95sBEQBRWhEARojaY2zY9F4yozAnwnhE8AASTJqEKjEpGNeNJ4OK2GielXAtlBAFU0HA0G06mL4G42gBGABMAGYACwAGk0iFo0DaACIAMKmHzG8AzMaMwkk45ELk8oXU8BPADujBI4FQeUIyDQTMWywYvJuAJ4QPF0GlsrajmQOsBtUQbjccdgbWqUejQKVquTubz+sNJtNr2QxqL0ZmRZmiXYIas0yFeKF32mCPZGmYKCV4f5tNxIqh0E782bYb7sDQ2wjByOp2g5xkE2g11u6-Kz35n3APz+kejILBEPAY9h9uRqJn7CHPDveKZ3dJ5P44Dv1N4T5FDJfLKIhyTrwNy06vPeWKPhB9IxgwmZyvUirKuA6ranqGgGka4BmhaGhWja1oAayxKOs6Yauvy7pegwPp+nAKBBrwU6fhBEY5tGEoIQmWhODWQJphmMpZseJbgAWqH8SWmFljhFawFWUnWvWjbgFO0y-k8d4TsyrJsr2-bJF+D6aWOnZAA)

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -25,12 +25,20 @@ export type RequestPolicy =
 /** How the operation has */
 export type CacheOutcome = 'miss' | 'partial' | 'hit';
 
+/** Serialisable GraphQL primitives */
+type Primitive = PrimitiveMap | number | boolean | string | null | undefined;
+
+/** A map of primitives and primitive maps */
+interface PrimitiveMap {
+  [key: string]: Primitive | Primitive[] | PrimitiveMap;
+}
+
 /** A Graphql query, mutation, or subscription. */
 export interface GraphQLRequest {
   /** Unique identifier of the request. */
   key: number;
   query: DocumentNode;
-  variables?: object;
+  variables?: PrimitiveMap;
 }
 
 /** Metadata that is only available in development for devtools. */
@@ -60,7 +68,7 @@ export interface Operation extends GraphQLRequest {
 }
 
 /** Resulting data from an [operation]{@link Operation}. */
-export interface OperationResult<Data = any> {
+export interface OperationResult<Data extends PrimitiveMap = PrimitiveMap> {
   /** The [operation]{@link Operation} which has been executed. */
   operation: Operation;
   /** The data returned from the Graphql server. */
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -30,6 +30,7 @@ type Primitive = PrimitiveMap | number | boolean | string | null | undefined;
 
 /** A map of primitives and primitive maps */
 interface PrimitiveMap {
+  __typename?: string;
   [key: string]: Primitive | Primitive[] | PrimitiveMap;
 }
 


packages/core/src/types.ts
```diff
@@ -60,7 +68,7 @@ export interface Operation extends GraphQLRequest {
 }
 
 /** Resulting data from an [operation]{@link Operation}. */
-export interface OperationResult<Data = any> {
+export interface OperationResult<Data extends PrimitiveMap = PrimitiveMap> {
```
kitten:
```suggestion
export interface OperationResult<Data = PrimitiveMap | null> {
```

Let's make sure that we also take `null` into account. I'd either add it here or below to `data?: Data | null`, since it can always happen when the entire GraphQL endpoint errors out.

Also, I don't think we can add `extends PrimitiveMap` because the result may not be serialisable after it passed through graphcache or other custom exchanges that may add special data, so let's assume the user knows what they're doing when they pass in a `Data` generic that doesn't extend `PrimitiveMap`.

Also, I realised that users can get the "legacy behaviour" back by just passing `any` to the `Data` generic, so no harm done I guess.

packages/core/src/types.ts
```diff
@@ -25,12 +25,20 @@ export type RequestPolicy =
 /** How the operation has */
 export type CacheOutcome = 'miss' | 'partial' | 'hit';
 
+/** Serialisable GraphQL primitives */
+type Primitive = PrimitiveMap | number | boolean | string | null | undefined;
+
+/** A map of primitives and primitive maps */
+interface PrimitiveMap {
+  [key: string]: Primitive | Primitive[] | PrimitiveMap;
```
kitten:
```suggestion
interface PrimitiveMap {
  __typename?: string;
  [key: string]: Primitive | Primitive[] | PrimitiveMap;
```

We can make the assumption that `__typename` is always either `undefined` or a string :tada:

packages/core/src/types.ts
```diff
@@ -25,12 +25,20 @@ export type RequestPolicy =
...
+}
+
 /** A Graphql query, mutation, or subscription. */
 export interface GraphQLRequest {
   /** Unique identifier of the request. */
   key: number;
   query: DocumentNode;
-  variables?: object;
+  variables?: PrimitiveMap;
```
kitten: I think this change should be made to the framework bindings, e.g. `useQuery` etc, in particular by adding it to the generic there like we do below for `OperationResult`. That's because someone may be using `variables` to add some non-serialisable metadata, like `Date`, which would then be stringified by `stringifyVariables`. It may also be used to pass variables that are only used offline and then filtered out, so I'm a little uneasy about making more assumptions here

packages/core/src/types.ts
andyrichardson: Good shout!

packages/core/src/types.ts
andyrichardson: Yeah I was thinking of the toString issue - would [adding something like this](http://www.typescriptlang.org/play/?ssl=1&ssc=1&pln=15&pc=38#code/PQKhAIGUFMCcEsCGAbeBnRAjZ1wHFZEAHACwEUAZcIhAW3gBd4A3aNcEYAWACgGBPIrgAKdRi1wBecOAB2AV1qY44AD7hMAe005EsteDQMEsgOYGFyZAfmyAJtABm8WdDsGA3uAabIxl6YAXOAAFACU4JIAfIb+ZuAAvgDcvLygEACC4LTE4JqO1GJMrOx67jTw9MW4OUTsnLwuDHCOiADGIkUSALK5HrwyANoA1tD8wUYmpgC6waKV4qwG81USg9PLXay9RCk8Cak8bciIaOzd-ADCJ2fg-TwyMkTy2PBt3r5xpuF3A4--sGgDHksH0ACILuBjqd2C4jHoOmC9v8DvtDm1NLIjOBELAguAVospHJoAB3cAXa4w8JJIA) to the primitives help?

I'm personally in the mindset that users should be making sure that their variables should be intentionally converted beforehand but I appreciate that enforcing that isn't a great idea.

kitten: It's unlikely 😅  We now also support passing `File` and `Blob` variables for `@urql/exchange-multipart-fetch` and cases like these make it hard for us to have a stricter type by default 🤔

PR #608, opened 2020-03-12
Add DebugEvent system to urql client for devtools and debugging

andyrichardson:
## About

There are a lot of moving parts going on in Urql exchanges, these include:

 - Operation specific events (i.e. “operation deduped”)
 - Exchange state events (i.e. “cache state has been changed”)

Currently we provide a means for communicating *some* operation specific events but not all. There are no means to communicate exchange specific events which are not coupled to an operation.

This proposal is the addition of an event target. It's a pub/sub mechanism for exchanges to publish the aforementioned events and for devtools, a user, or any other third party to consume.

## Usage

Listening to events in the console:

```js
const client = createClient({ ... });
client.debugTarget.addEventListener(console.log);
```

Passing the client to a third party library (such as devtools);

```js
myDevTool(client);
```

## Types

Types only apply if a known key is used. This prevents conflicts with core while also allowing for ambiguous types.

If third parties want to add a type, can use [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html)

```ts
import { DebugEventTypes } from 'urql'

declare module "urql" {
  export interface DebugEventTypes {
    newType: { dataProp: string };
  }
}
```

## Todo

 - [x] Disable in prod
 - [ ] Add docs on debugging events
 - [x] Add babel plugin to remove debug events in prod bundles (cc @JoviDeCroock)


packages/core/src/utils/Target.ts
```diff
@@ -0,0 +1,19 @@
+import { DebugEvent } from '../types';
+
+type EventListener = (e: DebugEvent) => void;
+
+export class Target {
+  private listeners: EventListener[] = [];
```
wgolledge: If we support it, could it make sense to use private class fields instead?

andyrichardson: I thought this was a private class field?

andyrichardson:
Cheers for the thoughts @wgolledge 

> I'm guessing this could, down the line, remove the need for the devtools exchange?

Yes totally, instead we could just pass the client to a devtools function.

> One more point on the event dispatching - I wonder whether it could make sense to have an optional dispatch system applied to every exchange that could be invoked (if the exchange desires) with a message and optional data (and the type if it can't be inferred)? It could save having to pass the client through exchanges 🤷‍♂

Initially I thought it would be a good idea to only pass the dispatch function but then realised that we already pass the client (which contains the EventTarget) so it felt pretty redundant.

I think most exchanges will end up with a function like this:

```tsx
const myExchange = ({ client, forward }) => {
  const publishMessage = (m) => client.eventTarget.dispatchEvent({ ...m, source: "myExchange" });
  // ...
}
```

packages/core/src/utils/Target.ts
wgolledge: Sorry I shoulda clarified - JS private class fields with `#`. Just had a look and they don't seem to be supported in Firefox yet. Feel free to resolve

wgolledge:
Awesome - yep looks good as far as I can see! Will wait for @kitten and/or @JoviDeCroock to weigh in

packages/core/src/utils/Target.ts
andyrichardson: TIL!

packages/core/src/utils/Target.ts
```diff
@@ -0,0 +1,19 @@
+import { DebugEvent } from '../types';
```
JoviDeCroock: Personally I'd move this to a global instead of saving it on the client, this supports debug warnings that could go across multiple roots which would be awesome.

packages/core/src/exchanges/cache.ts
```diff
@@ -131,10 +140,18 @@ export const afterMutation = (
...
   const { additionalTypenames } = response.operation.context;
 
-  [
+  const invalidTypeNames = [
     ...collectTypesFromResponse(response.data),
     ...(additionalTypenames || []),
-  ].forEach(typeName => {
+  ];
+
+  client.eventTarget.dispatchEvent({
```
JoviDeCroock: if we can reduce this to `dispatchEvent` we can easily write a `babel-plugin` that wraps this with `__dev__` like we use in [`graphCache`](https://github.com/FormidableLabs/urql/blob/master/scripts/babel/transform-invariant-warning.js)

packages/core/src/exchanges/cache.ts
```diff
@@ -131,10 +140,18 @@ export const afterMutation = (
...
+    type: 'cacheInvalidation',
+    message: `The following typenames have been invalidated: ${invalidTypeNames}`,
+    data: { value: invalidTypeNames },
```
sofiapoh: I think for debugging it might be useful to include the operation as part of the cache invalidation event

andyrichardson: Thanks sofia! Done :+1:

andyrichardson:
```suggestion
    data: { value: typeNamesToBeInvalidated },
```

packages/core/src/exchanges/cache.ts
```diff
@@ -131,10 +140,18 @@ export const afterMutation = (
...
+  });
+
+  invalidTypeNames.forEach(typeName => {
```
andyrichardson:
```suggestion
  typeNamesToBeInvalidated.forEach(typeName => {
```

packages/core/src/exchanges/cache.ts
andyrichardson:
```suggestion
    data: { value: typeNamesToBeInvalidated },
```

andyrichardson:
Update - 

A few example of an event we might see:
```js
{
  type: "fetch",
  message: "A network request has been made",
  operation,
}
{
  type: "refetch",
  message: "Attempting to refetch a failed network request",
  operation,
  data: { 
    retries: 1,
  }
}
```

Required attributes are now:
 - type (UID for event)
 - message (human readable description)
 - operation (associated operation)
 - data (optional additional data)


This also means we need to change our devtools implementation so that we no longer listen via an exchange but instead for events. That way we streamline every message to look similar to the above.

packages/core/src/types.ts
```diff
@@ -76,12 +76,29 @@ export interface OperationResult<Data = any> {
...
+    typenames: string[];
+    response: OperationResult;
+  };
+}
+
+export type DebugEvent<T extends keyof DebugEventTypes | string> = {
+  type: T;
+  message: string;
+  operation: Operation;
+} & (T extends keyof DebugEventTypes
```
andyrichardson: I'm not too sure why this was required but the following wouldn't work:

```tsx
export type DebugEvent<T extends keyof DebugEventTypes | string> = {
  type: T;
  message: string;
  operation: Operation;
  data: T extends keyof DebugEventTypes ? DebugEventTypes[T] : any | undefined;
}
```

kitten:
@andyrichardson this seems like the way to go :+1: in what cases do we need to provide additional data? It'd be great if we kept this more on the untyped side, so that adding new debugging events would be easy.

What I basically mean is, if we make this a `log` call (naming-wise as well as in how it _looks_ to contributors), then adding new calls could be useful in Jest / development as well as being very flexible when people go and add custom logging/debugging events in new exchanges.

So for instance, if someone goes and makes a `retryOrAuth` exchange, they may add 

```js
log('auth', 'Refreshing the current access token since it's expired);
```

So for them this would look just like `console.log`. It's also basically the same as what we currently do, we'd just allow for `type` to be raw strings as well.

andyrichardson:
Thanks for the response @kitten!

> What I basically mean is, if we make this a log call (naming-wise as well as in how it looks to contributors), then adding new calls could be useful in Jest / development as well as being very flexible when people go and add custom logging/debugging events in new exchanges.

One of the things I want to really make clear with this is that it's an event bus for the purpose of debugging - not for logging events. While making it work like _console.log_ does make things easier to get started, I think it sets a precedent that doesn't really reflect what's going on. 

I wouldn't advise people use this for testing as it exposes implementation details. In a browser however, events could very easily be logged to the console.

```tsx
const client = createClient({ ... });
client.eventTarget.addEventListener(console.log);
```

Every exchange action is associated with an operation, hiding that information when dispatching the event leaves room for ambiguity (not something you want when debugging).

The implementation of the event target is basically a 1-for-1 copy of [EventTarget](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/EventTarget). If you want to trim things down a little, we could pass `dispatchEvent` as an argument to exchanges but I don't see too much benefit for that.

> It'd be great if we kept this more on the untyped side, so that adding new debugging events would be easy.

Have you tried dispatching a custom event? There's nothing stopping you from doing that - the types are there for type safety and to prevent conflicts between third-party exchanges 👍

kitten:
So regardless of this being associated with testing or debugging (those were just other use-cases that came to mind), I believe the `client.eventTarget.dispatchEvent` naming doesn't immediately indicate that this isn't a core part of the package / client's architecture.

So while the terms "info" or "logging" (the latter indeed being inaccurate) may not be what we want here, I think it may make sense to replace the term "event" here, e.g. `client.debugEvents.dispatch` or `client.info.dispatch`, etc.

Overall I believe it should be clear to someone using this or reading our exchanges that this isn't something they need to learn, or something that we have to mention explicitly in the "Concepts" docs to avoid confusion.

To us this is just an addition, but I think using the term "events" in this case without other _hints_ in the code to its purpose could make it be confused with streams, since we use the term "events" frequently to explain what streams/observables are. (ref: https://formidable.com/open-source/urql/docs/concepts/stream-patterns/)

packages/core/src/types.ts
```diff
@@ -76,12 +76,29 @@ export interface OperationResult<Data = any> {
...
 }
 
 /** Function responsible for listening for streamed [operations]{@link Operation}. */
 export type Exchange = (input: ExchangeInput) => ExchangeIO;
 
 /** Function responsible for receiving an observable [operation]{@link Operation} and returning a [result]{@link OperationResult}. */
 export type ExchangeIO = (ops$: Source<Operation>) => Source<OperationResult>;
+
+/** Debug event types (interfaced for declaration merging). */
+export interface DebugEventTypes {
```
JoviDeCroock: I think strictly typing this defeats the purpose of this, what if a thirth party exchange is used that has a `cache successfully restored` event (hypothetical). Are we expecting them to move to this repository to add that type here and wait for a publish? We could strictly type this in integrations (like devtools) so we avoid having to support all event-types.

andyrichardson:
Sure thing @kitten I wasn't aware that your issue was with the use of the term `eventTarget` and I've renamed it to something less generic - `debugTarget`

packages/core/src/types.ts
kitten: So, `dispatchEvent` does allow for `string` I believe https://github.com/FormidableLabs/urql/pull/608/files#diff-65274cc049fe6e9b471d39b43d571e44R18

But I think we can address the events themselves in a separate PR, when we implement them. Specifically, `cacheInvalidation` will be different between Graphcache and the document cache, so I'd treat this as tentative events for now that will be replaced when we start adding this to the exchanges (?)

.changeset/early-camels-live.md
```diff
@@ -0,0 +1,7 @@
+---
+'@urql/exchange-graphcache': minor
```
JoviDeCroock: I'm not entirely sure if these are minor since we aren't exposing added functionality to the user apart from an added benefit when they bumped `@urql/core`

CC @kitten

andyrichardson: Feature addition justifies minor no? Those working on their own exchanges would likely more than a patch.

andyrichardson:
I've found a limitation of the current state notably being the lack of source.

We can either get this merged and add source support once I've had time to flesh it out or can hold this off for now.

.changeset/early-camels-live.md
JoviDeCroock: Thinking about this, we might have to update the babel plugin to also check for `client.target` just so people who update `graphCache`/... won't crash their app if `core` isn't on par

packages/core/src/types.ts
```diff
@@ -5,9 +5,6 @@ import { CombinedError } from './utils/error';
 
 export { ExecutionResult } from 'graphql';
 
-/** Utility type to Omit keys from an interface/object type */
-export type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
```
andyrichardson: This is a TS global now

packages/core/src/types.ts
```diff
@@ -113,17 +113,16 @@ export interface DebugEventTypes {
...
   };
-  // Graphcache Exchnage
-  graphcacheHit: {
-    policy: RequestPolicy;
-    value: OperationResult;
-  };
 }
 
-export type DebugEvent<T extends keyof DebugEventTypes | string> = {
+export type DebugEventArg<T extends keyof DebugEventTypes | string> = {
```
andyrichardson: Would have preferred to not require two DebugEvent types but this would not work

```ts
<T extends keyof DebugEventTypes | string>(arg: Omit<DebugEvent<T>, 'source'>) => void
```

packages/core/src/exchanges/compose.ts
```diff
@@ -8,7 +8,15 @@ export const composeExchanges = (exchanges: Exchange[]): Exchange => {
 
   return payload => {
     return exchanges.reduceRight((forward, exchange) => {
-      return exchange({ client: payload.client, forward });
+      return exchange({
+        client: payload.client,
+        forward,
+        dispatchDebug: e =>
+          payload.client.debugTarget!.dispatchEvent({
```
andyrichardson: @kitten is this what you were thinking?

```suggestion
          process.env.NODE_ENV === "production" ? noop : payload.client.debugTarget!.dispatchEvent({
```

.changeset/early-camels-live.md
kitten: I think we’d want to bump `core` by a minor, but we can get away with bumping everything else by `patch` 😇

packages/core/src/client.ts
```diff
@@ -73,11 +76,13 @@ export const createClient = (opts: ClientOptions) => new Client(opts);
...
 /** The URQL application-wide client library. Each execute method starts a GraphQL request and returns a stream of results. */
 export class Client {
+  // Event target for monitoring
+  debugTarget?: Target;
+
   // These are variables derived from ClientOptions
   url: string;
   fetch?: typeof fetch;
   fetchOptions?: RequestInit | (() => RequestInit);
-  exchange: Exchange;
```
andyrichardson: Assuming this is intended to be a private variable - let me know if that isn't the case.

.changeset/early-camels-live.md
andyrichardson: The exchange libraries take a new argument - [that's a minor](https://semver.org/#spec-item-7)

packages/core/src/exchanges/compose.ts
```diff
@@ -1,14 +1,20 @@
...
+  client,
+  forward,
+}: Omit<ExchangeInput, 'dispatchDebug'>) =>
+  exchanges.reduceRight(
+    (forward, exchange) =>
+      exchange({
+        client,
+        forward,
+        dispatchDebug: e =>
+          client.debugTarget!.dispatchEvent({
```
wgolledge: Will passing it like this still work with the babel plugin?

packages/core/src/exchanges/compose.ts
kitten: Yep :+1:

packages/core/src/exchanges/compose.ts
andyrichardson: It'll require changes but hopefully should make our babel plugin more accurate!

packages/core/src/exchanges/compose.ts
andyrichardson: This is handled in the babel plugin now :+1:

packages/core/src/types.ts
```diff
@@ -76,12 +73,59 @@ export interface OperationResult<Data = any> {
...
+  type: T;
+  message: string;
+  operation: Operation;
+} & (T extends keyof DebugEventTypes
+  ? { data: DebugEventTypes[T] }
+  : { data?: any });
+
+export type DebugEvent<
+  T extends keyof DebugEventTypes | string
```
andyrichardson:
```suggestion
  T extends keyof DebugEventTypes | string = string
```

packages/core/src/exchanges/fetch.ts
```diff
@@ -157,16 +175,40 @@ const executeFetch = (
...
       return makeResult(operation, result, response);
     })
     .catch((error: Error) => {
-      if (error.name !== 'AbortError') {
-        return makeErrorResult(
-          operation,
-          statusNotOk ? new Error(response.statusText) : error,
-          response
-        );
+      if (err.name === 'AbortError') {
```
andyrichardson, line 192:
```suggestion
      if (error.name === 'AbortError') {
```

packages/core/src/exchanges/fetch.ts
```diff
@@ -157,16 +175,40 @@ const executeFetch = (
...
+        return;
       }
+
+      dispatchDebug({
+        type: 'fetchError',
+        message: err.name,
```
andyrichardson, line 198:
```suggestion
        message: error.name,
```

packages/core/src/exchanges/fetch.ts
```diff
@@ -157,16 +175,40 @@ const executeFetch = (
...
+        operation,
+        data: {
+          url,
+          fetchOptions: opts,
+          value: err,
```
andyrichardson, line 203:
```suggestion
          value: error,
```

exchanges/graphcache/src/cacheExchange.ts
```diff
@@ -305,7 +306,14 @@ export const cacheExchange = (opts?: CacheExchangeOpts): Exchange => ({
           res.operation.context.requestPolicy !== 'cache-only'
         );
       }),
-      map(res => addCacheOutcome(res.operation, 'miss'))
+      map(res => {
+        dispatchDebug({
+          type: 'graphcacheMiss',
```
JoviDeCroock, line 311: Wondering if we need to generalise this more `cacheMiss` (because we also do just `cacheHit`)

exchanges/graphcache/src/cacheExchange.ts
```diff
@@ -338,6 +346,15 @@ export const cacheExchange = (opts?: CacheExchangeOpts): Exchange => ({
             );
           }
 
+          dispatchDebug({
+            type: 'cacheHit',
+            message: `A requested operation was found and returned from the cache.`,
+            operation: res.operation,
```
JoviDeCroock, line 352: What parts of operation are we using, just wondering if this is too much

exchanges/retry/src/retryExchange.ts
```diff
@@ -75,14 +75,23 @@ export const retryExchange = ({
           })
         );
 
+        dispatchDebug({
+          type: 'retryRetrying',
```
JoviDeCroock, line 79: Are we making it a standard to prefix it with the exchange or are we going to aim at very generic ones, so it's easier to integrate?

packages/core/src/exchanges/fetch.ts
```diff
@@ -118,7 +123,9 @@ const createFetchSource = (operation: Operation, shouldUseGet: boolean) => {
     let ended = false;
 
     Promise.resolve()
-      .then(() => (ended ? undefined : executeFetch(operation, fetchOptions)))
+      .then(() =>
+        ended ? undefined : executeFetch(dispatchDebug, operation, fetchOptions)
```
JoviDeCroock, line 127: this is a potentially undefined argument, let's pass it as the last one instead

exchanges/graphcache/src/cacheExchange.ts
kitten: I’m concerned it may not always be serialisable, ie fetch options being a function.

Does it make sense to just send over the request (query, variables, key) and then just display the rest from/on the events?

exchanges/graphcache/src/cacheExchange.ts
andyrichardson:
```suggestion
          type: 'cacheMiss',
```

Seems about right - both the [default cache](https://github.com/FormidableLabs/urql/pull/608/files#diff-6c7d190a26cd6cac287789abab335c32R70) and graphCache dispatch the same "event type"

packages/core/src/exchanges/fetch.ts
andyrichardson:
```suggestion
        ended ? undefined : executeFetch(operation, fetchOptions, dispatchDebug)
```

packages/core/src/exchanges/fetch.ts
```diff
@@ -137,13 +144,24 @@ const createFetchSource = (operation: Operation, shouldUseGet: boolean) => {
 };
 
 const executeFetch = (
+  dispatchDebug: ExchangeInput['dispatchDebug'],
```
andyrichardson, line 147:
```suggestion
```

packages/core/src/exchanges/fetch.ts
```diff
@@ -137,13 +144,24 @@ const createFetchSource = (operation: Operation, shouldUseGet: boolean) => {
...
   operation: Operation,
   opts: RequestInit
 ): Promise<OperationResult> => {
```
andyrichardson, line 150:
```suggestion
  dispatchDebug: ExchangeInput['dispatchDebug'],
): Promise<OperationResult> => {
```

exchanges/retry/src/retryExchange.ts
andyrichardson: Yeah this should have been amended now that we've added the `source` exchange to events.

Multiple exchanges can use the same type, as long as the semantic meaning and passed arguments are the same.

exchanges/graphcache/src/cacheExchange.ts
andyrichardson: We're currently [serialising the dispatched events](https://github.com/FormidableLabs/urql-devtools-exchange/pull/26/files#diff-76d6bc8407250594abb3ec8eb8a74cf8R97) in the devtools exchange so I'm not too concerned about this as long as we don't have / introduce circular data structures.

From a DX experience, this just means there's no need to destructure the operation every dispatch and gives us more flexibility down the road.

kitten: Ok; wasn’t too sure how this would be hooked up but sounds good to me 👍

packages/core/src/exchanges/compose.test.ts
```diff
@@ -1,14 +1,11 @@
...
 import { composeExchanges } from './compose';
+import { noop } from '../utils';
 
-const mockClient = {
-  debugTarget: {
-    dispatchEvent: jest.fn(),
-  },
-} as any;
-const forward = jest.fn();
+const mockClient = {} as any;
```
andyrichardson, line 6: @kitten if we're going ahead with not exposing the dispatchDebug can we go ahead and flag it as private rather than just changing our consumption of the client?

Issue #620, opened 2020-03-14
maskTypeName and Cannot use 'in' operator to search for '__typename' in null


zenios:
The problem should be here 
https://github.com/FormidableLabs/urql/blob/d0a3a5e55343f38c5acbaf7f57c299f81930bbff/packages/core/src/utils/maskTypename.ts#L13

PR #621, opened 2020-03-15
(core) - fix maskTypenames clauses

JoviDeCroock:
## Summary

Fixes https://github.com/FormidableLabs/urql/issues/620

`typeof null === 'object'` so we need to check if we have a truthy value before going into this branch.

## Set of changes

- add check for `maskTypenames` to see if we are dealing with a truthy value before checking the object properties.

### DIFF
--- a/.changeset/red-candles-decide.md
+++ b/.changeset/red-candles-decide.md
@@ -1,0 +1,5 @@
+---
+'@urql/core': patch
+---
+
+Add a guard to "maskTypenames" so a null value isn't considered an object.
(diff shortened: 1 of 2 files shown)


.changeset/red-candles-decide.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/core': patch
+---
+
+Add a guard to "maskTypenames" so a null value isn't considered an object.
```
kitten: Just for other changesets, let’s write them more in the mood of why sth has changed rather than what has changed :)

PR #622, opened 2020-03-16
(graphcache) - Change Store constructor to accept options object

kitten:
This changes `cacheExchange` to just pass on its options and `Store` to accept an options object as well to make it easier to use directly.

```js
// before:
new Store(
  undefined, // schema
  { /* resolvers */ },
  { /* updates */ },
  { /* optimistic */ },
  { /* keys */ },
);

// after:
new Store({
  schema: undefined,
  resolvers: {},
  updates: {},
  optimistic: {},
  keys: {},
});
```

`storage` is exempted since the rehydration logic is coupled to the `cacheExchange` (for now). Bundle size remains unchanged.

### DIFF
--- a/exchanges/graphcache/src/cacheExchange.ts
+++ b/exchanges/graphcache/src/cacheExchange.ts
@@ -88,13 +88,7 @@ export const cacheExchange = (opts?: CacheExchangeOpts): Exchange => ({
 }) => {
   if (!opts) opts = {};
 
-  const store = new Store(
-    opts.schema,
-    opts.resolvers,
-    opts.updates,
-    opts.optimistic,
-    opts.keys
-  );
+  const store = new Store(opts);
 
   let hydration: void | Promise<void>;
   if (opts.storage) {
--- a/exchanges/graphcache/src/cacheExchange.ts
+++ b/exchanges/graphcache/src/cacheExchange.ts
@@ -86,12 +86,10 @@ export const cacheExchange = (opts?: CacheExchangeOpts): Exchange => ({
   forward,
   client,
 }) => {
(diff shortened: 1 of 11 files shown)


exchanges/graphcache/src/cacheExchange.ts
```diff
@@ -86,12 +86,10 @@ export const cacheExchange = (opts?: CacheExchangeOpts): Exchange => ({
   forward,
   client,
 }) => {
-  if (!opts) opts = {};
```
andyrichardson: Good shout on removing this!

kitten: To be fair, I've just moved it 😆  but typically transpilers will jump through more hoops in the generated code with default params rather than reassigning the argument. It's a minor details though, since we do get the type safety from TypeScript in either cases :+1:

JoviDeCroock: `default params` are very expensive size wise, see [this REPL](https://babeljs.io/en/repl#?browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYewdgzgLgBAHjAvDAFAbwFAxgJwIZgAmIAtkjAKwA0WMANgKYQQBKBxJNAvgJRIB8MTNkbM2RUuVGt2kgD5yYANgDctHAygBXHGFyyyAKnpMZEkmq4YgA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Cenv&prettier=false&targets=&version=7.8.7&externalPlugins=)

PR #628, opened 2020-03-17
Compile away dispatchEvent

JoviDeCroock:


packages/core/src/exchanges/cache.ts
```diff
@@ -59,7 +59,7 @@ export const cacheExchange: Exchange = ({ forward, client }) => {
         const cachedResult = resultCache.get(operation.key);
 
         if (cachedResult) {
-          client.eventTarget.dispatchEvent({
+          client.eventTarget!.dispatchEvent({
```
kitten: Seems like we should enforce that `eventTarget` is defined and assume that it does, since the babel plugin takes care of creation for us?

scripts/babel/transform-event-target.js
```diff
@@ -0,0 +1,38 @@
+const warningDevCheckTemplate = `
+  if (process.env.NODE_ENV !== 'production') {
+    NODE;
+  }
+`.trim();
+
+const visited = 'visitedByEventTargetTransformer';
+const calleeProperty = 'dispatchEvent';
+const calleeObject = 'eventTarget';
+const client = 'client';
```
kitten: Minor nit: I know we want to enforce that it's called on the client, but the name `client` isn't necessarily the name of a local variables 😅  so I think we'll have to ignore the name of the `callee`

packages/core/src/exchanges/cache.ts
JoviDeCroock: I think this is fine since this essentially comes down to TS not knowing that this is undefined in prod and that the function can't be called in prod. I could also just strip everything from the class instead if you like?

kitten: Nah, I think this is a minor detail so either is actually fine on second thought :+1:

packages/core/src/client.ts
```diff
@@ -328,8 +332,4 @@ export class Client {
     const operation = this.createRequestOperation('mutation', query, opts);
     return this.executeRequestOperation(operation);
   };
-
-  public get debugTarget() {
```
andyrichardson: Doesn't removing this mean exchanges could rewrite the debugTarget?

JoviDeCroock: They can always do that if they're not using TypeScript, this class is transpiled to a function either way

kitten: We could set it as non-writeable, but I mean if we wanted to do that for every property and method, then we’d be doing this a lot 😅

I wouldn’t be too worried about this. When an exchange reassigns this they’ve already broken the API’s contract 🤷‍♀️

PR #632, opened 2020-03-17
Set logo fixed height for safari bug

wgolledge:
## Summary

- Add fixed height to fix Logo bug on safari

### DIFF
--- a/.changeset/spicy-flies-push.md
+++ b/.changeset/spicy-flies-push.md
@@ -1,0 +1,5 @@
+---
+'urql-docs': patch
+---
+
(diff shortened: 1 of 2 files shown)


.changeset/spicy-flies-push.md
```diff
@@ -0,0 +1,5 @@
+---
+'urql-docs': patch
+---
+
+Add fixed height to logo to fix Safari bug
```
kitten: So `changeset` has broken their CLI and the site is now showing up although it's set to `private`, but we won't need to create changesets for it btw :)

PR #650, opened 2020-03-20
(core) - Fix stringifyVariables (and thus operation keys) for Files

### LINKED ISSUE #649: mutations with different files in parallel, results lost
codesandbox links:
server: https://codesandbox.io/s/graphql-server-1-l6ry3
client: https://codesandbox.io/s/graphql-client-1-dtnkx

i have a schema with a single mutation:
```
type Mutation {
   createMessage:String!
```
the server always returns a random string (it contains a random uuid).

then i have an urql-client, where if you press a button,
i execute two `createMessage` mutations at the same time,
and display their result. 

expected behavior:
- i get two different messages

actual behavior:
- i get the same message two times

i looked into the code,and what seems to be happening is this:
- two ajax-requests are sent out (the two mutations), and both ajax-requests return with different messages. so far so good
- but, both ajax-requests contain the same operation-key, so (it seems) as soon as one finishes, URQL finishes both mutations with the same result.
- i tried to use a `fetchPolicy: 'network-only'`, but that did not help
- it seems the operation-key is a hash from the mutation and variables. so, when the mutation and variables are the same, the operation-key is the same.

am i doing something wrong here? for  me it seems mutations should not be deduplicated this way. 

i understand that i can just send some mutation-variable that always changes and it willl solve the problem, but that seems ugly :(

p.s: this is just a minimal example, the place where i found the issue was with using the new multipart-form-data-exchange, i was  uploading pictures, and only the picture changed in the mutations. i assume the key-hash does not take the separately traveling blobs into account.

kitten:
Resolve #649

## Summary

When two operations with a `File` in the variables are sent the operation key would likely be the same since we never take any `File` attributes into account.

We'd hence like to have special handling for objects without any enumerable keys that also aren't plain objects.

## Set of changes

- Add random (but stable) keying for objects without enumerable keys that aren't plain objects

### DIFF
--- a/.changeset/wicked-cows-heal.md
+++ b/.changeset/wicked-cows-heal.md
@@ -1,0 +1,5 @@
+---
+'@urql/core': patch
+---
+
+Fix stringifyVariables (and hence operation keys) for variables containing Files
--- a/.changeset/wicked-cows-heal.md
+++ b/.changeset/wicked-cows-heal.md
@@ -2,4 +2,4 @@
 '@urql/core': patch
 ---
 
-Fix stringifyVariables (and hence operation keys) for variables containing Files
+Add support for variables that contain non-plain objects without any enumerable keys, e.g. `File` or `Blob`. In this case `stringifyVariables` will now use a stable (but random) key, which means that mutations containing `File`s — or other objects like this — will now be distinct, as they should be.
--- a/packages/core/src/utils/stringifyVariables.test.ts
+++ b/packages/core/src/utils/stringifyVariables.test.ts
@@ -29,7 +29,19 @@ it('throws for circular structures', () => {
   }).toThrow();
 });
 
-it('stringifies date correctly', () => {
+it('stringifies dates correctly', () => {
   const date = new Date('2019-12-11T04:20:00');
   expect(stringifyVariables(date)).toBe(date.toJSON());
+});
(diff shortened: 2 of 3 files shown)


gabor:
thanks for looking into this problem. i don't want to complicate things too much, but you can assign `Blob` objects too,probably, and those do not have names or modification-dates :-(

(as i explained in https://github.com/FormidableLabs/urql/issues/649, i would prefer to just drop deduplication for mutations, but still, with this pull-requests, things will be more consistent, and that is a good thing)

JoviDeCroock:
> thanks for looking into this problem. i don't want to complicate things too much, but you can assign `Blob` objects too,probably, and those do not have names or modification-dates :-(
> 
> (as i explained in #649, i would prefer to just drop deduplication for mutations, but still, with this pull-requests, things will be more consistent, and that is a good thing)

Mutations aren't deduplicated it's a single stream of events, so the response is associated with an operationKey. We need a way to distinguish these things efficiently

kitten:
We could likely just use `”[Blob ${size}]` for now?

kitten:
Ok, we've found an alternative. Instead we now generate a random key for objects with non-enumerable keys that aren't "plain objects". So this should support `File`, `Blob`, and more

PR #652, opened 2020-03-20
fix match usage

Jay-A-McBee:
## Summary
I was solving a similar problem in the formidable.com rewrite, taking inspiration from this slick implementation and noticed that the value returned from `match` was not being treated as an array here.

## Set of changes
- add `parsePathname` function and handle returned value like an array


kitten:
Awesome! Nice catch 🙌

Issue #653, opened 2020-03-21
[graphcache] mutation update executed twice


zenios:
Hi All

I have a mutation update configured for graphcache that for some weired reason it is executed twice for a single mutation.

One execution with empty result
Second execution with filled result

I did some debugging and i can see that the first execution is done via writeOptimistic before the mutation is send to the server.

And the second is the correct execution done after the results arrives.

I dont have any optimistic configurations in place

zenios:
The mutation is

```
mutation CreateFriendRequestMutation($withUserId: ID!) {
  createFriendRequest(withUser: $withUserId)
}
```

I dont know what other information i can provide

JoviDeCroock:
A reproduction is the source of all information, it's very hard for us to make an assumption as to what variables you are calling with, ... For instance why is there no result is it an invalid call or the same call as the second?

zenios:
The mutation accept a string and returns a boolean.

And the update is called twice one with result as {} and the second with result {createFriendRequest: true}

JoviDeCroock:
So what I can derive from your current explanation is the following:

You have:

```
cacheExchange({
  updates: { Mutation: { createFriendRequest: () => smth  } },
  optimistic: { createFriendRequest: () => smth  },
})
```

And you are asking why `createFriendRequest` is called twice, right? The last argument of your updater function is `ctx` which has a property called `optimistic` this can help your determine if this is called because of an `optimistic` update or just your regular response.

This because even if it's an optimistic update you can have special logic, like adding an item to a list, we allow the user to utilise the updater function for the response they construct in `optimistic` and the one they get from their server-response.

zenios:
Here is a sandbox

https://codesandbox.io/s/focused-night-ztnmm

I have a console log inside update that you can see it logs twice

The only difference from my case is in sandbox case both result and args are null on first execution
and args null on second execution

JoviDeCroock:
Thanks for the reproduction @zenios that's appreciated, it had to do with the fact that the mutation had no selectionset, there's a fix PR up!

zenios:
You are welcome

kitten:
Fixed in https://github.com/FormidableLabs/urql/releases/tag/%40urql%2Fexchange-graphcache%402.3.0

PR #654, opened 2020-03-21
(graphcache) - fix no selectionset mutation updates in optimistic phase

JoviDeCroock:
## Summary

Their was a scenario where a mutation without selectionset would cause a call to its respective updater function.

```gql
mutation {
  someUpdate
}
```

As you can see the above has no selectionset, this would cause the `writeSelectionSet` not to check if there was an actual optimistic-entry for the current rootField

Fixes: https://github.com/FormidableLabs/urql/issues/653

## Set of changes

- fix no selectionset mutation updates in optimistic phase

### DIFF
--- a/exchanges/graphcache/src/operations/write.ts
+++ b/exchanges/graphcache/src/operations/write.ts
@@ -249,7 +249,7 @@ const writeSelection = (
       InMemoryData.writeRecord(entityKey || typename, fieldKey, fieldValue);
     }
 
-    if (isRoot) {
+    if (isRoot && (!ctx.optimistic || (ctx.optimistic && ctx.store.optimisticMutations[fieldName]))) {
       // We have to update the context to reflect up-to-date ResolveInfo
       updateContext(
         ctx,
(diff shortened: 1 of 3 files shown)


JoviDeCroock:
I fixed the codesandbox in this PR  https://github.com/FormidableLabs/urql/pull/656

exchanges/graphcache/src/operations/write.ts
```diff
@@ -247,7 +247,7 @@ const writeSelection = (
     } else if (entityKey && !isRoot) {
       // This is a leaf node, so we're setting the field's value directly
       InMemoryData.writeRecord(entityKey || typename, fieldKey, fieldValue);
-    }
+    } else if (ctx.optimistic && isRoot) continue;
```
kitten: Should we leave a TODO here for scalar optimistic results or just address that in a subsequent PR?
I’d have time to take a look at that, because I suppose it’s expected behaviour 🤷‍♀️

JoviDeCroock: I'd like to defer this to a later PR

Issue #655, opened 2020-03-21
[react] Can't perform a React state update on an unmounted component.


zenios:
Reproduction

https://codesandbox.io/s/wizardly-wright-f42o1

This has been discussed again and fixed

https://github.com/FormidableLabs/urql/issues/493

but i think it broke again when use-sbuscription was removed

JoviDeCroock:
I'm not seeing any errors, I just clicked the button a few times.

![image](https://user-images.githubusercontent.com/17125876/77237087-2dd6c280-6bc5-11ea-9395-3e96395ae175.png)

zenios:
![Screenshot from 2020-03-21 23-43-15](https://user-images.githubusercontent.com/89569/77237110-e99bf000-6bcd-11ea-87a6-e76849832d65.png)

JoviDeCroock:
Got it, your sandbox was missing the `{!data.todo.complete && <Button />}` part, was probably lagging behind for me. Thanks!

zenios:
You are on fire :) 

Two issues both fixed in less than an hour. Congrats

PR #656, opened 2020-03-21
Fix react warning

JoviDeCroock:
## Summary

when `executeMutation` still has to setState and the `useQuery` (or related) has reset the state already this could be an issue.

fixes: https://github.com/FormidableLabs/urql/issues/655

## Set of changes

- add `isMounted` checks to `useSource` and `useMutation`


packages/react-urql/src/hooks/useSource.ts
```diff
@@ -19,6 +19,8 @@ import { useClient } from '../context';
 let currentInit = false;
```
JoviDeCroock: should we move this inside the hook as well? all be it just to unmark this as a side-effect?

kitten: It needs to stay outside the hook for cross-component updates to be prevented

exchanges/suspense/package.json
```diff
@@ -19,8 +19,8 @@
     "react",
     "suspense"
   ],
-  "main": "dist/urql-exchange-suspense",
-  "module": "dist/urql-exchange-suspense.mjs",
+  "main": "dist/urql-exchange-suspense.js",
+  "module": "dist/urql-exchange-suspense.es.js",
```
kitten: I think we’ll also want “exports” removed. I’ll do another check and maybe a subsequent PR as I’m unsure whether we need the dependency on React or whether it’s just to indicate that it’s _for_ React 😅

PR #659, opened 2020-03-23
Add debug dispatches for fetch + fallback exchanges

andyrichardson:
## About

Add debug dispatches for key events in fetch and fallback exchanges

### DIFF
--- a/packages/core/src/exchanges/dedup.ts
+++ b/packages/core/src/exchanges/dedup.ts
@@ -3,6 +3,7 @@ import { Exchange, Operation, OperationResult } from '../types';
 
 /** A default exchange for debouncing GraphQL requests. */
 export const dedupExchange: Exchange = ({ forward, client }) => {
+  const dispatchEvent = client.debugTarget!.dispatchEvent;
   const inFlightKeys = new Set<number>();
 
   const filterIncomingOperation = (operation: Operation) => {
@@ -20,7 +21,7 @@ export const dedupExchange: Exchange = ({ forward, client }) => {
     inFlightKeys.add(key);
 
     if (isInFlight) {
-      client.debugTarget!.dispatchEvent({
+      dispatchEvent({
         type: 'dedup',
         message: 'An operation has been deduped.',
         operation,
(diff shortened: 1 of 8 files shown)


packages/core/src/exchanges/dedup.ts
```diff
@@ -3,6 +3,7 @@ import { Exchange, Operation, OperationResult } from '../types';
 
 /** A default exchange for debouncing GraphQL requests. */
 export const dedupExchange: Exchange = ({ forward, client }) => {
+  const dispatchEvent = client.debugTarget!.dispatchEvent;
```
JoviDeCroock: This won't be defined (read: will throw) and will bypass the babel plugin, let's keep doing this in full (or edit the babel plugin to remove this as well)

packages/core/src/exchanges/fallback.ts
```diff
@@ -1,18 +1,26 @@
...
-        operationName !== 'teardown' &&
+        operation.operationName !== 'teardown' &&
         process.env.NODE_ENV !== 'production'
       ) {
-        console.warn(
-          `No exchange has handled operations of type "${operationName}". Check whether you've added an exchange responsible for these operations.`
-        );
+        const message = `No exchange has handled operations of type "${operation.operationName}". Check whether you've added an exchange responsible for these operations.`;
+
+        client.debugTarget?.dispatchEvent({
```
JoviDeCroock: let's make this an exclamation mark since we're sure this is in dev

packages/core/src/exchanges/fetch.ts
```diff
@@ -132,12 +140,24 @@ const createFetchSource = (operation: Operation, shouldUseGet: boolean) => {
 };
 
 const executeFetch = (
+  client: Client,
   operation: Operation,
   opts: RequestInit
 ): Promise<OperationResult> => {
   const { url, fetch: fetcher } = operation.context;
+  const dispatchEvent = client.debugTarget!.dispatchEvent;
```
JoviDeCroock: ditto

packages/core/src/exchanges/fetch.ts
```diff
@@ -150,11 +170,37 @@ const executeFetch = (
...
+          url,
+          fetchOptions: opts,
+          value: result,
+        },
+      });
+
+      return makeResult(operation, result, response);
+    })
     .catch(err => {
-      if (err.name !== 'AbortError') {
```
JoviDeCroock: let's keep this condition and just move `dispatchEvent` in here

PR #660, opened 2020-03-23
Add debugEvent dispatches for retryExchange and graphcacheExchange

wgolledge:
## Changes

- As above
- Change the `retryCount` increment logic

--- 

Do you think it would be useful to log any other graphcache events that I've missed?

### DIFF
--- a/exchanges/graphcache/src/cacheExchange.ts
+++ b/exchanges/graphcache/src/cacheExchange.ts
@@ -352,7 +352,7 @@ export const cacheExchange = (opts?: CacheExchangeOpts): Exchange => ({
             (policy === 'cache-first' && outcome === 'partial')
           ) {
             debugMessage +=
-              ' but is being retried due to the policy or outcome';
+              outcome === 'partial' && policy === 'cache-first' ? ' but is being retried due to the result being partial.' : ' but is being retried due to "cache-and-network.';
             result.stale = true;
             client.reexecuteOperation(
               toRequestPolicy(operation, 'network-only')
--- a/exchanges/graphcache/src/cacheExchange.ts
+++ b/exchanges/graphcache/src/cacheExchange.ts
@@ -322,7 +322,7 @@ export const cacheExchange = (opts?: CacheExchangeOpts): Exchange => ({
       map(res => {
         client.debugTarget!.dispatchEvent({
           type: 'graphcacheMiss',
-          message: 'The operation could not be queried from the cache',
+          message: 'The result could not be retrieved from the cache',
           operation: res.operation,
         });
         return addCacheOutcome(res.operation, res.outcome);
--- a/exchanges/graphcache/src/cacheExchange.ts
+++ b/exchanges/graphcache/src/cacheExchange.ts
@@ -287,6 +287,7 @@ export const cacheExchange = (opts?: CacheExchangeOpts): Exchange => ({
   };
 
   return ops$ => {
+    const dispatchEvent = client.debugTarget!.dispatchEvent;
     const sharedOps$ = pipe(ops$, share);
 
     // Buffer operations while waiting on hydration to finish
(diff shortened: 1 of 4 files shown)


exchanges/graphcache/src/cacheExchange.ts
```diff
@@ -287,6 +287,7 @@ export const cacheExchange = (opts?: CacheExchangeOpts): Exchange => ({
   };
 
   return ops$ => {
+    const dispatchEvent = client.debugTarget!.dispatchEvent;
```
JoviDeCroock: this will require a rewrite of the babel plugin or use `client.debugTarget!.dispatchEvent` all the time since now this will crash in prod

exchanges/graphcache/src/cacheExchange.ts
```diff
@@ -338,16 +346,29 @@ export const cacheExchange = (opts?: CacheExchangeOpts): Exchange => ({
             extensions: res.extensions,
           };
 
+          let debugMessage = 'A cache hit occured';
           if (
             policy === 'cache-and-network' ||
             (policy === 'cache-first' && outcome === 'partial')
           ) {
+            debugMessage +=
+              ' but is being retried due to the policy or outcome';
```
JoviDeCroock:
```suggestion
              outcome === 'partial' && policy === 'cache-first' ? ' but is being retried due to the result being partial.' : ' but is being retried due to "cache-and-network.';
```

I hope this get DCE'd since pure strings are quite horrible to have in bundle

JoviDeCroock:
We might be missing a cache rehydration one and one where it's a cache miss so we're going for a fetch.

exchanges/retry/src/retryExchange.ts
```diff
@@ -51,7 +51,7 @@ export const retryExchange = ({
       retry$,
       mergeMap((op: Operation) => {
         const { key, context } = op;
-        const retryCount = context.retryCount || 0;
+        const retryCount = context.retryCount++ || 1;
```
andyrichardson: This will mutate `context.retryCount` - I'd try and avoid this for operations

```suggestion
        const retryCount = context.retryCount++ || 1;
```

exchanges/retry/src/retryExchange.ts
```diff
@@ -97,18 +106,30 @@ export const retryExchange = ({
...
           // Send failed responses to be retried by calling next on the retry$ subject
           // Exclude operations that have been retried more than the specified max
           nextRetryOperation(res.operation);
           return false;
-        } else {
-          return true;
         }
+
+        client.debugTarget!.dispatchEvent({
+          type: 'retryMaxAttemptsReached',
```
andyrichardson: Minor suggestion - how would you feel about the term _exhausted_?

```suggestion
          type: 'retryExhausted',
```

exchanges/retry/src/retryExchange.ts
```diff
@@ -75,14 +75,23 @@ export const retryExchange = ({
           })
         );
 
+        client.debugTarget!.dispatchEvent({
+          type: 'retryRetrying',
+          message: 'Retrying the operation',
```
andyrichardson: I think the verbosity of the other message below would be more useful.

I'm not totally sure what the cause of the retry is but would something like this work?

```suggestion
          message: `A network request has failed and a retry has been triggered (${retryCount} / ${MAX_ATTEMPTS})`,
```

exchanges/retry/src/retryExchange.ts
JoviDeCroock: Why is the mutation a bad thing? If we want to avoidit we would have to make it

```
 const retryCount = context.retryCount = context.retryCount + 1 || 1;
```

andyrichardson:
@wgolledge is it possible to add some kind of `graphcacheUpdate` for when values are updated by a mutation?

kitten:
@andyrichardson @wgolledge regarding `graphcacheUpdate`. This is very similar to the default cache’s update; so it’s just a reexecution. Can we make sure that the events match up so that the devtools don’t become coupled to our two cache implementations?

The metadata may be different but I believe the core principle matches up nicely 🙏

andyrichardson:
> The metadata may be different but I believe the core principle matches up nicely 🙏

I can see using the same event making sense if they are the same events  - even if it's only the source of the event that is differing.

Issue #663, opened 2020-03-24
Punch up the URQL site


jherr:
The URQL documentation site needs to be more immediately useful. Have a look at the Material-UI site:

![Screen Shot 2020-03-24 at 9 36 21 AM](https://user-images.githubusercontent.com/22392/77452478-37258200-6db3-11ea-9e9d-ef70eff78b0b.png)

Right on the home page they have installation and usage, right below the big logo. That's what people want to know; how do I install it, how do I make my first query. 

Following on from that should be a tight section about what distinguishes URQL from other GraphQL clients.

Also, the title needs to change from 'urql Documentation' to something like 'URQL: Fast, Easy, Reliable GraphQL client', or whatever your tag line is.

kitten:
We've discussed some options internally and I believe we have a couple of tasks for this:

- [ ] Add a "Quick Start" guide to the docs and make it the index page
- [ ] Move the current "Introduction" page to be "Overview"
- [ ] Add code snippets and up-to-date features to the homepage
- [ ] Add a "Comparison" page to the docs somewhere

All in all, the "Quick Start" guide should cover the most basic usage and also discuss the "Why someone would use urql" briefly as well as list some features, to enforce the homepage's message.

The homepage itself needs to simply be updated. The feature set isn't up-to-date and doesn't "sell" the library very well.

jherr:
Happy to review any PR you have up for that. This is a great library.

jherr:
Maybe a few more comments, having gone through the docs trying to understand exchanges.

* The wonka library is cool, but us as an end user of the library I'm not sure why I'm learning about it. I've written a working app that uses queries, mutations, and subscriptions, and I haven't used wonka at all. Can we take that information out? Or just make it a small sidebar?
* The exchanges information needs use cases for when and why I would alter them. For example, you might use an exchange to automatically add authentication, or something like that.
* There is a spelling error on the streams patterns page; `framewrk` should be `framework`.
* The `Operations stream and results stream` graphic could be made more operational by changing out the A, B, and C nodes with the default set of exchanges, then noting best practice places for extension. For example, here is where you would add an exchange to alter the payload before it goes to the client. Here is where you would add an exchange to alter the response from the client.
*  The site could use an FAQ. Topics could include; How queries could be merged? How do I get started with SSR?  That kind of thing

Sorry to just drop this all in here. Honestly trying to help out.

jakubriedl:
[...]

PR #665, opened 2020-03-24
(docs) - quick-start

JoviDeCroock:
This touches on the:

- installation (npm/yarn)
- adding the client
- adding a basic query

This is really the bare minimum to get started, we might want to move this to the homepage only though. This gives a url, ... That will just work (all though it could have a cold-start)

Relates to https://github.com/FormidableLabs/urql/issues/663

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -5,7 +5,11 @@ order: 0
 
 # Quick start
 
-Here we'll see the quickest way to get started with `urql`.
+`urql` is a blazingly fast, lightweight, and customisable GraphQL client. We believe in usability and adaptability, `urql` is built with extensibility at its
+core, allowing you to to include just the key logic for a basic app, in addition to being able to grow to support dynamic single-app applications and highly
+customised infrastructure.
+
+In this page we'll take a look at the quickest route to get up and running with `urql`.
 
 ## Installation
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,49 +1,54 @@
 ---
-title: Introduction
+title: Quick start
 order: 0
 ---
 
-# Introduction
+# Quick start
 
-`urql` is an implementation of a GraphQL client, built to be both easy to use for newcomers to
-GraphQL, as well as extensible, to grow to support dynamic single-app applications and highly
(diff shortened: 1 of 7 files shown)


docs/README.md
```diff
@@ -1,49 +1,54 @@
...
 order: 0
 ---
 
-# Introduction
+# Quick start
 
-`urql` is an implementation of a GraphQL client, built to be both easy to use for newcomers to
-GraphQL, as well as extensible, to grow to support dynamic single-app applications and highly
-customized GraphQL infrastructure. In short, `urql` prioritizes usability and adaptability.
+Here we'll see the quickest way to get started with `urql`.
```
wgolledge: I think we need to retain some sort of selling point / key factor as the first thing that users see:

>`urql` is a blazingly fast, lightweight, and customisable GraphQL client. We believe in usability and adaptability; `urql` is built with extensibility at its core, allowing you to to include just the key logic for a basic app, in addition to being able to grow to support dynamic single-app applications and highly customised infrastructure.
>
>In this page we'll take a look at the quickest route to get up and running with `urql`.

?

docs/overview.md
```diff
@@ -0,0 +1,49 @@
...
+
+As you're adopting GraphQL, `urql` becomes your primary data layer and can handle content-heavy
+pages through ["Document Caching"](./basics/document-caching.md) as well as dynamic and data-heavy
+apps through ["Normalized Caching"](./graphcache/normalized-caching.md).
+
+## Constituent Parts
+
+`urql` can be understood as a collection of connected parts and packages. When you [first get
+started](./basics/getting-started.md) you'll only need to install a single package for your
+framework of choice. We're then able to declaratively send GraphQL requests to our API.
```
wgolledge, line 20: Nit: I think this mix between `you` and `we` is a tiny bit strange in this paragraph. Maybe we could make it all third person?

```suggestion
`urql` can be understood as a collection of connected parts and packages. When [getting started](./basics/getting-started.md) we only need to install a single package for our
framework of choice. We're then able to declaratively send GraphQL requests to our API.
```

PR #666, opened 2020-03-24
(core) - Golf some bytes in typenames

JoviDeCroock:
This removes some redundant code from `core/utils/typenames.ts`

### DIFF
--- a/packages/core/src/utils/typenames.ts
+++ b/packages/core/src/utils/typenames.ts
@@ -36,7 +36,7 @@ export const collectTypesFromResponse = (response: object) =>
 const formatNode = (node: FieldNode | InlineFragmentNode) => {
   if (
     node.selectionSet &&
-    node.selectionSet.selections.some(
+    !node.selectionSet.selections.some(
       node => node.kind === Kind.FIELD && node.name.value === '__typename'
     )
   ) {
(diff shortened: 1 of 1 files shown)


packages/core/src/utils/typenames.ts
```diff
@@ -42,9 +41,7 @@ const hasTypenameField = (set: SelectionSetNode) => {
 };
 
 const formatNode = (node: FieldNode | InlineFragmentNode) => {
-  if (!node.selectionSet) {
-    return false;
```
kitten: The return false here isn’t required I believe hit it does stop `visit` from traversing the rest of the AST. Not sure if that even makes a difference here though

packages/core/src/utils/typenames.ts
```diff
@@ -20,11 +19,10 @@ const collectTypes = (obj: EntityLike | EntityLike[], types: string[] = []) => {
     });
   } else if (typeof obj === 'object' && obj !== null) {
     for (const key in obj) {
-      const val = obj[key];
-      if (key === '__typename' && typeof val === 'string') {
-        types.push(val);
+      if (key === '__typename' && typeof obj[key] === 'string') {
```
JoviDeCroock: can typeof obj[key] even be no string when we are dealing with a __typename?

packages/core/src/utils/typenames.ts
JoviDeCroock: There isn't much to traverse further if it has no selectionSet so I think removing it is acceptable

packages/core/src/utils/typenames.ts
kitten: Technically not. We just haven’t made this assumption for core

JoviDeCroock: let's jsut merge as is then :D

PR #671, opened 2020-03-25
update babel plugin

JoviDeCroock:
https://astexplorer.net/#/gist/7098a93b18f9000dafa4a38ec985ee96/96b512473db6f9c4cebcf8a544b44cd98436efcc

### DIFF
--- a/scripts/babel/transform-debug-target.js
+++ b/scripts/babel/transform-debug-target.js
@@ -5,8 +5,7 @@ const warningDevCheckTemplate = `
 `.trim();
 
 const visited = 'visitedByDebugTargetTransformer';
-const calleeProperty = 'dispatchEvent';
-const calleeObject = 'debugTarget';
+const dispatchProperty = 'dispatchDebug';
 
 const plugin = ({ template, types: t }) => {
   const wrapWithDevCheck = template(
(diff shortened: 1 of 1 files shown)


scripts/babel/transform-debug-target.js
```diff
@@ -16,15 +15,32 @@ const plugin = ({ template, types: t }) => {
 
   return {
     visitor: {
+      ObjectProperty(path) {
+        if (path.node.key && path.node.key.name === dispatchProperty && !path.node[visited]) {
+          path.node[visited] = true;
+		      path.node.value = t.conditionalExpression(
```
andyrichardson: Just curious - how come you went from the template syntax to this?

JoviDeCroock: templating will always add `semicolons` which isn't desirable for properties

kitten: there's alternatives in `@babel/template` for this, but I'm fine with getting a working version in first :) https://babeljs.io/docs/en/babel-template#templateexpression

andyrichardson: Removing the semicolon after NODE in the template string (`NODE;`) doesn't fix that?

JoviDeCroock: My template doesn't have a semi https://github.com/FormidableLabs/urql/pull/671/commits/c4e657c025afbe6de6b527942dd7a188086d4357#diff-21f64d2a8eadde32a78a9f45d0587966L9

andyrichardson: Woah, that's super unusual

PR #674, opened 2020-03-30
Implement new persistence logic compatible with commutativity

kitten:
## Summary

This contains a stable implementation of persistence that combines its run with the garbage collection run and takes commutativity into account using a new approach that squashes layers on-the-fly.

## Set of changes

- [x] Refactor to remove the old persistence logic
- [x] Refactor GC code to use data state
- [x] Add joined key batch to flag persistence
- [x] Implement new persistence deferred run next to `gc`
- [x] Implement new rehydration logic

### DIFF
--- a/exchanges/graphcache/src/store/data.ts
+++ b/exchanges/graphcache/src/store/data.ts
@@ -1,3 +1,5 @@
+import { stringifyVariables } from '@urql/core';
+
 import {
   Link,
   EntityField,
@@ -6,14 +8,15 @@ import {
   SerializedEntries,
 } from '../types';
 
-import { makeDict } from '../helpers/dict';
-import { invariant, currentDebugStack } from '../helpers/help';
 import {
   fieldInfoOfKey,
   joinKeys,
   serializeKeys,
   indexOfSeparator,
 } from './keys';
+
+import { makeDict } from '../helpers/dict';
+import { invariant, currentDebugStack } from '../helpers/help';
 import { scheduleTask } from './defer';
 
 type Dict<T> = Record<string, T>;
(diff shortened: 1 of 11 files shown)


exchanges/graphcache/src/store/data.ts
```diff
@@ -554,25 +538,53 @@ export const inspectFields = (entityKey: string): FieldInfo[] => {
...
-        break;
-      case 114:
-        writeRecord(entityKey, fieldKey, entries[key]);
-        break;
+    const value = entries[key];
+    const sepIndex = indexOfSeparator(key);
+    if (value && sepIndex > -1) {
+      const entityKey = key.slice(0, sepIndex);
+      const fieldKey = key.slice(sepIndex + 1);
+      if (value[0] === ':') {
```
wgolledge: Nit: Could ternary this to save a couple of bytes and avoid nested if

kitten: Closure & Terser should both ternary this automatically :)

exchanges/graphcache/src/store/data.ts
```diff
@@ -319,64 +330,38 @@ const extractNodeMapFields = <T>(
...
-          }
-
-          updateRCForLink(data.gcBatch, data.refCount, linkNode[fieldKey], -1);
-        }
+    // Delete the reference count, and delete the entity from the GC batch
+    delete currentData!.refCount[entityKey];
+    batch.delete(entityKey);
+    currentData!.records.base.delete(entityKey);
+    const linkNode = currentData!.links.base.get(entityKey);
+    if (linkNode) {
```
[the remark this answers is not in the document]
kitten: Yep, it can only be `undefined` or a dictionary :)

exchanges/graphcache/src/store/data.ts
wgolledge: Good to know! 🚀

Issue #693, opened 2020-04-12
Unable to invalidate null Connection type


olistic:
```
urql@1.9.6
@urql/exchange-graphcache@2.3.1
```

**Steps to reproduce**

1. Have a Connection type following the [spec](https://relay.dev/graphql/connections.htm) and a mutation that creates a Node type.

```graphql
type Query {
  exercisesConnection(page: PageInput!): ExercisesConnection
}

type Mutation {
  createExercise(input: CreateExerciseInput!): CreateExerciseResult
}
```

2. Configure the Graphcache's `cacheExchange` so that when creating a Node type, all queries for the corresponding Connection type (i.e. the different pages) are invalidated:

```js
cacheExchange({
  updates: {
    Mutation: {
      createExercise: (_result, _args, cache) => {
        cache
          .inspectFields({ __typename: 'Query' })
          .filter(({ fieldName }) => fieldName === 'exercisesConnection')
          .forEach(({ fieldName, arguments: args }) => {
            cache.invalidate('Query', fieldName, args);
          });
      },
    },
  },
})
```

3. Query the first page for which there are no results:

```graphql
query {
  exercisesConnection({ page: { after: null, first: 10 } }) {
    ...
  }
}
```

This returns `null` for the `exercisesConnection` field as there are no exercises:

<img width="720" alt="Screen Shot 2020-04-11 at 21 45 20" src="https://user-images.githubusercontent.com/5600126/79057973-bfbe7200-7c3d-11ea-97d8-9157b50d2691.png">

4. Execute the mutation and return to the list view.

**Expected behavior**

The query for the first page (3) results in a cache miss and a new request is made to the server, populating the list with the newly created exercise.

**Actual behavior**

The query for the first page results in a cache hit, and the list remains empty.

If the page is refreshed, the new exercise appears:

<img width="724" alt="Screen Shot 2020-04-11 at 21 51 04" src="https://user-images.githubusercontent.com/5600126/79058016-8d614480-7c3e-11ea-9ba5-c9d7d6979379.png">

The curious thing is that, at this point, if a second exercise is created, the cache is properly invalidated and the query is re-executed upon visiting the list view.

I'm gonna work on a reproduction ASAP, but figured I'd open this in the meantime just in case it's obvious what I'm doing wrong.

Thanks!

olistic:
If I add the following logging lines:

```diff
@@ -19,7 +19,10 @@ const cacheOptions = {
      .inspectFields({ __typename: 'Query' })
[...]

PR #694, opened 2020-04-12
(fix) - invalidate null link

JoviDeCroock:
## Summary

There was an issue in invalidation when a link was null, this lead to the cache invalidating a record instead of the link since we were checking for a truthy value rather than just undefined.

Fixes: https://github.com/FormidableLabs/urql/issues/693

## Set of changes

- Change link check to `!== undefined` so we account for `null`


ole.log(entity);
```diff
@@ -0,0 +1,62 @@
+warning: LF will be replaced by CRLF in exchanges/graphcache/src/operations/invalidate.ts.
```
kitten, line 1: Accidental commit?

exchanges/graphcache/src/operations/invalidate.ts
```diff
@@ -18,7 +18,7 @@ export const invalidateEntity = (
 
   for (let i = 0, l = fields.length; i < l; i++) {
     const { fieldKey } = fields[i];
-    if (InMemoryData.readLink(entityKey, fieldKey)) {
+    if (InMemoryData.readLink(entityKey, fieldKey) !== undefined) {
```
kitten, line 21: Oops 😅

ole.log(entity);
JoviDeCroock: Hmm, `eslint` and windows seemed to disagree there did not notice that my bad

olistic:
Thanks for the expedition here! Is there any chance of releasing a patch with this?

kitten:
@olistic I’ll get to releasing this tomorrow 👍 in the meantime you should be able to test this using CodeSandbox CI which publishes temporary packages per PR

olistic:
Thank you @kitten, this is a great tip! 🚀

kitten:
@olistic this has now been released 🌟

PR #697, opened 2020-04-14
(fix) - append query to the payload for multipart fetches

JoviDeCroock:
## Summary

In an issue on [spectrum](https://spectrum.chat/urql/help/help-with-a-file-upload-mutation~08f3413a-0ba5-4577-9fd2-625e10b551cc) there was talk about the `query` getting lost on a multipart-fetch. This is because when using the `clone` we only include the `variables`.

## Set of changes

- Add `query` to the `payload`

### DIFF
--- a/packages/core/src/exchanges/debug.ts
+++ b/packages/core/src/exchanges/debug.ts
@@ -2,7 +2,7 @@ import { pipe, tap } from 'wonka';
 import { Exchange } from '../types';
 
 export const debugExchange: Exchange = ({ forward }) => {
-  if (process.env && process.env.NODE_ENV === 'production') {
+  if (process.env.NODE_ENV === 'production') {
     return ops$ => forward(ops$);
   } else {
     return ops$ =>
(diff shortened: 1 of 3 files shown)


packages/core/src/exchanges/debug.ts
```diff
@@ -2,7 +2,7 @@ import { pipe, tap } from 'wonka';
 import { Exchange } from '../types';
 
 export const debugExchange: Exchange = ({ forward }) => {
-  if (process.env.NODE_ENV === 'production') {
+  if (process.env && process.env.NODE_ENV === 'production') {
```
kitten, line 5: Is this an accidental change? It likely won’t fix any issue 😅

JoviDeCroock: Oops, yes I was experimenting with a plugin to produce full-browser builds 😅

JoviDeCroock:
Closing because this bug was due to a discrepancy in the Elixir spec.

https://elixirforum.com/t/file-upload-mutation-with-abinthe-giving-no-query-document-supplied-error/30624

PR #710, opened 2020-04-17
Use source for debug events

andyrichardson:
## About

Use subject implementation and API for debug events


packages/core/src/client.ts
```diff
@@ -331,4 +332,8 @@ export class Client {
     const operation = this.createRequestOperation('mutation', query, opts);
     return this.executeRequestOperation(operation);
   };
+
+  get debugSource() {
```
andyrichardson, line 336: Open to omitting this and just using the debugSubject property 🤷

JoviDeCroock, line 336: This will need some adjustements to the babel plugin, not entirely sure how `bublé` handles getters might be worth loking into the compiled target here

kitten: I'd just use the property, so we don't even have the compiled getter; the property will be accessible anyway, so shouldn't make a usage difference :)

andyrichardson: Good point, I've moved the API around so it's back to `client.debugTarget` - does that solve our problem?

packages/core/src/client.ts
```diff
@@ -45,7 +46,7 @@ import {
 } from './utils';
 
 import { DocumentNode } from 'graphql';
-import { Target } from './utils/Target';
+import { subscriptionT } from 'wonka/dist/types/src/Wonka_types.gen';
```
JoviDeCroock, line 49: This feels like an incorrect import to me

packages/core/src/client.ts
JoviDeCroock: Yes double checked, it seems to do :)

packages/core/src/client.ts
andyrichardson: Good catch

PR #719, opened 2020-04-19
next-urql: Ensure ctx in clientOptions is available during all stages of SSR

### LINKED ISSUE #707: next-urql: ctx in clientConfig undefined during ssrPrepass
When `ssrPrepass` is ran in the `getInitialProps` function of the `withUrql` HOC, the clientOptions are initialized without app or page context.

This is an issue because I am passing an API token through to a GraphQL endpoint by including it in the request headers. During the SSR prepass phase, the endpoint will not receive a token and will not return the data I'm expecting for the current user.

urql version & exchanges:

- `next-urql@^0.3.6`
- `urql@^1.9.6`
- default exchanges

**Steps to reproduce**

1. Clone and run the reduced test case at https://github.com/nickrttn/next-urql-test-case
2. Take a look at the server logs

**Expected behavior**

I would expect `AppOrPageContext` to be available during all stages of server-side rendering so the environment can be configured correctly.

**Actual behavior**

During the prepass phase, the AppTree component does not receive a preconfigured client or the correct options for the `clientConfig` function, causing it to create a new client in the `useMemo` hook of the `withUrql` render function. That client is not configured correctly when `ctx` is used in the `clientConfig` function.

parkerziegler:
Resolves #707. 

## Summary

In #707, @nickrttn pointed out that `ctx` wasn't available during all stages of SSR with `next-urql`. After some investigation with `console.trace`, it was clear that `ctx` _was_ available during execution of `withUrql`'s `getInitialProps` function, but _not_ during the `useMemo` that gets executed as part of the server side render.

The steps to fix this are twofold:

1. Move the call to create the `clientOptions` _after_ the check for an existing `urqlClient`. If we receive `urqlClient` from `getInitialProps`, we return early from the `useMemo`. This saves us from having to call `initUrqlClient` twice on the server side (the first call being in `getInitialProps`, the second call being inside `useMemo`, where `ctx` is `undefined`).
2. The above requires that we return `urqlClient` in `getInitialProps` on the server side. To ensure that we don't use the same `urqlClient` instance in the browser, we use `toJSON` to ensure it gets serialized to `null`. That way, we don't mess with the separation b/t server and client. This is a trick I picked up from reading the [`next-with-apollo` source](https://github.com/lfades/next-with-apollo/blob/master/src/withApollo.tsx#L99-L105).

I don't envision there being regressions with this approach, since the serialization piece should still allow users to create different server and client `urqlClient` instances with `process.browser` or `typeof window` checks as before.

## Set of changes

**Packages effected:**
- `next-urql`

**Changes:**
- Check if `urqlClient` exists in `useMemo` on the server side to prevent calling `clientConfig` multiple times.
- Ensure `urqlClient` is serialized to `null` to maintain separate server and client instances.

### DIFF
--- a/.changeset/eighty-turkeys-jam.md
+++ b/.changeset/eighty-turkeys-jam.md
@@ -1,0 +1,5 @@
+---
+'next-urql': patch
+---
+
+Check if urqlClient exists in useMemo on the server side to prevent calling clientConfig and initUrqlClient multiple times. Ensure urqlClient is serialized to null to maintain separate server and client instances.
--- a/.changeset/eighty-turkeys-jam.md
+++ b/.changeset/eighty-turkeys-jam.md
@@ -2,4 +2,4 @@
 'next-urql': patch
 ---
 
-Check if urqlClient exists in useMemo on the server side to prevent calling clientConfig and initUrqlClient multiple times. Ensure urqlClient is serialized to null to maintain separate server and client instances.
+Ensure that the Next.js context is available during all stages of SSR. Previously a missing check in `useMemo` on the server-side caused `clientConfig` from being called repeatedly, and another issue may have caused the client from being serialized to initial props.
--- a/packages/next-urql/src/__tests__/with-urql-client.spec.tsx
+++ b/packages/next-urql/src/__tests__/with-urql-client.spec.tsx
@@ -66,9 +66,7 @@ describe('withUrqlClient', () => {
 
   describe('with ctx callback to create client options', () => {
     // Simulate a token that might be passed in a request to the server-rendered application.
-    const token = Math.random()
-      .toString(36)
-      .slice(-10);
+    const token = Math.random().toString(36).slice(-10);
 
     const mockContext: NextUrqlPageContext = {
       AppTree: MockAppTree,
--- a/packages/next-urql/src/__tests__/with-urql-client.spec.tsx
+++ b/packages/next-urql/src/__tests__/with-urql-client.spec.tsx
@@ -50,7 +50,7 @@ describe('withUrqlClient', () => {
       expect(spyInitUrqlClient).toHaveBeenCalledTimes(1);
     });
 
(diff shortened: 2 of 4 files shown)


packages/next-urql/src/__tests__/with-urql-client.spec.tsx
```diff
@@ -59,15 +59,16 @@ describe('withUrqlClient', () => {
       const tree = shallow(<Component {...props} />);
       const app = tree.find(MockApp);
 
-      expect(spyInitUrqlClient).toHaveBeenCalledTimes(2);
```
parkerziegler, line 62: This test description wasn't 💯 accurate. This test runs in a Node environment, so I'm not sure why I even included the "client-side in the component" piece. In any case, now that `urqlClient` is returned in `getInitialProps` on the server, `initUrqlClient` should only be called once on the server 👍

parkerziegler:
I'll leave this open until the morning tomorrow for others to peek at, then merge it then barring any issues. Thanks Jovi!

oorestisime:
I wonder whether this will also fix https://spectrum.chat/urql/help/help-with-setting-up-next-urql-and-getting-token-from-auth0~b1d74f5b-2b8f-4c39-87cb-3687f2d071f6?m=MTU4NzMyNDYwNzc3OA==

.changeset/eighty-turkeys-jam.md
```diff
@@ -0,0 +1,5 @@
+---
+'next-urql': patch
+---
+
+Check if urqlClient exists in useMemo on the server side to prevent calling clientConfig and initUrqlClient multiple times. Ensure urqlClient is serialized to null to maintain separate server and client instances.
```
kitten, line 5:
```suggestion
Ensure that the Next.js context is available during all stages of SSR. Previously a missing check in `useMemo` on the server-side caused `clientConfig` from being called repeatedly, and another issue may have caused the client from being serialized to initial props.
```

I try to keep these informative, but not too in-depth (basically answering "Does this affect me?" as quickly as possible)

PR #739, opened 2020-04-22
Add @urql/exchange-persisted-fetch exchange package

### LINKED ISSUE #196: Persisted queries
Relay and Apollo support a concept called "persisted queries" whereby the GraphQL queries are persisted on the server side, and the client only needs to send an identifier (eg. a hash) to execute it. This avoids having to re-upload and re-parse the query every time it's executed, and the server can store the query in some intermediate form (such as a serialized AST) to improve performance.

Does urql support sending query IDs rather than the entire query? How would I integrate persisted queries into urql? I guess it'd need a Babel transform of some sort? 

Reference: https://facebook.github.io/relay/docs/en/persisted-queries.html

Even better would be support for Apollo's "automatic persisted queries" (see https://blog.apollographql.com/automatic-persisted-queries-and-cdn-caching-with-apollo-server-2-0-bf42b3a313de)

kitten:
Resolve #196

This package implements Persisted Queries support and is a `persistedFetchExchange` that can be added in front of a regular `fetchExchange`.

The `persistedFetchExchange` only handles queries and sends an extra persisted `extensions` entry with a SHA-256 hash. The hash is generated minimally depending on the environment and a replacement library is only included (when possible) in Node.js environments, where size doesn't quite matter.


exchanges/persisted-fetch/src/persistedFetchExchange.ts
```diff
@@ -0,0 +1,112 @@
...
+  operation: Operation
+): Source<OperationResult> => {
+  const body = makeFetchBody(operation);
+  const query: string = body.query!;
+
+  body.query = undefined;
+  body.extensions = {
+    persistedQuery: {
+      version: 1,
+      sha256hash: hash(query),
```
lorenries, line 85: Awesome work!! This should be `sha256Hash` with a capital H for it to work with Apollo Server.

kitten: Oh wow! Thanks for having an eye on this! 🙌

exchanges/persisted-fetch/src/persistedFetchExchange.ts
```diff
@@ -0,0 +1,112 @@
...
+      sharedOps$,
+      filter(operation => operation.operationName === 'query'),
+      mergeMap(operation => {
+        const { key } = operation;
+        const teardown$ = pipe(
+          sharedOps$,
+          filter(op => op.operationName === 'teardown' && op.key === key)
+        );
+
+        if (!supportsPersistedQueries) {
```
JoviDeCroock, line 44: Wondering if this can be true for one query but not another (for example due to microservices)

zenios:
If the endpoint delegates to different graphql servers per request then it
might.

exchanges/persisted-fetch/src/persistedFetchExchange.ts
kitten: I'm just mirroring what the spec / Apollo Persisted Queries says. But generally, a merged schema should be responsible for the extension, since queries can overlap several micro-services. So by spec, this shouldn't happen. And if it does, this doesn't explode 😅

PR #740, opened 2020-04-22
Fix empty import cleanup script and move it to Rollup

kitten:
Move script that cleans up empty imports and requires to Rollup and fix it;


scripts/rollup/cleanup-plugin.js
```diff
@@ -0,0 +1,46 @@
...
+      ImportDeclaration(path) {
+        if (path.node.specifiers.length === 0) {
+          path.remove();
+        }
+      },
+      CallExpression(path) {
+        if (
+          t.isIdentifier(path.node.callee) &&
+          path.node.callee.name === 'require' &&
+          path.parent.type === 'ExpressionStatement'
```
kitten, line 16: This statement has been modified. Calls can also be on assignments or other statements, so the safest check is to check whether they're "freestanding", i.e. on `ExpressionStatement` nodes.

PR #741, opened 2020-04-22
Fix persistedFetchExchange import and add debug messages

kitten:


exchanges/persisted-fetch/src/persistedFetchExchange.ts
```diff
@@ -90,24 +99,85 @@ const makePersistedFetchSource = (
...
-  operation: Operation
+  operation: Operation,
+  dispatchDebug: ExchangeInput['dispatchDebug']
 ): Source<OperationResult> => {
   const body = makeFetchBody(operation);
+  const url = makeFetchURL(operation, body);
+  const fetchOptions = makeFetchOptions(operation, body);
 
-  return makeFetchSource(
+  dispatchDebug({
```
andyrichardson, line 153: The 1-1 mapping you've got here will definitely work - lets catch up on the other stuff this week!

PR #745, opened 2020-04-23
(graphcache) - Shift mutation results to top layer

kitten:
This makes a slight adjustment to the data layer reservation.
Optimistic Mutation layers are now not only cleared but will be
shifted to the front of the optimistic order queue when they're
written with a concrete result (non optimistic result).

The optimistic mutation layer is therefore still discarded but
the layer itself is reused, which allows for slight adjustments
in how layers are handled when they transition from being
optimistic to non-optimistic.

This also makes an adjustment that allows layers to transition
from being non-optimistic to optimistic. This shouldn't happen
in practice, but protects us from race-conditions.

### DIFF
--- a/.changeset/long-rocks-give.md
+++ b/.changeset/long-rocks-give.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-graphcache': patch
+---
+
+Adjust mutation results priority to always override query results as they arrive, similarly to subscriptions. This will prevent race conditions when mutations are slow to execute at the cost of some consistency.
--- a/exchanges/graphcache/src/cacheExchange.test.ts
+++ b/exchanges/graphcache/src/cacheExchange.test.ts
@@ -1545,7 +1545,7 @@ describe('commutativity', () => {
     });
 
     expect(reexec).toHaveBeenCalledTimes(2);
-    expect(data).toHaveProperty('node.name', 'query b');
+    expect(data).toHaveProperty('node.name', 'mutation');
   });
 
   it('applies optimistic updates on top of commutative queries', () => {
--- a/exchanges/graphcache/src/cacheExchange.ts
+++ b/exchanges/graphcache/src/cacheExchange.ts
@@ -220,9 +220,9 @@ export const cacheExchange = (opts?: CacheExchangeOpts): Exchange => ({
         optimisticKeysToDependencies.get(key)
       );
       optimisticKeysToDependencies.delete(key);
+    } else {
+      reserveLayer(store.data, operation.key);
     }
-
-    reserveLayer(store.data, operation.key);
 
     let queryDependencies: Set<string> | void;
     if (result.data) {
--- a/exchanges/graphcache/src/cacheExchange.ts
+++ b/exchanges/graphcache/src/cacheExchange.ts
@@ -140,22 +140,21 @@ export const cacheExchange = (opts?: CacheExchangeOpts): Exchange => ({
       reserveLayer(store.data, operation.key);
     } else if (operation.operationName === 'teardown') {
       noopDataState(store.data, operation.key);
-    } else if (operation.operationName === 'mutation') {
+    } else if (
+      operation.operationName === 'mutation' &&
+      operation.context.requestPolicy !== 'network-only'
+    ) {
       // This executes an optimistic update for mutations and registers it if necessary
-      if (operation.context.requestPolicy !== 'network-only') {
-        const { dependencies } = writeOptimistic(
-          store,
-          operation,
-          operation.key
-        );
-        if (dependencies.size !== 0) {
-          optimisticKeysToDependencies.set(operation.key, dependencies);
-          const pendingOperations = new Set<number>();
-          collectPendingOperations(pendingOperations, dependencies);
-          executePendingOperations(operation, pendingOperations);
-        }
-      } else {
-        reserveLayer(store.data, operation.key);
+      const { dependencies } = writeOptimistic(
+        store,
+        operation,
+        operation.key
+      );
+      if (dependencies.size !== 0) {
+        optimisticKeysToDependencies.set(operation.key, dependencies);
+        const pendingOperations = new Set<number>();
+        collectPendingOperations(pendingOperations, dependencies);
+        executePendingOperations(operation, pendingOperations);
       }
     }
 
@@ -225,9 +224,9 @@ export const cacheExchange = (opts?: CacheExchangeOpts): Exchange => ({
         optimisticKeysToDependencies.get(key)
       );
       optimisticKeysToDependencies.delete(key);
-    } else {
-      reserveLayer(store.data, operation.key);
     }
+
+    reserveLayer(store.data, operation.key);
 
     let queryDependencies: Set<string> | void;
     if (result.data) {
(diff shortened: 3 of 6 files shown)


frederikhors:
@kitten is this somehow related to https://spectrum.chat/urql/help/create-mutation-with-a-cache-and-network-list~2d06beed-6d13-4dd6-b977-a78bbc50d82a?

JoviDeCroock:
> @kitten is this somehow related to https://spectrum.chat/urql/help/create-mutation-with-a-cache-and-network-list~2d06beed-6d13-4dd6-b977-a78bbc50d82a?

We're trying some stuff out related to offline

kitten:
@frederikhors Yes! I'm trying to see whether it's better to strike a balance with commutativity, where we sacrifice some consistency but make mutations take precedence, so updates will always show up on top of query results that have been started after the mutations (but arrive before the mutation result).

You can try this branch out with the Graphcache build from CodeSandbox CI, and let us know if that helps :+1:

frederikhors:
I tried.

Something is changed.

I recorded a GIF; I'll comment again in a moment to explain.

frederikhors:
Now it "reloads"/"refetches" todos after the slow mutation but it shows also the optimistic one.

Maybe because I'm using:

```js
const optimistic = {
  addTodo: (variables, cache, info) => {
    return {
      id: 0 - Date.now(), // <--- this
      text: variables.text,
      complete: false,
      __typename: "Todo"
    };
  }
};
```

and the `ID` from server is different.

I use `0 - Date.now()` to make it negative for this: `<tr class:newTodo={todo.id < 0}>` to make `tr` `background` green.

And again now when I come back to `cache-and-network` list my optimistic todo one disappear!

kitten:
@frederikhors Hm, I think I accidentally had a change here that converted your optimistic result to a non-optimistic one: https://github.com/FormidableLabs/urql/pull/745/commits/b943c09dab95c2baf93b60046636dc625cc6026c#diff-59e023625accfe756fba94b9832ba785L228-R230

Could you try the latest change on `c91506e` please, once it finishes on CodeSandbox CI? It seems that we had only loose protections against `reserveLayer` converting an optimistic layer to a non-optimistic one, which I accidentally did.

frederikhors:
One problem fixed! 😄

One problem left! 😢

**GIF**:



---

It works now the refetch after slow mutation.

And again now when I come back to `cache-and-network` list - during slow mutation - my optimistic todo one disappear!

kitten:
That's expected behaviour. You may be overusing `cache-and-network`. If you think about the executed operations what happens, if I'm not mistaken is the following:

- You add a todo, which triggers the mutation. This adds your optimistic todo and updates related queries
- You return to the todo list; `cache-and-network` is used, which means that all todos refetch.
- The todo list query completes; this overrides your optimistic todo
- The todo mutation completes; this overrides, which means that's now back, since it's reflected in server-state.

There are several ways to address this and the two are:

- Consider whether this use of optimistic updates is valid. In some cases explicit actions by the user that have wide-ranging changes (typically adding/removing an item) should maybe not be optimistic and instead have explicit progress. I recognise however that this is just an example app, so let's move on from this suggestion
- Don't overuse `cache-and-network`. Ultimately any query that is optimistically altered runs the risk of being overwritten by actual data refetches, which is intentional.

What we're currently discussing though is to "block" `cache-and-network` from sending queries, when an optimistic mutation is in progress for it.

frederikhors:
I've thought about it a lot.

I use optimistics for the most part to signal the user that something is happening; I know I could use Svelte (or React) "simply", but I noticed that it is much simpler and more immediate to use something that I already have available: the optimistic.

In Apollo I had no problem because:

- a slow mutation starts which adds an element
- I create an optimistic element
- back to the list; the optimisitc element is present; the list updates (because `cache-and-network`) but does not delete the optimistic element (or more than one) in progress
- the user notes that they are still in progress
- as soon as the slow mutations end, they update what they need to update

I don't know if Apollo forbids `cache-and-networks` to leave if there are optimistic mutations in progress, but what could be the negative consequences?

In the meantime, I'm trying the news with the example of the mutation slow delete.

_Just everything works and we agree the most idiomatic way, it would be great to create a Wiki or FAQ to document these recipes._

kitten:
@frederikhors it just happens to work in Apollo, because in this case your slow-down happens to be in-order, i.e. the mutation completing last, and optimistic layers always taking precedence in Apollo 😉 

I've been also thinking about `cache-and-network`, and it's simply a bad default to have with `@urql/exchange-graphcache`, since a normalised cache seeks to _avoid_ excessive network requests and share cache updates as much as possible, without sending more requests. It makes more sense to have an exchange that _upgrades_ `cache-first` queries to `cache-and-network` if they haven't been sent in a while (TTL-based).

Anyway, I've pushed a WIP-commit (no tests yet) of an initial implementation of blocking stale requests from being refetched when an optimistic update is in progress. That'll need some tests if it works and maybe some tweaks before it's ready, but it may make this behaviour more intuitive.

kitten:
I’ll probably branch off the changes, where we block network refetches for data that has been altered by optimistic mutations, and will first get the optimistic layer change merged in

frederikhors:
I tried both with #4b63410.

It works now the adding and the `cache-and-network`. But as soon I try to add 2 new todos some problem again:

frederikhors:
Interesting what is happening with deletions here (https://codesandbox.io/s/urql-svelte-cacheinvalidate-many-items-j89ru):

kitten:
> But as soon I try to add 2 new todos some problem again

As far as I can tell, that looks like expected behaviour, your `updateQuery` is being run when the actual result comes back, but either updaters (whichever executes first) will see the optimistic todo of the other, so you need to account for that manually.

For us, there's not much we can do automatically for lists. If we ignore optimistic data in `updateQuery` then your new layer will lose the old optimistic references. So that's something that has to be handled manually; that goes back to what I was saying about optimistic updates. For lists they're inherently tricky, and for us there's only so much we can do to prevent duplication. At some point those heuristics are API-specific and hence on you 😅 

> Interesting what is happening with deletions here

I think that warning is simply at the wrong position; I added a quick fix for that.

kitten:
Collecting some future ideas and issues;

Seems like we need to add a `popDebugNode` in development for some places, which I simply forgot to add before.

Optimistic duplication can actually be addressed. But to do that we’d have to check how many optimistic mutations are in progress, then enforce that they’re _settled together_. That means if multiple optimistic updates have been applied, we wait until the last of their mutations have settled and clear the optimistic data & write the results as one batch.

frederikhors:
> I think that warning is simply at the wrong position; I added a quick fix for that.

@kitten the warning is fixed with https://github.com/FormidableLabs/urql/pull/745/commits/be80a04715d7023ac7928c548c98874906a06f64 but I was referring to the behavior: after the first slow mutation (a delete one) all optimistic todos (with a deleting mutation in progress) are removed and after each mutation finish with `invalidate` them are removed. But I loose optimistic green background.

It seems we are close.

Is this what are you talking about in your latest comment?

> Optimistic duplication can actually be addressed. But to do that we’d have to check how many optimistic mutations are in progress, then enforce that they’re settled together. That means if multiple optimistic updates have been applied, we wait until the last of their mutations have settled and clear the optimistic data & write the results as one batch.

kitten:
Yep, it seems that the blocking there doesn't quite work. But as that's just an experiment, I think there are some bugs related to that.

Because these improvements will take some time they'll definitely be separate from this PR, so I opened an RFC for them: https://github.com/FormidableLabs/urql/issues/747

And I've got another for a warning improvement: https://github.com/FormidableLabs/urql/issues/748

What I'm talking about in the last two paragraphs there is a solution to the problem where an update after a mutation result may commit other optimistic updates to a permanent, non-optimistic layer. That's all described in the RFC though, which also contains a spec for the refetch blocking.

kitten:
This has been released in [`@urql/exchange-graphcache@2.4.0`](https://github.com/FormidableLabs/urql/releases/tag/%40urql%2Fexchange-graphcache%402.4.0)

PR #755, opened 2020-04-26
(docs) - graphcache gotcha's

JoviDeCroock:
This documents the measures for:
- commutativity
- deferring refetches

Any other ones popping up for any of you?

### DIFF
--- a/docs/graphcache/gotchas.md
+++ b/docs/graphcache/gotchas.md
@@ -37,7 +37,7 @@ To prevent this, Graphcache will temporarily pause refetches that may overwrite 
 Once all mutations with optimistic updates complete however, all results will be applied at once,
 and refetched that may update the mutation data will be rerun.
 
-Let's look at an example, the has a list of an entity (for instance `Todos`), we click delete on four of them and we use `optimistic` to
+Let's look at an example, the cache has a list of an entity (for instance `Todos`), we click delete on four of them and we use `optimistic` to
 delete these before the network-request actually returns. This puts our UI in an inconsistent state since the list is `partial`
 so it would trigger a refetch, if these mutations would be slow and the fetch would be quick this would make our UI jump
 to prevent this, this fetch will be deferred until those four mutations complete.
--- a/docs/graphcache/gotchas.md
+++ b/docs/graphcache/gotchas.md
@@ -1,0 +1,37 @@
+---
+title: Gotchas
+order: 5
+---
+
+# Gotchas
+
+When using this normalised cache there are a few gotcha's we need to point out,
+this will make it easier to define best practices and to avoid pitfalls.
+
+Most of the measures we'll cover are brought to life to ensure we can build out
+to a fully offline normalised cache.
+
+## Commutativity
+
+By default we try to ensure idempotent and commutative results, the last part of this
+is the most relevant to be aware of for our users.
+
+Commutative internals mean that we'll keep the order of operations consistent, let's look
+at an example to clear this up.
+
+We are in our application and we queue up two queries, an `authorsQuery` and a `todosQuery`.
+
+When we dispatch `todosQuery` first and `authorsQuery` later but the `authorsQuery` returns a result
+before the `todosQuery` we'll put this result in a layer (this means it isn't actually in the real data)
+until the `todosQuery` arrives. When the `todosQuery` arrives we'll commit that first and then the `authorsQuery`
+to ensure our results are accurate. 
+
+## Optimistic results & refetches
+
+When we have mutations in-flight that have optimistic entries we defer refetches due to `cache-and-network` and
+`partial` this ensures we don't put our data in an inconsistent state.
+
+Let's look at an example, we have a list of an entity, we click delete on four of them and we use `optimistic` to
+delete these before the network actually returns. This puts our UI in an inconsistent state since the list is `partial`
+so it would trigger a refetch, if these mutations would be slow and the fetch would be quick this would make our UI jump
+to prevent this, this fetch will be deferred until those four mutations complete.
--- a/docs/graphcache/gotchas.md
+++ b/docs/graphcache/gotchas.md
@@ -37,7 +37,8 @@ To prevent this, Graphcache will temporarily pause refetches that may overwrite 
 Once all mutations with optimistic updates complete however, all results will be applied at once,
 and refetched that may update the mutation data will be rerun.
 
-Let's look at an example, the cache has a list of an entity (for instance `Todos`), we click delete on four of them and we use `optimistic` to
+Let's look at an example. Suppose our app has a list of entities (for instance `Todo`s).
+We've written an optimistic update so that when the user deletes an item from the list, we make it disappear immediately.
 Usually, if the list is using a query with `cache-and-network`, this means that the optimistic update would automatically trigger a refetch of the list, which would make the deleted item reappear, although it shouldn't.
 
 To prevent this, Graphcache waits for the mutation to complete instead, as it detects that the query overlaps with the optimistic update, and retriggers the refetch only when all mutations have completed.
--- a/docs/graphcache/gotchas.md
+++ b/docs/graphcache/gotchas.md
@@ -30,7 +30,7 @@ then the data from the `authorsQuery` to ensure the actual data is consistent.
 ## Optimistic results & refetches
 
 Optimistic updates can temporarily update the data after incoming mutations, which will trigger on screen queries to update.
-However, if we also use `cache-and-network` at the same time, some queries can refetch and overwrite our optimistic data,
+However, if we also use [`cache-and-network`](https://formidable.com/open-source/urql/docs/basics/queries/#request-policies) at the same time, some queries can refetch and overwrite our optimistic data,
 causing an unintended state where the intended optimistic update is destroyed.
 Such an unintended refetch can also happen if after an optimistic update a query is refetched when it’s not or
 only partially available in the cache.
--- a/docs/graphcache/gotchas.md
+++ b/docs/graphcache/gotchas.md
@@ -5,12 +5,10 @@ order: 5
 
 # Gotchas
 
-When using the normalised cache there are a few out of the box features
-we need to be aware of, being aware of these features will make it easier
-to define best practices and to avoid pitfalls.
+Graphcache has a couple of hidden features behind the scenes, that handle how it deals with certain edge cases.
+Being aware of these features and edge cases makes it easier to define best practices and to avoid pitfalls.
 
-Most of the measures we'll cover are brought to life to ensure we can build out
-to a fully offline normalised cache.
+Most of these hidden features and mechanisms in this section exist to ensure consistent behavior for Graphcache's (future) offline and persistence features.
 
 ## Commutativity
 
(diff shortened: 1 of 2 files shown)


kitten:
Love the document! I’d like to find a different name for it possibly and rephrase it to indicate the capabilities and mechanics that Graphcache provides automatically, if that makes sense? 😇

docs/graphcache/gotchas.md
```diff
@@ -0,0 +1,37 @@
...
+## Commutativity
+
+By default the cache ensures idempotent and commutative results, the last part of this
+is the most relevant one to be aware of.
+
+Commutative guarantees mean that we'll keep the order of operations consistent, let's look
+at an example to clear this up.
+We are in our application and we queue up two queries, an `authorsQuery` and a `todosQuery`.
+
+When we dispatch `todosQuery` first and `authorsQuery` later but the `authorsQuery` returns a result
```
andyrichardson, line 24:
```suggestion
We dispatch `todosQuery` first and `authorsQuery` later, but the `authorsQuery` returns a result
```

docs/graphcache/gotchas.md
```diff
@@ -0,0 +1,37 @@
...
+before the `todosQuery` we'll put this result in a layer (this means it isn't actually in the real data)
```
andyrichardson, line 25:
```suggestion
before the `todosQuery`. We'll put this result in a layer (this means it isn't actually in the real data)
```

docs/graphcache/gotchas.md
```diff
@@ -0,0 +1,37 @@
...
+until the `todosQuery` arrives. When the `todosQuery` arrives we'll commit that first and then the `authorsQuery`
```
andyrichardson, line 26:
```suggestion
until the `todosQuery` arrives. When the `todosQuery` does arrive, we commit it first and then the `authorsQuery` after,
```

Some minor structural changes which might make it easier to read / process

docs/graphcache/gotchas.md
```diff
@@ -0,0 +1,37 @@
...
+to ensure our results are accurate. 
+
+## Optimistic results & refetches
+
+When we have mutations in-flight that have optimistic entries we defer refetches due to `cache-and-network` and
```
andyrichardson, line 31:
```suggestion
When mutations with optimistic entries are in-flight, refetches are deferred due to `cache-and-network` and
```

Minor tweak to reduce the amount uses of "we"

docs/graphcache/gotchas.md
```diff
@@ -0,0 +1,37 @@
+---
+title: Gotchas
+order: 5
+---
+
+# Gotchas
```
andyrichardson, line 6: Spewing some suggestions for titles

 - Tips & tricks 
 - Under the hood
 - Additional info
 - Things to note
 - Special updates
 - Functional nuances

kitten: I think “Under the good” gets the gist right, as it indicates “hidden features” that work in the background, but sounds a little slangy 😄

docs/graphcache/gotchas.md
kitten: We can probably avoid this by phrasing this with an explanation first?

> Optimistic updates can temporarily update the data after incoming mutations, which will trigger on screen queries to update. However, if we also use `cache-and-network` at the same time, some queries can refetch and overwrite our optimistic data, causing an unintended state where the intended optimistic update is destroyed.
> Such an unintended refetch can also happen if after an optimistic update a query is refetched when it’s not or only partially available in the cache.
> 
> To prevent this, Graphcache will temporarily pause refetches that may overwrite optimistic updates. Once all mutations with optimistic updates complete however, all results will be applied at once, and refetched that may update the mutation data will be rerun.

docs/graphcache/gotchas.md
```diff
@@ -0,0 +1,37 @@
...
+
+Most of the measures we'll cover are brought to life to ensure we can build out
+to a fully offline normalised cache.
+
+## Commutativity
+
+By default the cache ensures idempotent and commutative results, the last part of this
+is the most relevant one to be aware of.
+
+Commutative guarantees mean that we'll keep the order of operations consistent, let's look
```
kitten, line 20: I think the use of “we” to mean “Graphcache” is a little confusing in this particular document.

Since the docs use “we” for instructions, we’ll likely have to resort to explicitly saying that “Graphcache” does something rather than using “we” for “us / the library” too

docs/graphcache/gotchas.md
```diff
@@ -17,21 +17,29 @@ to a fully offline normalised cache.
...
+layer (this means it isn't actually in the real data) until the `todosQuery` arrives.
+When the `todosQuery` does arrive. The cache will commit the data from the `todosQuery` first and
+then the data from the `authorsQuery` to ensure the actual data is consistent.
 
 ## Optimistic results & refetches
 
-When we have mutations in-flight that have optimistic entries we defer refetches due to `cache-and-network` and
-`partial` this ensures we don't put our data in an inconsistent state.
+Optimistic updates can temporarily update the data after incoming mutations, which will trigger on screen queries to update.
+However, if we also use `cache-and-network` at the same time, some queries can refetch and overwrite our optimistic data,
```
andyrichardson, line 33: This reads really well! 

```suggestion
However, if we also use [`cache-and-network`](https://formidable.com/open-source/urql/docs/basics/queries/#request-policies) at the same time, some queries can refetch and overwrite our optimistic data,
```

docs/graphcache/gotchas.md
```diff
@@ -0,0 +1,45 @@
...
+We dispatch the `todosQuery` first and the `authorsQuery` later, the `authorsQuery` returns a result
+before the `todosQuery`. The cache will put this result from the `authorsQuery` in a
+layer (this means it isn't actually in the real data) until the `todosQuery` arrives.
+When the `todosQuery` does arrive. The cache will commit the data from the `todosQuery` first and
+then the data from the `authorsQuery` to ensure the actual data is consistent.
+
+## Optimistic results & refetches
+
+Optimistic updates can temporarily update the data after incoming mutations, which will trigger on screen queries to update.
+However, if we also use [`cache-and-network`](https://formidable.com/open-source/urql/docs/basics/queries/#request-policies) at the same time, some queries can refetch and overwrite our optimistic data,
```
kitten, line 33: Careful here :) this should just be a relative link

docs/graphcache/gotchas.md
```diff
@@ -0,0 +1,45 @@
+---
+title: Gotchas
+order: 5
+---
+
+# Gotchas
+
+When using the normalised cache there are a few out of the box features
+we need to be aware of, being aware of these features will make it easier
+to define best practices and to avoid pitfalls.
```
kitten, line 10:
```suggestion
Graphcache has a couple of hidden features behind the scenes, that handle how it deals with certain edge cases.
Being aware of these features and edge cases makes it easier to define best practices and to avoid pitfalls.
```

docs/graphcache/gotchas.md
```diff
@@ -0,0 +1,45 @@
...
+
+Most of the measures we'll cover are brought to life to ensure we can build out
+to a fully offline normalised cache.
```
kitten, line 13:
```suggestion
Most of these hidden features and mechanisms in this section exist to ensure consistent behavior for Graphcache's (future) offline and persistence features.
```

docs/graphcache/gotchas.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Commutativity
+
+By default the cache ensures idempotent and commutative results, the last part of this
+is the most relevant one to be aware of.
```
kitten, line 18:
```suggestion
By default the cache ensures that results from the API are applied commutatively.
```

- removing `idempotent` as it's technically not reapplying results, so this doesn't apply
- "last part" is ambiguous but also not needed anymore
- fixes comma separating two distint sentences

docs/graphcache/gotchas.md
```diff
@@ -0,0 +1,45 @@
...
+
+Commutative guarantees means that the cache will keep the order of operations consistent, let's look
+at an example to clear this up.
+We are in our application and we queue up two queries, an `authorsQuery` and a `todosQuery`.
```
kitten, line 19:
```suggestion
Commutativity is a guarantee that means that Graphcache will attempt to keep the order of cache changes consistent, when results arrive out-of-order from the API, for instance due to network latency. When a query is dispatched first and another second, their results can still arrive in reverse, but Graphcache will still make sure that the cache results are applied in order, as if the results arrived in the right order.
Let's look at an example to understand this.
Suppose our application queues up two queries, an `authorsQuery` and a `todosQuery`.
```

docs/graphcache/gotchas.md
```diff
@@ -0,0 +1,45 @@
...
+
+We dispatch the `todosQuery` first and the `authorsQuery` later, the `authorsQuery` returns a result
+before the `todosQuery`. The cache will put this result from the `authorsQuery` in a
```
kitten, line 22:
```suggestion
We dispatch the `todosQuery` first and the `authorsQuery` later, but because of network latency the `authorsQuery`'s result arrived before the `todosQuery`'s result.
The cache will put the out-of-order result from the `authorsQuery` onto a layer.
```

docs/graphcache/gotchas.md
```diff
@@ -0,0 +1,45 @@
...
+layer (this means it isn't actually in the real data) until the `todosQuery` arrives.
+When the `todosQuery` does arrive. The cache will commit the data from the `todosQuery` first and
+then the data from the `authorsQuery` to ensure the actual data is consistent.
```
kitten, line 25:
```suggestion
This layer contains the out-of-order result's data, but hasn't been committed yet to the permanent in-memory data.
When the `todosQuery` does arrive and all out-of-order results have been stored on layers, the cache will start committing in order to ensure the actual data is consistent.
This layering approach means that Graphcache is able to temporarily reorder results as they arrive, ensuring that the results don't overwrite each other in a random order.
```

docs/graphcache/gotchas.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Optimistic results & refetches
+
+Optimistic updates can temporarily update the data after incoming mutations, which will trigger on screen queries to update.
```
kitten, line 29:
```suggestion
We've previously learned about [how to create "Optimistic Updates" in an earlier section.](./custom-updates.md#optimistic-updates)
Optimistic updates can temporarily update cached data after incoming mutations, which will trigger on screen queries to update.
```

docs/graphcache/gotchas.md
```diff
@@ -0,0 +1,45 @@
...
+However, if we also use [`cache-and-network`](https://formidable.com/open-source/urql/docs/basics/queries/#request-policies) at the same time, some queries can refetch and overwrite our optimistic data,
+causing an unintended state where the intended optimistic update is destroyed.
+Such an unintended refetch can also happen if after an optimistic update a query is refetched when it’s not or
+only partially available in the cache.
+
+To prevent this, Graphcache will temporarily pause refetches that may overwrite optimistic updates.
+Once all mutations with optimistic updates complete however, all results will be applied at once,
+and refetched that may update the mutation data will be rerun.
+
+Let's look at an example, the has a list of an entity (for instance `Todos`), we click delete on four of them and we use `optimistic` to
```
kitten:
```suggestion
Let's look at an example. Suppose our app has a list of entities (for instance `Todo`s).
We've written an optimistic update so that when the user deletes an item from the list, we make it disappear immediately.
```

docs/graphcache/gotchas.md
```diff
@@ -0,0 +1,45 @@
...
+delete these before the network-request actually returns. This puts our UI in an inconsistent state since the list is `partial`
+so it would trigger a refetch, if these mutations would be slow and the fetch would be quick this would make our UI jump
+to prevent this, this fetch will be deferred until those four mutations complete.
```
kitten, line 42:
```suggestion
Usually, if the list is using a query with `cache-and-network`, this means that the optimistic update would automatically trigger a refetch of the list, which would make the deleted item reappear, although it shouldn't.

To prevent this, Graphcache waits for the mutation to complete instead, as it detects that the query overlaps with the optimistic update, and retriggers the refetch only when all mutations have completed.
```

docs/graphcache/gotchas.md
kitten: Let's go with "Under the hood" for now if we can't come up with anything more apt yet?

PR #761, opened 2020-04-27
(docs) - Compress Urql logo

andyrichardson:
## About

Super minor tweak to reduce logo pop-in.

PNG resized to 300px wide, SVG would be way better if we have one.


kitten:
We can probably use the SVG, but failing that, was the PNG run through `pngquant` after resizing? That always saves a couple of kBs :)

andyrichardson:
Good find @wgolledge

PR #775, opened 2020-04-29
Fix debug event sources

### LINKED ISSUE #776: Name mangling breaks debug events
## About

The current build process mangles the `name` attributes on functions causing the `source` attribute of debug events to be broken.

andyrichardson:
## About

Amend name mangling which breaks debug source info

Fix #776

### DIFF
--- a/exchanges/graphcache/src/cacheExchange.ts
+++ b/exchanges/graphcache/src/cacheExchange.ts
@@ -441,3 +441,5 @@ export const cacheExchange = (opts?: CacheExchangeOpts): Exchange => ({
     ]);
   };
 };
+// @ts-ignore
+cacheExchange.name = 'cacheExchange';
--- a/scripts/rollup/plugins.js
+++ b/scripts/rollup/plugins.js
@@ -144,6 +144,7 @@ const terserMinified = terser({
   ie8: false,
   toplevel: true,
   compress: {
+    keep_fnames: true,
     keep_infinity: true,
     pure_getters: true,
     passes: 10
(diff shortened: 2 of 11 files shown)


JoviDeCroock:
Manging is done by `closure-compiler` and this won't fix it for `exchange-factories` like `graphCache`

andyrichardson:
> Manging is done by `closure-compiler` and this won't fix it for `exchange-factories` like `graphCache`

Okay no worries, I've manually added the function names in the meantime as a temporary fix

exchanges/graphcache/src/cacheExchange.ts
```diff
@@ -441,3 +441,5 @@ export const cacheExchange = (opts?: CacheExchangeOpts): Exchange => ({
     ]);
   };
 };
+// @ts-ignore
+cacheExchange.name = 'cacheExchange';
```
JoviDeCroock, line 445: This will crash in IE11 I think

andyrichardson: > This will crash in IE11 I think

Do we have a way of testing that?

scripts/rollup/plugins.js
```diff
@@ -144,6 +144,7 @@ const terserMinified = terser({
   ie8: false,
   toplevel: true,
   compress: {
+    keep_fnames: true,
```
kitten, line 147: This is only used for the production bundle, so this should be reverted 😅 also, Closure is responsible for most of the function name mangling (in both production or development) and this option is turned off on Terser in development

PR #778, opened 2020-04-29
Update fetchSuccess/Error payload to match raw GraphQL JSON results

### LINKED ISSUE #777: Incorrect data attribute on fetch debug events
## About

A `fetchResponse` event should return the successful JSON response from a successful fetch request. Currently it returns an _OperationResult_.

A `fetchError` event should return the JSON response from a failed fetch request. Currently it returns an _OperationResult_.

kitten:
Fix #777

### DIFF
--- a/exchanges/multipart-fetch/src/multipartFetchExchange.ts
+++ b/exchanges/multipart-fetch/src/multipartFetchExchange.ts
@@ -80,20 +80,33 @@ export const multipartFetchExchange: Exchange = ({
         makeFetchSource(operation, url, fetchOptions),
         takeUntil(teardown$),
         onPush(result => {
-          const error = !result.data ? result.error : undefined;
-
-          dispatchDebug({
-            type: error ? 'fetchError' : 'fetchSuccess',
-            message: `A ${
-              error ? 'failed' : 'successful'
-            } fetch response has been returned.`,
-            operation,
-            data: {
-              url,
-              fetchOptions,
-              value: error || result,
-            },
-          });
+          if (!result.data && result.error && result.error.networkError) {
+            dispatchDebug({
+              type: 'fetchError',
+              message: 'A failed fetch response has been returned.',
+              operation,
+              data: {
+                url,
+                fetchOptions,
+                value: result.error.networkError,
+              },
+            });
+          } else {
+            dispatchDebug({
+              type: 'fetchSuccess',
+              message: 'A successful fetch response has been returned.',
+              operation,
+              data: {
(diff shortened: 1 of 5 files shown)


exchanges/multipart-fetch/src/multipartFetchExchange.ts
```diff
@@ -80,20 +80,33 @@ export const multipartFetchExchange: Exchange = ({
...
+            });
+          } else {
+            dispatchDebug({
+              type: 'fetchSuccess',
+              message: 'A successful fetch response has been returned.',
+              operation,
+              data: {
+                url,
+                fetchOptions,
+                value: {
```
andyrichardson, line 102: Thanks for doing this @kitten

I'm not sure sending the debug message here is going to work as we don't get the raw http response (i.e. response.json()). If I'm reading this right, this ends up being much more like an OperationResponse which will be dispatched regardless.

The goal is to have network information (similar to the network tab) but on our timeline and colocated with the source operation.

andyrichardson: I think we're going to have to do this in `makeFetchSource` but the good news is that we won't need to manually add these dispatches for every exchange that makes use of the makeFetchSource function

andyrichardson: I'll make a branch off of this

kitten: Yes, that's on purpose. I'd very much like to keep the separation in-tact and keep `dispatchDebug` to exchange-related files, e.g. out of `@urql/core/internal`.
Passing around the function just seems like an unnecessary coupling of everything that is just vaguely related to the exchange to me.

The serialisation in this PR is only different to the raw JSON in two ways:

- `GraphQLErrors` are now normalized, since they've come back from the `CombinedError` (which in my book, is a good thing 🎉  People have mistakenly tried to set arbitrary properties on their error without using `extensions` a lot, so displaying their raw results in urql-related tools would cause this behaviour to potentially come back)
- keys that aren't `data`, `errors`, `extensions` won't be passed through, which is by the spec good

Also, moving `dispatchDebug` to `fetchSource` will cause issues with the `persistedFetchExchange` as we can't be granular enough with events anymore. A missing persisted fetch response would therefore become `fetchSuccess`, which just isn't right, and we'd have no good way to prevent `fetchSource` from issuing that event, although we know that we've just received a specific Persisted Error response.

PR #785, opened 2020-04-30
Fix example not working without a page reload on first open.

### LINKED ISSUE #782: First load on example repo doesn't work
## About

When loading [this example](https://github.com/FormidableLabs/urql/tree/master/packages/react-urql/examples/1-getting-started), the first load doesn't work.

## Notes
 - This looks to be related to a limitation on service workers intercepting calls immediately after installation.
 - Working round that limitation would be nice (if possible) or even just reloading the page after the service worker has been first installed.

amyboyd:
## Summary

Resolves #782.

## Set of changes

The crucial change is adding `self.clients.claim();`. Without this, overriding `fetch` in the service worker does not work until the page is reloaded.

### DIFF
--- a/packages/react-urql/examples/1-getting-started/src/service-worker.ts
+++ b/packages/react-urql/examples/1-getting-started/src/service-worker.ts
@@ -8,6 +8,10 @@ self.addEventListener('install', (event: any) => {
   event.waitUntil(self.skipWaiting());
 });
 
+self.addEventListener('activate', (event) => {
+  self.clients.claim();
+});
+
 self.addEventListener('fetch', (event: any) => {
   const request = event.request;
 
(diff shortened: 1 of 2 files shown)


packages/react-urql/examples/1-getting-started/src/service-worker.ts
```diff
@@ -8,6 +8,10 @@ self.addEventListener('install', (event: any) => {
   event.waitUntil(self.skipWaiting());
 });
 
+self.addEventListener('activate', (event) => {
```
andyrichardson, line 11: Was that it? 😱

andyrichardson:
@amyboyd  I just tried this out but it looks like the problem still persists. Did you manage to unregister the service worker before reloading?

Kapture 2020-04-30 at 14 21 31

PR #789, opened 2020-05-03
(suspense) - Fix suspenseExchange affecting subscriptions and add deprecation note

kitten:
**NOTE: This doesn't affect server-side / SSR support! It's only for the `suspenseExchange` of the `@urql/exchange-suspense` package**

This adds a fix that would affect `subscription` operations, although they remain untouched by suspense-mode.
It also adds a deprecation note, since client-side suspense in `urql` isn't as stable as it should be and overall not very useful.

This PR will remain opened for discussion briefly, if anyone has good arguments as to why this package _shouldn't_ be deprecated.

### DIFF
--- a/.changeset/red-brooms-drive.md
+++ b/.changeset/red-brooms-drive.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-suspense': minor
+---
+
+Add deprecation notice, as exchange isn't as useful as expected and unstable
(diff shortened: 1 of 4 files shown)


micabe:
Hi all, I confirm my problem disappeared with the sandbox build of suspense exchange.
Applause to @kitten !

PR #796, opened 2020-05-06
(feat) - @urql/exchange-utils

JoviDeCroock:
This kickstarts a package providing some common util-exchanges.


kitten:
@wgolledge That's true! the name is a little ambiguous for what we may add. Maybe `@urql/exchange-tools` would be nicer, i.e. `exchanges/tools/`?

Issue #754, opened 2020-04-25
Is it possible to select between getStaticProps or getServerSideProps instead of getInitialProps in next-urql?


fillipvt:
Hello!

The reason of the question is so that urql don't necessarily opt-out us of [automatic static optimization](https://nextjs.org/docs/advanced-features/automatic-static-optimization).

According to the docs:

> This determination is made by the absence of `getServerSideProps` and `getInitialProps` in the page.

So if it were possible to select `getStaticProps` we would have  the best of both world. 😊 But I'm just guessing for now!

kitten:
Not yet! But I see that this is an important feature we should add 👍

balazsorban44:
Looking forward to this! If it was in place, I would have jumped to try this out right away!

jakubriedl:
This [RFC in next.js](https://github.com/zeit/next.js/issues/11897) is related to this request and if that gets in it would make new options for this change.

PR #797, opened 2020-05-07
(next-urql) - getStaticProps & getServerProps

JoviDeCroock:
## Summary

Relates to: https://github.com/FormidableLabs/urql/issues/754

We need to implement a switch between `getInitialProps`, `getStaticProps` and `getServerProps` I've been browsing some examples but have not yet succeeded in finding a reasonable approach. This is a first step towards some common guidelines.

## Set of changes

- Optionally bind `getInitialProps` to `withUrql`

### DIFF
--- a/docs/advanced/server-side-rendering.md
+++ b/docs/advanced/server-side-rendering.md
@@ -238,7 +238,6 @@ export default withUrqlClient(ssrExchange => ({
 }))(Index);
 ```
 
-The `withUrqlClient` higher-order component function accepts a third parameter in case you are using
-[`getStaticProps`](https://nextjs.org/docs/basic-features/data-fetching#getstaticprops-static-generation),
-this third `boolean` parameter allows you to opt-out of binding `getInitialProps` to the component-tree.
-This means that only the `client` will be provided.
+The `withUrqlClient` higher-order component function accepts a second parameter called the `options`,
+these have one option for now, `ssr`, we can use this when the wrapped `page` does not implement
+`getInitialProps` but children down the tree do.
--- a/docs/advanced/server-side-rendering.md
+++ b/docs/advanced/server-side-rendering.md
@@ -238,6 +238,5 @@ export default withUrqlClient(ssrExchange => ({
 }))(Index);
 ```
 
-The `withUrqlClient` higher-order component function accepts a second parameter called the `options`,
-these have one option for now, `ssr`, we can use this when the wrapped `page` does not implement
-`getInitialProps` but children down the tree do.
+Unless the component that is being wrapped already has a `getInitialProps` method, `next-urql` won't add its own SSR logic, which automatically fetches queries during
+server-side rendering. This can be explicitly enabled by passing the `{ ssr: true }` option as a second argument to `withUrqlClient`.
--- a/docs/advanced/server-side-rendering.md
+++ b/docs/advanced/server-side-rendering.md
@@ -238,7 +238,7 @@ export default withUrqlClient(ssrExchange => ({
 }))(Index);
 ```
 
-The `withUrqlClient` higher-order component function accepts a thirth parameter in case you are using
+The `withUrqlClient` higher-order component function accepts a third parameter in case you are using
 [`getStaticProps`](https://nextjs.org/docs/basic-features/data-fetching#getstaticprops-static-generation),
-this thirth `boolean` parameter allows you to opt-out of binding `getInitialProps` to the component-tree.
+this third `boolean` parameter allows you to opt-out of binding `getInitialProps` to the component-tree.
 This means that only the `client` will be provided.
--- a/docs/advanced/server-side-rendering.md
+++ b/docs/advanced/server-side-rendering.md
@@ -237,3 +237,8 @@ export default withUrqlClient(ssrExchange => ({
   exchanges: [dedupExchange, cacheExchange, ssrExchange, fetchExchange],
 }))(Index);
 ```
+
+The `withUrqlClient` higher-order component function accepts a thirth parameter in case you are using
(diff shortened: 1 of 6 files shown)


packages/next-urql/src/with-urql-client.tsx
```diff
@@ -23,7 +23,8 @@ export function withUrqlClient(
     cacheExchange,
     ssrExchange,
     fetchExchange,
-  ]
+  ],
+  noSsr?: boolean
```
JoviDeCroock, line 27: The name isn't ideal yet but this will be used to basically indicate that the subtree is using `getInitialProps` but the wrapped page itself isn't using it.

packages/next-urql/src/with-urql-client.tsx
```diff
@@ -50,54 +51,56 @@ export function withUrqlClient(
...
-
-      // Run the prepass step on AppTree. This will run all urql queries on the server.
-      await ssrPrepass(<AppTree {...appTreeProps} />);
-
-      // Serialize the urqlClient to null on the client-side.
-      // This ensures we don't share client and server instances of the urqlClient.
-      (urqlClient as any).toJSON = () => {
-        return null;
-      };
+    if (AppOrPage.getInitialProps || !noSsr) {
```
JoviDeCroock, line 54: if we don't see a `getInitialProps` or we haven't disabled this subtree from `getInitialProps` we bind it to `withUrql` else the consumer of our library can export `getServersideProps` or `getStaticProps`

ryan-gilb: hey Jovi -- so as a consumer i would have to opt-in to automatic static optimization?

JoviDeCroock: Hey, so the problem is that before we had no check so I added an opt-in check for backwards compatability but the goal would be to just rely in them opting in to tree-traversal.

This opt-in step is meant for the user to declare they are using `getStaticProps` or `getServersideProps`

ryan-gilb: oh i see so it's for those on versions of Next without the automatic optimization to opt in to it?

JoviDeCroock: yes

JoviDeCroock: So the thing is that if we add `getInitialProps` in the tree and the user exports `getStaticProps` or `getServerSideProps` next will throw an error saying that these shouldn't be combined. So we allow for the user to tell us to skip out on us adding this to `withUrql`

ryan-gilb: oh that makes sense, thanks for clarifying :)

docs/advanced/server-side-rendering.md
```diff
@@ -237,3 +237,8 @@ const mergeExchanges = ssrExchange => [dedupExchange, cacheExchange, ssrExchange
 
 export default withUrqlClient({ url: 'http://localhost:3000/graphql' }, mergeExchanges)(Index);
 ```
+
+The `withUrqlClient` higher-order component function accepts a thirth parameter in case you are using
```
parkerziegler, line 241:
```suggestion
The `withUrqlClient` higher-order component function accepts a third parameter in case you are using
```

docs/advanced/server-side-rendering.md
```diff
@@ -237,3 +237,8 @@ const mergeExchanges = ssrExchange => [dedupExchange, cacheExchange, ssrExchange
...
+[`getStaticProps`](https://nextjs.org/docs/basic-features/data-fetching#getstaticprops-static-generation),
+this thirth `boolean` parameter allows you to opt-out of binding `getInitialProps` to the component-tree.
```
parkerziegler, line 243:
```suggestion
this third `boolean` parameter allows you to opt-out of binding `getInitialProps` to the component-tree.
```

packages/next-urql/src/with-urql-client.tsx
parkerziegler: Gotcha – what would you think about `preferStaticRendering` or `preferStatic`?

JoviDeCroock:
Next up: consolidate with https://github.com/FormidableLabs/urql/pull/816

packages/next-urql/src/with-urql-client.tsx
```diff
@@ -50,54 +51,58 @@ export function withUrqlClient(
...
-      await ssrPrepass(<AppTree {...appTreeProps} />);
-
-      // Serialize the urqlClient to null on the client-side.
-      // This ensures we don't share client and server instances of the urqlClient.
-      (urqlClient as any).toJSON = () => {
-        return null;
-      };
+    // TODO: this should probably become (AppOrPage.getInitialProps || ssr) in
+    // the next major bump.
+    if (AppOrPage.getInitialProps || !preferStatic) {
```
JoviDeCroock, line 56: I wonder if now that we're major bumping either way if we want to do the following:

```
if (AppOrPage.getInitialProps || ssr) {}
```

This would mean that only if the `_app` or the root `page` itself has `getInitialProps` OR if the user passes `ssr` as an option this will initiate. This means that the user would have to decide that if one of the page-children has `getInitialProps` they have to add the `ssr` option.

This seems linear with [next-apollo](https://www.npmjs.com/package/next-apollo)

parkerziegler: I'd be open to that! I like the idea that SSR is now opt-in with static rendering as the default. If you supply a page that defines `getInitialProps`, it'll opt you in to SSR, but if you just implement `getStaticProps` you'll be off to the races w/ SSG. I'm good with it!

sunpietro:
I'm still not sure how to use that feature so I can start using urql client in `getServerSideProps` method. Can you provide some example?

Issue #801, opened 2020-05-10
urql-next _app bundle size


jakubriedl:
While analyzing _app bundle size I noticed that some packages are shipped to the client even though they are never used there.

### react-ssr-prepass (5KB gzip)
React SSR prepass is used [only on server](https://github.com/FormidableLabs/urql/blob/master/packages/next-urql/src/with-urql-client.tsx#L87) (after `if (typeof window !== 'undefined') return ..`) so it doesn't need to be shipped to client and can safe 5KB of bundle on every page. One option how to fix it could be use of `const ssrPrepass = await import('react-ssr-prepass')` which is supported by since Node 9.7 but conflicts with current TS configuration (requires module flag to `es2020 | esnext | commonjs | amd | system | umd`) while current is `es2015`.

### js-sha256 (4KB gzip) (from persisted-fetch exchange)
Because of [import on top of sha256.ts](https://github.com/FormidableLabs/urql/blob/master/exchanges/persisted-fetch/src/sha256.ts#L1) client bundle with persisted fetch contains `js-sha256` which is used only on server side and doesn't need to be shipped to client. Possible fix would be similar as above using dynamic import and native node module.
[sha256.ts#L49](https://github.com/FormidableLabs/urql/blob/master/exchanges/persisted-fetch/src/sha256.ts#L49)
```js
    const { createHash } = await import('crypto')
    return createHash('sha256').update(query).digest('base64')
```

urql version & exchanges:
```
"urql": "^1.9.7",
"@urql/exchange-persisted-fetch": "^0.1.3",
"next-urql": "^0.3.7",
```

**Steps to reproduce**
Create production bundle using next and [analyze the bundles](https://flaviocopes.com/nextjs-analyze-app-bundle/). 

**Expected behavior**
Client bundles contain only code executed on client

**Actual behavior**
At least 2 significant packages are included in client bundle, possibly some others I haven't notice.
- react-ssr-prepass
- js-sha256

**Monkey patch**
As temporary fix it is possible to remove those packages from client bundle by using webpack [null-loader](https://webpack.js.org/loaders/null-loader/). To do so add following code to `next.config.js` webpack configuration.
```js 
    if (!isServer) {
      config.module.rules.push({
        test: /.*react-ssr-prepass.*|.*js-sha256.*/,
        use: 'null-loader',
      })
    }
```

JoviDeCroock:
Hey,

[...]

PR #802, opened 2020-05-10
(next-urql) - only include prepass in server-build

JoviDeCroock:
## Summary

Relates to https://github.com/FormidableLabs/urql/issues/801
Blocked by: https://github.com/FormidableLabs/react-ssr-prepass/pull/54

When [Next.js makes the client-side bundle](https://github.com/zeit/next.js/blob/canary/packages/next/build/webpack-config.ts#L810) it exposes a `process` variable to determine we are on the server or not.

The good thing about this is that we can use this to remove code for certain bundles, now this will transform to `ìf (false)` for the client-side bundle, when that happens `terser` can perform `dead code removal` and effectively remove this whole dep from the client-side bundle.

## Set of changes

- utilize `process.browser` to remove the `react-ssr-prepass` dependency on the client-side bundle

## Size impact

Before

```
Page                                                           Size     First Load
─ λ /                                                          37.3 kB      106 kB
```

After

```
Page                                                           Size     First Load
─ λ /                                                          29 kB       97.5 kB
```

### DIFF
--- a/.changeset/quiet-weeks-notice.md
+++ b/.changeset/quiet-weeks-notice.md
@@ -1,0 +1,5 @@
+---
+'next-urql': patch
+---
+
+Use `process.browser` to remove `react-ssr-prepass` from the client-build
--- a/.changeset/quiet-weeks-notice.md
+++ b/.changeset/quiet-weeks-notice.md
@@ -2,4 +2,4 @@
 'next-urql': patch
 ---
 
-Use `process.browser` to remove `react-ssr-prepass` from the client-build
+Use `typeof window !== 'undefiend'` to remove `react-ssr-prepass` from the client-build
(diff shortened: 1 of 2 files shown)


jakubriedl:
It seems that [`process.browser` is deprecated in favour of `typeof window`](https://github.com/zeit/next.js/pull/7651). So I guess the `else` branch was the fix.

JoviDeCroock:
@jakubriedl Yes but there is a fundamental difference between the two `typeof window !== 'undefined'` is a runtime concept, while `process.browser` is a build-time concept. Utilizing `typeof window !== 'undefined'` won't remove the `react-ssr-prepass` dependency since this conditional won't be evaluated.

JoviDeCroock:
Using `typeof window !== 'undefined` with Next 9.2

```
Page                                                           Size     First Load
─ λ /                                                          37.4 kB      106 kB
```

Using `typeof window !== 'undefined` with Next latest

```
Page                                                           Size     First Load JS
┌ λ /                                                          34.5 kB        91.8 kB
```

JoviDeCroock:
This PR should not be needed as long as we merge in https://github.com/FormidableLabs/react-ssr-prepass/pull/54 and bump the version.

Issue #799, opened 2020-05-10
RFC: Persisted Fetch - support GET for hashed queries 


jakubriedl:
## Summary
New Persisted fetch exchange is awesome, but it is missing support for GET for hashed queries that prevents us from using it.

In our use-case we want to use standard POST queries when query is not persisted (to avoid super long urls which cause issues on some devices) but GET for persisted (hashed) queries so it can be cached on CDN. Apollo Client supports this behaviour with [useGETForHashedQueries option](https://www.apollographql.com/docs/apollo-server/performance/apq/#using-get-requests-with-apq-on-a-cdn)

## Proposed Solution
Implement `useGETForHashedQueries` option similar to Apollo.

JoviDeCroock:
Hey,

Since we are using our [builder](https://github.com/FormidableLabs/urql/blob/master/exchanges/persisted-fetch/src/persistedFetchExchange.ts#L117) which already supports using [`GET`](https://github.com/FormidableLabs/urql/blob/ec9b8cf77bfb113c54fac11167e6e70316685a84/packages/core/src/internal/fetchOptions.ts#L38).

You can enable this with the [`client.preferGETMethod`](https://formidable.com/open-source/urql/docs/api/core/#client) option

jakubriedl:
I tried that but unfortunately it changes also the normal non-persisted queries to GET. That leads to huge URLs that are causing troubles on some devices. The `useGETForHashedQueries` behaviour works the way that only persisted queries are GET.

JoviDeCroock:
To exclusively use `GET` for hashed queries we'd need to transform that `exchange` to be a Higher Order Function I'd guess. That would set [`operation.context.preferGETMethod`](https://github.com/FormidableLabs/urql/blob/ec9b8cf77bfb113c54fac11167e6e70316685a84/packages/core/src/internal/fetchOptions.ts#L24) or something similar.

jakubriedl:
That is exactly what I did in my local fork and it is working fine (not configurable though).
https://github.com/FormidableLabs/urql/blob/master/exchanges/persisted-fetch/src/persistedFetchExchange.ts#L75

```
            return makePersistedFetchSource(
              {
                ...operation,
                context: { ...operation.context, preferGetMethod: true },
              },
              body,
              dispatchDebug,
            )
```

JoviDeCroock:
[...]

PR #804, opened 2020-05-10
(persisted-fetch) - support persisted get

JoviDeCroock:
## Summary

Implements #799 

## Set of changes

- Makes the `persistedFetchExchange` accept options

### DIFF
--- a/.changeset/big-worms-punch.md
+++ b/.changeset/big-worms-punch.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-persisted-fetch': major
+---
+
+make the persistedFetchExchange accept arguments, this to allow for GET for persisted-queries
--- a/.changeset/big-worms-punch.md
+++ b/.changeset/big-worms-punch.md
@@ -2,4 +2,24 @@
 '@urql/exchange-persisted-fetch': major
 ---
 
-make the persistedFetchExchange accept arguments, this to allow for GET for persisted-queries
+Make the persistedFetchExchange accept an optional argument, this argument consists of one option `preferGetForPersistedQueries`.
+
+To migrate you will have to change from
+
+```js
+import { persistedFetchExchange } from '@urql/exchange-persisted-fetch';
+
+createClient({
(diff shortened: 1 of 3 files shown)


.changeset/big-worms-punch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/exchange-persisted-fetch': major
+---
+
+make the persistedFetchExchange accept arguments, this to allow for GET for persisted-queries
```
kitten, line 5: We should probably explain the options here and add a small migration guide

exchanges/persisted-fetch/src/persistedFetchExchange.ts
```diff
@@ -72,6 +77,8 @@ export const persistedFetchExchange: Exchange = ({
               },
             };
 
+            operation.context.preferGetMethod = !!(options as PersistedFetchExchangeOptions)
```
kitten, line 80: This isn't reset later on, so all queries become GET after this if the option is turned on.
Instead we could add another argument to `makePersistedFetchSource` to "force GET" and then pass it only here, https://github.com/FormidableLabs/urql/blob/c98cb111e0e14a88ebde25475f188cb36b4a7e90/exchanges/persisted-fetch/src/persistedFetchExchange.ts#L120-L123

exchanges/persisted-fetch/src/persistedFetchExchange.ts
```diff
@@ -81,11 +92,22 @@ export const persistedFetchExchange: Exchange = ({
...
+              );
             } else if (result.error && isPersistedMiss(result.error)) {
               // Add query to the body but leave SHA256 hash intact
               body.query = query;
-              return makePersistedFetchSource(operation, body, dispatchDebug);
+              return makePersistedFetchSource(
+                operation,
+                body,
+                dispatchDebug,
+                !!(options as PersistedFetchExchangeOptions)
```
kitten, line 108: I believe this one should be set to `false` as well, since it sends the entire `query`

PR #808, opened 2020-05-11
(chore) - Fix up react-urql examples

amyboyd:
* Upgrade example dependencies to match latest URQL dependencies.
* Add README.md files where missing.
* Misc other tweaks to get the examples running.

Examples 1, 2 and 4 should now work first-time following the README instructions.

Example 3 still doesn't work in the browser. An error message shows: 'Error: Invalid hook call. Hooks can only be called inside of the body of a function component.'.


packages/react-urql/examples/1-getting-started/README.md
```diff
@@ -0,0 +1,34 @@
+## React & service worker example
```
andyrichardson, line 1: Minor suggestion - I would avoid adding any kind of info about the service worker if possible (I think it adds more complexity and the reason behind it is to save us having to maintain a GraphQL server)

packages/react-urql/examples/1-getting-started/README.md
```diff
@@ -0,0 +1,34 @@
...
+## Usage
+
+#### 1. Install dependencies in repo root
+
+To get started with the example, make sure to install depdencies for `urql` and `react-urql` directories. Running `yarn build` in the root will also go into each `packages/*` folder, install dependencies, and build. (The example's `package.json` will create a symlink to the built `react-urql`.)
+
+```bash
+# In root directory
+yarn
+yarn build
```
andyrichardson, line 16: This should have been done from the get go - thanks so much for sorting this out!

How would you feel about doing this in a postInstall script for each example? That way users could just do `yarn` and `yarn start` from the working directory.

packages/react-urql/examples/1-getting-started/package.json
```diff
@@ -15,11 +15,11 @@
   },
   "dependencies": {
     "@urql/devtools": "latest",
-    "graphql": "^14.6.0",
+    "graphql": "^15.0.0",
```
andyrichardson, line 18: Heads up, I had some issues with breaking changes for this release - if you've tested it on your end then it's probably fine!

packages/react-urql/examples/1-getting-started/package.json
```diff
@@ -15,11 +15,11 @@
...
     "graphql-tag": "^2.10.3",
     "idb": "^5.0.1",
     "react": "^16.13.0",
     "react-dom": "^16.13.0",
-    "urql": "latest"
+    "urql": "link:../.."
```
andyrichardson, line 23: You managed to get this working? Educate me 😅 

I always have issues with conflicting react versions and end up having to fight with links.

packages/react-urql/examples/2-using-subscriptions/src/app/index.tsx
```diff
@@ -12,14 +12,15 @@ import {
 } from 'urql';
 import './index.css';
 import { Messages } from './Messages';
+import { SERVER_PORT, WS_PORT } from '../config';
```
andyrichardson, line 15: +1 for env variables!

packages/react-urql/examples/2-using-subscriptions/src/config.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
```
andyrichardson, line 1: Would adding these to a .env file work out of the box with webpack?


#### .env
```
SERVER_PORT=4000
WS_PORT=4001
```

#### index.tsx
```ts
const subscriptionClient = new SubscriptionClient(
  `ws://localhost:${process.env.WS_PORT}`,
  {}
)
```

packages/react-urql/examples/1-getting-started/package.json
andyrichardson: On second thought, we have this [running in codesandbox](https://codesandbox.io/s/github/FormidableLabs/urql/tree/master/packages/react-urql/examples/1-getting-started) so this might break usage. In the past I've used a sketchy sed script to switch to "link" for development usage but I think there might be a better way

packages/react-urql/examples/4-core-node/package.json
```diff
@@ -6,6 +6,6 @@
   "dependencies": {
     "isomorphic-fetch": "2.2.1",
     "graphql-tag": "2.10.1",
-    "wonka": "3.2.1"
+    "wonka": "4.0.10"
```
andyrichardson, line 9: For the sake of removing maintenance requirements, how do folks feel about

```suggestion
    "wonka": "4"
```

or

```suggestion
    "wonka": "latest"
```

Also open to us having something like renovate update these examples every couple of weeks.

packages/react-urql/examples/2-using-subscriptions/src/config.js
amyboyd: Webpack doesn't pickup .env files automatically sadly. A package like https://github.com/motdotla/dotenv is needed, and the JS has to start with `require('dotenv').config()`

IMO not worth switching to a .env file for the example because it's another package (`motdotla/dotenv`) for people to learn when their real goal is to learn URQL.

packages/react-urql/examples/1-getting-started/README.md
amyboyd: Done.

packages/react-urql/examples/1-getting-started/README.md
amyboyd: Done.

packages/react-urql/examples/1-getting-started/package.json
amyboyd: I think my changes break it in codesandbox. Debugging...

packages/react-urql/examples/4-core-node/package.json
amyboyd: I guess it depends on @kitten's plans for handling breaking changes in `wonka`. Unless there's special circumstances, I think we should use `4.0.10` or `^4.0.10`.

kitten: I feel like no examples should use pinning or lockfiles. Given our limited dependencies a simple minor-range (so for wonka `^4.x.x`) is completely sufficient to make sure that the exampel continues to work.
It's more restrictive than `latest` (I mean, API may change after all), but will ensure that we don't have to constantly keep all examples up-to-date.

More specifically though, why is `wonka` even installed though in this example? Is it used outside of `urql`? If not it can be removed from the `package.json`

packages/react-urql/examples/1-getting-started/package.json
andyrichardson: No worries, my guess would be that it's this line - the codesandbox root directory is the example folder so relative urls (such as in this link) outside this folder will cause issues.

packages/react-urql/examples/2-using-subscriptions/src/config.js
andyrichardson: I usually don't use a package in order to source env files. Just checked and it looks like CRA and Parcel both support .env files without any additional plugins needed. 

I think this might be another reason to use Parcel/CRA and remove the noise that comes from having to manage a webpack config. If not though, it looks like there are [webpack plugins here](https://webpack.js.org/plugins/environment-plugin/#dotenvplugin).

packages/react-urql/examples/4-core-node/package.json
andyrichardson: I've never used this syntax before but looking at the npm docs, I believe `"4"` is equivalent to `"^4.x.x"` (i.e. no pinning).

> More specifically though, why is wonka even installed though in this example? Is it used outside of urql? If not it can be removed from the package.json

Wonka used to be an Urql peer dependency right? I think that is why.

amyboyd:
Building `packages/*` and `exchanges/*` in parallel seems to double the speed on my machine. There might be unintended consequences I haven't thought about -- someone who knows the build system inside-out might know?

On CodeSandbox, 1 works, 2 and 3 don't work (same as `master` branch):

https://codesandbox.io/s/github/FormidableLabs/urql/tree/example-tweaks/packages/react-urql/examples/1-getting-started

https://codesandbox.io/s/github/FormidableLabs/urql/tree/example-tweaks/packages/react-urql/examples/2-using-subscriptions

https://codesandbox.io/s/github/FormidableLabs/urql/tree/example-tweaks/packages/react-urql/examples/3-ssr-with-nextjs

packages/react-urql/examples/4-core-node/package.json
amyboyd: Good point about not having lock files in examples -- removed.

Re. why is `wonka` installed, `pipe` and `subscribe` are used here:

```
pipe(
  client.executeQuery(request),
    if (error) {
      console.warn('Error', error);
    }

    console.log('Data', JSON.stringify(data, ' ', 2));
  })
);
```

packages/react-urql/examples/1-getting-started/README.md
```diff
@@ -0,0 +1,26 @@
+## React
+
+This shows how to use URQL in the browser with React.
+
+A page displays a to-do list, and you can add items or mark them as complete.
+
+(Behind the scenes, we use a browser service worker to pretend to be a GraphQL server. This is done to avoid setting up a real GraphQL server for the example.)
+
+## Usage
+
```
andyrichardson, line 10: Something like this would be great for those not looking to pull the repo and build locally 

```suggestion
> Note: The quickest way to get started is to visit the live [codesandbox example](https://codesandbox.io/s/github/FormidableLabs/urql/master/packages/react-urql/examples/1-getting-started)
```

amyboyd: nice idea, added, thanks! :)

packages/react-urql/examples/4-core-node/package.json
kitten: That example above could also be simplified to use our shorthands (`client.query(...).toPromise()`), which is I guess what we'd recommend for Node.js usage anyway :)

packages/react-urql/examples/2-using-subscriptions/package.json
```diff
@@ -35,6 +35,6 @@
     "react": "^16.9.0",
     "react-dom": "^16.9.0",
     "subscriptions-transport-ws": "^0.9.16",
-    "urql": "link:../../"
+    "urql": "latest"
```
kitten, line 38: Another note, if none of our examples use `link:` anymore, their Webpack configs can be simplified and we won't have to build the entire repo anymore;
I think the important thing is that examples just work and secondly that they reflect the repo state :)

scripts/rollup/build.js
```diff
@@ -18,11 +18,11 @@ if (process.env.CIRCLE_NODE_TOTAL) {
...
 
-(async () => {
-  for (const package of packages) {
-    const packageName = path.relative(workspaceRoot, package);
-    console.log('> Building', packageName);
+const builds = packages.map(package => {
+  const packageName = path.relative(workspaceRoot, package);
+  console.log('> Building', packageName);
 
+  return new Promise(async function(resolve, reject) {
```
kitten, line 25: The `new Promise` wrapper here doesn't seem to be necessary

amyboyd: Changed!

PR #812, opened 2020-05-13
(graphcache) - Skip resolvers when writing to the cache

### LINKED ISSUE #811: Re-running resolvers for updated cache data
```
urql@1.9.7
@urql/exchange-graphcache@2.4.2
```

**Steps to reproduce**

1. Have a Graphcache configured with `resolvers` and `updates`:

```js
cacheExchange({
  resolvers: {
    Foo: {
      date: ({ date }) => parseISO(date), // Convert string `date` to Date object.
    },
  },
  updates: {
    createMemberWorkout: (result, args, cache) => {
      const {
        createFoo: { foo },
      } = result;
      cache.updateQuery(
        {
          query: FooListQuery,
        },
        (data) =>
          data && {
            ...data,
            foos: [foo, ...data.foos],
          },
      );
    },
  },
});
```

2. Execute `createFoo` mutation with at least one other foo already in Graphcache's cache.

**Expected behavior**

No errors are thrown, both the existing foo and the new foo end up in the cache.

**Actual behavior**

An error is thrown because `parseISO` in the resolver config tries to parse a date that is already a Date object. This happens because the `data` argument provided to `updateQuery` [has already been modified by the custom resolvers](https://github.com/FormidableLabs/urql/blob/53d775f39aa156bfcc2fc31de2e7be6d7d5662d6/exchanges/graphcache/src/operations/query.ts#L293-L304), and it's then written to the cache in the modified form.

As a workaround, I'm converting the date back to a string before writing to the cache in `updateQuery`, but this doesn't seem ideal.

If this is not the intended behavior, I'd like to submit a PR. Would it make sense to skip the custom resolvers when [reading the cache for `updateQuery`](https://github.com/FormidableLabs/urql/blob/caedd0fcb381038773947e7c682b6bfc90cba375/exchanges/graphcache/src/store/store.ts#L171)?

olistic:
## Summary

This PR is a quick draft to test https://github.com/FormidableLabs/urql/issues/811#issuecomment-628000289. Disabling the resolvers for `updateQuery`'s read operation didn't cause the existing tests to fail

**TODO:**

- [ ] Polish API.
- [ ] Write new tests for the edge case this is addressing.

Resolves #811.

## Set of changes

TBD

## Further comments

Please let me know if these changes are in the right direction. Happy to make any changes!


JoviDeCroock:
Hey @olistic 

Thanks for the PR! This is a good start but I would do it a bit differently, this at the moment introduces a performance dive since we are introducing a polymorphic environment. I also think this won't work with `readFragment`, ...

I think we could introduce something like `currentOperation` which shows what instantiated the current process we're in, when this is set we can decide to not use our `resolvers` when the `currentOperation` is `write` rather than `read`.

kitten:
To the above,
In all `write` operations we can set `currentOperation = 'write';`. There are two entrypoints to writing;

- https://github.com/FormidableLabs/urql/blob/53d775f39aa156bfcc2fc31de2e7be6d7d5662d6/exchanges/graphcache/src/operations/write.ts#L51-L56
- https://github.com/FormidableLabs/urql/blob/53d775f39aa156bfcc2fc31de2e7be6d7d5662d6/exchanges/graphcache/src/operations/write.ts#L93-L97

And in `read` operations we can set `currentOperation = 'read';`, which only has one entrypoint:
https://github.com/FormidableLabs/urql/blob/53d775f39aa156bfcc2fc31de2e7be6d7d5662d6/exchanges/graphcache/src/operations/query.ts#L60-L64

Then the resolvers can just be skipped if `currentOperation !== 'read'`: https://github.com/FormidableLabs/urql/blob/master/exchanges/graphcache/src/operations/query.ts#L288

This would be beneficial, because we'd be likely to reuse this information in multiple places, and it's easier to maintain as it doesn't add explicit new arguments and simply uses the fact that a `write` operation may cascade into a read operation. This may also give us some room in the future to improve our dependency-tracking, since we can filter some dependencies out.

olistic:
@JoviDeCroock @kitten Thanks for your feedback and pointers! I'll make the suggested changes soon.

olistic:
@JoviDeCroock @kitten Here goes an update:

1. Created a new `OperationType` enum that is either read or write
2. Added `currentOperation` to the global state in `store/data.ts`
3. Made `initDataState` receive the operation type as the first parameter
    - I wasn't sure about making it the first parameter or an optional one, but [this comment](https://github.com/FormidableLabs/urql/blob/86a8343634682080e904164f53e301d7f4a0ce4f/exchanges/graphcache/src/store/data.ts#L68) made me go with the former. If I misinterpreted that and the data state can be initialized for reasons other than a read/write operation, this will probably need to be changed. Please let me know if you would do this differently.
4. Updated the two entry points for writing and the entry point for reading to pass the corresponding operation type to `initDataState`
5. Updated the logic in `readSelection` to skip resolvers if the current operation is not a read

There are a couple of changes I needed to make in addition to the above:

6. Updated three calls to `initDataState` that were not done by operations entry points ([here](https://github.com/FormidableLabs/urql/blob/86a8343634682080e904164f53e301d7f4a0ce4f/exchanges/graphcache/src/store/data.ts#L138), [here](https://github.com/FormidableLabs/urql/blob/86a8343634682080e904164f53e301d7f4a0ce4f/exchanges/graphcache/src/store/data.ts#L159) and [here](https://github.com/FormidableLabs/urql/blob/86a8343634682080e904164f53e301d7f4a0ce4f/exchanges/graphcache/src/store/data.ts#L578))
    - Passed read as the operation type to preserve existing behavior, although I'm not sure this is what we want (the last call at least seems to be in the context of a write)

I haven't updated the tests yet as I wanted to first validate this approach.

Looking forward to hearing from you!

exchanges/graphcache/src/store/data.ts
```diff
@@ -156,10 +161,22 @@ export const noopDataState = (
...
   clearDataState();
 };
 
+export const getCurrentOperation = (): OperationType => {
+  invariant(
+    currentOperation !== null,
+    'Invalid Cache call: The cache may only be accessed or mutated during' +
+      'operations like write or query, or as part of its resolvers, updaters, ' +
+      'or optimistic configs.',
+    2
```
JoviDeCroock, line 174: We'll need to add an entry to `docs/graphcache/errors.md` for this

exchanges/graphcache/src/types.ts
```diff
@@ -183,3 +183,8 @@ export interface StorageAdapter {
 }
 
 export type Dependencies = Record<string, true>;
+
+export enum OperationType {
```
JoviDeCroock, line 187: Let's make this a const enum so it's not reflected in our outputted bundle

```suggestion
export const enum OperationType {
```

exchanges/graphcache/src/store/data.ts
```diff
@@ -135,14 +139,15 @@ export const clearDataState = () => {
   if (process.env.NODE_ENV !== 'test' && !data.defer) {
     data.defer = true;
     scheduleTask(() => {
-      initDataState(data, null);
+      initDataState(OperationType.Read, data, null);
```
JoviDeCroock, line 142: I think this should be a write since we are writing data to the external store and potentially deleting unconnected links with `gc`

exchanges/graphcache/src/types.ts
```diff
@@ -183,3 +183,8 @@ export interface StorageAdapter {
 }
 
 export type Dependencies = Record<string, true>;
+
+export const enum OperationType {
+  Read = 'READ',
+  Write = 'WRITE',
```
kitten, line 189: I’d simply use a union type of `”read | “write”` here; we can keep it lowercase to keep it simple, because it’ll just be used directly and skip the use of enums.

Realistically we don’t need to use enums which transpile to objects and just stick to simple strings

exchanges/graphcache/src/operations/query.ts
```diff
@@ -285,7 +287,11 @@ const readSelection = (
     if (resultValue !== undefined && node.selectionSet === undefined) {
       // The field is a scalar and can be retrieved directly from the result
       dataFieldValue = resultValue;
-    } else if (resolvers && typeof resolvers[fieldName] === 'function') {
+    } else if (
+      getCurrentOperation() !== OperationType.Read &&
```
kitten, line 291: Is this flipped?

exchanges/graphcache/src/store/data.ts
kitten: I think this is a duplication of an existing error so it’s fine 👍

exchanges/graphcache/src/types.ts
olistic: I like it!

exchanges/graphcache/src/store/data.ts
olistic: Updated, thanks! What about the other two calls to `initDataState` (in `noopDataState` and `hydrateData`)?

JoviDeCroock:
I was thinking about this the first time I bounced into this, the only reason I see for a breaking change is if users are relying on `readFragment` or `readQuery` to get the "resolved" version of their field. I'm open to a breaking change, we could group this with the work for `offline`.

olistic:
@kitten Thank you! 🙌 Unless you're in a hurry to get this in for a particular milestone, I'd prefer to finish the work I started myself :-) I updated the TODO list in the description and will be addressing it soon.

olistic:
Hey @kitten. I don't want to be blocking this either. Feel free to grab it, I will otherwise work on it if I have time this week and it hasn't been taken!  Sorry, my schedule has been a bit more tight than anticipated 😢

Issue #813, opened 2020-05-14
RFC: next-urql mergeExchangeOption should receive ctx


lionelB:
## Summary

In a context of a jwt auth based application, best practices enforce using a short live token (jwt-token) and a longer live token (refresh_token). 
A common practice si to store the refreshToken as a serverSide cookie. It's generally do in a `refreshToken()` method that could be happen clientSide or serverSide(on initial access) 
To having a authExchange that take care of refreshing the token, we need to access ctx so authExchange can pass `ctx` to refreshToken when it happen on serverside so cookie header can be set.
Since it's not the case,  we need to use fetchOption to bind context to the refreshToken method
```js
 return withUrqlClient(
    (ctx) => ({
      url: "/graphql",
      fetchOptions: {
        refreshToken: () => refreshToken(ctx),
      },
    }),
```

## Proposed Solution

It would be simpler to also have ctx from the mergeExchange
```
(ssrExchange, ctx) => [
      dedupExchange,
      cacheExchange,
      ssrExchange,
      () => authExchange(ctx),
      fetchExchange,
    ]
```
@kitten wrote an authExchange example https://gist.github.com/kitten/6050e4f447cb29724546dd2e0e68b470#file-authexchangewithteardown-js
and an auth example is here https://gist.github.com/lionelB/65c0472304b18f9417665b5bce85991e

kitten:
I’m starting to think that it may be interesting to attempt to change the API entirely, if possible.

The second `mergeExchanges` argument has lead to confusion and we could steer people in the right direction by simply making the entire HOC accept a single option, e.g.

```ja
withUrqlClient((ctx, ssrExchange) => ({
  /* ... options */
  url: ‘/api/GraphQL’,
  exchanges: [
    dedupExchange,
    cacheExchange,
    ssrExchange,
    fetchExchange
  ],
}))(Page);
```

PR #816, opened 2020-05-18
(next-urql) - streamline client creation

JoviDeCroock:
## Summary

This is a breaking change for `next-urql`, we were seeing a lot of confusion because of the `mergeExchanges` argument. This proposal streamlines it to the normal client.
 
However this does impose us to always specify a function for the creation of a client

Implements https://github.com/FormidableLabs/urql/issues/813

Still have to rewrite a lot of the testing.

TODO:
- [x] update examplees
- [x] update docs

### DIFF
--- a/packages/next-urql/src/__tests__/init-urql-client.spec.ts
+++ b/packages/next-urql/src/__tests__/init-urql-client.spec.ts
@@ -9,13 +9,4 @@ describe('initUrqlClient', () => {
     expect(urqlClient).toHaveProperty('url', 'http://localhost:3000');
     expect(urqlClient).toHaveProperty('suspense', true);
   });
-
-  it('should accept an optional mergeExchanges function to allow for exchange composition', () => {
-    const urqlClient = initUrqlClient({
-      url: 'http://localhost:3000',
-    });
-
-    expect(urqlClient).toHaveProperty('url', 'http://localhost:3000');
-    expect(urqlClient).toHaveProperty('suspense', true);
-  });
 });
--- a/packages/next-urql/src/__tests__/init-urql-client.spec.ts
+++ b/packages/next-urql/src/__tests__/init-urql-client.spec.ts
@@ -1,7 +1,7 @@
 import { initUrqlClient } from '../init-urql-client';
 
 describe('initUrqlClient', () => {
-  it('should return the urqlClient instance and ssrCache', () => {
+  it('should return the urqlClient instance', () => {
     const urqlClient = initUrqlClient({
       url: 'http://localhost:3000',
     });
--- a/packages/next-urql/src/__tests__/init-urql-client.spec.ts
+++ b/packages/next-urql/src/__tests__/init-urql-client.spec.ts
@@ -1,27 +1,21 @@
-import { ssrExchange } from 'urql';
-
 import { initUrqlClient } from '../init-urql-client';
 
 describe('initUrqlClient', () => {
   it('should return the urqlClient instance and ssrCache', () => {
-    const [urqlClient, ssrCache] = initUrqlClient({
+    const urqlClient = initUrqlClient({
       url: 'http://localhost:3000',
     });
 
     expect(urqlClient).toHaveProperty('url', 'http://localhost:3000');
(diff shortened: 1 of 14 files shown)


kitten:
How does this look to you, @parkerziegler and @ryan-gilb? 😇

packages/next-urql/src/init-urql-client.ts
```diff
@@ -1,42 +1,20 @@
...
     urqlClient = createClient({
       ...clientOptions,
       suspense: isServer || clientOptions.suspense,
-      exchanges: mergeExchanges(ssrCache),
     });
   }
 
   // Return both the Client instance and the ssrCache.
-  return [urqlClient, ssrCache];
+  return urqlClient;
```
ryan-gilb, line 19: what's the reasoning behind no longer returning the cache?

JoviDeCroock: Opted to handle this in the `withUrqlClient` this was needed to incorporate the `ssrExchange` in the original client options.

ryan-gilb: ahh gotcha, makes sense!

packages/next-urql/src/__tests__/init-urql-client.spec.ts
```diff
@@ -1,63 +1,21 @@
...
-  debugExchange,
-  dedupExchange,
-  cacheExchange,
-  fetchExchange,
-} from 'urql';
-
 import { initUrqlClient } from '../init-urql-client';
 
 describe('initUrqlClient', () => {
   it('should return the urqlClient instance and ssrCache', () => {
```
parkerziegler, line 4:
```suggestion
  it('should return the urqlClient instance', () => {
```

packages/next-urql/src/__tests__/init-urql-client.spec.ts
```diff
@@ -1,63 +1,21 @@
...
       url: 'http://localhost:3000',
     });
 
     expect(urqlClient).toHaveProperty('url', 'http://localhost:3000');
     expect(urqlClient).toHaveProperty('suspense', true);
-    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    expect(ssrCache!.toString()).toEqual(ssrExchange().toString());
   });
 
   it('should accept an optional mergeExchanges function to allow for exchange composition', () => {
```
parkerziegler, line 13: Looks like we can fully yeet away this test now that this is all being initialized in a single function 🎉

packages/next-urql/src/__tests__/with-urql-client.spec.tsx
```diff
@@ -79,39 +81,34 @@ describe('withUrqlClient', () => {
...
-        { url: 'http://localhost:3000' },
-        mockMergeExchanges
-      )(MockApp);
+      Component = withUrqlClient(() => ({
+        url: 'http://localhost:3000',
+        exchanges: [exchange] as any[],
+      }))(MockApp);
     });
 
     it('calls the user-supplied mergeExchanges function', () => {
```
parkerziegler, line 114: It looks like we could fully get rid of this particular test and alter the `it` description below it from:

```
'uses exchanges returned from mergeExchanges'
```

to:

```
'uses exchanges defined in the client config'
```

or something of that nature.

kitten:
I believe we need to update the docs anyway, so batching the two PRs may make sense 😅

packages/next-urql/src/with-urql-client.tsx
```diff
@@ -46,7 +57,17 @@ export function withUrqlClient(clientConfig: NextUrqlClientConfig) {
         : (appOrPageCtx as NextPageContext);
 
       const ssrCache = ssrExchange({ initialState: undefined });
-      const urqlClient = initUrqlClient(clientConfig(ssrCache, ctx));
+      const client = clientConfig(ssrCache, ctx);
```
kitten, line 60: this naming is a little confusing 😅  let's maybe keep calling this `clientConfig` and replace the function with `get*` or `make*`?

Issue #810, opened 2020-05-12
RFC: Graphcache warnings for configs


kitten:
## Summary

Currently a couple of warnings in Graphcache are just dedicated to checking the user's input when "schema awareness" is active. In other words, when we have a schema, we check whether some fields even exist on the schema, and if they don't the user will see a warning: https://github.com/FormidableLabs/urql/blob/ecb58bfdd97ed660a2dd248fd51e44c07b7350b5/exchanges/graphcache/src/ast/schemaPredicates.ts#L62-L113

Such a warning doesn't exist for fields in the config, like `opts.updates.Mutation.oopsTypo`; It'd be great to add those and take this opportunity to also _test_ whether our existing warnings still work correctly!

## Proposed Solution

- Add checks for all `config` fields that relate to the `schema`, if one has been passed
- Add tests for `query` and `write` operations to check whether warnings don't mis-trigger

## Requirements

- When `opts.schema` has been passed, use new schema predicates to warn whether `opts.updates`, `opts.resolvers`, and `opts.optimistic` contain fields that don't exist on the schema
- When `opts.schema` has been passed, use new schema predicates to warn whether `opts.keys` contains typenames that don't exist on the schema

PR #820, opened 2020-05-20
810: Graphcache warnings for `opts.keys`

amyboyd:
## Summary

As described in #810, when a schema is known, we want to validate `opts.keys`, `opts.updates`, `opts.resolvers`, and `opts.optimistic`.

When any of the validation fails, we want to warn the user (either via console or throwing an error).

## Set of changes

*This PR only deals with `opts.keys`*, not the others (updates/resolvers/optimistic), to get feedback.

The issue I'd like feedback on is: If the new validation calls `console.warn`, we preserve backwards compatibility because no error is thrown that would change the real functionality. If the new validation throws an error, that is backwards incompatible and we would need to do a major version bump.

IMO, throwing an error is probably better, but until we are ready to do a major version bump, I think we should use `console.warn`.

Thoughts?

I will finish the rest of #810 (opts.updates, opts.resolvers, and opts.optimistic) in another PR -- just want to get feedback on this b/compat issue first.

### DIFF
--- a/exchanges/graphcache/src/ast/schemaPredicates.test.ts
+++ b/exchanges/graphcache/src/ast/schemaPredicates.test.ts
@@ -1,6 +1,10 @@
 import { buildClientSchema } from 'graphql';
 import * as SchemaPredicates from './schemaPredicates';
 
+let warnSpy: jest.SpyInstance;
+beforeEach(() => (warnSpy = jest.spyOn(console, 'warn')));
+afterEach(() => warnSpy.mockRestore());
+
 describe('SchemaPredicates', () => {
   // eslint-disable-next-line
   const schema = buildClientSchema(require('../test-utils/simple_schema.json'));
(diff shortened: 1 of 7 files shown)


exchanges/graphcache/src/ast/schemaPredicates.ts
```diff
@@ -111,3 +112,20 @@ function expectAbstractType(
     5
   );
 }
+
+export function expectValidKeyingConfig(
+  schema: GraphQLSchema,
+  keys: KeyingConfig
+): void {
+  const types = Object.keys(schema.getTypeMap());
```
kitten, line 120: Awesome! Didn’t realise it’d be this simple!
All this is missing is a quick `if (process.env.NODE_ENV !== “production”)` check so this function compiles away.
Afaik, empty function bodies (since `warn` does compile away) may not be optimised away as the property reads themselves could have side-effects?

I’m not sure about the above, but we can simply check the production min bundle for this `Object.keys` signature.

Also, I think `warn` instead of `invariant` is perfect! We have quite a lot of warnings in Graphcache for unexpected conditions that _won’t_ stop it from working. A misnamed configuration item would be part of that imho. Especially since the schema can go out of sync with the config/queries in development.

exchanges/graphcache/src/ast/schemaPredicates.test.ts
```diff
@@ -1,6 +1,10 @@
 import { buildClientSchema } from 'graphql';
 import * as SchemaPredicates from './schemaPredicates';
 
+let warnSpy: jest.SpyInstance;
+beforeEach(() => (warnSpy = jest.spyOn(console, 'warn')));
```
kitten, line 5: Small note. Warnings were an afterthought in Graphcache unfortunately 😅 so in our tests we started testing for them similarly to this and prevented the console from being flooded.
So we already have a global spy: https://github.com/FormidableLabs/urql/blob/5295e33e6d17016912e12f70acd79ac4a51b3a95/scripts/jest/setup.js#L18

So it’s possible now to just run assertions against `console.warn` and `console.warn.mock` directly

exchanges/graphcache/src/ast/schemaPredicates.ts
amyboyd: The Webpack/Rollup optimisations are very cool! Have added that `if NODE_ENV` and the new code is now removed in `.min.js`:

image

exchanges/graphcache/src/ast/schemaPredicates.test.ts
amyboyd: Done - thanks for the pointer!

JoviDeCroock:
We forgot to add a `minor` changset for this

kitten:
Retroactively added here: https://github.com/FormidableLabs/urql/commit/6044b536dbd295a9e2cdf3b98efcca885348ec39

PR #826, opened 2020-05-22
(graphcache) Validation for Graphcache's opts.updates, opts.resolvers, opts.optimistic

### LINKED ISSUE #810: RFC: Graphcache warnings for configs

amyboyd:
## Summary

As described in #810, when a schema is known, we want to validate `opts.keys`, `opts.updates`, `opts.resolvers`, and `opts.optimistic`.

When any of the validation fails, we want to warn the user via `console.warn`.

Validation for `opts.keys` has already been added in PR https://github.com/FormidableLabs/urql/pull/820

## Set of changes

* `opts.updates.Mutation` and `opts.updates.Subscription` are now validated against the schema.
* `opts.resolvers` is now validated against the schema.
* `opts.optimistic` is now validated against the schema.

Resolves #810.

### DIFF
--- a/docs/graphcache/errors.md
+++ b/docs/graphcache/errors.md
@@ -289,3 +289,24 @@ This error occurs when an unknown type is found in `opts.keys`.
 
 Check whether your schema is up-to-date, or whether you're using an invalid
 typename in `opts.keys`, maybe due to a typo.
+
+## (21) Invalid mutation
+
+> Invalid mutation field `???` is not in the defined schema but the `updates` option is referencing it.
+
+When you're passing an introspected schema to the cache exchange, it is
+able to check whether your `opts.updates.Mutation` is valid.
+This error occurs when an unknown mutation field is found in `opts.updates.Mutation`.
+
+Check whether your schema is up-to-date, or whether you've got a typo in `opts.updates.Mutation`.
+
+## (22) Invalid subscription
+
+> Invalid subscription field: `???` is not in the defined schema but the `updates` option is referencing it.
+
+When you're passing an introspected schema to the cache exchange, it is
+able to check whether your `opts.updates.Subscription` is valid.
+This error occurs when an unknown subscription field is found in `opts.updates.Subscription`.
+
+Check whether your schema is up-to-date, or whether you're using an invalid
+subscription name in `opts.updates.Subscription`, maybe due to a typo.
--- a/docs/graphcache/errors.md
+++ b/docs/graphcache/errors.md
@@ -310,3 +310,14 @@ This error occurs when an unknown subscription field is found in `opts.updates.S
 
 Check whether your schema is up-to-date, or whether you're using an invalid
 subscription name in `opts.updates.Subscription`, maybe due to a typo.
+
+## (23) Invalid resolver
+
+> Invalid resolver: `???` is not in the defined schema, but the `resolvers`
+> option is referencing it.
+
+When you're passing an introspected schema to the cache exchange, it is
+able to check whether your `opts.resolvers` is valid.
+This error occurs when an unknown query, type or field is found in `opts.resolvers`.
+
+Check whether your schema is up-to-date, or whether you've got a typo in `opts.resolvers`.
--- a/docs/graphcache/errors.md
+++ b/docs/graphcache/errors.md
@@ -321,3 +321,14 @@ able to check whether your `opts.resolvers` is valid.
 This error occurs when an unknown query, type or field is found in `opts.resolvers`.
 
 Check whether your schema is up-to-date, or whether you've got a typo in `opts.resolvers`.
+
+## (24) Invalid optimistic mutation
+
+> Invalid optimistic mutation field: `???` is not a mutation field in the defined schema,
+> but the `optimistic` option is referencing it.
+
+When you're passing an introspected schema to the cache exchange, it is
+able to check whether your `opts.optimistic` is valid.
+This error occurs when a field in `opts.optimistic` is not in the schema's `Mutation` fields.
+
+Check whether your schema is up-to-date, or whether you've got a typo in `Mutation` or `opts.optimistic`.
--- a/exchanges/graphcache/src/store/store.ts
+++ b/exchanges/graphcache/src/store/store.ts
@@ -86,6 +86,13 @@ export class Store implements Cache {
         if (hasUpdates) {
           SchemaPredicates.expectValidUpdatesConfig(this.schema, this.updates);
         }
+
+        if (this.resolvers) {
+          SchemaPredicates.expectValidResolversConfig(
+            this.schema,
+            this.resolvers
+          );
+        }
       }
     }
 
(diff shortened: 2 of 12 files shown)


exchanges/graphcache/src/store/store.ts
```diff
@@ -78,6 +78,13 @@ export class Store implements Cache {
       if (this.keys) {
         SchemaPredicates.expectValidKeyingConfig(this.schema, this.keys);
       }
+
```
JoviDeCroock, line 81: This is looking awesome already, just a minor NIT I would wrap this whole validation part in `if (process.env.NODE_ENV !== 'production)` since these are intended as developer-time warnings rather than production ones

kitten:
On a side note before a review since this is WIP 😇 typically for WIP PRs we use GitHub’s new draft PRs and not mention necessarily in the title that they’re WIP. We sometimes still forget but PRs can be converted to draft PRs. Since those are blocking we also don’t need to add any additional notes to the description then 👍

For the title we don’t need to mention the ticket number, and can instead just add “Related to #810” or in this case “Resolves #810” so it auto closes on merge.

For the title we also often start them with “(lib)” so “(graphcache)” 😅

These aren’t conventions yet or written down as I mostly just edit things every now and then and we’ve been pretty consistent, but if it ever diverges more I’ll make sure to write this all down :)

amyboyd:
@kitten Thanks for info - converted to draft and updated title/description!

exchanges/graphcache/src/store/store.ts
amyboyd: Agreed - done!

docs/graphcache/errors.md
```diff
@@ -284,3 +284,27 @@ This error occurs when an unknown type is found in `opts.keys`.
 
 Check whether your schema is up-to-date, or whether you're using an invalid
 typename in `opts.keys`, maybe due to a typo.
+
+## (21) Invalid mutation <a id="21"></a>
+
+> Invalid mutation: `???` is not a mutation in the defined schema
```
kitten, line 290: Small nit: The wording here is colloquially correct ("mutation" referring to part of the `Mutation` type, etc) but it may be more specific to say "Mutation field" here and "is not a Mutation field" / "is not a field on the Mutation type"

amyboyd: Updated!

amyboyd:
PR ready for review (no longer WIP).

.changeset/beige-kiwis-begin.md
```diff
@@ -0,0 +1,6 @@
+---
+'@urql/exchange-graphcache': minor
+---
+
+Issue warnings when an unknown query, type or field has been included in Graphcache's opts.resolvers configuration to help spot typos.
+See: [#820](https://github.com/FormidableLabs/urql/pull/820)
```
kitten, line 6: I’d remove this since we’ll already have the other changeset pointing to this previous PR :)
The reference should be added automatically. There’s also two more duplicate changesets I believe

exchanges/graphcache/src/store/store.ts
```diff
@@ -75,8 +75,31 @@ export class Store implements Cache {
...
 
-      if (this.keys) {
-        SchemaPredicates.expectValidKeyingConfig(this.schema, this.keys);
+      if (process.env.NODE_ENV !== 'production') {
+        if (this.keys) {
+          SchemaPredicates.expectValidKeyingConfig(this.schema, this.keys);
+        }
+
+        const hasUpdates =
+          Object.keys(this.updates.Mutation).length > 0 ||
```
kitten, line 84: Nit: these will always be defined so the check here can go :)

exchanges/graphcache/src/store/store.ts
```diff
@@ -75,8 +75,31 @@ export class Store implements Cache {
...
+          Object.keys(this.updates.Subscription).length > 0;
+        if (hasUpdates) {
+          SchemaPredicates.expectValidUpdatesConfig(this.schema, this.updates);
+        }
+
+        if (this.resolvers) {
```
kitten, line 90: Nit: this and the check below are redundant as well

scripts/jest/preset.js
```diff
@@ -3,6 +3,7 @@ module.exports = {
   setupFiles: [
     require.resolve('./setup.js')
   ],
+  clearMocks: true,
```
kitten, line 6: Oh that’s amazing! I didn’t know that existed 😍

exchanges/graphcache/src/store/store.ts
amyboyd: I'm a fan of overly-defensive checks 😊. Partly because it guards against future code changes, but also lines like `if (hasUpdates)` communicate intent and IMO help devs understand what the author was thinking.

exchanges/graphcache/src/store/store.ts
amyboyd: See https://github.com/FormidableLabs/urql/pull/826#discussion_r430765116

I'd rather keep if you/others are OK with that reasoning?

exchanges/graphcache/src/store/store.ts
kitten: I’m ok with that, hence it’s just a nit 👍 Personally, I believe the types convey the intention here as our configs remain non-partial and guaranteed, so if we forgot the defaults there’d still be the tests and the types on this class that’d stand in the way of that

amyboyd: Fair point. Unless anyone else has comments on this, I'll merge PR as it is?

kitten: I think it’s fine as is 👍 just need to get the changeset fixed up and it’s ready to be merged when you’re ready 🎢

.changeset/beige-kiwis-begin.md
amyboyd: Changesets merged into `.changeset/thick-avocados-enjoy.md`

PR #827, opened 2020-05-22
(babel) - esm-support

JoviDeCroock:
Add a transform that gets rid of all the `process.env.NODE_ENV !== 'production'` checks and transforms them into `typeof process !== 'undefined' && process.env.NODE_ENV !== 'production'`.

This is an issue we initially faced in `devtools` where we tried to introduce conditionals `process?.env?.NODE_ENV` which will result in the same esm-incompatability where the global isn't correctly referenced.

I've tested and this check is also correctly DCE'd by `terser` and `closure`.

[Try it out](https://try.terser.org/)

```
const x = () => {
  if (typeof process !== 'undefined' && 'production' !== 'production') {
    console.log('y')    
  }
}

x();
```

### DIFF
--- a/scripts/babel/transform-esm-compatability.js
+++ b/scripts/babel/transform-esm-compatability.js
@@ -1,0 +1,36 @@
+const checkForTypeCheck = (node) => {
+  if (node.type === "LogicalExpression") {
+    return checkForTypeCheck(node.left) || checkForTypeCheck(node.right);
+  } else if (node.left && node.left.operator === "typeof" && node.left.argument.name === "process") {
+    return true;
+  }
+  return false;
+};
+
+const plugin = ({ template, types: t }) => {
+  const typeCheckNode = template.expression.ast`typeof process !== 'undefined'`;
+
+  return {
+    visitor: {
+      BinaryExpression(path) {
+        const { node } = path;
+        const { left } = node;
+        if (left && left.property && left.property.name === "NODE_ENV") {
+          const logicalExpression = path.findParent((path) => path.isLogicalExpression());
+          if (!logicalExpression) {
+            // This is a normal singular if-statement
+            path.replaceWith(t.logicalExpression("&&", typeCheckNode, node));
+          } else {
+            // This is a logical expression, we need to find out whether or not we're already using the check.
+            const { left, right } = logicalExpression.node;
+            console.log(left, right);
+            if (checkForTypeCheck(left) || checkForTypeCheck(right)) return;
+            logicalExpression.replaceWith(t.logicalExpression("&&", typeCheckNode, logicalExpression.node));
+          }
+        }
+      }
+    }
+  };
+};
+
+export default plugin;
--- a/scripts/rollup/plugins.js
+++ b/scripts/rollup/plugins.js
@@ -15,6 +15,7 @@ import cleanup from './cleanup-plugin.js'
 import babelPluginTransformPipe from '../babel/transform-pipe';
 import babelPluginTransformInvariant from '../babel/transform-invariant-warning';
 import babelPluginTransformDebugTarget from '../babel/transform-debug-target';
+import babelPluginTransformEsmCompatability from '../babel/transform-esm-compatability';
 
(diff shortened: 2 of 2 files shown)


andyrichardson:
This is awesome @JoviDeCroock! Really glad to see we're adding in esmodule support.

One thing worth emphasizing is that we don't currently have a way of switching between prod / dev builds when using esmodules (not necessarily for optimization purposes but useful for  removing developer warnings and prompts).

As it stands right now, we default to dev for esmodule which IMHO makes sense. If there's some kind of esmodule standard for flagging the environment the user is working in, we could definitely support that in the future :+1:

Super basic example of how someone people might be doing this when deploying

```
// index.js
window._ES_PROD_ = false;

// On prod deployment script
sed -i 's/window._ES_PROD_ = false/window._ES_PROD_ = true/' index.js
// ...
```

lukejacksonn:
Great work @JoviDeCroock 🥳 and yes.. andy raises a good point. There currently isn't a there isn't any kind of esmodule standard for flagging the environment the user is working in unfortunately.

This issue has plagued a few other repos and prevented them becoming fully browser esm compatible like what is being proposed here.

Most of the research I have done into this issue is documented in this PR on graphql-js itself https://github.com/graphql/graphql-js/pull/2409 which was the last piece in this puzzle https://github.com/graphql/graphql-js/pull/2277.

Perhaps we should consider speccing out what a standard for this could look like!

scripts/babel/transform-esm-compatability.js
```diff
@@ -0,0 +1,38 @@
+const checkForTypeCheck = (node) => {
```
kitten, line 1: Small typo in the filename here: `compatability` => `compatibility`
Otherwise only the changeset is missing, but I suppose we can just `patch` bump everything

JoviDeCroock: I've been trying to test some stuff and it seems to work but we haven't got sandboxes for everything 😅 I think this code can't actually negatively impact any places judging from the astexplorer though. If anyone comes up with an edge case please throw it at me.

I'll add the changesets later

amyboyd:
This looks like a neat addition!

Does it also handle a `===` as well as `!==`? For example `if (process.env.NODE_ENV === 'production') { return; } doStuff();`

JoviDeCroock:
@amyboyd this isn't really part of the scope, commonly in bundlers the `__dev__` optimization is `process.env.NODE_ENV !== 'production'` this because when the `replacePlugin` of a bundler kicks in this becomes `'production' !== 'production'` which when your minification comes into play (`Terser`, ...) is removed because this code is unreachable all of a sudden.

This optimization is commonly used in `urql` to give developer warnings to the user, these developer warnings however are useless in production and bloat the bundle size of the app in question (since most of the times these are long strings).

- Removal of the debug-events code can be found [here](https://github.com/FormidableLabs/urql/blob/master/scripts/babel/transform-debug-target.js)
- Transformation of the `invariants` and `warnings` can be found [here](https://github.com/FormidableLabs/urql/blob/master/scripts/babel/transform-invariant-warning.js)

In libraries there probably won't be a case for `process.env.NODE_ENV === 'production'` since it doesn't really matter for `ESM` or serves no purpose of being cut out in development.

amyboyd:
OK, I think I follow. With lines like [144 here](https://github.com/FormidableLabs/urql/blob/f12b477ec89234b9b17ffef4bcbcc886b265b55c/exchanges/graphcache/src/ast/schemaPredicates.ts#L144), will the `if` need inverted? Same for line 197, 238?

JoviDeCroock:
It shouldn't be a hard requirement, I do it because specifying the env isn't as standard in for instance `svelte` (we've had a couple of issues about that) which means that those could trigger in a published application.

kitten:
The only case here that's uncovered is that it's not transpiling `process.env.NODE_ENV !== 'production'` to `typeof process === 'undefined || process.env.NODE_ENV !== 'production'`

PR #838, opened 2020-05-26
(core) - Expose client.subscription shortcut method

### LINKED ISSUE #837: Calling `client.query` with subscription is not correctly recognized
I am using `@urql/core` together with XState and I don't need any other abstraction lib. I would have to call `client.executeSubscription` directly in the client code to mitigate the issue which is a bit more verbose.

I would like to propose the addition of `client.subscription`. Alternatively, the subscription exchange could try to identify the actual subscription document(s) instead of relying on operation type.

urql version & exchanges: 1.11.0

**Steps to reproduce**

`client.query(SubscriptionDocument)`

**Expected behavior**

The subscription exchange should be used.

**Actual behavior**

The subscription goes over to fetch instead.

FredyC:
Resolves #837 

## Summary

Expose `client.subscription` to be correctly recognized by the exchange.

## Set of changes

An only small addition to `@urql/core` API.

### DIFF
--- a/.changeset/hot-dryers-drive.md
+++ b/.changeset/hot-dryers-drive.md
@@ -1,0 +1,5 @@
+---
+'@urql/core': minor
+---
+
+Expose `client.subscription` to be correctly recognized by exchange
--- a/.changeset/hot-dryers-drive.md
+++ b/.changeset/hot-dryers-drive.md
@@ -2,4 +2,4 @@
 '@urql/core': minor
 ---
 
-Expose `client.subscription` to be correctly recognized by exchange
+Expose a `client.subscription`shortcut method, similar to `client.query` and `client.mutation`
--- a/docs/advanced/subscriptions.md
+++ b/docs/advanced/subscriptions.md
@@ -48,10 +48,7 @@ package](https://github.com/apollographql/subscriptions-transport-ws).
 import { Client, defaultExchanges, subscriptionExchange } from 'urql';
 import { SubscriptionClient } from 'subscriptions-transport-ws';
 
-const subscriptionClient = new SubscriptionClient(
-  'wss://localhost/graphql',
-  { reconnect: true }
-);
+const subscriptionClient = new SubscriptionClient('wss://localhost/graphql', { reconnect: true });
 
 const client = new Client({
   url: '/graphql',
@@ -132,3 +129,9 @@ As we can see, the `result.data` is being updated and transformed by
 the `handleSubscription` function. This works over time, so as
 new messages come in, we will append them to the list of previous
 messages.
+
+## Directly with @urql/core
+
+To consume subscriptions without any level of abstraction you should
+use `client.subscription` method for the subscription exchange to
+correctly pick it up. Or you can use `client.executeSubscription(createRequest(query, variables))`.
--- a/docs/advanced/subscriptions.md
+++ b/docs/advanced/subscriptions.md
@@ -130,8 +130,28 @@ the `handleSubscription` function. This works over time, so as
 new messages come in, we will append them to the list of previous
 messages.
 
-## Directly with @urql/core
+## One-off Subscriptions
 
-To consume subscriptions without any level of abstraction you should
-use `client.subscription` method for the subscription exchange to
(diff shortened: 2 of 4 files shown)


FredyC:
@kitten I did not bother adding tests because `query/mutation` are not tested as well.

packages/core/src/client.ts
```diff
@@ -309,6 +309,18 @@ export class Client {
...
 
+  subscription<Data = any, Variables extends object = {}>(
+    query: DocumentNode | string,
+    variables?: Variables,
+    context?: Partial<OperationContext>
+  ): Source<OperationResult<Data>> {
+    if (!context || typeof context.suspense !== 'boolean') {
+      context = { ...context, suspense: false };
+    }
+
```
kitten, line 320:
```suggestion
```

This isn't necessary as `suspense` is not applicable to subscriptions anyway

docs/api/core.md
```diff
@@ -103,6 +103,12 @@ This is similar to [`client.query`](#clientquery), but dispatches mutations inst
 [Read more about how to use this API on the "Core Package"
 page.](../concepts/core-package.md#one-off-queries-and-mutations)
 
+### client.subscription
+
+This is similar to [`client.query`](#clientquery), but provides raw stream without `toPromise` method.
+
+[Read more about how to use setup subscriptions](../advanced/subscriptions.md)
```
kitten, line 110:
```suggestion
This is similar to [`client.query`](#clientquery), but does not provide a `toPromise()` helper method on the streams it returns.

[Read more about how to use this API on the "Subscriptions" page.](../advanced/subscriptions.md)
```

docs/advanced/subscriptions.md
```diff
@@ -132,3 +129,9 @@ As we can see, the `result.data` is being updated and transformed by
 the `handleSubscription` function. This works over time, so as
 new messages come in, we will append them to the list of previous
 messages.
+
+## Directly with @urql/core
+
+To consume subscriptions without any level of abstraction you should
+use `client.subscription` method for the subscription exchange to
+correctly pick it up. Or you can use `client.executeSubscription(createRequest(query, variables))`.
```
kitten, line 137:
```suggestion
## One-off Subscriptions

Whe you're using subscriptions directly without `urql`'s framework bindings, you can use the `Client`'s `subscription` method for one-off subscriptions. This method is similar to the ones for mutations and subscriptions [that we've seen before on the "Core Package" page.](../concepts/core-package.md#one-off-queries-and-mutations)

This method will always [returns a Wonka stream](../concepts/stream-patterns.md#the-wonka-library) and doesn't have a `.toPromise()` shortcut method, since promises won't return the multiple values that a subscription may deliver. Let's convert the above example to one without framework code, as we may use subscriptions in a Node.js environment.

```js
import { pipe, subscribe } from 'wonka';

const newMessages = `
  subscription MessageSub {
    newMessages {
      id
      from
      text
    }
  }
`;

const { unsubscribe } = pipe(
  client.subscription(MessageSub),
  subscribe(result => {
    console.log(result); // { data: ... }
  })
);
```

.changeset/hot-dryers-drive.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/core': minor
+---
+
+Expose `client.subscription` to be correctly recognized by exchange
```
kitten, line 5:
```suggestion
Expose a `client.subscription`shortcut method, similar to `client.query` and `client.mutation`
```

packages/core/src/client.ts
FredyC: I was thinking about that but wasn't sure as I don't have much experience with it yet.

FredyC:
@kitten Can you please publish this soon? I realized this is also problem with a web project when running a subscription from XState machine. I know I can workaround that easily, but I would rather have the `client.subscription` directly instead of replacing it manually later.

kitten:
@FredyC We’re currently batching up several changes for a larger release cycle. If you’d like to use this change right now you can use it from the cache review packages that CodeSandbox CI publishes to its CDN

PR #793, opened 2020-05-05
(graphcache) - Implement offlineExchange

kitten:
Resolve #683

PR #840, opened 2020-05-29
(graphcache) - Remove deprecated Grapcache APIs

kitten:
This change is queued for the next **major** release, since it contains major breaking changes. Hence it should be merged and released together with #793.

- Remove the deprecated `populateExchange` export
- Remove the deprecated `store.invalidateQuery` method

Bundlesize went from `6.3kB` to `6.19kB` minzipped (delta `-0.11kB`)


.changeset/tricky-avocados-enjoy.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/exchange-graphcache': major
+---
+
+The deprecated `cache.invalidateQuery()` method has been removed. Please migrate over to `cache.invalidate()` instead, which operates on individual fields instead of queries.
```
kitten, line 5: Note: this is a separate log because it’s an extremely important point for migration, so we should separate it out from the other changeset

PR #855, opened 2020-06-09
(next-urql) - reuse the ssrExchange when there's one present

JoviDeCroock:
## Summary

Restores former behavior where we reused the [`ssrExchange`](https://github.com/FormidableLabs/urql/blob/next-urql%400.3.8/packages/next-urql/src/init-urql-client.ts)

## Set of changes

- Reuse the ssr exchange

### DIFF
--- a/packages/next-urql/src/with-urql-client.tsx
+++ b/packages/next-urql/src/with-urql-client.tsx
@@ -22,6 +22,8 @@ function getDisplayName(Component: React.ComponentType<any>) {
   return Component.displayName || Component.name || 'Component';
 }
 
+let ssr;
+
 export function withUrqlClient(
   getClientConfig: NextUrqlClientConfig,
   options?: WithUrqlClientOptions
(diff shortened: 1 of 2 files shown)


packages/next-urql/src/with-urql-client.tsx
```diff
@@ -35,7 +37,7 @@ export function withUrqlClient(
           return urqlClient;
         }
 
-        const ssr = ssrExchange({ initialState: urqlState });
+        if (!ssr) ssr = ssrExchange({ initialState: urqlState });
```
kitten, line 40: Wouldn't this then also be reused on the server-side?

JoviDeCroock: Seems so yes, I might've worded this wrongly 

```
  const isServer = typeof window === 'undefined';
  if (isServer || !urqlClient) {
    ssrCache = ssrExchange({ initialState });

    urqlClient = createClient({
      ...clientOptions,
      exchanges: mergeExchanges(ssrCache),
    });
  }
```

wgolledge: Looks okay to me with the server check

PR #857, opened 2020-06-11
(next) - prevent client-serialization even without getInitialProps

JoviDeCroock:
## Summary

We shouldn't serialize the urql-client down to the client, we used to rely on this happening within `getInitialProps`, now that this isn't always present it'd be better to rely on `initUrqlClient` to handle this for us.

## Set of changes

- add `client.toJSON = () => null` to `initUrqlClient` to prevent serialization between server and client.

### DIFF
--- a/.changeset/curly-bags-tap.md
+++ b/.changeset/curly-bags-tap.md
@@ -1,0 +1,5 @@
+---
+'next-urql': patch
+---
+
+Prevent serialization even if we don't have getInitialProps
--- a/.changeset/curly-bags-tap.md
+++ b/.changeset/curly-bags-tap.md
@@ -2,4 +2,4 @@
 'next-urql': patch
 ---
 
-Prevent serialization even if we don't have getInitialProps
+Prevent serialization of the `Client` for `withUrqlClient` even if the target component doesn't have a `getInitialProps` method. Before this caused the client to not be initialised correctly on the client-side.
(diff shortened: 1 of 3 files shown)


packages/next-urql/src/init-urql-client.ts
```diff
@@ -13,6 +13,9 @@ export function initUrqlClient(clientOptions: ClientOptions): Client | null {
       ...clientOptions,
       suspense: isServer || clientOptions.suspense,
     });
+    // Serialize the urqlClient to null on the client-side.
+    // This ensures we don't share client and server instances of the urqlClient.
+    (urqlClient as any).toJSON = () => null;
```
kitten, line 18: This in general is a really odd hack 😆  But I hope it's working. Is this a critical hotfix?

.changeset/curly-bags-tap.md
```diff
@@ -0,0 +1,5 @@
+---
+'next-urql': patch
+---
+
+Prevent serialization even if we don't have getInitialProps
```
kitten, line 5:
```suggestion
Prevent serialization of the `Client` for `withUrqlClient` even if the target component doesn't have a `getInitialProps` method. Before this caused the client to not be initialised correctly on the client-side.
```

packages/next-urql/src/init-urql-client.ts
JoviDeCroock: Not critical, more anticipating someone to walk into this pitfall

PR #859, opened 2020-06-12
(graphcache) - Add root operation type traversal warning

### LINKED ISSUE #858: Unexpected data: null in useQuery result, possibly due to __typename: 'Subscription'
It looks like _something_ in graphcache fails when my nested field with type: 'Subscription' is queried. See sample app in gist, linked below.

urql version & exchanges:
```
"urql": "1.9.8",
"@urql/exchange-graphcache": "3.0.0",
```

**Steps to reproduce**
See [sample app in gist](https://gist.github.com/stevehollaar/d1bb8be85e1b40bcb809a43292be35ca)

**Expected behavior**
`data` returned by `useQuery` should be populated by query result

**Actual behavior**
`data` is `null`

Prior conversation about this issue: https://spectrum.chat/urql/help/infinite-loop-of-query-re-fetching~53171239-62b5-4fc0-9a6e-6f0c6d5a7464

kitten:
Resolve #858

## Summary

Add warning for queries that traverse an Operation Root Type (Mutation / Subscription types occuring in a query result)

It's possible for someone to _explicitly_ use `Mutation` or `Subscription` inside `Query`, which by convention is discouraged and shouldn't be done. The worst that we do in such a case is now issue a warning, but I believe it's more important to prevent this mistake than to ignore it silentily, especially since it's extremely uncommon.

## Set of changes

- Add check to `readSelection` for non-Query root types and issue warning
- Add explanation of the warning to `errors.md`
- Ensure that the warning logs out the relevant typenames

### DIFF
--- a/docs/graphcache/errors.md
+++ b/docs/graphcache/errors.md
@@ -332,3 +332,36 @@ able to check whether your `opts.optimistic` is valid.
 This error occurs when a field in `opts.optimistic` is not in the schema's `Mutation` fields.
 
 Check whether your schema is up-to-date, or whether you've got a typo in `Mutation` or `opts.optimistic`.
+
+## (25) Invalid root traversal
+
+> Invalid root traversel: A selection was being read on `???` which is an uncached root type.
+> The `Mutation` and `Subscription` types are special Operation Root Types and cannot be read back
+> from the cache.
+
+In GraphQL every schema has three [Operation Root
+Types](https://spec.graphql.org/June2018/#sec-Root-Operation-Types). The `Query` type is the only
+one that is cached in Graphcache's normalized cache, since it's the root of all normalized cache
+data, i.e. all data is linked and connects back to the `Query` type.
+
+The `Subscription` and `Mutation` types are special and uncached; they may link to entities that
+will be updated in the normalized cache data, but are themselves not cached, since they're never
+directly queried.
+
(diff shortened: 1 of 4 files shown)


docs/graphcache/errors.md
```diff
@@ -332,3 +332,36 @@ able to check whether your `opts.optimistic` is valid.
 This error occurs when a field in `opts.optimistic` is not in the schema's `Mutation` fields.
 
 Check whether your schema is up-to-date, or whether you've got a typo in `Mutation` or `opts.optimistic`.
+
+## (25) Invalid root traversal
+
+> Invalid root traversel: A selection was being read on `???` which is an uncached root type.
```
amyboyd, line 338: Small typo, should be `traversal`

exchanges/graphcache/src/operations/query.ts
```diff
@@ -237,6 +237,21 @@ const readSelection = (
   const isQuery = key === store.rootFields['query'];
 
   const entityKey = (result && store.keyOfEntity(result)) || key;
+  if (!isQuery && !!ctx.store.rootNames[entityKey]) {
+    warn(
+      'Invalid root traversel: A selection was being read on `' +
```
amyboyd, line 242: Small typo, should be `traversal`

docs/graphcache/errors.md
kitten: Nice catch!

PR #862, opened 2020-06-12
(svelte) - Expand Readables with factory and PromiseLike methods

### LINKED ISSUE #706: svelte: previous data undefined on variable update
Unlike with the react version of library, when updating variable components of query, the current data (before request returns) is no longer present but is `undefined`. So only relying on `fetching` is possible and thus we get a flash of content, having to branch off of fetching / not fetching alone.

I would like to do this pattern to only show top-level loading when there is no (previous) data.

```svelte
{#if $result.fetching && !$result.data}
    <Loading />
{:else}
<!-- show data UI -->
   <!-- some data table, etc -->    
  
  <!-- subsequent queries with new data can show loading state within the UI, not top level -->
  <Pagination>
    <button>Prev</button>
    <button>Next</button>
    {#if $result.fetching}
      <Loading/>
    {/if}
  </Pagination>
{/if}
```

urql version: `@urql/svelte@0.1.3`

**Expected behavior**
svelte version works like React

**Actual behavior**
data is reset to undefined each time variables change
 
React: (expected)
https://codesandbox.io/s/nervous-herschel-wqpz0?file=/src/Page.js

Svelte: (data undefined)
https://codesandbox.io/s/affectionate-euclid-25gxp?file=/page.svelte

I'm new to this library, but just tested this out with React and it works as I'm expecting. It is likely that hooks caching intermediate values doesn't map 1:1 with the svelte observable code.. or something.

### LINKED ISSUE #795: svelte: error when mutation is in function
Hi guys,

I'm having a hard time trying make a simple login mutation work, here's the code:

```
import {
    initClient,
    mutate
} from '@urql/svelte';

initClient({
    url: 'https://graphql.fauna.com/graphql',
});

let promise;
let username = '';
let password = '';

const login = () => {
    promise = mutate({
        query: `
            mutation (
                $username: String!,
                $password: String!
            ) {
                loginUser(input: {
                    username: $username,
                    password: $password
                })
            }
        `,
        variables: {
            username,
            password,
        }
    });
}
```

urql version & exchanges: v0.2.3

I don't understand why it just launches the mutation once and then it throws the error `Function called outside component initialization`, the only way to make it work is to do something like `$: result = mutation(...)` but it doesn't make sense, I don't want to run the mutation after each keystroke.

So please tell me, what I am doing wrong ?

Thank you !

### LINKED ISSUE #806: svelte: getClient() returns undefined
Hi guys,

there is a little bug that I wanted to report.

Here's my component:

```
<script>
    import {
        getClient,
        mutate,
    } from '@urql/svelte';
    import { push } from 'svelte-spa-router';

    import Input from 'src/components/Input';

    let username = '';
    let password = '';

    const login = async () => {
        const client = getClient();
        console.log(client);
    }
</script>

<form
    on:submit|preventDefault={() => { login() }}
    method="POST"
    class="
    space-y-4
">
    <Input
        on:change={(event) => { username = event.detail }}
        label="Username"
        value={username}
    />
    <Input
        on:change={(event) => { password = event.detail }}
        label="Password"
        value={password}
        type="password"
    />
    <Button
        type="submit"
        class="
        mt-10
    ">
        Submit
    </Button>
</form>
```

If the first thing I do after the page loads is to click on the submit button, getClient() will return undefined even though it has been before BUT if I just write a single letter in one of the inputs before clicking on the button it will return the actual client. Of course I can add an if but I don't think this a normal behavior haha.

Not sure what the problem is.

Anyway,

thanks for this great library 👍

### LINKED ISSUE #854: Expose toPromise to @urql/svelte
When using the Svelte package, to access a Promise version of a query, I need to ditch the usual "query" way:

```js
import { query } from '@urql/svelte';

const result = query({ query: ... });

result.subscribe(...)
```
to a more internal, prone-to-breaking-change way

```js
import { getClient } from '@urql/svelte';

const result = await getClient().query(...).toPromise();
```

Is it possible to expose `toPromise` to the svelte implementation of `query(query: ...)`?

Thanks!

kitten:
Resolve #706 
Resolve #854
Resolve #806
Resolve #795

## Summary

The previous store implementations for `query`, `mutate`, and `subscription` were not as capable as we'd hoped. Two problems surfaced mainly:

- The state of `query` and `subscription` could not be preserved. Passing new arguments meant that the old state was lost
- The APIs were confusing. Svelte has many different ways to consume data and we only provided one that worked and was limited

To solve this there are now several more ways of using these bindings. To start with a simple one `mutate` now returns a `MutationStore`, which can be subscribed to, can be called itself to start a mutation, and can be used as a `PromiseLike`

```js
// Standard Usage:
mutate({ query, variables })()

// Subscribable Usage:
$: result = mutate({ query, variables });

// Curried Usage
const executeMutation = mutate({ query, variables });
const onClick = () => executeMutation();

// Curried Usage with overrides
const executeMutation = mutate({ query });
const onClick = () => await executeMutation({ variables });
```

Similarly the `query` and `subscription` helpers support new use-cases. But they're also recursively callable to create new sub-stores!

```js
// Subscribable Usage (as before):
$: result = query({ query: TestQuery, variables });

// Subscribable Usage which preserves state over time:
const testQuery = query({ query: TestQuery });
// - this preserves the state even when the variables change!
$: result = testQuery({ variables });

// Promise-based callback usage:
const testQuery = query({ query: TestQuery });
const doQuery = async () => await testQuery;

// Promise-based usage updates the subscribables!
const testQuery = query({ query: TestQuery });
const doQuery = async () => await testQuery;
// - doQuery will also update this result
$: result = query({ query: TestQuery, variables });
```

### DIFF
--- a/.changeset/strange-donuts-listen.md
+++ b/.changeset/strange-donuts-listen.md
@@ -1,0 +1,39 @@
+---
+'@urql/svelte': minor
+---
+
+Refactor all operations to allow for more use-cases which preserve state and allow all modes of Svelte to be applied to urql.
+
+```
+// Standard Usage:
+mutate({ query, variables })()
+
+// Subscribable Usage:
+$: result = mutate({ query, variables });
+
+// Curried Usage
+const executeMutation = mutate({ query, variables });
+const onClick = () => executeMutation();
+
+// Curried Usage with overrides
+const executeMutation = mutate({ query });
+const onClick = () => await executeMutation({ variables });
+
+// Subscribable Usage (as before):
+$: result = query({ query: TestQuery, variables });
+
+// Subscribable Usage which preserves state over time:
+const testQuery = query({ query: TestQuery });
+// - this preserves the state even when the variables change!
+$: result = testQuery({ variables });
+
+// Promise-based callback usage:
+const testQuery = query({ query: TestQuery });
+const doQuery = async () => await testQuery;
+
+// Promise-based usage updates the subscribables!
+const testQuery = query({ query: TestQuery });
+const doQuery = async () => await testQuery;
+// - doQuery will also update this result
+$: result = query({ query: TestQuery, variables });
+```
--- a/packages/svelte-urql/example/src/App.svelte
+++ b/packages/svelte-urql/example/src/App.svelte
@@ -5,7 +5,7 @@
 
   let i = 0;
 
-  $: todos = query({
+  const todosQuery = query({
     query: `
       query {
         todos {
@@ -15,16 +15,19 @@
         }
       }
     `,
-    variables: { i },
   });
 
-  const increment = () => i = i + 1;
+  $: todos = todosQuery({ pause: true });
+
+  const execute = () => todosQuery().then();
 </script>
 
 {#if $todos.fetching}
   Loading...
 {:else if $todos.error}
   Oh no! {$todos.error.message}
+{:else if !$todos.data}
+  No data
 {:else}
   <ul>
     {#each $todos.data.todos as todo}
@@ -33,4 +36,4 @@
   </ul>
 {/if}
 
-<button on:click={increment}>Increment {i}</button>
+<button on:click={execute}>Execute</button>
--- a/packages/svelte-urql/src/operations/mutate.ts
+++ b/packages/svelte-urql/src/operations/mutate.ts
@@ -12,7 +12,8 @@ export interface MutationArguments<V> {
 }
 
 export interface MutationStore<T = any, V = object>
-  extends Readable<OperationResult<T>> {
+  extends Readable<OperationResult<T>>,
+    PromiseLike<OperationResult<T>> {
   (additionalArgs?: Partial<MutationArguments<V>>): Promise<OperationResult<T>>;
 }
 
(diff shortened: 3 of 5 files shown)


kitten:
@frederikhors You’re a Svelte user right? Not sure if I can inconvenience you, but it’d be amazing to have more eyes on this ✌️

frederikhors:
> @frederikhors You’re a Svelte user right? Not sure if I can inconvenience you, but it’d be amazing to have more eyes on this ✌️

Yes of course! No problem! Already on my ToDo list! Give me a few days, please...

Maybe @cdock1029 can help here, too. 😄

frederikhors:
@kitten, I could not wait and in addition to having reread all the 4 open issues, I also read in depth what the changes are.

I can't wait to try it in the CRUD example (on codesnadbox) that I always use to identify problems and in my projects.

Can you release?

kitten:
@frederikhors You can test it against the CodeSandbox [packages that CodeSandbox CI releases](https://ci.codesandbox.io/status/FormidableLabs/urql/pr/862/builds/32063). I'm currently just convincing myself that it works as expected, then I think we'll be ready to merge and push this out :tada:

Issue #864, opened 2020-06-14
With makeDefaultStorage and offlineExchange: error: "Cannot read property '0' of null"


frederikhors:
**Steps to reproduce**

1. go on this CRUD example here: https://codesandbox.io/s/urql-svelte-crud-with-indexeddb-cqg5i
1. click on "New todo"
1. save it with button
1. click on "Reload" button
1. it gives an error: `Cannot read property '0' of null`

**GIF:**

![issue](https://user-images.githubusercontent.com/41120635/84592914-3e57ab80-ae49-11ea-9511-dffe6f05913a.gif)

**Code:**

```js
import gql from "graphql-tag";
import { initClient, dedupExchange, fetchExchange } from "@urql/svelte";
import { offlineExchange } from "@urql/exchange-graphcache";
import { makeDefaultStorage } from "@urql/exchange-graphcache/default-storage";
import { TODOS_QUERY, TODO_FRAG } from "./gql";

const updates = {
  Mutation: {
    addTodo: (result, args, cache, info) => {
      const allFields = cache.inspectFields("Query");
      const todoQueries = allFields.filter(x => x.fieldName === "todos");
      todoQueries.forEach(x => {
        cache.updateQuery(
          { query: TODOS_QUERY, variables: x.arguments },
          data => {
            return {
              ...data,
              todos: [...data.todos, result.addTodo]
            };
          }
        );
      });
    }
  }
};

const optimistic = {
  addTodo: (variables, cache, info) => {
    return {
      ...variables,
      __typename: "Todo",
      id: "TEMP_ID_" + Date.now(),
      complete: false,
      createdAt: Date.now()
    };
  }
};

const storage = makeDefaultStorage({
  idbName: "graphcache-v3",
  maxAge: 7
});

const exchanges = [
  dedupExchange,
  offlineExchange({ storage, updates, optimistic }),
  fetchExchange
];

function OnInit() {
  initClient({ url: "https://h1pcl.sse.codesandbox.io", exchanges });
}

export default {
  OnInit
};
```

frederikhors:
I thought it was due to this: https://spectrum.chat/urql/help/is-there-a-way-to-invalidate-queries-with-variables-without-using-variables~5e4fc1b8-849e-46e9-a9ca-72698a8433d0?m=MTU4NTkyMDE5ODMwOA==.

But it's happening also with `cache.invalidate` for the `deleteTodo` resolver:

```js
deleteTodo: (result, args, cache, info) => {
[...]

PR #865, opened 2020-06-14
(graphcache) - should not store undefined values in storage

JoviDeCroock:
## Summary

Should not store undefined values since these aren't valid in GraphQL

Fixes: https://github.com/FormidableLabs/urql/issues/864
Sandbox: https://codesandbox.io/s/urql-svelte-crud-with-indexeddb-cqg5i?file=/package.json

## Set of changes

- exclude translation from undefined to null in storage

### DIFF
--- a/exchanges/graphcache/src/default-storage/index.ts
+++ b/exchanges/graphcache/src/default-storage/index.ts
@@ -54,12 +54,10 @@ export const makeDefaultStorage = (opts?: StorageOptions): DefaultStorage => {
   const serializeBatch = (): string => {
     let data = '';
     for (const key in batch) {
-      const value = batch[key];
-      data += `${stringifyVariables(key)}:${
-        value !== undefined ? stringifyVariables(value) : 'null'
-      },`;
+      if (batch[key] !== undefined) {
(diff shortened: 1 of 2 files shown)


frederikhors:
I tried with codesandbox package. It works!

Can you release it please?

exchanges/graphcache/src/default-storage/index.ts
```diff
@@ -54,12 +54,10 @@ export const makeDefaultStorage = (opts?: StorageOptions): DefaultStorage => {
   const serializeBatch = (): string => {
     let data = '';
     for (const key in batch) {
-      const value = batch[key];
-      data += `${stringifyVariables(key)}:${
-        value !== undefined ? stringifyVariables(value) : 'null'
-      },`;
+      if (batch[key] !== undefined) {
```
kitten, line 57: This isn’t quite right I believe, because we do need to ensure that this deletes / overrides old values.
So we may want to specially encode `undefined`

frederikhors:
@kitten, I could not wait and in addition to having reread all the 4 open issues, I also read in depth what the changes are.

I can't wait to try it in the CRUD example (on codesnadbox) that I always use to identify problems and in my projects.

Can you release?

PR #870, opened 2020-06-17
(persisted) - Add generateHash option to the persistedFetchExchange

lorenries:
## Summary

This PR adds support for a custom hash function in the persisted query exchange. This is useful for non-browser environments like react-native, where you might want to use a different crypto API ([like Expo's](https://docs.expo.io/versions/v37.0.0/sdk/crypto/), for instance).

Usage:

```ts
import sha256 from "my-library";

const client = createClient({
  url: 'http://localhost:1234/graphql',
  exchanges: [
    dedupExchange,
    cacheExchange,
    persistedFetchExchange({ generateHash: sha256 }),
    fetchExchange
  ],
});

```

## Set of changes

Adds an argument to the `persistedFetchExchange` factory. No breaking changes.

### DIFF
--- a/.changeset/green-lies-help.md
+++ b/.changeset/green-lies-help.md
@@ -2,4 +2,4 @@
 '@urql/exchange-persisted-fetch': minor
 ---
 
-Adds support for custom hash functions by adding a `generateHash` option in the `createPersistedFetch` factory.
+Adds support for custom hash functions by adding a `generateHash` option to the exchange.
--- a/.changeset/green-lies-help.md
+++ b/.changeset/green-lies-help.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-persisted-fetch': minor
+---
+
+Adds support for custom hash functions by adding a `generateHash` option in the `createPersistedFetch` factory.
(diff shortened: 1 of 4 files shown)


.changeset/green-lies-help.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/exchange-persisted-fetch': minor
+---
+
+Adds support for custom hash functions by adding a `generateHash` option in the `createPersistedFetch` factory.
```
kitten, line 5: Sorry, just a small nit here. It says “in the `createPersistedFetch` factory”, but it should probably just say something like “to the exchange”, right?

It’s a slightly important nit since the changesets are combined into the next CHANGELOG entries for each release, so they’ll need to be clear and descriptive

lorenries: Ah, good call 👍

kitten:
Thanks again! This has now been published as `1.1.0` https://github.com/FormidableLabs/urql/blob/main/exchanges/persisted-fetch/CHANGELOG.md#110

lorenries:
Cool! Thank you both ❤️

PR #896, opened 2020-07-06
Wrong link, I think

frederikhors:
Can you check if I'm wrong, please?


kitten:
That's intentional :) it's rather obvious that it's on the same page, but Basics has more information on the topic.

frederikhors:
Ops, @kitten. The actual link doesn't work!

kitten:
Oh yea, it's not a relative link! I'll fix that up in a jiffy

PR #897, opened 2020-07-09
(graphcache) - Add nodes support to relayPagination helper

mrflip:
## Summary

Some APIs allow you to requests pages of nodes, rather than individually-cursored items -- here is a good [writeup of why to do this](https://medium.com/javascript-in-plain-english/graphql-pagination-using-edges-vs-nodes-in-connections-f2ddb8edffa0). This patch handles pagination for either style of response. 

See example app and server at

* https://codesandbox.io/s/urql-relay-pagination-server-cuuc2?file=/example.graphql (server) and 
* https://codesandbox.io/s/urql-demo-client-t0d8o (app)

## Set of changes

Added a helper just like relayPagination, but which iterated over a field called nodes, a direct list element of the page. It's appreciably simpler, as is node-style pagination.

This is not a breaking change.

### DIFF
--- a/docs/graphcache/computed-queries.md
+++ b/docs/graphcache/computed-queries.md
@@ -197,7 +197,7 @@ your queries.
 Given you have a [relay-compatible schema](https://facebook.github.io/relay/graphql/connections.htm)
 on your backend we offer the possibility of endless data resolving.
 This means that when you fetch the next page in your data
-received in `useQuery` you'll see the previous pages as well. This is usefull for
+received in `useQuery` you'll see the previous pages as well. This is useful for
 endless scrolling.
 
 You can achieve this by importing `relayPagination` from `@urql/exchange-graphcache/extras`.
@@ -236,6 +236,11 @@ last: 1, before: c => node 89, startCursor: d
 
 With inwards merging the nodes will be in this order: `[1, 2, ..., 89, 99]`
 And with outwards merging: `[..., 89, 99, 1, 2, ...]`
+
+The helper happily supports schemata that return nodes rather than
+individually-cursored edges. For each paginated type, you must either
+always request nodes, or always request edges -- otherwise the lists
+cannot be stiched together.
 
 ### Reading on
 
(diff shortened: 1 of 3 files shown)


kitten:
Hiya, 👋  sorry for the late reply. I was investigating this field and it's not widely supported as it's not in the Relay Pagination spec, which is rather strict about how pagination should be implemented in a given schema. However, GraphQL Nexus has built-in support for generating paginated structures with the `nodes` fields so that's good enough for me.

I don't believe though that we need to copy the `relayPagination` helper 1:1 and just make minor modifications. Could you propose a change to `relayPagination` that instead introduces a boolean option or something similar for this please? That seems more fitting I believe 😃

frederikhors:
@kitten sorry for my little intrusion.

@mrflip is right. Although the `nodes` are not officially in the Relay Pagination spec they are increasingly used (it's a trend I see daily).

Since these days I am putting on a new project with cursor pagination it would not be bad to have it in the already excellent URQL.

And I agree with you, @kitten, on the flag (boolean) or something else in the same code.

mrflip:
I had thought about doing this with a flag of some sort, but the ways I could see to do it would either (a) have if statements in the code, or (b) have some extra abstraction to choose whether pagination would be done by nodes or edges. I'm worried that the code would become bulky in the one case and obtuse in the other.

However, I now think that the right thing to do is to accept either nodes or edges as they come in, and I've pushed a new version that does so. Given that a person would per-type choose to receive either nodes or edges, we don't really need conditionals; the absent fields just have no impact. This also means that if a person's application delegated part of its API to a strict implementation while other parts allowed nodes or edges, they could do so. I've added a caution in the docs to not sometimes request a type with nodes and other times request that type with edges -- or is it too obvious to say so?

mrflip:
Given that it's silly to request nodes and edges at the same time, I can split the tests to do one or the other if you prefer it.

kitten:
Hiya :wave: This LGTM, so we're ready to get this merged. Can you add a changeset please? Maybe with this message as a minor change:

>  Add support for a paginated `nodes` field (rather than the standard `edges`) to the `relayPagination` helper

We have a guide on how to do this here: https://github.com/FormidableLabs/urql/blob/main/CONTRIBUTING.md#how-do-i-document-a-change-for-the-changelog

PR #899, opened 2020-07-10
(exchanges) - implement requestPolicyExchange

JoviDeCroock:
## Summary

Implements an exchange that monitors all incoming operations and checks them vs an internal map, if this operation is `cache-first` or `cache-only` and has exceeded the TTL it will upgrade this operation to a `cache-and-network` policy. This helps us refresh stale data.

The user can specify a custom TTL in ms

## Set of changes

- add new exchange

### DIFF
--- a/exchanges/request-policy/src/requestPolicyExchange.ts
+++ b/exchanges/request-policy/src/requestPolicyExchange.ts
@@ -1,0 +1,44 @@
+import { Operation, Exchange } from '@urql/core';
+import { pipe, map } from 'wonka';
+
+const defaultTTL = 5 * 60 * 1000;
+
+export interface Options {
+  ttl: number;
+}
+
+export const requestPolicyExchange = (options): Exchange => ({ forward }) => {
+  const operations = new Map();
+
+  const TTL = (options || {}).ttl || defaultTTL;
+
+  const processIncomingOperation = (operation: Operation): Operation => {
+    if (
+      operation.context.requestPolicy !== 'cache-first' &&
+      operation.context.requestPolicy !== 'cache-only'
+    )
+      return operation;
+    if (!operations.has(operation.key)) {
+      operations.set(operation.key, new Date());
+      return operation;
+    } else {
+      const lastOccurrence = operations.get(operation.key);
+      const currentTime = new Date().getTime();
+      if (currentTime - lastOccurrence.getTime() > TTL) {
+        operations.set(operation.key, new Date());
+        return {
+          ...operation,
+          context: {
+            ...operation.context,
+            requestPolicy: 'cache-and-network',
+          },
+        };
+      }
+      return operation;
+    }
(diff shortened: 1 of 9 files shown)


exchanges/request-policy/src/requestPolicyExchange.ts
```diff
@@ -0,0 +1,44 @@
...
+  const processIncomingOperation = (operation: Operation): Operation => {
+    if (
+      operation.context.requestPolicy !== 'cache-first' &&
+      operation.context.requestPolicy !== 'cache-only'
+    )
+      return operation;
+    if (!operations.has(operation.key)) {
+      operations.set(operation.key, new Date());
+      return operation;
+    } else {
```
andyrichardson, line 24: Non blocking, we could just yeet this else out of the file

exchanges/request-policy/tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+  "extends": "../../tsconfig.json",
+  "include": ["src"],
+  "compilerOptions": {
+    "baseUrl": "./",
+    "paths": {
```
andyrichardson, line 6: I thought this would have been unnecessary - assuming it breaks if you don't add the mappings?

exchanges/request-policy/src/requestPolicyExchange.ts
```diff
@@ -0,0 +1,44 @@
+import { Operation, Exchange } from '@urql/core';
+import { pipe, map } from 'wonka';
+
+const defaultTTL = 5 * 60 * 1000;
+
+export interface Options {
+  ttl: number;
+}
+
+export const requestPolicyExchange = (options): Exchange => ({ forward }) => {
```
kitten, line 10: The type annotation for `Options` is missing here ✌️

exchanges/request-policy/tsconfig.json
kitten: Each package needs this file. I've found that global type checking only works with a `tsconfig` in the monorepo's root. Local type checking for a package requires the local config, but it can inherit from the root one. The paths needs to be defined in both places and need to be different for the typings output to be correct.

This is from memory, so the explanations of why this is this way may not be 100% accurate 😅

stevehollaar:
Thanks for doing this, I'd love to use this exchange 👍 
Can I ask for one change? I'd like to _only_ upgrade certain operations (in my case, I _don't_ want to upgrade `cache-only` queries, which I use for optimistic partial rendering)
Perhaps another option that we can pass to the exchange:
```
  upgradeOperation: (operation: Operation) => boolean
```
to be used like:
```
  upgradeOperation: operation => operation.context.requestPolicy === 'cache-first'
```
Happy to do this in user-land if it's just me that needs this, or make a separate PR after this one lands. Thanks again! 🙏

PR #903, opened 2020-07-15
(exchanges) - implement the focusExchange

JoviDeCroock:
This exchange is meant to avoid stale queries, everytime the window reenters focus the currently active queries will be refetched with a `cache-and-network` policy.

TODO:
- tests
- docs

### DIFF
--- a/docs/api/refocus-exchange.md
+++ b/docs/api/refocus-exchange.md
@@ -5,7 +5,7 @@ order: 10
 
 # Refocus exchange
 
-`@urql/exchange-refocus` is an exchange for the [`urql`](../../README.md) GraphQL client that tracks currently active operations and redispatches them when the
+`@urql/exchange-refocus` is an exchange for the `urql` that tracks currently active operations and redispatches them when the
 window regains focus
 
 ## Quick Start Guide
--- a/docs/api/refocus-exchange.md
+++ b/docs/api/refocus-exchange.md
@@ -1,4 +1,9 @@
-# @urql/exchange-refocus
+---
+title: '@urql/exchange-refocus'
+order: 10
+---
+
+# Refocus exchange
 
 `@urql/exchange-refocus` is an exchange for the [`urql`](../../README.md) GraphQL client that tracks currently active operations and redispatches them when the
 window regains focus
--- a/exchanges/refocus/src/refocusExchange.ts
+++ b/exchanges/refocus/src/refocusExchange.ts
@@ -1,0 +1,50 @@
+import { pipe, tap } from 'wonka';
+import { Exchange, Operation } from '@urql/core';
+
+export const refocusExchange = (): Exchange => {
+  return ({ client, forward }) => ops$ => {
+    const watchedOperations = new Map<number, Operation>();
+    const observedOperations = new Map<number, number>();
+    const keys: Array<number> = [];
+
+    window.addEventListener('focus', () => {
+      keys.forEach(key => {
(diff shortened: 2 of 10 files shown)


exchanges/refocus/src/refocusExchange.ts
```diff
@@ -0,0 +1,50 @@
...
+      if (op.operationName !== 'query' && op.operationName !== 'teardown')
+        return;
+
+      if (op.operationName === 'query' && !observedOperations.has(op.key)) {
+        observedOperations.set(op.key, 1);
+        watchedOperations.set(op.key, op);
+        keys.push(op.key);
+      } else if (op.operationName === 'query') {
+        const observedCount = observedOperations.get(op.key) as number;
+        observedOperations.set(op.key, observedCount + 1);
```
kitten, line 33: Shouldn't the count be irrelevant since `teardown` only kicks in when the observing count has already reached zero? Apart from that `reexecuteOperation` should be a nice second protection against this since it's a noop for unobserved queries

exchanges/refocus/src/refocusExchange.ts
```diff
@@ -0,0 +1,50 @@
...
+import { Exchange, Operation } from '@urql/core';
+
+export const refocusExchange = (): Exchange => {
+  return ({ client, forward }) => ops$ => {
+    const watchedOperations = new Map<number, Operation>();
+    const observedOperations = new Map<number, number>();
+    const keys: Array<number> = [];
+
+    window.addEventListener('focus', () => {
+      keys.forEach(key => {
```
kitten, line 11: If we're already using `forEach` here we could also call `watchedOperations.forEach`?

docs/api/refocus-exchange.md
```diff
@@ -0,0 +1,32 @@
+---
+title: '@urql/exchange-refocus'
+order: 10
+---
+
+# Refocus exchange
+
+`@urql/exchange-refocus` is an exchange for the [`urql`](../../README.md) GraphQL client that tracks currently active operations and redispatches them when the
```
kitten, line 8: This is an invalid link for the docs site 😅 

```suggestion
`@urql/exchange-refocus` is an exchange for the `urql` that tracks currently active operations and redispatches them when the
```

exchanges/refocus/src/refocusExchange.ts
```diff
@@ -0,0 +1,38 @@
...
+  return ({ client, forward }) => ops$ => {
+    const watchedOperations = new Map<number, Operation>();
+    const observedOperations = new Map<number, number>();
+
+    window.addEventListener('focus', () => {
+      watchedOperations.forEach(op => {
+        client.reexecuteOperation(
+          client.createRequestOperation(
+            'query',
+            { key: op.key, query: op.query, variables: op.variables },
```
kitten, line 14: Fun fact, I think `Operation` does extend `GraphQLRequest`, so you can pass the entire `op` here to save some bytes

exchanges/refocus/src/refocusExchange.ts
```diff
@@ -0,0 +1,38 @@
...
+            { requestPolicy: 'cache-and-network' }
+          )
+        );
+      });
+    });
+
+    const processIncomingOperation = (op: Operation) => {
+      if (op.operationName !== 'query' && op.operationName !== 'teardown')
+        return;
+
```
kitten, line 24:
```suggestion
```

Dead code?

kitten (approved):

PR #904, opened 2020-07-16
(react-hooks) prefer exported function instead of block-scoped variable

dotansimha:
## Summary

This PR changes the definition of React Hooks to use `export function` instead of `export const`. It's a simple change that allow developers later to interact better with the signature of the hooks.

In TS, these `const` are defined as block-scoped variables, and cannot be overwritten/redefined/overloaded by external TS libraries or consumers. 

You can see a [real-time example for that here](https://www.typescriptlang.org/play/?ssl=1&ssc=1&pln=12&pc=2#code/CYUwxgNghgTiAEBbA9sArhBByALiAzjlvAN4CwAUPNfAPS3wgAeADsjDvAGZoB2YOAJbJe8AO7sA1vko1Grdpx78hIpAE8AYnzAAKFrCiIAXPEIxBvAOYBKU+ctWA3JVk168th3hgRheMDIBLy44lJu1MxenL68-ojqAGqwpvqGJmY4FtY28AC8AHyZ2c6UAL6uFKCQsAgo6JjwuAREpBGeitw6qqIJ2vxpMEamvGiIAEYgMHbwoxNTLlQ0UZ2x8Ukp8IPDs2OT0-lFc-uLZUA)  , and it's [also reported as issue](https://github.com/microsoft/TypeScript/issues/39622) 

I'm working on a new plugin for https://github.com/dotansimha/graphql-code-generator , and I would like to extend `urql` behaviour with module augmentation - but at the moment it's not possible because of the way the hooks are defined.

## Set of changes

Changed `export const` to be `export function` in all React Hooks.

### DIFF
--- a/.changeset/six-shoes-kneel.md
+++ b/.changeset/six-shoes-kneel.md
@@ -1,0 +1,5 @@
+---
+'urql': patch
+---
+
+prefer exported function instead of block-scoped variable
--- a/packages/react-urql/src/hooks/useMutation.ts
+++ b/packages/react-urql/src/hooks/useMutation.ts
@@ -28,9 +28,9 @@ export type UseMutationResponse<T, V> = [
   ) => Promise<OperationResult<T>>
 ];
 
-export const useMutation = <T = any, V = object>(
+export function useMutation<T = any, V = object>(
   query: DocumentNode | string
-): UseMutationResponse<T, V> => {
+): UseMutationResponse<T, V> {
   const isMounted = useRef(true);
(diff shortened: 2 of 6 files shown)


dotansimha:
@JoviDeCroock `changeset` seems to work fine now, thank you for fixing it :) I updated the PR.

dotansimha:
Thanks @wgolledge @JoviDeCroock ! 
Any chance you plan to release a new version soon? :)

Issue #907, opened 2020-07-16
RFC: Reduce whitespace in queries by using `stripIgnoredCharacters`


dhinklexo:
## Summary

Currently when we make graphql requests using GET, the length of the request is much longer than anticipated because whitespace is included in the request itself. From my research, this appears to occur when we transform the AST from `graphql-tag` into a string via `graphql`. Aside from being a longer (and therefore slower) request, this also hits max length issues which are most apparent on GET requests.

## Proposed Solution

graphql `14.3.0` adds a utility function called `stripIgnoredCharacters` ( https://github.com/graphql/graphql-js/pull/1802 ) that we can call on the stringified request to remove whitespace characters that don't impact the query itself. We should make use of this utility function.

_I'll attach a sample request and the formatted query as a follow up_

dhinklexo:
Here's the repro of the underlying behavior this proposal aims to correct:

https://github.com/dhinklexo/urql-whitespace-issue-test-bench

PR #908, opened 2020-07-16
(core) - Strip whitespace from query for GET requests

JoviDeCroock:
## Summary

Strips redundant whitespaces when using `GET`

This solves the case where the user provides their own queries, ....

Fixes: https://github.com/FormidableLabs/urql/issues/907

## Set of changes

- Strips redundant whitespaces when using `GET`

### DIFF
--- a/packages/core/src/internal/fetchOptions.ts
+++ b/packages/core/src/internal/fetchOptions.ts
@@ -49,7 +49,9 @@ export const makeFetchURL = (
   }
 
   if (body.query) {
-    search.push('query=' + encodeURIComponent(body.query));
+    search.push(
+      'query=' + encodeURIComponent(body.query.replace(/[\s,]+/g, ' ').trim())
+    );
   }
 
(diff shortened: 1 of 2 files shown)


packages/core/src/internal/fetchOptions.ts
```diff
@@ -49,7 +49,9 @@ export const makeFetchURL = (
   }
 
   if (body.query) {
-    search.push('query=' + encodeURIComponent(body.query));
+    search.push(
+      'query=' + encodeURIComponent(body.query.replace(/[\s,]+/g, ' ').trim())
```
wgolledge, line 53: How come we replace with a space? Also is it risky to remove commas?

JoviDeCroock: The Graphql Query spec only needs the text to be seperated so we go for the smallest amount of space possible (commas aren't enforced)

PR #923, opened 2020-07-28
(next-urql) - Never suspend option

JoviDeCroock:
## Summary

This adds the `neverSuspend` option, this feature is intended to support the newly added `getStaticProps`, ... The problem area is that when someone defines one of these functions exported at the same level of a component we can't assume what props will be injected since our `prepass` doesn't have access to the functions.

Imagine a scenario where we fetch an entity based on a query parameter injected with `getServerProps`, during our prepass this prop will be `undefined` and we'll query the entity with `id: undefined`. After that Next.js will dispatch the `renderToString` and this will result in them calling `useQuery` again but with different variables since this prop will now be a real id. This will trigger urql to throw a `Suspense` boundary which currently isn't supported in React 16.

## Set of changes

- Add `neverSuspend` option


packages/next-urql/src/with-urql-client.tsx
```diff
@@ -29,9 +29,10 @@ export function withUrqlClient(
   options?: WithUrqlClientOptions
 ) {
   if (!options) options = {};
+
   return (AppOrPage: NextPage<any> | typeof NextApp) => {
-    const shouldBindGetInitialprops = Boolean(
-      AppOrPage.getInitialProps || options!.ssr
+    const shouldEnableSuspense = Boolean(
+      AppOrPage.getInitialProps || (options!.ssr && !options!.neverSuspend)
```
wgolledge, line 35: This might cause a situation where `AppOrPage.getInitialProps` truthy and will override the `neverSuspend`?

JoviDeCroock: Good spot, prettier goofed up brackets

docs/advanced/server-side-rendering.md
```diff
@@ -240,3 +240,8 @@ export default withUrqlClient(ssrExchange => ({
 
 Unless the component that is being wrapped already has a `getInitialProps` method, `next-urql` won't add its own SSR logic, which automatically fetches queries during
 server-side rendering. This can be explicitly enabled by passing the `{ ssr: true }` option as a second argument to `withUrqlClient`.
+
+When you are using `getStaticProps`, `getServerProps`, ... You should opt-out of `Suspense` by setting the `neverSuspend` option to `true` in
+your `withUrqlClient`.
```
parkerziegler, line 245:
```suggestion
When you are using `getStaticProps`, `getServerSideProps`, or `getStaticPaths`,  you should opt-out of `Suspense` by setting the `neverSuspend` option to `true` in your `withUrqlClient` configuration.
your `withUrqlClient`.
```

docs/advanced/server-side-rendering.md
```diff
@@ -240,3 +240,8 @@ export default withUrqlClient(ssrExchange => ({
...
+This because when we are doing a prepass of your component tree we can't know what these functions will inject in the props and this injection
```
parkerziegler, line 246:
```suggestion
During the prepass of your component tree `next-urql` can't know what these functions will inject in the props of your page. This injection
```

docs/advanced/server-side-rendering.md
```diff
@@ -240,3 +240,8 @@ export default withUrqlClient(ssrExchange => ({
...
+could change the `variables` used in your `useQuery`, ... Making these throw during the subsequent `toString` pass which isn't supported in React 16.
```
parkerziegler, line 247:
```suggestion
could change the `variables` used in your `useQuery`. This will lead to error being thrown during the subsequent `toString` pass, which isn't supported in React 16.
```

packages/next-urql/README.md
```diff
@@ -215,3 +215,8 @@ Using `withUrqlClient` on a page that has `getInitialProps` will opt that compon
 component has children using `getInitialProps` but that component itself is not and you want to opt in to this behavior you'll have to set the second argument of
 `withUrqlClient`, this means `withUrqlClient(() => clientOptiosn, { ssr:true })`.
 This measure is available so we can support `getStaticProps`, ...
+
+When you are using `getStaticProps`, `getServerProps`, ... You should opt-out of `Suspense` by setting the `neverSuspend` option to `true` in
```
parkerziegler, line 219: See comments above, just wanted to help make the docs a little bit clearer!

andyrichardson (approved):

OscarBarrett:
Looking forward to this being merged 🙏  is there anything holding it up?

kitten:
@OscarBarrett Rhe Next.js APIs for data fetching are a little troublesome for us. There are a lot of cases where automatic fetching isn't an option which is why we've explored this, but overall this doesn't really help to deliver a solid experience with urql, both from a DX standpoint and with SSR.

So we've been waiting a little to see whether this API change actually makes sense and whether there are alternatives (although it's not looking like it at the moment)

kitten (approved):

PR #934, opened 2020-08-06
(persisted-fetch) - Don't send persisted queries if the hash operation fails

lorenries:
## Summary

In insecure browser environments, the `window.crypto.subtle` api isn't available. The default hashing function in the persisted query exchange will resolve an empty string in this case, and send that empty string as the persisted query, which leads to a `PersistedQueryNotFound` exception. This is fine, because the exchange will follow up with non-persisted query, but we can improve the default behavior by not sending a persisted query at all if the hash fails.

## Set of changes

Updates the persisted query exchange to only send a persisted query if there's a non-falsey hash value. Let me know if there's a better way of implementing this with Wonka!

### DIFF
--- a/exchanges/persisted-fetch/src/persistedFetchExchange.ts
+++ b/exchanges/persisted-fetch/src/persistedFetchExchange.ts
@@ -71,22 +71,32 @@ export const persistedFetchExchange = (
           // Hash the given GraphQL query
           fromPromise(hashFn(query, operation.query)),
           mergeMap(sha256Hash => {
-            // Attach SHA256 hash and remove query from body
-            body.query = undefined;
-            body.extensions = {
-              persistedQuery: {
-                version: 1,
-                sha256Hash,
-              },
-            };
+            // if the hashing operation was successful, add the persisted query exception
+            if (sha256Hash) {
+              // Attach SHA256 hash and remove query from body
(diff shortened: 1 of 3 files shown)


exchanges/persisted-fetch/src/persistedFetchExchange.ts
```diff
@@ -71,22 +71,32 @@ export const persistedFetchExchange = (
...
+                dispatchDebug,
+                !!(options as PersistedFetchExchangeOptions)
+                  .preferGetForPersistedQueries
+              );
+            } else {
+              return makePersistedFetchSource(
+                operation,
+                body,
+                dispatchDebug,
+                false
```
kitten, line 97: We could move up the if-block and let this condition be `!!((options as PersistedFetchExchangeOptions)
                  .preferGetForPersistedQueries && sha256Hash)` 🤔  may not be worth it though

lorenries: Good idea, that's definitely cleaner!

kitten:
This has now been released 🙌 Sorry for the delay and thanks for the contribution! ❤️ https://github.com/FormidableLabs/urql/releases/tag/%40urql%2Fexchange-persisted-fetch%401.2.1

PR #937, opened 2020-08-14
(docs) - add example of sub-field inspection

JoviDeCroock:
Adds an example of inspecting a non-rootfield

### DIFF
--- a/docs/graphcache/custom-updates.md
+++ b/docs/graphcache/custom-updates.md
@@ -237,6 +237,31 @@ const cache = cacheExchange({
 
 Now when we come onto a list we'll know that this list needs to be refetched.
 
+### Inspecting a sub-field
+
+We've seeen how to inspect fields for a root-field in the above, but what if your query looks like this:
+
+```
+query {
+  todo(id: "x") {
+    id
+    authors {
+      id
+      name
+    }
+  }
+}
+```
+
+Now we'd need to traverse all the `todos` to find which we need, but there's another solution.
+Rather than `cache.inspectFields('Query') which would give us all queried`todo`fields with their arguments we'll ask for all`Todo` types for a given id.
+
+```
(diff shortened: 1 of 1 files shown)


docs/graphcache/custom-updates.md
```diff
@@ -237,6 +237,31 @@ const cache = cacheExchange({
...
+```
+
+Now we'd need to traverse all the `todos` to find which we need, but there's another solution.
+Rather than `cache.inspectFields('Query') which would give us all queried`todo`fields with their arguments we'll ask for all`Todo` types for a given id.
+
+```
+cache.inspectFields({ __typename: 'Todo', id: args.id })
+```
+
+Now wee'll get all fields for the given `todo` and can freely update the `authors`.
```
wgolledge, line 263:
```suggestion
Now we'll get all fields for the given `todo` and can freely update the `authors`.
```

docs/graphcache/custom-updates.md
```diff
@@ -237,6 +237,31 @@ const cache = cacheExchange({
...
+    authors {
+      id
+      name
+    }
+  }
+}
+```
+
+Now we'd need to traverse all the `todos` to find which we need, but there's another solution.
+Rather than `cache.inspectFields('Query') which would give us all queried`todo`fields with their arguments we'll ask for all`Todo` types for a given id.
```
wgolledge, line 257:
```suggestion
Rather than use `cache.inspectFields('Query')`, which would give us all queried `todo` fields with their arguments, we can instead provide an object as the argument to `inspectFields` asking for all `Todo` types for a given id.
```

docs/graphcache/custom-updates.md
```diff
@@ -237,6 +237,31 @@ const cache = cacheExchange({
...
+
+```
```
wgolledge, line 259:
```suggestion
```js
```

PR #940, opened 2020-08-18
Add urql-custom-scalars-exchange to showcase

clentfort:
## Summary

Add the urql-custom-scalars-exchange to the list of community packages

### DIFF
--- a/docs/showcase.md
+++ b/docs/showcase.md
@@ -46,3 +46,5 @@ maintainers, and supporters. This page is specifically dedicated to all of you!
   compute fields on-the-fly using the `@computed` directive.
 - [`graphql-code-generator`](https://graphql-code-generator.com/docs/plugins/typescript-urql): A plugin
   that helps you make typesafe hooks/components with urql.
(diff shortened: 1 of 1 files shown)


JoviDeCroock:
Just a question about this statement

> Should this exchange be listed before or after the cache exchange?
> I don't know. I have it listed before the grahp-cache exchange, but maybe that's stupid?

With `graphCache` this can also be solved with the `resolver` pattern but I guess you want to resolve on a type-basis rather than a field basis. I think it's safer to put before the `cacheExchange` else exchanges like graphCache won't reflect the server-side state which is its intention.

clentfort:
Yeah, I saw you can add custom resolvers but that is tiresome for things like dates. `urql-computed-exchange` plus custom directives could also solve this problem, but again that is tiresome with very common types.

Thanks for the insight regarding the order of exchanges, I'll update the README accordingly.

JoviDeCroock (approved):
Awesome work! Love the concept

PR #944, opened 2020-08-23
fix(preact): handle case where `prev.current` is null

jlengstorf:
after upgrading to `@urql/preact@1.2.0`, I started getting an error where `prev.current` is sometimes `null`, which throws `TypeError: Cannot read property 'key' of null`

I wasn't able to test this bix with your build step, but it should transpile properly

we could probably drop the `prev.current !== undefined` with this fix, but I didn't want to change more than I had to because I don't have a ton of context on this project

### DIFF
--- a/packages/preact-urql/src/hooks/useRequest.ts
+++ b/packages/preact-urql/src/hooks/useRequest.ts
@@ -12,7 +12,7 @@ export const useRequest = (
   return useMemo(() => {
     const request = createRequest(query, variables);
     // We manually ensure reference equality if the key hasn't changed
-    if (prev.current !== undefined && prev.current.key === request.key) {
+    if (prev.current !== undefined && prev.current?.key === request.key) {
       return prev.current;
     } else {
       return (prev.current = request);
--- a/packages/preact-urql/src/hooks/useRequest.ts
+++ b/packages/preact-urql/src/hooks/useRequest.ts
@@ -12,7 +12,7 @@ export const useRequest = (
   return useMemo(() => {
     const request = createRequest(query, variables);
     // We manually ensure reference equality if the key hasn't changed
-    if (prev.current !== undefined && prev.current?.key === request.key) {
+    if (prev.current != null && prev.current.key === request.key) {
       return prev.current;
     } else {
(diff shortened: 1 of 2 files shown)


packages/preact-urql/src/hooks/useRequest.ts
```diff
@@ -12,7 +12,7 @@ export const useRequest = (
   return useMemo(() => {
     const request = createRequest(query, variables);
     // We manually ensure reference equality if the key hasn't changed
-    if (prev.current !== undefined && prev.current.key === request.key) {
+    if (prev.current !== undefined && prev.current?.key === request.key) {
```
JoviDeCroock, line 15: We could relax this check to:

```suggestion
    if (prev.current != null && prev.current.key === request.key) {
```

JoviDeCroock (commented):
Hey and thanks for taking your time to PR this fix (and bring it to our attention). It seems like we're missing a changeset indicating we'll need a patch release. If you need help doing this I'm up for doing that, else you can do this by running `yarn changeset`

packages/preact-urql/src/hooks/useRequest.ts
kitten: is this potentially a big somewhere in Preact / Preact Compat? I see that we explicitly never have `current: null` set as a possible state for this ref, but since that's the default value I'm wondering whether it's an external bug?

JoviDeCroock: This is a bug in latest Preact (solved in master)

jlengstorf: wait, does that mean we can drop this entirely if Preact solved it upstream?

JoviDeCroock: imo, we can include this fix and support the version. It's not needed though but I see no harm in `!= null`

jlengstorf:
added a changeset and switched from optional chaining to a `null` check

kitten (approved):

JoviDeCroock (approved):

PR #947, opened 2020-08-27
(exchanges) implement error exchange

kadikraman:
## Summary

An exchange that allows inspecting all errors within the urql client.

```js
import { errorExchange, dedupExchange, fetchExchange } from 'urql';

const client = new Client({
  exchanges: [
    dedupExchange,
    errorExchange({
      onError: (error: CombinedError, operation: Operation) => {
        console.log("An error!", error);
      }
    }),
    fetchExchange
  ],
});
```

## Set of changes

- adding an error exchange to urql core
- adding a note about the error exchange in the documentation site
- tests! 👏

### DIFF
--- a/.changeset/many-comics-grin.md
+++ b/.changeset/many-comics-grin.md
@@ -1,0 +1,5 @@
+---
+'@urql/core': minor
+---
+
+Adds an error exchange to urql-core. This allows tapping into all graphql errors within the urql client. Useful for logging, debugging, handling authentication errors etc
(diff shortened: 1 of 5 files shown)


kitten (approved):
Let's just slap a change set onto this and send it on its merry way! 🥳🙌

wereHamster:
Saw this exchange documented on the homepage, but when I tried to use it I noticed it's not available yet in the npm package :(

kitten:
@wereHamster Yep, it's not quite released yet

jamesie:
When using this I find it just console.errors the error in my browser window and doesnt actually run any methods I call inside the callback. I run the code exactly like the documentation and this still happens. Does it also seem to not be working fine for other people or is this just a me problem?

Issue #938, opened 2020-08-14
RFC: Clarify terminology and edge cases in Graphcache docs


kitten:
- Add explanation for "entities," "links," and "records" to "Normalized Caching" introduction document, with diagrams if necessary (See: https://spectrum.chat/urql/help/solved-trying-to-update-a-relay-connection-edges-fields-with-invalidate-inspectfields~b86a1c05-2e73-44d5-bf4c-c9b605cd658b)
- Clarify that optimistic updates reflect the current cache state and can't query more data dynamically as they reflect offline optimistic changes and not real updates (See: https://spectrum.chat/urql/help/solved-cant-read-the-cache-inside-of-an-optimistic-response~860f1064-246e-46a8-9dc7-9b5aab19e4a2)
- Clarify what happens to entities without IDs in regards to how that can be worked around and how these entities can be updated (See: https://spectrum.chat/urql/help/optimistic-responses-without-an-id~1d76fe6b-3971-43f1-be69-80f0f1393ec5)

earksiinni:
Another idea based on https://spectrum.chat/urql/help/how-do-you-access-entities-directly-from-graphcache~76d4f5f5-1fdc-4ae7-8056-fc18fe2eb014:

Clarify cache manipulations:

- Default cache update after mutations
  - Attempts to automatically merge retrieved results based on `__typename` and key (but how? under what specific conditions? edge cases? etc.)
  - Clarify that the schema can be carefully designed to take advantage of this default behavior, give an example.
- Reading/writing query results
  - `readQuery`
  - `updateQuery`, which incidentally updates entities (explain under what conditions an entity update happens, how the merge works, etc.)
  - Clarify the difference and relationship between query results vs. cache entities (for example, when does updating one update the other?)
- Reading/writing whole entities
  - `readFragment`
  - `writeFragment`
- Reading/writing entity properties
  - `resolve`
  - `writeFragment`

PR #950, opened 2020-09-01
(docs) - elaborate on graphcache documentation

JoviDeCroock:
Resolves: https://github.com/FormidableLabs/urql/issues/938

## Summary

It clarifies some frequent misunderstandings within the graphCache documentation.

### DIFF
--- a/docs/graphcache/custom-updates.md
+++ b/docs/graphcache/custom-updates.md
@@ -5,10 +5,10 @@ order: 3
 
 # Custom Updates
 
-Every time a `subscription` triggeres or a `mutation` receives a response GraphCache will look at the response and see
+Every time Graphcache sees a result from the API for a `subscription` or a `mutation` it will look at the response and traverse it.
 if it can get a key from the response. This key is by default derived from the `__typename` and `id`/`_id`, unless
 customized in your `keys` configuration, this entity will then be added to the cache this means that
-an existing entity with the found key will be updated.. While it will update all normalized entities that it finds in
+This means that existing entities with a found key will be updated and new ones will be written. While it will update all normalized entities that it finds in
 those results, it can't for instance tell whether a new item should be appended or removed from a
 list.
 
@@ -271,7 +271,7 @@ The _Optimistic Updates_ configuration allows us to set up "temporary" results f
 will be applied immediately. This is a great solution to reduce the waiting time for the user.
 
 > Note that an optimistic response is meant to be a temporary update to an entity until the server responds to your mutation.
-> This means that what you return here should reflect the shape of what the server would return.
+> This means that what you return here should reflect the shape of what the server will return.
 
 This technique is often used with one-off mutations that are assumed to succeed, like starring a
 repository, or liking a tweet. In such cases it's often desirable to make the interaction feel
--- a/docs/graphcache/custom-updates.md
+++ b/docs/graphcache/custom-updates.md
@@ -5,8 +5,10 @@ order: 3
 
 # Custom Updates
 
-_Graphcache_ will attempt to automatically react to your mutations' and subscriptions' results
-but sometimes this isn't possible. While it will update all normalized entities that it finds in
+Every time a `subscription` triggeres or a `mutation` receives a response GraphCache will look at the response and see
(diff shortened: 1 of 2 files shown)


docs/graphcache/normalized-caching.md
```diff
@@ -62,6 +62,29 @@ Multiple results may refer to the same piece of data — a mutation for instance
...
+A few terms that will be used throughout the GraphCache documentation that are important to know
+to get a full understanding.
+
+### Entity
+
+This is an object for which the cache can generate a key, like `Todo:1`.
+
+### Records
+
+These are properties that relate to an Entity, in the above case this would be `title`, ...
```
wgolledge, line 76:
```suggestion
These are properties that relate to an entity, in the above case this would be `title`, ...
```

wgolledge (commented):
Nice updates!!

docs/graphcache/custom-updates.md
```diff
@@ -5,8 +5,10 @@ order: 3
 
 # Custom Updates
 
-_Graphcache_ will attempt to automatically react to your mutations' and subscriptions' results
-but sometimes this isn't possible. While it will update all normalized entities that it finds in
+Every time a `subscription` triggeres or a `mutation` receives a response GraphCache will look at the response and see
```
kitten, line 8:
```suggestion
Every time Graphcache sees a result from the API for a `subscription` or a `mutation` it will look at the response and traverse it.
```

docs/graphcache/custom-updates.md
```diff
@@ -5,8 +5,10 @@ order: 3
...
+if it can get a key from the response. This key is by default derived from the `__typename` and `id`/`_id`, unless
```
kitten, line 9: > This process is the same as for queries but instead of starting at the `Query` root type, it will start searching for keyable entities, where an object's `__typename` and `id` or `_id` fields (or a custom `keys` config) are provided, so it can write normalized entities to the cache.

docs/graphcache/custom-updates.md
```diff
@@ -5,8 +5,10 @@ order: 3
...
+customized in your `keys` configuration, this entity will then be added to the cache this means that
+an existing entity with the found key will be updated.. While it will update all normalized entities that it finds in
```
kitten, line 11:
```suggestion
This means that existing entities with a found key will be updated and new ones will be written. While it will update all normalized entities that it finds in
```

docs/graphcache/normalized-caching.md
```diff
@@ -62,6 +62,29 @@ Multiple results may refer to the same piece of data — a mutation for instance
 later on in the app. This would automatically cause _Graphcache_ to update any related queries in
 the entire app, because all references to each each entity are shared.
 
+## Terminology
+
+A few terms that will be used throughout the GraphCache documentation that are important to know
```
kitten, line 67: *Graphcache is what we use consistently

docs/graphcache/normalized-caching.md
```diff
@@ -62,6 +62,29 @@ Multiple results may refer to the same piece of data — a mutation for instance
...
+to get a full understanding.
+
+### Entity
```
kitten, line 70: Does this look good in the markdown document or should this be a list with emphasis marks?

docs/graphcache/normalized-caching.md
```diff
@@ -62,6 +62,29 @@ Multiple results may refer to the same piece of data — a mutation for instance
...
+These are properties that relate to an Entity, in the above case this would be `title`, ...
+internally these will be represented as `Todo:1.title`
+
+### Link
+
+This is the connection between entities or the base `Query` field.
+
+a few examples would be:
+
+- `Query.todos: an array of ids`
```
kitten, line 85: I'd prefer to not highlight this as code, use the word "keys," and instead say "this may be an array or a single string which are keys that refer to other entities"

kitten:
I think we're still missing some notes (in normalized cache?) how the updating process works and what happens to unkeyable entities and records. That'd be a more seamless way to introduce the terminology and would explain how the entire logic works under the hood, traversing entities, generating embedded keys where necessary, and reading it back the same way if possible

kitten (approved):

PR #972, opened 2020-09-09
Fix some typos in auth exchange readme

Marcisbee:
## Summary

Just fixing some typos.

As for `errorExchange` change, typescript throws type error, so I presume that it should be changed too:
image


kitten (approved):
Whoops! Nice catch!

PR #973, opened 2020-09-09
(docs) - Add page for more detailed auth docs

kadikraman:
🔐

### DIFF
--- a/docs/advanced/authentication.md
+++ b/docs/advanced/authentication.md
@@ -1,0 +1,290 @@
+---
+title: Authentication
+order: 8
+---
+
+Most APIs include some type of authentication, usually in the form of an auth token that is sent with each request header.
+
+The purpose of the [`authExchange`](../api/auth-exchange) is to provide a flexible api that facilitates the typical
+JWT-based authentication flow.
+
+## Typical Authentication Flow
+
+1. Initial login
+    The user opens the application and authenticates for the first time. They enter their credentials and receive an auth token.
+    The token is saved to storage that is persisted though sessions (e.g. `localStorage` on the web or `AsyncStorage` in React Native).
+    The token is added to each subsequent request in an auth header.
+
+2. Resume
+    The user opens the application after having authenticated in the past. In this case, we should already have the token in persisted storage.
+    We fetch the token from storage and add to each request, usually as an auth header.
+
+3. Forced log out due to invalid token
+    The user's session could become invalid for a variety reasons:
+    - their token expired
+    - they requested to be signed out of all devices
+    - their session was invalidated remotely
+
+    In this case, we would want to also log them out in the application so they could have the opportunity to log in again. To do this,
+    we want to clear any persisted storage, and redirect them to the application home or login page.
+
+4. User initiated log out
+    When the user chooses to log out of the application, we usually send a logout request to the api, then clear any tokens from persisted
+    storage, and redirect them to the application home or login page.
+
+5. Refresh (optional)
+    This is not always implemented, but given that your API supports it, the user will receive both an auth token and a refresh token, where the auth token
+    is valid for a shrter duration of time (e.g. 1 week) than the refresh token (e.g. 6 months) and the latter can be used to request a new
+    auth token if the auth token has expired.
+
+    The refresh logic is triggered either when the JWT is known to be invalid (e.g. by decoding it and inspecting the expiry date), or when an api
+    request returns with an unauthorized response. For graphQL APIs, it is usually an error code, instead of a 401 HTTP response, but both can be
+    supported. When the token as been successfully refreshed (this can be done as a mutation to the graphQL api or a request to a different api
+    endpoint, depending on implementation), we will save the new token in persisted storage, and retry the failed request with the new auth header.
+    The user should be logged out and persisted storage cleared if the refresh fails or if the re-executing the query with the new token fails with
+    an auth error for the second time.
+
+## Implementation
+
+First, install the `@urql/exchange-auth` alongside `urql`:
+
+```sh
+yarn add @urql/exchange-auth
+# or
+npm install --save @urql/exchange-auth
+```
+
+You'll then need to add the `authExchange`, that this package exposes to your Client. The `authExchange` is an asynchronous exchange, so it must be placed
+in front of all `fetchExchanges` but after all other synchronous exchanges, like the `cacheExchange`.
+
+```js
+import { createClient, dedupExchange, cacheExchange, fetchExchange } from 'urql';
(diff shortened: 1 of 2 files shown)


JoviDeCroock (approved):
This is so detailled, awesome! 💯

docs/advanced/authentication.md
```diff
@@ -0,0 +1,328 @@
...
+const didAuthError = ({ error }) => {
+  return error.graphQLErrors.some(
+    e => error.response.status === 401,
+  );
+},
+```
+
+Then `didAuthError` returns `true`, it will trigger the exchange to trigger the logic for asking for reauthentication via `getAuth`.
+
+### Configuring `getAuth` (after `didAuthError` returns `true`)
```
kitten, line 219: If "after `didAuthError` returns `true` is important information and there's no corresponding section that for instance says `false` instead, let's move this into the text below as a full explanation?

> `getAuth` will be triggered to refresh the authentication state if `didAuthError` returned `true` previously, due to at least one GraphQL request erroring out with an authentication error, as we defined previously.

docs/advanced/authentication.md
```diff
@@ -0,0 +1,328 @@
...
+    localStorage.setItem('refreshToken', result.data.refreshLogin.refreshToken);
+
+    return {
+      token: result.data.refreshLogin.token,
+      refreshToken: result.data.refreshLogin.refreshToken,
+    };
+  }
+
+  localStorage.clear();
+  logout();
```
kitten, line 270:
```suggestion
  // This is where auth has gone wrong and we need to clean up and redirect to a login page
  localStorage.clear();
  logout();
```

Just adding a comment here since this isn't reusable for the user's app

docs/advanced/authentication.md
```diff
@@ -0,0 +1,328 @@
...
+
+  logout();
+
+  return null;
+}
+```
+
+Here, `logout()` should clear your persisted storage of any tokens and set the `isLoggedIn` property as described above to `false`.
+
+If you do have a token refresh endpoint, and a refresh token, you can attempt to get a new token for the user:
```
kitten, line 242:
```suggestion
Here, `logout()` is a placeholder that is called when we got an error, so that we can redirect to a login page again and clear our tokens from local storage or otherwise.

If we had a way to refresh our token using a refresh token, we can attempt to get a new token for the user first:
```

Just rewriting to keep the first-person plural consistent

docs/advanced/authentication.md
```diff
@@ -0,0 +1,328 @@
...
+  }
+
+  localStorage.clear();
+  logout();
+
+  return null;
+}
+```
+
+Here we use the special mutate function provided by the auth exchange to do the token refresh. If your auth is not handled via graphQL, you can
```
kitten, line 276:
```suggestion
Here we use the special mutate function provided by the auth exchange to do the token refresh. If your auth is not handled via GraphQL but a REST endpoint, you can
```

docs/advanced/authentication.md
```diff
@@ -0,0 +1,328 @@
...
+use `fetch` in this function instead of a mutation. All other requests will be paused while `getAuth` returns, so you won't get multiple auth failures
```
kitten, line 277:
```suggestion
use `fetch` in this function instead of a mutation. All other requests will be paused while `getAuth` returns, so we never have to handle multiple auth failures
```

docs/advanced/authentication.md
```diff
@@ -0,0 +1,328 @@
...
+at the same time.
+
+### Configuring `willAuthError`
+
+`willAuthError` is an optional parameter and is run _before_ a network request is made. You can use it to trigger the logic in
+`getAuth` without the need to get a graphql error first. For example, you could use this indicate that the auth will fail because the JWT is invalid:
```
kitten, line 283:
```suggestion
`willAuthError` is an optional parameter and is run _before_ a network request is made. We can use it to trigger the logic in
`getAuth` without the need to send a request and get a GraphQL Error back. For example, we can use this to predict that the authentication will fail because our JWT is invalid already:
```

docs/advanced/authentication.md
```diff
@@ -0,0 +1,328 @@
...
+const willAuthError = ({ authState }) => {
+  if (!authState || /* JWT is expired */) return true;
+  return false;
+}
+```
+
+## Handling Logout with the Error Exchange
+
+It is also possible to handle your logout logic in the error exchange instead of in the auth exchange. To do this, you'll need to add the
+`errorExchange` to the exchanges array, _before_ the `authExchange`. The order is very important here:
```
kitten, line 295:
```suggestion
We can also handle authentication errors in an `errorExchange` instead of the `authExchange`. To do this, we'll need to add the
`errorExchange` to the exchanges array, _before_ the `authExchange`. The order is very important here:
```

docs/advanced/authentication.md
```diff
@@ -0,0 +1,328 @@
...
+import { createClient, dedupExchange, cacheExchange, fetchExchange, errorExchange } from 'urql';
+import { authExchange } from '@urql/exchange-auth';
+
+const client = createClient({
+  url: '/graphql',
+  exchanges: [
+    dedupExchange,
+    cacheExchange,
+    errorExchange({
+      onError: ({ error }) => {
```
kitten, line 307:
```suggestion
      onError: (error) => {
```

docs/advanced/authentication.md
```diff
@@ -0,0 +1,328 @@
...
+    }),
+    fetchExchange,
+  ],
+});
+```
+
+The `errorExchange` will only receive an auth error when the auth exchange as tried and failed to handle it. This means we have
+either failed to refresh the token, or there is no token refresh functionality. If you get an auth error in the `errorExchange` (as defined in
+the `didAuthError` configuration section above), then you can be confident that it is an auth error, the `authExchange` isn't able to recover
+from it, and the user should be logged out.
```
kitten, line 328:
```suggestion
The `errorExchange` will only receive an auth error when the auth exchange has already tried and failed to handle it. This means we have
either failed to refresh the token, or there is no token refresh functionality. If we receive an auth error in the `errorExchange` (as defined in
the `didAuthError` configuration section above), then we can be confident that it is an auth error that the `authExchange` isn't able to recover
from, and the user should be logged out.
```

docs/advanced/authentication.md
kadikraman: The two sections are "Configuring `getAuth` (fetch from storage)" and "Configuring `getAuth` (after `didAuthError` returns `true`)".

I could rephrase it to "after an auth error has happened"?

kitten (approved):

Issue #974, opened 2020-09-11
Schema awareness: can I avoid using it? How can I optimize schema.json? Can I code-split it?


frederikhors:
# Summary

- [ ] Use URQL CacheExchange without Schema awareness

  - [ ] Fix the infinite reloading issue

  - [ ] Document how to

- [ ] Document ways we can optimize Introspection Schema

  - [x] Amazing tool from urql team (https://github.com/FormidableLabs/urql/pull/976)

    - [ ] Strip out "descriptions" from final `.json`

    - [ ] Sort alphabetically (or such) final `.json` to check for any changes

  - [ ] Document how to wrap the entire introspection result in a string and `JSON.parse(it)`; update example (maybe the svelte one) with this
  
  - [ ] Is there something else to remove from the final schema?

- [ ] Generate schema

  - [x] Generate it from API introspection call (docs OK!)

  - [ ] Generate it from `*.graphql` files (needs script)

- [ ] How to load schema the async way with code-splitting, docs

## Story

Schema awareness is the most loved and hated thing for those who use URQL.

<details>
<summary>Read the long story</summary>

As soon as the application becomes a little bigger than a simple "hello world" you need to use different queries for the same entities that you can first see as part of a list and then as a detail.

Some fields of these queries may be `null` but URQL cannot know this. He needs someone to "show him the way".

That someone is our GraphQL Schema.

One advantage of the awareness schema is that urql doesn't have to wait for new query data to fetch and can show the data it has already cached for that entity.

One "disadvantage" is that for some projects the schema is very large and this is a problem both for the final bundle size and for the browser's parsing time (TTI, time to interactive).

</details>

## Use URQL CacheExchange without Schema awareness:

- [We started talking about this here](https://spectrum.chat/urql/help/how-about-a-convention-for-list-detail-that-speeds-up-the-ui-a-lot~4bcb6019-8ada-4b6d-b325-72e9244217d8)

- @kitten suggested a method that might work, but he also made it clear that it is not "sponsored" by the URQL team because quite rightly it is error-prone and may not make 100% sense:

  <details>
  <summary>See the suggested code</summary>

  ```js
  const maybePartial = (parent, args, cache, info) => {
    const fieldValue = cache.resolve(info.parentKey, info.fieldName, Object.keys(args).length ? args : undefined);
    if (fieldValue === null) {
      info.partial = true;
      return null;
    } else {
      return fieldValue;
    }
  };

  const resolvers = {
    Query: {
[...]

PR #976, opened 2020-09-11
(introspection) - Implement initial @urql/introspection package

kitten:
Related #974

The new `@urql/introspection` utilities aim to contain all kinds of functions that may be used to treat client-side GraphQL schemas, currently: to minify them. This may be used to minify an `IntrospectionQuery` to get it down to a minimum possible size.

This has been tested against the GitHub GraphQL API schema and it brings it down from `49.4kB` minzipped to `29.6kB` minzipped. The uncompressed sizes are `1.1MB` and `551KB`.

We've determined that it's _not_ effective to translate the GraphQL Introspection Query to a schema string and turning that back to a `GraphQLSchema`. The reduction is about a third of size, but this does pull in about 15-20KB more from `graphql`. Since we're dealing with a worst-case scenario sized schema here, this doesn't seem worth it, since that extra code increases the parsing time much more (including the GraphQL schema parsing time) than pure JSON.

### DIFF
--- a/packages/introspection/src/minifyIntrospectionQuery.ts
+++ b/packages/introspection/src/minifyIntrospectionQuery.ts
@@ -88,14 +88,18 @@ const minifyIntrospectionType = (
                 })),
             } as any)
         ),
-        interfaces: type.interfaces.map(int => ({
-          kind: 'INTERFACE',
-          name: int.name,
-        })),
-        possibleTypes: type.possibleTypes.map(type => ({
-          kind: type.kind,
-          name: type.name,
-        })),
+        interfaces:
+          type.interfaces &&
+          type.interfaces.map(int => ({
+            kind: 'INTERFACE',
+            name: int.name,
+          })),
+        possibleTypes:
+          type.possibleTypes &&
+          type.possibleTypes.map(type => ({
+            kind: type.kind,
+            name: type.name,
+          })),
       };
     }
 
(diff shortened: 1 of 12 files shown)


packages/introspection/src/minifyIntrospectionQuery.ts
```diff
@@ -0,0 +1,153 @@
...
+        `Unrecognized type reference of type: ${(fromType as any).kind}.`
+      );
+  }
+};
+
+const minifyIntrospectionType = (
+  type: IntrospectionType
+): IntrospectionType => {
+  switch (type.kind) {
+    case 'OBJECT': {
```
JoviDeCroock, line 48: Can we maybe use the `Kind` exported from graphql for safety?

JoviDeCroock (approved):
Nit, not blocking. Awesome work Phil

packages/introspection/src/minifyIntrospectionQuery.ts
kitten: I'm not sure if it's the same 🤔

JoviDeCroock: Oh all right! thanks for checking 👍

frederikhors:
Amazing! Now it would be nice to understand how I use it! LOL! :smile:

Issue #901, opened 2020-07-11
RFC: GCG Typegen for Graphcache's config


Andrew-Talley:
## Summary
Right now, the cache exchange's types seem to do a good job at preventing any type errors without manual casting. That being said, it doesn't have the same support for type-hinting that other parts of the package (e.g., `useQuery`) do. This also seems to have confused some people (e.g., #851). I think the cacheExchange could provide better type support by being transformed into a generic function that lets users provide the types that should appear in their cache, so that, e.g., when writing an optimistic response, the types of arguments available will be listed.

## Proposed Solution
I can think of two ways to achieve this. First, provide manual overrides of the `CacheExchangeOpts` types. This would enable graphql code generating tools to build these types. However, this makes it highly unlikely that someone would build all these types by hand, since that'd be just as slow as using the existing types. And, until the code generators tools exist, this feature would be mostly useless.

The second way would be to let something representing the schema be passed to it, and determine the types that way. I think the best candidate would be something like graphql-codegen's [typescript resolvers](https://graphql-code-generator.com/docs/plugins/typescript-resolvers). Then, the cache exchange could derive the types it needs from that. I built out types that at least show it's possible to do this for all of the fields on the cache exchange options – I'm happy to post all of those if that's helpful, but to give an example of just one:

```typescript
type ResolverMap = Record<string, Record<string, (...args) => any>>; // Simplified representation of graphql codegen's resolver type
type MapParent<T extends Record<string, AnyFunction>> = {
  [P in keyof T]: ReturnType<T[P]>
}
type NewResolverConfig<T extends ResolverMap> = {
  [P in keyof T]: {
    [R in keyof T[P]]: (parent: MapParent<T[P]>, args: Parameters<T[P][R]>[1], cache: Cache, info: ResolverInfo) => DataField | undefined; // The return type could likely be improved – maybe Partial<ReturnType<T[P][R]>> | undefined
  }
}


export interface CacheExchangeOpts<T extends ResolverMap | undefined = undefined> {
    resolvers?: T extends undefined ? ResolverConfig : NewResolverConfig<T>;
    // ... updates, optimistic, etc.
}
```
Then, a user would be able to define a type representing their schema and pass it to cache exchange to get type suggestions:
```typescript
type SchemaResolver = {
  UserDefinedType: {
    field: (parent: UserDefinedType, args: { id: string }) => DataField
  }
}
const myCacheExchange = cacheExchange<SchemaResolver>({ ... });
// Or, of course
[...]

PR #978, opened 2020-09-13
allow use of generics in the cacheExchange

JoviDeCroock:
## Summary

Adds generic typings to the cacheExchange and add support for alternative mutation types.

This issue was found by a [Spectrum user](https://spectrum.chat/urql/help/typenames-for-root-types-in-schema-json~4b01bc46-8315-4aaa-ae56-9d06b049e215)

The goal would be to automatically generate these generics from the graphql-schema later one with a tool like graphql-code-generator.

## Set of changes

- Can now use `cacheExchange<UpdaterConfig, ...>`
- We support alternative mutation typenames on the updatersConfig now

### DIFF
--- a/exchanges/graphcache/src/cacheExchange.ts
+++ b/exchanges/graphcache/src/cacheExchange.ts
@@ -73,21 +73,24 @@ const toRequestPolicy = (
   },
 });
 
-export interface CacheExchangeOpts {
-  updates?: Partial<UpdatesConfig>;
-  resolvers?: ResolverConfig;
-  optimistic?: OptimisticMutationConfig;
-  keys?: KeyingConfig;
+export interface CacheExchangeOpts<Updaters, Resolvers, Optimistic, Keys> {
+  updates?: Partial<Updaters>;
+  resolvers?: Resolvers;
+  optimistic?: Optimistic;
+  keys?: Keys;
   schema?: IntrospectionQuery;
   storage?: StorageAdapter;
 }
 
-export const cacheExchange = (opts?: CacheExchangeOpts): Exchange => ({
-  forward,
-  client,
-  dispatchDebug,
-}) => {
-  const store = new Store(opts);
+export const cacheExchange = <
+  Updaters = UpdatesConfig,
+  Resolvers = ResolverConfig,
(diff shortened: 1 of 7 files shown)


exchanges/graphcache/src/store/store.ts
```diff
@@ -70,18 +70,26 @@ export class Store implements Cache {
       const queryType = schema.getQueryType();
       const mutationType = schema.getMutationType();
       const subscriptionType = schema.getSubscriptionType();
+
       if (queryType) queryName = queryType.name;
       if (mutationType) mutationName = mutationType.name;
       if (subscriptionType) subscriptionName = subscriptionType.name;
 
+      this.updates = {
+        [mutationName]: (opts.updates && opts.updates[mutationName]) || {},
```
kitten, line 79: I'm kind of confused by this; we should be doing this normalisation in the traversal logic but internally all root types should have the standard names once they leave the schema; so I'm wondering if this is 100% correct

kitten:
I think the generics are maybe a bit problematic to be added at this point? I know we must have some kind of normalisation issue with the root type names somewhere again, but should we first fix that? I assume the generics are more needed once we jump onto #901

exchanges/graphcache/src/cacheExchange.ts
```diff
@@ -73,21 +73,24 @@ const toRequestPolicy = (
   },
 });
 
-export interface CacheExchangeOpts {
-  updates?: Partial<UpdatesConfig>;
-  resolvers?: ResolverConfig;
-  optimistic?: OptimisticMutationConfig;
-  keys?: KeyingConfig;
+export interface CacheExchangeOpts<Updaters, Resolvers, Optimistic, Keys> {
```
andyrichardson, line 76: This should keep that type safety in there

```suggestion
export interface CacheExchangeOpts<Updaters extends UpdatesConfig, Resolvers extends ResolverConfig, Optimistic extends OptimisticMutationConfig, Keys extends KeyingConfig> {
```

kitten:
Closed for now so we can tackle this / reopen again when we work on the GraphQL Code Generator counterpart

PR #981, opened 2020-09-17
Use ramdisk in CI

andyrichardson:


kitten (approved):
This should really be the default 😂

PR #985, opened 2020-09-18
(graphcache) - Fix offline queries not delivering any results on cache-miss and offline

kitten:
## Summary

Previously when the user was offline and the cache would fail with a complete cache miss for offline queries then no result would be delivered to the client and instead would fetch infinitely. Instead the `offlineExchange` now retries with `cac he-only`.

### DIFF
--- a/.changeset/pretty-pans-doubt.md
+++ b/.changeset/pretty-pans-doubt.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-graphcache': patch
+---
+
+Fix operation results being obstructed by the `offlineExchange` when the network request has failed due to being offline and no cache result has been issued. Instead the `offlineExchange` will now retry with `cache-only` policy
--- a/docs/graphcache/offline.md
+++ b/docs/graphcache/offline.md
@@ -12,9 +12,6 @@ Updates](./custom-updates.md#optimistic-updates) this can be used to build an ap
 serves cached data entirely from memory when a user's device is offline and still display
 optimistically executed mutations.
 
-> **NOTE:** Offline Support is currently experimental! It hasn't been extensively tested yet and
-> may not always behave as expected. Please try it out with caution!
-
 ## Setup
 
 Everything that's needed to set up offline-support is already packaged in the
--- a/exchanges/graphcache/src/cacheExchange.ts
+++ b/exchanges/graphcache/src/cacheExchange.ts
@@ -27,6 +27,7 @@ import {
 
 import { query, write, writeOptimistic } from './operations';
 import { makeDict, isDictEmpty } from './helpers/dict';
+import { addCacheOutcome, toRequestPolicy } from './helpers/operation';
 import { filterVariables, getMainOperation } from './ast';
 import { Store, noopDataState, hydrateData, reserveLayer } from './store';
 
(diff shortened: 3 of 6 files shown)


frederikhors:
Amazing! Please merge it! LOL :smile:

kitten:
@frederikhors ha! I was just about to send it over to you 😂 have you tested whether it behaves as expected for you now?

frederikhors:
Nope. In a few minutes. I'll write here.

wgolledge (approved):

frederikhors:
@kitten I'm sad.

It works just the first time I come back to list page.

The second time it's as before: `data: undefined`.

:((((((

kitten:
@frederikhors That may be expected in some cases though if somehow some data is uncached. If `fetching: false, stale: false` then it's working but has a cache miss. You can also check `operation.context.meta.cacheOutcome`

frederikhors:
Nope. fetching is true now!

frederikhors:
Oh wow! I just discovered it works removing `dedupExchange` and `retryExchange`. I'm still trying to understand what the problem is.

frederikhors:
The problem is not `retryExchange`. Is` dedupExchange`.

I don't know why. I'm trying to understand.

kitten:
Ok, I believe that's a timing issue depending on where the operation is coming from, which would in some cases be too soon if `cache-and-network` hasn't issued a prior result. So this commit should fix that.

frederikhors:
I will try now.

frederikhors:
Tried it with `dedupExchange`, it's the same: `{ fetching: true, stale: false, data: undefined }`.

:(

kitten:
@frederikhors So, let's break this down, that's really odd; it works without the `dedupExchange` even with the latest commit?

frederikhors:
> it works without the `dedupExchange` even with the latest commit?

Yep. Withtou dedup the `players` is:

`{ fetching: false, stale: true, data: [...] }`

kitten:
Ok, I'm not even sure if that's a related thing then, since the new commit circumvents the full pipeline as we usually do when we explicitly want an operation to skip `dedup` with an added subject in `offlineExchange`. So it doesn't even go through the `dedupExchange` anymore to run the additional `cache-only` operation.

frederikhors:
This is the codesandbox: https://codesandbox.io/s/urql-svelte-crud-offline-fetching-stale-data-rj22w.

I can't find a way to try the commit package in codesandbox, do you know how to do?

Otherwise you can download it and try locally.

If you visit `/todos` and then `/todo/<ID>` it will cache data.

Then you can try to put your browser offline and go again back to list and again on that todo.

And again.

And you'll see the problem.

kitten:
@frederikhors That's fixed now as far as I can tell. We're not able to return the result synchronously in that case, because it's already gone out to the `fetchExchange` so things are getting a little confused. Anyway, I've checked this against your sandbox and all is well

frederikhors:
> things are getting a little confused

I'm sad about this.

What do you think to do?

I'm trying it now.

Tests are failing...

kitten:
@frederikhors I mean, like I said, this should be working as expected now; so I believe there were some timing issues in my previous commit that didn't let the operation to be reprocessed, but it should be fine now.

frederikhors:
I'm trying it now.

frederikhors:
I doesn't work with dedup enabled, @kitten.

frederikhors:
Maybe I'm wrong.

I'm changing in my node_modules the content of `@urql/exchange-graphcache` using https://pkg.csb.dev/FormidableLabs/urql/commit/08510251/@urql/exchange-graphcache.

Right?

kitten:
@frederikhors i just tested it by pasting the latest commit's CodeSandbox CI build into the sandbox, and switching my browser to offline-mode, and it does behave as expected for me, so I'm not sure if all the steps are reflected on your end; but I'm pretty certain that it's working

frederikhors:
@kitten. I understand now.

The codesandbox is working right now with your latest commit in package.json. No problem.

The problem I have now is with `retryExchange`: this was my issue with your latest commit. And now dedup is enabled.

The result is as before (after one back and forth, list > detail):  `{ fetching: true, stale: false, data: undefined }`.

frederikhors:
> I think there's a subtle bug either in the Svelte integration or in how the client is initialised when starting from the start page. When I go to the app's homepage, then offline, then go to the list it doesn't even send any network requests, which is unrelated to this PR. I'm not sure why that's happening but that'd explain why the changes in principle are working but there's still an issue in the test app.

It's codesandbox and IndexedDB. If you remove it manually from browser it works.

If you try it in Incognito mode it works.

frederikhors:
So now the problem is with `retryExchange`.

I'm very tired... :smile:

PR #1001, opened 2020-09-24
[Proposal] Add getQueryNames utility to Urql

andyrichardson:
## About

There are a number of use cases where this is useful - mostly in testing/snapshot environments.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -10,4 +10,5 @@ export {
   makeErrorResult,
   formatDocument,
   maskTypename,
+  getQueryNames,
 } from './utils';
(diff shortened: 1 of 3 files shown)


kitten:
Temporarily closing this out so that #1005 can be tackled in its entirety in one PR.
Small side note, once that work starts this `getOperationName` utility can simply be moved and exported. We already "implicitly enforce" (by convention and by how we run requests) a single operation being sent so I think there's also no need for us to expose a utility to extract multiple operation names; https://github.com/FormidableLabs/urql/blob/8c74924c1d49588f265cbc3776e98fbff4995199/packages/core/src/internal/fetchOptions.ts#L13-L20

Issue #1007, opened 2020-09-28
RFC: Intent to implement stable Svelte bindings


kitten:
## Summary

Currently I see our current `@urql/svelte` package and its past and current implementation as nice experiments which helped check the boundaries of Svelte and test APIs and the understanding of users.

The latest PR allowed _all_ modes of Svelte's stateful stores to be used somehow, with similar semantics to the other bindings, See #862 

It's becoming increasingly clear that this is confusing and doesn't feel idiomatic to Svelte.
The next implementation of `@urql/svelte` should start from a clean slate and be idiomatic to Svelte and intuitively avoid confusion. It is allowed to stray away from feeling like the `urql` and `@urql/preact` bindings, even forgoing permanent state memory if necessary.

## Problems with the approach

The current approach attempts to allow three modes of usage:

- Store usage without memoised state
- Store usage with memoised state via an intermediate factory function
- Async-await/Promise usage

The last option of these feels idiomatic to Svelte but _should be discouraged_. We never aim to show results that can never reactively update. This is objectively bad.

The memoised usage is confusing and Svelte simply does not have a full reactivity model that allows us to observe changes. The user can do this manually and this may feel wrong to us, trying to implement complex hooks flows in Svelte, however this is not our problem to solve.

## Bugs to avoid

- Unmounting issues, which may be inherent to our implementation: #948, #979
- Incorrect promise behaviour #932

frederikhors:
❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️

kitten:
An update has been published as `@urql/svelte@1.0.0`

PR #1011, opened 2020-09-29
(graphcache) - Fix offline queries not flowing through the entire cacheExchange

kitten:
Test against: https://codesandbox.io/s/urql-crud-router-doesnt-update-forked-3nf19

This change updates when queries that have received an offline / network error are dispatched as `cache-only` operations again, and adds support for them being retried automatically (in case they're still active) when the client comes back online.

A small change has also been made to `@urql/core` to allow `client.reexecuteOperation` to be called with mutations.

### DIFF
--- a/.changeset/strange-waves-confess.md
+++ b/.changeset/strange-waves-confess.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-graphcache': patch
+---
+
+Fix queries that have erroed with a `NetworkError` (`isOfflineError`) not flowing back completely through the `cacheExchange`.
--- a/.changeset/strange-waves-confess.md
+++ b/.changeset/strange-waves-confess.md
@@ -3,3 +3,4 @@
 ---
 
 Fix queries that have erroed with a `NetworkError` (`isOfflineError`) not flowing back completely through the `cacheExchange`.
+These queries should also now be reexecuted when the client comes back online.
--- a/.changeset/tidy-bobcats-sniff.md
+++ b/.changeset/tidy-bobcats-sniff.md
@@ -1,0 +1,5 @@
+---
+'@urql/core': patch
+---
+
+Allow `client.reexecuteOperation` to be called with mutations which skip the active operation minimums.
--- a/exchanges/graphcache/src/offlineExchange.test.ts
+++ b/exchanges/graphcache/src/offlineExchange.test.ts
@@ -229,10 +229,7 @@ describe('offline', () => {
     );
 
     next(queryOp);
-    expect(result).toBeCalledTimes(0);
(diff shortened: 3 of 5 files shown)


kitten:
Tested against: https://codesandbox.io/s/urql-crud-router-doesnt-update-forked-7c622?file=/src/urql.js

**Working as intended!** The todos page starts fetching and receives an empty result with the passed operation as expected.

Screenshot 2020-09-29 at 16 17 20

JoviDeCroock (approved):

frederikhors:
@urql/svelte is not working because I think @urql/core is not updated to latest one.

Am I wrong?

kitten:
@frederikhors If you've upgraded `@urql/exchange-graphcache` then you'll need to upgrade `@urql/core`. It may now also be duplicated, so just bump its version or check the lockfile and bump it there.

frederikhors:
I tried. Also with @urql/core 1.13.1 it doesn't work.

I'll updated my CodeSandBox reproduction now and I will post it here or where you want.

frederikhors:
Also in the [Preact reproduction on CodeSandBox](https://codesandbox.io/s/urql-crud-router-doesnt-update-5x5tr) only updating:

1. `@urql/exchange-graphcache` from `3.1.1` to `3.1.2` and

1. `@urql/core` to `1.13.1`

**doesn't fix.**

So I think with Svelte is the same: only upgrading `@urql/exchange-graphcache` to `3.1.2` and `@urql/core` to `1.13.1` **is not enough**.

Something in `@urql` is changed that doesn't work anymore with `@urql/svelte` today.

frederikhors:
Here the broken reproduction with Svelte today: https://codesandbox.io/s/urql-svelte-crud-offline-fetching-stale-data-rj22w.

kitten:
@frederikhors ive got a fixed up sandbox in the original issue that's been tested to work.
I'm afraid any discrepancy here may be that the expected behaviour or conditions are flawed in the first place?

I think with Svelte we'll need to focus on new bindings. There's too many things that can break, be misused, or are buggy in the state that they're in.

frederikhors:
> ive got a fixed up sandbox in the original issue that's been tested to work.

Yep because you updated `@urql/preact` too. This is what I'm saying. Svelte is broken today.

However, I'm very tired of all this.

Maybe I'll leave Svelte and I'll use Vue with Apollo.

We'll see.

It has been a great adventure with you so far, but I feel I bother you a lot every time.

Thanks for everything.

kitten:
@frederikhors that's alright; that said though, `@urql/preact` is now expected to work but it'll take #1007 for `@urql/svelte`'s issues to be ironed out; although we do have a good list of issues with it now

PR #1019, opened 2020-09-30
(graphcache) - Fix offlineExchange timing after rehydration

kitten:


exchanges/graphcache/src/cacheExchange.ts
```diff
@@ -249,10 +250,15 @@ export const cacheExchange = (opts?: CacheExchangeOpts): Exchange => ({
     // If no hydration takes place we replace this stream with an empty one
     const bufferedOps$ = hydration
       ? pipe(
-          sharedOps$,
-          buffer(fromPromise(hydration)),
+          combine(
```
kitten, line 253: This works because `combine` can retroactively still resolve an operation synchronously once the first operation comes in. This lets `bufferedOps$` hang in its active state for as long or short as needed.

`buffer` on the other hand may not emit when its list of values is empty, but would wait for the next signal if then a new event comes in. Because we'd before only wait for one buffer output though and have one input event, there are conditions where `bufferedOps$` would stay active indefinitely.

JoviDeCroock (approved):

PR #1023, opened 2020-10-02
(vue) - initial implementation

JoviDeCroock:
## Summary

This is an initial implementation for the vue-urql integration. I found out that we could provide backwards compatibility for version 2 of Vue as well by using [this package](https://github.com/antfu/vue-demi) this does essentially not change our implementation but imports from said package rather than from `vue` itself.

This is currently a work in progress and is subject to change, the `vue` api for now feels very lenient to our subscription model.

TODO:

- [x] useQuery
- [x] useMutation
- [x] useSubscription
- [x] example
- [x] find out why TS bugs out while the imports are def available [docs](https://v3.vuejs.org/guide/composition-api-provide-inject.html#using-inject) thanks @kitten 
- [x] thennable on `useQuery` result for `vue suspense`.

Example: https://codesandbox.io/s/wispy-glade-n813p?file=/src/components/HelloWorld.vue


packages/vue-urql/src/useQuery.ts
```diff
@@ -0,0 +1,127 @@
...
+        result.value.stale = !!stale;
+        result.value.error = error;
+        result.value.fetching = false;
+        result.value.extensions = extensions;
+        result.value.operation = operation;
+      })
+    ).unsubscribe;
+  });
+
+  if (!args.pause) {
```
kitten, line 80: Is the `args.pause` if here correct? If this is initialised with `pause = true` and I call `resume()` won't this watcher be inactive?

packages/vue-urql/src/useQuery.ts
```diff
@@ -0,0 +1,127 @@
...
+      onPush(({ stale, data, error, extensions, operation }) => {
+        result.value.data = data;
+        result.value.stale = !!stale;
+        result.value.error = error;
+        result.value.fetching = false;
+        result.value.extensions = extensions;
+        result.value.operation = operation;
+      }),
+      publish
+    ).unsubscribe();
```
kitten, line 110: We should maybe just use `readQuery` here and abstract the assignment of `OperationResult` for reuse into a function?

packages/vue-urql/src/useQuery.ts
JoviDeCroock: Yeah, I should change to isPaused ref

packages/vue-urql/example/src/components/HelloWorld.vue
```diff
@@ -0,0 +1,36 @@
...
+<script>
+import { useQuery } from '@urql/vue';
+
+const query = `
+  query {
+    Country {
+      name
+    }
+  }
+`;
```
andyrichardson, line 18: This is declared twice

andyrichardson (approved):
🔥

packages/vue-urql/src/useQuery.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  const req = createRequest(args.query, args.variables || {});
+  if (req.key !== request.value.key) request.value = req;
+
+  let fetchOnMount = true;
+  onMounted(() => {
+    // Checks for synchronous data in the cache.
+    pipe(
+      client.executeQuery(request.value as GraphQLRequest),
+      onPush(res => {
```
kitten, line 110: That's superfluous over `subscribe`

packages/vue-urql/src/useQuery.ts
```diff
@@ -0,0 +1,177 @@
...
+    fetching,
+    executeQuery,
+    isPaused,
+    pause,
+    resume,
+  };
+
+  return {
+    ...result,
+    async then() {
```
kitten, line 160: Let's double check whether this transpiles well since writing it by hand is also petty easy 😅

packages/vue-urql/src/useQuery.ts
```diff
@@ -0,0 +1,177 @@
...
+  extensions: Ref<Record<string, any> | undefined>;
+  operation: Ref<Operation | undefined>;
+  resume: () => void;
+  pause: () => void;
+  isPaused: Ref<boolean>;
+  executeQuery: () => void;
+}
+
+export type UseQueryResponse<T> = UseQueryState<T> & {
+  then: () => Promise<UseQueryState<T>>;
```
kitten, line 39: Isn't there a `PromiseLike` built-in in TypeScript so that async/await and all other stuff works correctly?

packages/vue-urql/src/useQuery.ts
```diff
@@ -0,0 +1,177 @@
...
+    pause,
+    resume,
+  };
+
+  return {
+    ...result,
+    async then() {
+      fetchOnMount = false;
+
+      const res = await client
```
kitten, line 163: I think what's interesting here is the question of whether this will start the subscription anyway? If there's a way to clean up on in mount this could be interesting as like React Suspense it will delivery updates results.

Otherwise we'll have to explain people the difference pretty often

packages/vue-urql/src/useQuery.ts
JoviDeCroock: Subscribe had issues when it came down to TypeScript I'll double check

packages/vue-urql/src/useQuery.ts
JoviDeCroock: Hmm, I'll try to check it out, so we would then resolve once and let the stream handle updated results? So `resolve(result); executeQuery();`

kitten (approved):

PR #1030, opened 2020-10-03
Fix Windows overflow-x style

frederikhors:
## Summary

On Windows I can see this:

image

## Set of changes

I think it is better to use `auto`, see:

image


kitten (approved):
Cheers!

PR #1045, opened 2020-10-06
(core) - Deprecate the "operationName" property and move to "kind"

### LINKED ISSUE #1005: RFC: Deprecate `operationName` in favour of `kind`
## Summary

Wording is tricky, and there's currently some confusion between the following two

Screenshot 2020-09-28 at 14 55 32

### A - Operation type/kind

A finite set of operation types (query, mutation, subscription)

Currently this is indicated in `urql` on `Operation.operationName`. It's widely used in the `Client`'a and `Exchange`s' control flows and is confusing since it implies to refer to `OperationDefinitionNode.name.value`, which is ambiguous.

```gql
query GetUser # <- `query` is the kind
```

### B - Operation name

One or more **user provided** names of an operation.

While this is an optional name on the `OperationDefinitionNode` and not actively observed or used in `urql` it is sent sometimes by the `fetchExchange` and `subscriptionExchange` for instance. This is commonly what's also referred to as the "Operation Name."

```gql
query GetUser # <- `GetUser` is the name
```

> Note: A `DocumentNode` on `urql`'s `Operation` may contain multiple `OperationDefinitionNode`s (i.e. multiple queries) with multiple names but that's not supported by convention and it's encouraged, as in other clients, to only pass one, hence an operation will only have one possible name or none.

### Existing naming

Deprecate the [`operationName` property](https://github.com/FormidableLabs/urql/blob/main/packages/core/src/types.ts#L57) and introduce a `kind` property. The deprecated property would be removed in a subsequent major update but should for the time being issue a warning (as a getter) for the next minor bump in development and at most once.

<details>
  <summary>Ruled out</summary>

#### Option 2

Do option 1 and also rename the Operation interfaces ([Operation](https://github.com/FormidableLabs/urql/blob/main/packages/core/src/types.ts#L56), [OperationContext](https://github.com/FormidableLabs/urql/blob/main/packages/core/src/types.ts#L42), [OperationDebugMeta](https://github.com/FormidableLabs/urql/blob/main/packages/core/src/types.ts#L34) to use the term `Request`/`UrqlRequest` to prevent confusion with the Operation node type in GraphQL (of which, an Urql Operation can have multiple).

Re-introduce the legacy naming scheme as a deprecated alias and subsequently remove in a future major release.

```ts
/* @deprecated use `Request` instead */
export type Operation = Request;
/* @deprecated use `RequestContext` instead */
export type OperationContext = RequestContext;
```

> Note: This would take substantially more work including updates to documentation~

</details>

### Adding support for operation name

Expose [the `getOperationName` utility](https://github.com/FormidableLabs/urql/blob/372b0e3fa4da90d4acdfa4e53db068986e36f6af/packages/core/src/internal/fetchOptions.ts#L13) on `@urql/Core` by moving it, to allow users to get the operation name of a GraphQL `DocumentNode` more easily.

<details>
  <summary>Ruled out</summary>

#### Option 2

Add a `operationNames` property to the existing [Operation type](https://github.com/FormidableLabs/urql/blob/main/packages/core/src/types.ts#L58)

</details>

bkonkle:
Resolves #1005

## Summary

It's currently tricky to differentiate between the urql-specific `Operation.operationName` property, and the more general `OperationDefinitionNode.name.value` property. This PR deprecates the "operationName" property in favor of "kind", issues a warning the first time the deprecated property is accessed, and updates all exchanges.

## Set of changes

* Added `packages/core/src/utils/deprecation.ts` - a utility for issuing deprecation warnings only once.
* Added `packages/core/src/utils/operation.ts` with `makeOperation` - a utility for creating Operation objects with a getter/setter for the `operationName` property that issues a deprecation warning.
* Updated `packages/core/src/client.ts` to use `makeOperation` and move property references from "operationName" to "kind".
* Updated the core `cache`, `dedup`, `fallback`, `fetch`, `ssr`, and `subscription` exchanges to move references from "operationName" to "kind".
* Updated references in the `exchanges` packages.

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -54,8 +54,11 @@ export interface OperationContext {
 
 /** A [query]{@link Query} or [mutation]{@link Mutation} with additional metadata for use during transmission. */
 export interface Operation extends GraphQLRequest {
-  operationName: OperationType;
+  kind: OperationType;
   context: OperationContext;
+
+  /** @deprecated use Operation.kind instead */
+  operationName: OperationType;
 }
 
 /** Resulting data from an [operation]{@link Operation}. */
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -54,11 +54,11 @@ export interface OperationContext {
 
 /** A [query]{@link Query} or [mutation]{@link Mutation} with additional metadata for use during transmission. */
 export interface Operation extends GraphQLRequest {
-  kind: OperationType;
+  readonly kind: OperationType;
   context: OperationContext;
 
   /** @deprecated use Operation.kind instead */
-  operationName: OperationType;
+  readonly operationName: OperationType;
 }
 
 /** Resulting data from an [operation]{@link Operation}. */
--- a/packages/core/src/utils/operation.ts
+++ b/packages/core/src/utils/operation.ts
@@ -17,16 +17,12 @@ const DEPRECATED: Record<string, Warning> = {
 
 export const makeOperation = (
   kind: OperationType,
-  {
-    key,
-    query,
-    variables,
-  }: Pick<GraphQLRequest, 'key' | 'query' | 'variables'>,
+  request: Pick<GraphQLRequest, 'key' | 'query' | 'variables'>,
   context: OperationContext
 ): Operation => ({
-  key,
-  query,
-  variables,
+  key: request.key,
+  query: request.query,
+  variables: request.variables,
   kind,
   context,
 
--- a/packages/core/src/utils/operation.ts
+++ b/packages/core/src/utils/operation.ts
@@ -1,0 +1,44 @@
+import {
+  GraphQLRequest,
+  Operation,
+  OperationContext,
+  OperationType,
+} from '../types';
+import { deprecationWarning } from './deprecation';
+
+// TODO: Remove when the deprecated `operationName` property is removed
+const DEPRECATED = {
+  operationName: {
+    key: 'Operation.operationName',
+    message:
+      'The "Operation.operationName" property has been deprecated and will be removed in a future release of urql. Use "Operation.kind" instead.',
+  },
+};
+
+export const makeOperation = (
+  kind: OperationType,
+  {
+    key,
+    query,
+    variables,
+  }: Pick<GraphQLRequest, 'key' | 'query' | 'variables'>,
+  context: OperationContext
+): Operation => ({
+  key,
+  query,
+  variables,
+  kind,
+  context,
+
+  get operationName(): OperationType {
+    deprecationWarning(DEPRECATED.operationName);
+
+    return this.kind;
+  },
+
+  set operationName(kind: OperationType) {
+    deprecationWarning(DEPRECATED.operationName);
+
+    this.kind = kind;
+  },
+});
--- a/packages/core/src/utils/operation.ts
+++ b/packages/core/src/utils/operation.ts
@@ -17,7 +17,7 @@ const DEPRECATED: Record<string, Warning> = {
 
 export const makeOperation = (
   kind: OperationType,
-  request: Pick<GraphQLRequest, 'key' | 'query' | 'variables'>,
+  request: GraphQLRequest,
   context: OperationContext
 ): Operation => ({
   key: request.key,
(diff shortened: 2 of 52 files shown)


packages/core/src/utils/operation.ts
```diff
@@ -0,0 +1,44 @@
...
+    deprecationWarning(DEPRECATED.operationName);
+
+    return this.kind;
+  },
+
+  set operationName(kind: OperationType) {
+    deprecationWarning(DEPRECATED.operationName);
+
+    this.kind = kind;
+  },
```
kitten, line 43: I'd just drop this, as setting isn't ever expected; we have a couple of edge cases of course that we went through (e.g. spreading) but even then `operationName` is pretty much readonly. I think this also hints at the fact that we should update the `Operation` type to make `kind` competely `readonly`

packages/core/src/utils/operation.ts
```diff
@@ -0,0 +1,44 @@
...
+    query,
+    variables,
+  }: Pick<GraphQLRequest, 'key' | 'query' | 'variables'>,
+  context: OperationContext
+): Operation => ({
+  key,
+  query,
+  variables,
+  kind,
+  context,
```
kitten, line 31:
```suggestion
  kind: OperationType,
  request: GraphQLRequest,
  context: OperationContext
): Operation => ({
  kind,
  key: request.key,
  query: request.query,
  variables: request.variables,
  context,
```

Rearranging here; also putting `kind` first for debugging. I really like this pattern :+1: pretty unintrusive

packages/core/src/types.ts
```diff
@@ -54,8 +54,11 @@ export interface OperationContext {
 
 /** A [query]{@link Query} or [mutation]{@link Mutation} with additional metadata for use during transmission. */
 export interface Operation extends GraphQLRequest {
-  operationName: OperationType;
+  kind: OperationType;
```
kitten, line 57: Let's add `readonly` here to document our contract of not modifying this

packages/core/src/utils/operation.ts
bkonkle: Thanks! I used the `Pick<>` with `GraphQLRequest` so that I could also pass an existing Operation there that I want to copy. For example, if you have an existing query Operation that you want to issue a 'teardown' for:

```ts
makeOperation('teardown', query, query.context)
```

bkonkle: I can definitely see how it's more readable without the destructuring, though. I'll switch that back. :+1:

kitten: I think the `Pick` isn't necessary though there, right? An `Operation` already extends `GraphQLRequest` so it should also be accepted there without `Pick` as interfaces in TS are open types

packages/core/src/utils/operation.ts
```diff
@@ -0,0 +1,34 @@
...
+  context: OperationContext
+): Operation => ({
+  key: request.key,
+  query: request.query,
+  variables: request.variables,
+  kind,
+  context,
+
+  get operationName(): OperationType {
+    deprecationWarning(DEPRECATED.operationName);
```
kitten, line 30: Last thing, I swear 😅 Let's wrap this in `if (process.env.NODE_ENV !== 'production')`

We can then check in the output `min.js` bundles whether all warnings successfully compile away in our output

packages/core/src/utils/operation.ts
bkonkle: Oh, good point!

packages/core/src/utils/operation.ts
bkonkle: Should I wrap it here, or inside the `deprecationWarning` util function itself? The question is essentially "one place, or many places"?

bkonkle: Nice! So - `(name === 'warn' || name === 'deprecationWarning')` then?

JoviDeCroock (approved):
Awesome work Brandon!

kitten (approved):

kitten:
Let's figure out when we want to release this 🤔 I don't expect anyone to have any major issues with this change, since it won't be breaking unless someone does something absolutely weird, but it's worth double checking.

kitten:
Update: I loaded the sandbox as a quick smoke test and it all seems to work fine but it _is_ triggering the warning, which is odd. https://codesandbox.io/s/urql-issue-template-client-forked-nsrde

PR #1048, opened 2020-10-08
(docs) - Update Comparison page with relay-tools related projects

morrys:
In this PR I have added some of the libraries that I have implemented and currently supported:
 * [relay-hooks](https://github.com/relay-tools/relay-hooks): hooks for Relay for both concurrent and legacy modes
 * [react-relay-offline](https://github.com/morrys/react-relay-offline): it's a Relay extension for offline capabilities
 * [relay-angular](https://github.com/morrys/react-relay-offline): Angular Bindings including offline capabilities
 * [wora/apollo-offline](https://morrys.github.io/wora/docs/apollo-offline) is an Apollo extension for offline functionality

### DIFF
--- a/docs/comparison.md
+++ b/docs/comparison.md
@@ -81,7 +81,7 @@ hoisting all necessary data requirements to a page-wide query.
 |                               | urql                             | Apollo             | Relay             |
 | ----------------------------- | -------------------------------- | ------------------ | ----------------- |
 | React Bindings                | ✅                               | ✅                 | ✅                |
-| React Concurrent Hooks Support| 🛑                               | ✅                 | ✅ (experimental) |
+| React Concurrent Hooks Support| ✅                               | 🛑                 | ✅ (experimental) |
 | React Legacy Hooks Support    | ✅                               | ✅                 | 🟡 `relay-hooks`  |
 | React Suspense (Experimental) | ✅ (experimental on client-side) | 🛑                 | ✅                |
 | Next.js Integration           | ✅ `next-urql`                   | 🟡                 | 🔶                |
--- a/docs/comparison.md
+++ b/docs/comparison.md
@@ -81,12 +81,13 @@ hoisting all necessary data requirements to a page-wide query.
 |                               | urql                             | Apollo             | Relay             |
 | ----------------------------- | -------------------------------- | ------------------ | ----------------- |
 | React Bindings                | ✅                               | ✅                 | ✅                |
-| React Hooks Support           | ✅                               | ✅                 | ✅ (experimental) |
+| React Hooks Support           | ✅                               | ✅                 | ✅ (experimental), 🟡 `relay-hooks` |
 | React Suspense (Experimental) | ✅ (experimental on client-side) | 🛑                 | ✅                |
 | Next.js Integration           | ✅ `next-urql`                   | 🟡                 | 🔶                |
 | Preact Support                | ✅                               | 🔶                 | 🔶                |
-| Svelte Bindings               | ✅                               | 🟡 `svelte-apollo` | 🛑                |
+| Svelte Bindings               | ✅                               | 🟡 `svelte-apollo` | 🟡 `svelte-relay` |
 | Vue Bindings                  | 🛑 (planned)                     | 🟡 `vue-apollo`    | 🟡 `vue-relay`    |
+| Angular Bindings              | 🛑                               | 🟡 `apollo-angular`| 🟡 `relay-angular`|
 | Initial Data on mount         | ✅                               | ✅                 | ✅                |
 
 Interestingly all three libraries heavily support React as they were all started from the React
@@ -117,7 +118,7 @@ for instance.
 | Commutativity Guarantees                                | ✅                                                                    | 🛑                 | 🛑                                             |
 | Partial Results                                         | ✅                                                                    | ✅                 | 🛑                                             |
 | Safe Partial Results (schema-based)                     | ✅                                                                    | 🛑                 | 🛑                                             |
-| Offline Support                                         | ✅                                                                    | 🛑                 | 🛑                                             |
+| Offline Support                                         | ✅                                                                    | 🟡 `@wora/apollo-offline`| 🟡 `react-relay-offline` `relay-angular`|
 
 `urql` is the only of the three clients that doesn't pick [normalized
 caching](./graphcache/normalized-caching.md) as its default caching strategy. Typically this is seen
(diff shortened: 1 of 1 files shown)


sibelius (approved):
nice

docs/comparison.md
```diff
@@ -81,12 +81,13 @@ hoisting all necessary data requirements to a page-wide query.
 |                               | urql                             | Apollo             | Relay             |
 | ----------------------------- | -------------------------------- | ------------------ | ----------------- |
 | React Bindings                | ✅                               | ✅                 | ✅                |
-| React Hooks Support           | ✅                               | ✅                 | ✅ (experimental) |
+| React Hooks Support           | ✅                               | ✅                 | ✅ (experimental), 🟡 `relay-hooks` |
```
JoviDeCroock, line 84: I'd say to let the first-class integration like it is

docs/comparison.md
```diff
@@ -117,7 +118,7 @@ for instance.
 | Commutativity Guarantees                                | ✅                                                                    | 🛑                 | 🛑                                             |
 | Partial Results                                         | ✅                                                                    | ✅                 | 🛑                                             |
 | Safe Partial Results (schema-based)                     | ✅                                                                    | 🛑                 | 🛑                                             |
-| Offline Support                                         | ✅                                                                    | 🛑                 | 🛑                                             |
+| Offline Support                                         | ✅                                                                    | 🟡 `@wora/apollo-offline`| 🟡 `react-relay-offline` `relay-angular`|
```
JoviDeCroock, line 121: Why is angular mentioned here as well

morrys: Because the relay-angular library also supports offline

docs/comparison.md
morrys: Exactly but it is marked as experimental. While relay-hooks is not experimental and supports both react concurrent and legacy modes

kitten:
I'm not really comfortable here to right away merge this; there are some things we could correct here and it'd be formally correct, but at the end of the day we made one restriction to third-party libs:

> 🟡 Supported, but as an unofficial 3rd-party library. (Provided it's commonly used)

Ultimately, we know that all these missing features are solvable given the architecture of Relay, even if a feature is missing. There are several solutions for all of these features, different approaches, and a user may even implement what they need themselves, e.g. custom-tailored offline support in just their app.

But everything we link here is kind of an endorsement of tools that we know exist and are circulated solutions, so either they'll have been endorsed by the teams themselves or are defacto standards. This is why the explanation adds "Provided it's commonly used."

Now, this isn't a personal attack to you or anyone who makes these libraries, but we don't have an active link to the Relay team or the Relay community, so we can't go ahead and merge tools that are written by you personally immediately without any metrics of usage demonstrating that it's widely used 😅 

I know this isn't ideal, but we do have to set some limits first to be very careful about the mentions to other projects we add.
I'll reach out to some people who know more about Relay and will see what we can come up with

docs/comparison.md
kitten: We sort and filter these comparison entries in order of priority, so if there's a first party solution, nothing else will be added,
If there's a custom entry in the docs no third-party solution will be added, if there's a third party solution that's commonly used or in third party docs, that's added, and so on.

That's because we can't add every possible note here regarding _how_ somethings achieved

morrys: Two entries can be created:

|                                            | urql                                | Apollo                                        | Relay                          |
| ------------------------------------------ | ----------------------------------- | --------------------------------------------- | ------------------------------ |
| React concurrent Hooks Support     | 🛑                               | 🛑                 | ✅ (experimental) |
| React Legacy Hooks Support           | ✅                               | ✅                 |  🟡 `relay-hooks` |

kitten:
Okay, I just talked to some people and they'd say the closest thing to an official endorsement is `relay-tools`. Since you're involved in that let's just move on and get these changes reviewed and merged 😅

I do have to say that everything around Relay is a little confusing

docs/comparison.md
kitten: That's a good option 👍 let's do that. `urql` does support concurrent mode but I'm not 100% sure about Apollo. Let's just assume for now that they do

docs/comparison.md
kitten: Let's remove `relay-angular` from here since we have the other option already, since you already added a Relay option 😅

kitten: Also, I'd love to wait for the Apollo offline entry and get that removed first. We've spoken to the contributor before (and you of course! 😅) on that other issue, but I've sent this document over to some people at Apollo and would like them to decide whether it should be added here or not

morrys: You did well :)

I added it because there are third party libraries that enable Apollo to support offline

docs/comparison.md
```diff
@@ -81,12 +81,14 @@ hoisting all necessary data requirements to a page-wide query.
 |                               | urql                             | Apollo             | Relay             |
 | ----------------------------- | -------------------------------- | ------------------ | ----------------- |
 | React Bindings                | ✅                               | ✅                 | ✅                |
-| React Hooks Support           | ✅                               | ✅                 | ✅ (experimental) |
+| React Concurrent Hooks Support| 🛑                               | ✅                 | ✅ (experimental) |
```
kitten, line 84:
```suggestion
| React Concurrent Hooks Support| ✅                               | ✅                 | ✅ (experimental) |
```

Just as stated in the other note; concurrent mode and suspense is supported in `urql`, so I'm more worried that the Apollo entry here is inaccurate, but I'll figure that one out later :)

morrys:
Hi @kitten,
I understand your perplexities but every project if not followed carefully makes you confused, not just Relay :)
I certainly can't convince you that my libraries are the best (don't judge only by the number of downloads) but it will always be possible to update this comparison when the features are natively supported or implemented by more used libraries 👍 

I just updated the PR to include your suggestions.

kitten:
@morrys Totally agreed! Apologies from my end for the misunderstanding. I didn't see that you worked on `relay-tools` and before talking to some others I didn't know that `relay-tools` holds a large significance in the community in the absence of `facebook/relay` supporting these extra use-cases. 

Just added one more comment regarding the "React Concurrent Hooks Support" entry for `urql` and then we'll be good to go

docs/comparison.md
morrys: I will wait for your feedback and then I will update the pr 👍

morrys: So I update that urql supports react concurrent hooks and apollo doesn't?

kitten: I've just added the above suggestion, since I can directly add that to the batch 😅

kitten (approved):
Thanks! :tada: ❤️

morrys:
Thanks to you too 👍

PR #1052, opened 2020-10-08
Update Relay supported features

kesne:
## Summary

This updates a few Relay features to give them green checkmarks, which I believe is a more accurate representation of the Relay client. My justification for this is that Relay Networking layer is entirely user-provided. Because of this, I'd consider anything to touch networking (file uploads, switching to GET, and batched queries) to be technically supported. File uploads are also supported by `react-relay-network-layer`, so I added it there.

## Set of changes

Changing the emojis used in the comparison.


kitten:
I think I'll have to unfortunately reject this 😅 Sorry, but we did lay out criteria for this in the legend:

- ✅ Supported 1st-class and documented.
- 🔶 Supported and documented, but requires custom user-code to implement.
- 🟡 Supported, but as an unofficial 3rd-party library. (Provided it's commonly used)

If there were documented instructions how to set these things up we'd go with the second marker, and the third is when 3rd-party libraries support this and the docs don't mention this at all.

I know that this isn't reflective of what you _can_ do with all three libraries; but in fact if we were talking about what you could implement yourself on top of these libraries then everything & anything would have a green checkmark 😅

PR #866, opened 2020-06-14
(graphcache) - Reimplement default-storage serializer/deserializer

kitten:
Resolve #864 

## Summary

This fixes an issue where deleted values wouldn't be persisted correctly, since the `default-storage` serializer coerced both `null` and `undefined` to `"null"` JSON values. The `@urql/exchange/default-storage` implementation now comes with a new serializer which is incompatible with the old one but should be faster and more correct. It's incompatible with the old one, so the default `idbName` has been bumped to `graphcache-v4`.

## Set of changes

- Prevent optimistic layers from queueing up unnecessary keys onto the `data.persist` set
- Implement a new serializer and deserializer on `default-storage`

PR #750, opened 2020-04-24
(graphcache) - Implement RFC for optimistic behavior enhancements

kitten:
## Summary

This PR implements both recommendations from RFC #747.

These changes have been validated and tested so far against @frederikhors' unexpected behaviour reproductions in CodeSandboxes:

- https://codesandbox.io/s/urql-svelte-cacheinvalidate-many-items-j89ru
- https://codesandbox.io/s/urql-svelte-add-mutation-back-to-list-t98y7

## Set of changes

- Convert dependencies data structure from `Set<string>` to `type Dependencies = Record<string, true>`
- Keep track of optimistically updated dependencies (of type `Dependencies`)
- Block refetching of operations that have active optimistic updates and keep track of them (see above)
- Allow `executePendingOperations` to trigger refetch for previously blocked `cache-and-network` operations
- Keep track of optimistic mutation that are pending at the same time, and flush their results simultaneously, also erasing the optimistically updated (i.e. blocked) dependencies at the same time

PR #1074, opened 2020-10-20
(graphcache) - Reenable cumulative optimistic updates

kitten:
This was discovered in this Discussions thread: https://github.com/FormidableLabs/urql/discussions/1073

## Summary

In #866 we previously disallowed optimistic updates to read other optimistically written data, which meant to prevent conflicts between these data layers that will later trickle down into permanent layers. This isn't a concern anymore because optimistic layers will never contribute to concrete, non-optimistic data and are thrown away, if they're for optimistic mutations.

The edge case we aimed to prevent is also not possible anymore because we now never apply concrete updates on top of optimistic ones since #750, namely:

> - Block refetching of operations that have active optimistic updates and keep track of them (see above)
> - Keep track of optimistic mutation that are pending at the same time, and flush their results simultaneously, also erasing the optimistically updated (i.e. blocked) dependencies at the same time

I'm not 100% sure anymore at this time, why we deemed this unsafe before, but it shouldn't be.

## Set of changes

- Allow optimistic updates to see other optimistically written data
- Rename `currentIgnoreOptimistic` to just `currentOptimistic`
- Fix `getNode` to allow `currentOptimistic` to remain `true` so that `updatePersist` doesn't queue up noop updates for optimistic mutations

### DIFF
--- a/.changeset/ten-turkeys-appear.md
+++ b/.changeset/ten-turkeys-appear.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-graphcache': patch
+---
+
+Fix optimistic updates not being allowed to be cumulative and apply on top of each other. Previously in [#866](https://github.com/FormidableLabs/urql/pull/866) we explicitly deemed this as unsafe which isn't correct anymore given that concrete, non-optimistic updates are now never applied on top of optimistic layers.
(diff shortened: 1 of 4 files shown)


JoviDeCroock (approved):
Great find!

PR #1075, opened 2020-10-20
(auth) - Fix duplicate operation on willAuthError

kitten:
See: https://github.com/FormidableLabs/urql/discussions/1070

## Summary

Previously the `willAuthError` method would trigger the operation that triggered it twice. First for the active operation `mergeMap` function, and second for the `retryQueue` which is caused by `refreshAuth(operation)`.


JoviDeCroock (approved):
Nice find

kadikraman (approved):
🚫 🐞

PR #1077, opened 2020-10-21
(refocus) - Switch to visibilitychange api

### LINKED ISSUE #1063: RFC: Update refocusExchange to utilize Page Visibility API
See: https://github.com/FormidableLabs/urql/discussions/1055

## Summary

There are known issues that occur with the `focus` event:

- `iframe` can cause the `focus` event to fire numerous times if the user interacts with the `iframe` and then back with the current document
- DevTools can cause the `focus` event to fire numerous times if the user/developer interacts with the DevTools and then back with the document. (A very common case for developers)
- A developer fires a `focus` event (ie. `node.dispatchEvent(new CustomEvent('focus', { bubbles: true}))`). This case seems less likely but still a side-effect of the current implementation.

We may also want to check what other libraries, like `swr`, are doing.

## Proposed Solution

- Check whether the Page Visiblity API can help with this
- Reimplement `refocusExchange` with the new API
- Preserve IE11 support with fallbacks, polyfills, or workarounds (or an alternative approach entirely?)

tatchi:
Resolves #1063 

## Summary

Use the `visibilitychange` api.

## Set of changes

* Switch from `focus` to `visibilitychange` api

I looked at [SWR](https://github.com/vercel/swr/blob/4d6be3df441f925d235a4b26914efcc85ae335a3/src/use-swr.ts#L61) and [react-query](https://github.com/tannerlinsley/react-query/blob/master/src/core/setFocusHandler.ts#L14) code and noticed that they refetch on both `focus` and `visibilitychange` events. As you can see in the video below, `visibilitychange` does not fire when we switch from one window to another (while `focus` does). It actually makes sense since the visibility of the document does not change. According to #1063, `focus` has some issues so I guess it's fine to only refetch on `visibilitychange`?

ezgif com-gif-maker (1)

### DIFF
--- a/.changeset/dirty-dragons-try.md
+++ b/.changeset/dirty-dragons-try.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-refocus': minor
+---
+
+Switch from focus to visibilitychange api
--- a/.changeset/dirty-dragons-try.md
+++ b/.changeset/dirty-dragons-try.md
@@ -2,4 +2,4 @@
 '@urql/exchange-refocus': minor
 ---
 
-Switch from focus to visibilitychange api
+Switch from a `focus-event` triggering the refetch to a change in [`page-visbility`](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API). This means that interacting with an `iframe` and then going back to the page won't trigger a refetch, interacting with Devtools won't cause refetches and a bubbled `focusEvent` won't trigger a refetch.
--- a/exchanges/refocus/src/refocusExchange.ts
+++ b/exchanges/refocus/src/refocusExchange.ts
@@ -1,12 +1,36 @@
 import { pipe, tap } from 'wonka';
 import { Exchange, Operation } from '@urql/core';
 
+// Source: https://github.com/vercel/swr/blob/4d6be3df441f925d235a4b26914efcc85ae335a3/src/libs/is-document-visible.ts
+const isDocumentVisible = (): boolean => {
+  if (
+    typeof document !== 'undefined' &&
+    typeof document.visibilityState !== 'undefined'
+  ) {
+    return document.visibilityState !== 'hidden';
+  }
+  // always assume it's visible
+  return true;
+};
+
+// Source: https://github.com/vercel/swr/blob/4d6be3df441f925d235a4b26914efcc85ae335a3/src/libs/is-online.ts
+const isOnline = (): boolean => {
+  if (typeof navigator.onLine !== 'undefined') {
+    return navigator.onLine;
+  }
+  // always assume it's online
+  return true;
+};
+
 export const refocusExchange = (): Exchange => {
   return ({ client, forward }) => ops$ => {
     const watchedOperations = new Map<number, Operation>();
     const observedOperations = new Map<number, number>();
 
-    window.addEventListener('focus', () => {
+    window.addEventListener('visibilitychange', () => {
+      if (!isDocumentVisible() || !isOnline()) {
+        return;
+      }
       watchedOperations.forEach(op => {
         client.reexecuteOperation(
           client.createRequestOperation('query', op, {
--- a/exchanges/refocus/src/refocusExchange.ts
+++ b/exchanges/refocus/src/refocusExchange.ts
@@ -1,43 +1,24 @@
 import { pipe, tap } from 'wonka';
 import { Exchange, Operation } from '@urql/core';
 
-// Source: https://github.com/vercel/swr/blob/4d6be3df441f925d235a4b26914efcc85ae335a3/src/libs/is-document-visible.ts
-const isDocumentVisible = (): boolean => {
-  if (
-    typeof document !== 'undefined' &&
-    typeof document.visibilityState !== 'undefined'
-  ) {
-    return document.visibilityState !== 'hidden';
-  }
-  // always assume it's visible
-  return true;
-};
-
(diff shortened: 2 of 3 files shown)


exchanges/refocus/src/refocusExchange.ts
```diff
@@ -1,12 +1,36 @@
...
+};
+
+// Source: https://github.com/vercel/swr/blob/4d6be3df441f925d235a4b26914efcc85ae335a3/src/libs/is-online.ts
+const isOnline = (): boolean => {
+  if (typeof navigator.onLine !== 'undefined') {
+    return navigator.onLine;
+  }
+  // always assume it's online
+  return true;
+};
```
kitten, line 23: This check should be removed as we have different systems that care about offline; so generally, if we're offline, we should still issue the reexecution and let other exchanges deal with offline/online behaviour so this remains super predictable.

exchanges/refocus/src/refocusExchange.ts
```diff
@@ -1,12 +1,36 @@
...
+
 export const refocusExchange = (): Exchange => {
   return ({ client, forward }) => ops$ => {
     const watchedOperations = new Map<number, Operation>();
     const observedOperations = new Map<number, number>();
 
-    window.addEventListener('focus', () => {
+    window.addEventListener('visibilitychange', () => {
+      if (!isDocumentVisible() || !isOnline()) {
```
kitten, line 31: Let's just inline `isDocumentVisible`. There's also not really any reason to enter the source, since it's a common API with just a one liner, so this can just become:

```suggestion
      if (typeof document !== 'undefined' && document.visibilityState === 'hidden') {
```

tatchi: > Thanks for picking this up! First on a small side-note, we document our changes for our automatically generated changelogs using "changeset" (that's what the comment above is about). Just check these instructions here on how to document a change here and probably mark this one as "minor": https://github.com/FormidableLabs/urql/blob/main/CONTRIBUTING.md#how-do-i-document-a-change-for-the-changelog

Sorry, I should have read the contributing doc before submitting my PR 😅

exchanges/refocus/src/refocusExchange.ts
tatchi: Oh, I didn't know that there are other exchanges already dealing with the offline/online. Good to know 👍

tatchi:
Thanks for picking this up! First on a small side-note, we document our changes for our automatically generated changelogs using "changeset" (that's what the comment above is about). Just check these instructions here on how to document a change here and probably mark this one as "minor": https://github.com/FormidableLabs/urql/blob/main/CONTRIBUTING.md#how-do-i-document-a-change-for-the-changelog

Sorry, I should have read the contributing doc before submitting my PR 😅

exchanges/refocus/src/refocusExchange.ts
tatchi: Shouldn't it be: `document.visibilityState === 'visible'` (and not 'hidden') ?

exchanges/refocus/src/refocusExchange.ts
kitten: Specifically the `offlineExchange` as part of Graphcache adds a lot of offline support. But I actually just realised why you added this. I would still assume though that it's not this exchange's responsibility to care about offline. Especially since by the time the `refocusExchange` re-issues the operation and the request starts, the device may go offline.

exchanges/refocus/src/refocusExchange.ts
kitten: I don't think so? You made this `return` so it's the reverse, if we're hidden we don't re-issue the operation. Maybe it's more intuitive to just wrap all following statements in this if-statement instead to avoid this confusion (then it'd be `typeof document !== 'object' || document.visibilityState === 'visible'`

tatchi: Oh yes, indeed I forgot I made an early return here. Thanks :)

exchanges/refocus/src/refocusExchange.ts
```diff
@@ -6,14 +6,19 @@ export const refocusExchange = (): Exchange => {
...
-      watchedOperations.forEach(op => {
-        client.reexecuteOperation(
-          client.createRequestOperation('query', op, {
-            requestPolicy: 'cache-and-network',
-          })
-        );
-      });
+    window.addEventListener('visibilitychange', () => {
+      if (
+        typeof document !== 'object' ||
```
JoviDeCroock, line 11: At the moment this will only execute on the server with this check

```suggestion
        typeof window !== 'undefined' ||
```

.changeset/dirty-dragons-try.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/exchange-refocus': minor
+---
+
+Switch from focus to visibilitychange api
```
JoviDeCroock, line 5:
```suggestion
Switch from a `focus-event` triggering the refetch to a change in [`page-visbility`](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API). This means that interacting with an `iframe` and then going back to the page won't trigger a refetch, interacting with Devtools won't cause refetches and a bubbled `focusEvent` won't trigger a refetch.
```

exchanges/refocus/src/refocusExchange.ts
kitten: No the condition is an or 😅 so it says "document isn't an object (server) OR visible"

JoviDeCroock (approved):
Awesome work, thanks for picking this up!

kitten (approved):

PR #1080, opened 2020-10-21
(graphcache) - Clear optimistic updates before retrying in offlineExchange

### LINKED ISSUE #1079: Should custom updates triggered by optimistic mutation be repeated?
Hi! First of all thanks for your support so far 🙏 

As I mentioned in some [previous](https://github.com/FormidableLabs/urql/discussions/1073) [discussions](https://github.com/FormidableLabs/urql/discussions/1068) (repeating here for reference) I'm trying to support a scenario whereby users of a native app would be able to create some "Recommendation" entities while the device is offline.
I'm using `@urql/exchange-graphcache`'s `offlineExchange`, with an optimistic mutation and a custom updater which updates a query which lists recommendations

```ts
offlineExchange({
    storage, // <--- I created this using AsyncStorage
    schema,
    updates: {
        Mutation: {
            createRecommendation: (result, args, cache, info) => {
                const recommendation = .... // builds recommendation from args
                const fieldId = ... // gets field Id from args

                cache.updateQuery({ query: QUERY_RECOMMENDATIONS, variables: { fieldId } }, (data) => {
                    data.recommendations.push(recommendation);
                    return data;
                });
            },
        },
    },
    optimistic: {
        createRecommendation: ... // creates a Recommendation entry with a temporary ID
    }
});
```

Everything is working as expected except _maybe_ one thing which I though I would report.
Suppose the device is offline, and I create a Recommendation. The optimistic mutation runs, followed by the custom updater. After this `QUERY_RECOMMENDATIONS` has these entries

```json
[
  { "id": "temporary_id_001" }
]
```

At some point the device goes back online, so the mutation that was saved as "pending" in the offlineExchange's queue gets re-executed. We again run the optimistic mutation and the custom updater. At this point `QUERY_RECOMMENDATIONS` has these entries

```json
[
  { "id": "temporary_id_001" },
  { "id": "temporary_id_002" }
]
```

The mutation is successful as the device is online, so when a successful response is returned from the API the custom updater runs again. At this point `QUERY_RECOMMENDATIONS` has these entries

```json
[
  { "id": "1144cee7-5d0d-4589-a26c-e1322e429b2d" }
]
```

For a very brief time (between when the second optimistic mutation is applied and when the response come back from the API) the query has two entries for the "createRecommendation" mutation, but the user only ever meant to create one.
What I see on a "Recommendations list" screen is a brief flickering between 1 reco, 2 reco, 1 reco when the device gets back online.

I implemented a workaround whereby I assign a random number to the mutation's payload and when the second optimistic mutation runs I check in the custom updater whether an entry already exists in the cache with that random number. If one exists, I `return` from the updater. This solves the issue with the flickering.

I nevertheless thought I would open this discussion as conceptually I'm not sure we should run the optimistic mutation twice as it would cause a brief inconsistent state unless the mutation is "idempotent". Would it make more sense to run the optimistic migration once when the device is offline, and then when the device gets back online we only apply the custom updater on the actual response from the API?

_Originally posted by @codazzo in https://github.com/FormidableLabs/urql/discussions/1078_

kitten:
Fix #1079

## Summary

> ~**NOTE:** This changes an earlier (but rare) case's behaviour!~
> ~We expect that this changes how a "toggling mutation" may look in the UI. Specifically it will never be able now to actually toggle if we trigger it repeatedly in optimistic updates. That's probably a good thing however as between a bad network connection and going offline/online we can't tell whether all toggling mutations will be delivered and applied in-order and atomically. So it's invalid to prioritise this case over the following fix.~

This fix clears all optimistic updates as a single batch by sending `'teardown'` operations in the `offlineExchange` when the mutations are ready to be retried. This uses the existing functionality that clears layers using `noopDataState` when a `teardown` operation comes in.

This causes all optimistic layers to be erased before optimistic updates are reapplied.

## Set of changes

- Clear all optimistic updates when they're retried by the `offlineExchange` in a whole batch, by sending a `teardown` operaton for each failed mutation.
- Add test case for `noopDataState` in combination with a past and future optimistic update.

### DIFF
--- a/exchanges/graphcache/src/offlineExchange.ts
+++ b/exchanges/graphcache/src/offlineExchange.ts
@@ -73,32 +73,42 @@ export const offlineExchange = (opts: CacheExchangeOpts): Exchange => input => {
     storage.writeMetadata
   ) {
     const { forward: outerForward, client, dispatchDebug } = input;
+    const { source: reboundOps$, next } = makeSubject<Operation>();
     const optimisticMutations = opts.optimistic || {};
     const failedQueue: Operation[] = [];
 
     const updateMetadata = () => {
       const requests: SerializedRequest[] = [];
       for (let i = 0; i < failedQueue.length; i++) {
-        const op = failedQueue[i];
-        if (op.operationName === 'mutation') {
+        const operation = failedQueue[i];
+        if (operation.operationName === 'mutation') {
           requests.push({
-            query: print(op.query),
-            variables: op.variables,
+            query: print(operation.query),
+            variables: operation.variables,
           });
         }
       }
       storage.writeMetadata!(requests);
     };
 
-    let _flushing = false;
+    let isFlushingQueue = false;
     const flushQueue = () => {
-      if (!_flushing) {
-        _flushing = true;
-        let operation: void | Operation;
-        while ((operation = failedQueue.shift()))
-          client.reexecuteOperation(operation);
+      if (!isFlushingQueue) {
+        isFlushingQueue = true;
+
+        for (let i = 0; i < failedQueue.length; i++) {
+          const operation = failedQueue[i];
+          if (operation.operationName === 'mutation') {
+            next({ ...operation, operationName: 'teardown' });
+          }
+        }
+
+        for (let i = 0; i < failedQueue.length; i++)
+          client.reexecuteOperation(failedQueue[i]);
+
+        failedQueue.length = 0;
+        isFlushingQueue = false;
         updateMetadata();
-        _flushing = false;
       }
     };
 
@@ -145,7 +155,6 @@ export const offlineExchange = (opts: CacheExchangeOpts): Exchange => input => {
 
     return ops$ => {
       const sharedOps$ = share(ops$);
-      const { source: reboundOps$, next } = makeSubject<Operation>();
       const opsAndRebound$ = merge([reboundOps$, sharedOps$]);
 
       return pipe(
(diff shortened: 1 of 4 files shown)


kitten:
Tested in: https://github.com/FormidableLabs/urql/issues/1079#issuecomment-713599604

exchanges/graphcache/src/offlineExchange.ts
```diff
@@ -73,32 +73,42 @@ export const offlineExchange = (opts: CacheExchangeOpts): Exchange => input => {
...
+
+        for (let i = 0; i < failedQueue.length; i++) {
+          const operation = failedQueue[i];
+          if (operation.operationName === 'mutation') {
+            next({ ...operation, operationName: 'teardown' });
+          }
+        }
+
+        for (let i = 0; i < failedQueue.length; i++)
+          client.reexecuteOperation(failedQueue[i]);
```
wgolledge, line 107: I think this can be added to the `for` above?

wgolledge (approved):
Nice change! This seems logical to me

exchanges/graphcache/src/offlineExchange.ts
kitten: It's so that it's in order and torn down first so that it's all done before the first optimistic updats is performed :)

wgolledge: Ah cool. Looks great then!

JoviDeCroock (approved):
Great decision on this approach, awesome work!

PR #1094, opened 2020-10-28
(build) - Fix graphql imports for Webpack 5

### LINKED ISSUE #1086: Webpack 5 resolution issues
Please update your package.json correctly for Webapck Resolution in mjs/esm

```
[0] ModuleNotFoundError: Module not found: Error: Can't resolve 'graphql/language/visitor' in '/Volumes/lulu_dev/ecom-web-app/node_modules/@urql/core/dist'
[0] Did you mean 'visitor.mjs'?
[0] BREAKING CHANGE: The request 'graphql/language/visitor' failed to resolve only because it was resolved as fully specified
[0] (probably because the origin is a '*.mjs' file or a '*.js' file where the package.json contains '"type": "module"').
[0] The extension in the request is mandatory for it to be fully specified.
[0] Add the extension to the request.
```

urql version & exchanges: latest

**Steps to reproduce**

build on webpack v5

kitten:
Resolve #1086 

## Summary

Another week another small ESM issue in Webpack 5. In the upgrade Webpack now supports `package.json:exports` maps. However, a small problem with this support is how it handles sub-packages without a `package.json:exports`. There's a small quirk in Webpack that will require all import paths to be exact (including extensions) even when the sub-package doesn't have a `package.json:exports` map.

We're affected by this because we use our `babel-plugin-modular-graphql` plugin to convert some `graphql` imports into more specific ones. But instead of specifying files like `graphql/language/visitor` we need to add the full path like `graphql/language/visitor.mjs`. While I'm not convinced that this is the correct behaviour, it's an easy fix.

## Set of changes

- Upgrade to `babel-plugin-modular-graphql@1.0.0` which adds an `extension` option
- Refactor Rollup build pipeline to move production/non-production plugins to `output.plugins`. This allows us to merge the two build runs into a single one.
- Move `babel-plugin-modular-graphql` to `cleanup()` plugin.

### DIFF
--- a/.changeset/thin-rings-tell.md
+++ b/.changeset/thin-rings-tell.md
@@ -1,4 +1,5 @@
 ---
+'@urql/preact': patch
 '@urql/exchange-execute': patch
 '@urql/exchange-graphcache': patch
 '@urql/exchange-populate': patch
--- a/.changeset/thin-rings-tell.md
+++ b/.changeset/thin-rings-tell.md
@@ -1,0 +1,9 @@
+---
+'@urql/exchange-execute': patch
+'@urql/exchange-graphcache': patch
+'@urql/exchange-populate': patch
+'@urql/core': patch
+'@urql/introspection': patch
+---
+
+Add missing `.mjs` extension to all imports from `graphql` to fix Webpack 5 builds, which require extension-specific import paths for ESM bundles and packages.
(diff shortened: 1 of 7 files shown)


JoviDeCroock (approved):
Looks amazing!

PR #1095, opened 2020-10-28
(next) - Remove explicit dependency on next for next-urql's types

kitten:
This removes the dev-dependency on `next` from `next-urql` by providing vague types around what the `NextContext` and component types look like. It's provided so that the enhancer function still assumes and infers the correct type if possible.

This is to reduce the dependencies in our monorepo and it's been tested against the first example in the `next-urql` folder, which passes a `tsc` run successfully even with the new package.

### DIFF
--- a/packages/next-urql/src/types.ts
+++ b/packages/next-urql/src/types.ts
@@ -1,32 +1,45 @@
 import { GraphQLError } from 'graphql';
-import { NextPageContext } from 'next';
+import { ComponentType } from 'react';
 import { ClientOptions, Exchange, Client } from 'urql';
-import { AppContext } from 'next/app';
+
+export interface PartialNextContext {
+  res?: any;
+  AppTree?: NextComponentType<PartialNextContext>;
+  Component?: NextComponentType<PartialNextContext>;
+  ctx?: PartialNextContext;
+  [key: string]: any;
+}
+
+export type NextComponentType<
+  C extends PartialNextContext = PartialNextContext,
+  IP = {},
+  P = {}
+> = ComponentType<P & WithUrqlProps> & {
+  getInitialProps?(context: C): IP | Promise<IP>;
+};
 
 export type NextUrqlClientConfig = (
   ssrExchange: SSRExchange,
-  ctx?: NextPageContext
+  ctx?: PartialNextContext
 ) => ClientOptions;
 
-export interface NextUrqlPageContext extends NextPageContext {
+export interface NextUrqlContext extends PartialNextContext {
   urqlClient: Client;
 }
 
-export interface NextUrqlAppContext extends AppContext {
-  urqlClient: Client;
-}
-
-export type NextUrqlContext = NextUrqlPageContext | NextUrqlAppContext;
+export type NextUrqlPageContext = NextUrqlContext;
+export type NextUrqlAppContext = NextUrqlContext;
 
 export interface WithUrqlState {
   urqlState?: SSRData;
 }
 
 export interface WithUrqlClient {
-  urqlClient: Client;
+  urqlClient?: Client;
 }
 
 export interface WithUrqlProps extends WithUrqlClient, WithUrqlState {
+  resetUrqlClient?: () => void;
   [key: string]: any;
 }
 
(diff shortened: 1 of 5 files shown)


JoviDeCroock (approved):
Woooooow :tada:

packages/next-urql/src/types.ts
```diff
@@ -1,32 +1,45 @@
...
-import { NextPageContext } from 'next';
+import { ComponentType } from 'react';
 import { ClientOptions, Exchange, Client } from 'urql';
-import { AppContext } from 'next/app';
+
+export interface PartialNextContext {
+  res?: any;
+  AppTree?: NextComponentType<PartialNextContext>;
+  Component?: NextComponentType<PartialNextContext>;
+  ctx?: PartialNextContext;
```
parkerziegler, line 9: Is this to handle the case that `PartialNextContext` could be _either_ `NextPageContext` or `AppContext`? Looks fine to me, just wish there was a good way to limit recursive types in TS. I don't pretend to have enough knowledge of something like conditional types to make this cleaner, so looks good to me!

packages/next-urql/src/with-urql-client.tsx
```diff
@@ -72,27 +71,29 @@ export function withUrqlClient(
...
 
     // Set the displayName to indicate use of withUrqlClient.
-    withUrql.displayName = `withUrqlClient(${getDisplayName(AppOrPage)})`;
+    const displayName = AppOrPage.displayName || AppOrPage.name || 'Component';
+    withUrql.displayName = `withUrqlClient(${displayName})`;
 
     if (AppOrPage.getInitialProps || options!.ssr) {
-      withUrql.getInitialProps = async (appOrPageCtx: NextUrqlContext) => {
-        const { AppTree } = appOrPageCtx;
+      withUrql.getInitialProps = async (_ctx: any) => {
```
parkerziegler, line 88: Does this need to be `any` here? I believe we could just do:

```suggestion
      withUrql.getInitialProps = async (_ctx: PartialNextContext) => {
```

directly unless I'm missing something? It doesn't look like the type cast on L89 is necessary based on the way we access `appOrPageCtx`.

packages/next-urql/src/with-urql-client.tsx
```diff
@@ -72,27 +71,29 @@ export function withUrqlClient(
...
+        const appOrPageCtx = _ctx as PartialNextContext;
+        const AppTree = appOrPageCtx.AppTree!;
```
parkerziegler, line 90: Does this need to be `any` here? I believe we could just do:

```suggestion
      withUrql.getInitialProps = async (_ctx: PartialNextContext) => {
        const appOrPageCtx = _ctx;
        const AppTree = appOrPageCtx.AppTree!;
```

directly unless I'm missing something? It doesn't look like the type cast on L89 is necessary based on the way we access `appOrPageCtx`.

packages/next-urql/src/types.ts
```diff
@@ -1,32 +1,45 @@
...
+export type NextUrqlPageContext = NextUrqlContext;
+export type NextUrqlAppContext = NextUrqlContext;
 
 export interface WithUrqlState {
   urqlState?: SSRData;
 }
 
 export interface WithUrqlClient {
-  urqlClient: Client;
+  urqlClient?: Client;
```
parkerziegler, line 38: Are we just being cautious and making all properties optional, or is `urqlClient` only conditionally available? I can't quite recall 🤔

parkerziegler (approved):
Nice, love to see dependencies go away!

packages/next-urql/src/types.ts
kitten: I think we could also make this a union, but I thought since it's mostly used internally that won't be necessary. Instead the enhancer itself infers that the inout type is the output type 🤔

packages/next-urql/src/with-urql-client.tsx
kitten: That doesn't quite work for a reason I don't understand. There seems to be a conflict when it refers back to this at a later point 😩 but could've been a fluke. I'll try to adapt this again!

packages/next-urql/src/types.ts
kitten: The same type is now used for the input and output. Given that this is passed on and received I found it unsafe to type it as always being there. In case we know it's there (or the user does) it can always easily be forced using `!`

PR #1099, opened 2020-10-29
Update @urql/core to 1.14.1

frederikhors:
Can you please release it ASAP?

Thanks.


kitten:
Our versioning is automatic and this change specifically doesn't apply. `^1.13.1` is a minor range, meaning that `1.14.x` is included. If you need a visualisation of this, you can check this here :+1: https://semver.npmjs.com/

PR #1104, opened 2020-10-30
(core) - Refactor useSource and improve hooks edge case behaviour

kitten:
## Summary

The motivation of this change is to simplify the `useSource` and `useBehaviourSubject` implementations and to remove various additional edge cases from the implementation, like shallow equal updates. In theory it'd be best to also figure out a way to carry over the subscription from the mount phase to the effect phase, which we attempted before but which very deliberately is impossible in React.

This will not change the behaviour of these hooks dramatically but avoid unnecessary state updates when any updates are obviously equivalent and the hook will furthermore improve continuation from mount to effects, which will fix cases where the state between the mounting and effect phase may slightly change.

## Changes

- Combine `useSource` and `useBehaviourSubject` into just `useSource`
- Prevent shallow-equal updates from being triggered by `useSource` when the state is an object
- Improve continuation from mount to effect phase by re-using previous state values
- Fix `Client.onOperationEnd` to remove cancelled operations from its queue

### DIFF
--- a/.changeset/gorgeous-radios-dream.md
+++ b/.changeset/gorgeous-radios-dream.md
@@ -1,0 +1,7 @@
+---
+'@urql/core': patch
+'@urql/preact': patch
+'urql': patch
+---
+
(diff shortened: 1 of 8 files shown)


JoviDeCroock (approved):
This is such a good find!

kitten:
Tested against a previously failing sandbox here: https://codesandbox.io/s/modest-heyrovsky-7bnq7
(note that the same `useQuery` is used in the index page twice)

JoviDeCroock (approved):

PR #1108, opened 2020-11-02
(chore) - Create GH Action for automated releases

JoviDeCroock:
This automates our release process, we're already using changesets but what changed?

- Every merged PR will make a PR from changesets that increments all versions (`yarn changeset version`) and updates changelogs
- When we merge it it will release all the packages to the NPM-registry

TODO: add npm-token to this repo's secrets

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,30 @@
+name: Release
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  release:
+    name: Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@master
+        with:
+          fetch-depth: 0
+      - name: Setup Node.js 12.x
+        uses: actions/setup-node@master
+        with:
+          node-version: 12.x
+      - name: Install Dependencies
+        run: yarn --frozen-lockfile
+      - name: Create PR or Publish to npm
+        id: changesets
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,30 @@
...
+          fetch-depth: 0
+      - name: Setup Node.js 12.x
+        uses: actions/setup-node@master
+        with:
+          node-version: 12.x
+      - name: Install Dependencies
+        run: yarn --frozen-lockfile
+      - name: Create PR or Publish to npm
+        id: changesets
+        uses: changesets/action@master
```
kitten, line 25: Can we pin this so that it cannot be changed underneath our feet so to speak?

.github/workflows/release.yml
```diff
@@ -0,0 +1,30 @@
...
+        with:
+          publish: yarn changeset publish
```
kitten, line 27: I'll double-check what hooks this triggers, if any.

.github/workflows/release.yml
JoviDeCroock: They don't seem to have any tags/releases https://github.com/changesets/action

kitten: I suppose we could try to pin it to the latest commit? not sure if the syntax would be `changeset/action#bfeb9e0` or `changeset/action@bfeb9e0`

kitten (approved):

Issue #1092, opened 2020-10-27
storybook: Add devtools exchange to storybook plugin


andyrichardson:
## About

Add feature to dynamically import `@urql/devtools` and include the exchange in the storybook decorator if it exists.

## Where to start

 - Check out how we create a client for storybook fixtures [here](https://github.com/FormidableLabs/urql/blob/main/packages/storybook-addon/src/decorator/react.ts#L13)
 - Visit the [urql devtools exchange README](https://github.com/FormidableLabs/urql-devtools-exchange) for how to add the exchange to an Urql client

## Notes

Shoutout to @kiraarghy for the suggestion

kitten:
@andyrichardson quick note 😅 I suppose this is a change suggestion but there isn't enough details in here for someone to pick this up. It's ok if there's already someone (like yourself or someone else) that you'll assign to this, but given that it's tagged as "good first issue" it'll probably need more instructions, e.g. by following the RFC template.

andyrichardson:
@kitten sure thing I'll update it

kitten:
Published in `@urql/storybook-addon@1.0.4`

andyrichardson:
Doppeeeeee

PR #1112, opened 2020-11-02
(storybook) - Add devtools exchange

r281GQ:
## Summary

This for the issue #1092 .

## Set of changes

It adds the `devToolsExchange` to the exchanges if it can be found.

### DIFF
--- a/.changeset/rotten-buckets-smell.md
+++ b/.changeset/rotten-buckets-smell.md
@@ -1,0 +1,5 @@
+---
+'@urql/storybook-addon': patch
+---
+
+- Add `@urql/devtools` to `@urql/storybook-addon`. Previously it was not included (See [#1092](https://github.com/FormidableLabs/urql/issues/1092)), but now the exchange works out of the box with Storybook, by [@r281gq](https://github.com/r281GQ)
--- a/.changeset/rotten-buckets-smell.md
+++ b/.changeset/rotten-buckets-smell.md
@@ -2,4 +2,4 @@
 '@urql/storybook-addon': patch
 ---
 
-- Add `@urql/devtools` to `@urql/storybook-addon`. Previously it was not included (See [#1092](https://github.com/FormidableLabs/urql/issues/1092)), but now the exchange works out of the box with Storybook, by [@r281gq](https://github.com/r281GQ)
+- Add `@urql/devtools` to `@urql/storybook-addon`. Previously it was not included (See [#1092](https://github.com/FormidableLabs/urql/issues/1092)), but now the exchange works out of the box with Storybook
--- a/packages/storybook-addon/src/decorator/react.ts
+++ b/packages/storybook-addon/src/decorator/react.ts
@@ -4,13 +4,21 @@ import { createElement } from 'react';
 import { Provider, createClient } from 'urql';
 import { getStorybookExchange } from '../exchange';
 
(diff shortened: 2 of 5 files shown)


packages/storybook-addon/src/decorator/react.ts
```diff
@@ -4,13 +4,22 @@ import { createElement } from 'react';
 import { Provider, createClient } from 'urql';
 import { getStorybookExchange } from '../exchange';
 
+/* eslint-disable */
+const { devtoolsExchange } = require('@urql/devtools');
```
kitten, line 8: I see what the intention here is, e.g. along the lines of `try { require } catch (_e) {}` but since all decorators/addons run only on the client-side this will be going through Webpack, which will likely complain about missing dependencies. Not sure if we have an example repository to test this out on.

I think it's perfectly fine just to make our storybook addon depend on the devtools exchange and to then add it explicitly :+1:

yarn.lock
```diff
@@ -13027,11 +13044,16 @@ react-inlinesvg@^1.2.0:
...
     react-from-dom "^0.3.0"
 
-"react-is@^16.12.0 || ^17.0.0", react-is@^16.13.1, react-is@^16.6.0, react-is@^16.7.0, react-is@^16.8.1, react-is@^16.8.6, react-is@^17.0.1:
+"react-is@^16.12.0 || ^17.0.0", react-is@^17.0.1:
   version "17.0.1"
   resolved "https://registry.yarnpkg.com/react-is/-/react-is-17.0.1.tgz#5b3531bd76a645a4c9fb6e693ed36419e3301339"
   integrity sha512-NAnt2iGDXohE5LI7uBnLnqvLQMtzhkiAOLXTmv+qnF9Ky7xAPcX8Up/xWIhxvLVGJvuLiNc4xQLtuqDRzb4fSA==
 
+react-is@^16.13.1, react-is@^16.6.0, react-is@^16.7.0, react-is@^16.8.1, react-is@^16.8.6:
+  version "16.13.1"
```
kitten, line 13053: Just a quick note, we currently unfortunately have to use a Yarn resolution to force all libraries to use `react@^17.0.1` (mainly because `react-static` etc al haven't upgraded yet) so you'll need to rerun `yarn` in the root of the repo due to a yarn quirk 😅

r281GQ: Thanks for spotting it! 🙇  Corrected.

kitten (approved):

.changeset/rotten-buckets-smell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/storybook-addon': patch
+---
+
+- Add `@urql/devtools` to `@urql/storybook-addon`. Previously it was not included (See [#1092](https://github.com/FormidableLabs/urql/issues/1092)), but now the exchange works out of the box with Storybook, by [@r281gq](https://github.com/r281GQ)
```
kitten, line 5: Quick note again 😅 sorry... the changeset changelog generator will add the attribution ("by ...") automatically once it generates the changelog updates

r281GQ: No problemo! 👍  Corrected it.

andyrichardson:
Yoyoyo holup! This makes urql devtools a peer dependency and is breaking for anyone who doesn't have it installed

PR #1113, opened 2020-11-02
(core) - Add support for TypedDocumentNode

kitten:
Resolve https://github.com/dotansimha/graphql-typed-document-node/issues/36

## Summary

This comprehensively adds `Data` and `Variables` generics with untyped fallbacks to all of our results and operations, containing either `DocumentNode` or variables.

It uses the `TypedDocumentNode` type from `@graphql-typed-document-node/core` which is just a stand-in for generated types for a given GraphQL document. So other tooling can step in and fill this out. The advantage of using it throughout is that all of our APIs will be able to accept a `TypedDocumentNode` which then subsequently makes their results typed as well, similar to `graphql_ppx` and `reason-urql`.

## Set of changes

- Reexport `TypedDocumentNode` from `@urql/core` (It's technically a core type and that limits where the dependency is installed)
- Add `TypedDocumentNode` to all related types in `@urql/core`, so `GraphQLRequest`, `Operation`, `OperationResult`, several client methods, `makeOperation`, and `createRequest` now accept generics with fallbacks
- Add support for `TypedDocumentNode` to `urql`
- Add support for `TypedDocumentNode` to `@urql/preact`
- Add support for `TypedDocumentNode` to `@urql/svelte`
- Add `TypedDocumentNode` to appropriate store methods on `@urql/exchange-graphcache`

> **Note:** These changes have been marked as patches as we usually do for changes that only affect TypeScript typings.

### DIFF
--- a/.changeset/fifty-fans-dream.md
+++ b/.changeset/fifty-fans-dream.md
@@ -1,0 +1,9 @@
+---
+'@urql/exchange-graphcache': patch
+'@urql/core': patch
+'@urql/preact': patch
+'urql': patch
+'@urql/svelte': patch
+---
+
+Add support for `TypedDocumentNode` to infer the type of the `OperationResult` and `Operation` for all methods, functions, and hooks that either directly or indirectly accept a `DocumentNode`. See [`graphql-typed-document-node` and the corresponding blog post for more information.](https://github.com/dotansimha/graphql-typed-document-node)
--- a/exchanges/graphcache/src/store/store.ts
+++ b/exchanges/graphcache/src/store/store.ts
@@ -4,7 +4,8 @@ import {
   IntrospectionQuery,
   GraphQLSchema,
 } from 'graphql';
-import { createRequest } from '@urql/core';
+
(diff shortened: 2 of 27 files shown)


dotansimha:
This looks awesome!!! 
Btw, TypedDocumentNode is now supported directly from graphql-js (since 15.4), but i guess getting it from an external library is still the best way to go, since graphql-js is a peer dependency, so it might not be latest all times.

kitten:
@dotansimha Oh good hint! That's awesome 👏 I suppose we should still wait for a major bump on `graphql` to switch to the official one then, since that requires us to raise our peer dependency range.

JoviDeCroock (approved):

dotansimha:
@kitten @JoviDeCroock any plan for a release? or just ping me when it's published and I'll update the docs ;)

kitten:
@dotansimha we're currently looking at resolving #1115 before pushing out a release batch 😅 hopefully that won't take us too long

PR #1117, opened 2020-11-04
(retry) - Add operation argument to retryIf predicate

### LINKED ISSUE #1116: (retry) - Allow retryIf to avoid retrying subscriptions on non-network error
i use both retry-exchange and subscriptions. if there is an error on the server, while generating a subscription-event, then the error correctly arrives to the client, retry-middleware does a retry, and from this point on, `useSubscription` will be triggered multiple times for the same event.

urql version & exchanges:
urql=1.10.2 or 1.10.3
retry-exchange=0.1.10

**Steps to reproduce**

i created a codesandbox-example, this is the server:
https://codesandbox.io/s/urql-sub-retry-problem-server-6rrhd (nothing special here)
this is the client:
https://codesandbox.io/s/urql-sub-retry-problem-client-h8ie7

open the client, open the url with the running example, and open devtools, watch the console-output. (do not leave this running too long, or the tab might become unresponsive)
the server is set up so that it generates a new event every second, and has a 60% chance to `throw` while doing so.
while you get no errors, the new events arrive correctly, but after the first error happens, retry happens, and things get to arrive duplicated. when more errors happen, it becomes even more duplicated.

if you change the codesandbox example, to switch `urql` to version `1.10.1`, then it is suddenly fixed. it now works correctly.

unfortunately, in my local codebase reverting to `urql=1.10.1` does not fix the problem. i do not understand why, maybe a combination of package-versions?

**Expected behavior**
actually i am not even sure what exactly should happen, a subscription-event cannot really be retried. but duplicate events should not happen.

**Actual behavior**
i am getting duplicate events.

kitten:
Resolve #1116 

## Summary

Suppose there's a `error.graphqlErrors` error and that triggers a retry. This type of error won't have broken the operation pipe for subscriptions, so retrying will start an identical operation in that case for subscriptions, depending on the transport protocol and method that's used.

We could also consider adding a small dedup predicate to `subscriptionExchange` but for now we should allow the `retryExchange` to be configured to ignore certain cases per operation kind.

## Set of changes

- Add `operation: Operation` argument to `retryIf` predicate

### DIFF
--- a/.changeset/gentle-grapes-love.md
+++ b/.changeset/gentle-grapes-love.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-retry': minor
+---
+
+Add a second `Operation` input argument to the `retryIf` predicate, so that retrying can be actively avoided for specific types of operations, e.g. mutations or subscriptions, in certain user-defined cases.
--- a/exchanges/retry/src/retryExchange.test.ts
+++ b/exchanges/retry/src/retryExchange.test.ts
@@ -116,7 +116,7 @@ it(`retries if it hits an error and works for multiple concurrent operations`, (
   next(op);
 
   expect(mockRetryIf).toHaveBeenCalledTimes(1);
-  expect(mockRetryIf).toHaveBeenCalledWith(queryOneError);
+  expect(mockRetryIf).toHaveBeenCalledWith(queryOneError, op);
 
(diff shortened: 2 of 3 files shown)


wgolledge (approved):
Looks good 🥳

PR #1118, opened 2020-11-04
(docs) - add `graphql-ws` as a subscriptions client

enisdenjo:
👋, I'd like to introduce a subscriptions transport alternative [`graphql-ws`](https://github.com/enisdenjo/graphql-ws).

### DIFF
--- a/docs/advanced/subscriptions.md
+++ b/docs/advanced/subscriptions.md
@@ -69,6 +69,45 @@ we return to the `subscriptionExchange` inside `forwardSubscription`.
 
 [Read more about `subscription-transport-ws` on its README.](https://github.com/apollographql/subscriptions-transport-ws/blob/master/README.md)
 
+
+## Setting up `graphql-ws`
+
+If your GraphQL API server is using [graphql-ws](https://github.com/enisdenjo/graphql-ws),
+you'll be able to use it here too!
+
+```js
+import { createClient, defaultExchanges, subscriptionExchange } from 'urql';
+import { createClient as createGraphqlWsClient } from 'graphql-ws';
+
+const wsClient = createGraphqlWsClient({
+  url: 'wss://localhost/graphql',
+});
+
+const client = createClient({
+  url: '/graphql',
+  exchanges: [
+    ...defaultExchanges,
+    subscriptionExchange({
+      forwardSubscription(operation) {
+        return {
+          subscribe: (sink) => {
+            const dispose = wsClient.subscribe(operation, sink);
+            return {
+              unsubscribe: dispose,
+            };
+          },
+        };
+      },
+    }),
+  ],
+});
+```
+
+We create a WebSocket client with the necessary options and use the `subscribe` method from it to
+create a Subscription Observable, which we return to the `subscriptionExchange` inside `forwardSubscription`.
+
+[Read more on the `graphql-ws` README.](https://github.com/enisdenjo/graphql-ws/blob/master/README.md)
+
 ## React & Preact
 
 The `useSubscription` hooks comes with a similar API to `useQuery`, which [we've learned about in
--- a/docs/advanced/subscriptions.md
+++ b/docs/advanced/subscriptions.md
@@ -77,9 +77,9 @@ you'll be able to use it here too!
 
 ```js
 import { createClient, defaultExchanges, subscriptionExchange } from 'urql';
-import { createClient as createGraphqlWsClient } from 'graphql-ws';
+import { createClient as createWSClient } from 'graphql-ws';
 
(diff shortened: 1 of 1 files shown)


docs/advanced/subscriptions.md
```diff
@@ -69,6 +69,45 @@ we return to the `subscriptionExchange` inside `forwardSubscription`.
...
+
+const client = createClient({
+  url: '/graphql',
+  exchanges: [
+    ...defaultExchanges,
+    subscriptionExchange({
+      forwardSubscription(operation) {
+        return {
+          subscribe: (sink) => {
+            const dispose = wsClient.subscribe(operation, sink);
```
kitten, line 94: You can just return the result of `.subscribe(...)` directly since it's a Subscription-like object 🙌

docs/advanced/subscriptions.md
```diff
@@ -69,6 +69,45 @@ we return to the `subscriptionExchange` inside `forwardSubscription`.
...
 
+
+## Setting up `graphql-ws`
+
+If your GraphQL API server is using [graphql-ws](https://github.com/enisdenjo/graphql-ws),
+you'll be able to use it here too!
+
+```js
+import { createClient, defaultExchanges, subscriptionExchange } from 'urql';
+import { createClient as createGraphqlWsClient } from 'graphql-ws';
```
kitten, line 80: nit: I'd just name this `createWSClient` or `createSubscriptionClient` since it's all equivalent but more easily readable; basically because the casing on the import is a little inconsistent otherwise

kitten (commented):
This is awesome! I saw the announcement post but didn't quite get to updating the docs back then 😅 Just small small nits, but let's get this merged. Cheers! ❤️

docs/advanced/subscriptions.md
enisdenjo: Hmm, are you sure? Your subscription forwarder interface wouldn't agree:
https://github.com/FormidableLabs/urql/blob/85a0946ed83559988b228a605df0b26e49d40c51/packages/core/src/exchanges/subscription.ts#L46-L48

You require an "observable" to be returned. The observable must have a `subscribe` method and an `unsubscribe` method on the resulting subscription.

Our `subscribe` method intentionally returns **just** a cleanup function because it is intended to be a "subscriber" of an [Observable](https://github.com/tc39/proposal-observable#observable). This way you can use _any_ observable library as long as it conforms to the [ECMAScript Observable proposal](https://github.com/tc39/proposal-observable).
```ts
const wsClient = createWsClient({
  url: 'wss://localhost/graphql',
});
const observable = Observable.create(
  function subscriberFunction(observer: SubscriptionObserver) {
    // Subscriber func returns a cleanup function which will cancel the event stream/dispose of the subscription
    return wsClient.subscribe(operation, observer);
  }
);
```
_*You can check the [Client usage with Observable recipe](https://github.com/enisdenjo/graphql-ws#recipes) to get a better understanding_

JoviDeCroock (approved):
Nice work!

docs/advanced/subscriptions.md
enisdenjo: [I settled with `createWSClient`.](https://github.com/FormidableLabs/urql/pull/1118/commits/d37ca3f45aea54b59422ccebe334f120dd1e56d1)

Mainly because you can transport queries and mutations over `graphql-ws` too, but not only that - you can also stream executions for supporting directives like `@defer` or `@stream`.

docs/advanced/subscriptions.md
kitten: Oh, sorry, I just read that section. Any reason why you don't just return a spec-compliant observable? I get that the API is simpler on your end this way, but it seems rather contrarian to have no adapter for this 😅 that's just on a side-note though

enisdenjo: No worries, one reason is stated above:

>This way you can use any observable library as long as it conforms to the ECMAScript Observable proposal.

Another reason is that, if I add it, I simply have to maintain one more superfluous thing...

You're free to choose your own flavour! GQL libraries tend to use their own Observables. Apollo implements its own (which is actually zen-observable), Relay implements its own, you can stream the subscription using RxJS, etc. etc.

enisdenjo: You can also choose to **not** use observables, you can simply promisify the `subscribe` method. For single results, this is how you'd go by:
```ts
async function execute<T>(payload: SubscribePayload) {
  return new Promise((resolve, reject) => {
    let result: T;
    client.subscribe<T>(payload, {
      next: (data) => (result = data),
      error: reject,
      complete: () => resolve(result),
    });
  });
}
```

PR #1119, opened 2020-11-05
(fix) - use Record over object type for subscription operation variables

enisdenjo:
As suggested by TS and recommended ESLint rules - don't use `object` as a type. The `object` type is currently hard to use ([see this issue](https://github.com/microsoft/TypeScript/issues/21732)).

This also solves the type compatibility for `graphql-ws` where you can use the lib in TS as you would in plain ol' JS like in #1118.

Previously, because of this `object` type, you have to do this:
```ts
...
subscriptionExchange({
  forwardSubscription({ query, variables }) {
    return {
      subscribe: (sink) => {
        const dispose = wsClient.subscribe(
          {
            query,
            variables: variables as Record<string, unknown>, // 😞
          },
          sink,
        );

        return {
          unsubscribe: dispose,
        };
      },
    };
  },
}),
...
```

### DIFF
--- a/.changeset/breezy-ghosts-sin.md
+++ b/.changeset/breezy-ghosts-sin.md
@@ -1,0 +1,5 @@
+---
+'@urql/core': patch
+---
+
+Use `Record` over `object` type for subscription operation variables. The `object` type is currently hard to use ([see this issue](https://github.com/microsoft/TypeScript/issues/21732)).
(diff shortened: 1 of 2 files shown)


JoviDeCroock (approved):
This looks good to me! Would you mind running `yarn changeset`, marking this as a patch for `core` and adding a little explainer as to why this change is needed?

`changesets` are our way to automatically generate changelogs.

enisdenjo:
Awesome! [Is this what you had in mind?](https://github.com/FormidableLabs/urql/pull/1119/commits/d3c43a82b0fa5b3bf197d3c5dbf291cf11aea729)

JoviDeCroock:
That looks great @enisdenjo, thank you!

PR #1123, opened 2020-11-06
(core) - Improve Suspense implementation and fix client-side Suspense

### LINKED ISSUE #1115: (suspense) - Suspense Exchange doesn't hold on to results and errors
Originally reported by @StevenLangbroek 

## Summary

**Steps to reproduce**

1. Create a client in suspense mode
2. Add a query that always errors
3. (optional) add `suspenseExchange`

See: https://github.com/StevenLangbroek/urql-suspense-loop-repro/tree/c8c88f42687e4f2c9ccc7839fa1526e13540769d

**Expected behavior**

The component suspends, the fallback renders, then on resolution of the query it displays the error.

**Actual behavior**

The component suspends, rerenders, then suspends again. This repeats indefinitely

## Root Cause

The `@urql/exchange-suspense` package has not been kept up-to-date and we should consider moving suspense support back into mainline stable, and not experimental. @StevenLangbroek: Let me know if anything else is still causing trouble that we can address with this issue too, e.g. unexpected results after the latest changes on `main`.

The exchange should hold on to a result for one tick until the component re-renders at which point it can then reuse the result. I suspect that the double-subscription approach to initialise the effect subscription erases the result stored by the `suspenseExchange` before the component fully mounts, thus triggering suspense again.

kitten:
Resolve #1115 

## Summary

This reimplementation of our React Suspense support should improve edge-cases, support the `context.suspense === false` switch consistently, and fix the issue described in #1115 where uncached results would cause an infinite update loop when client-side Suspense was used.

The suspense logic has effectively moved to `useQuery` itself and caches result sources temporarily. In its implementation it assumes that all requests eventually are remounted, which is a safe assumption to make since after React Suspense triggers there's no way for a user to abort its eventual remount. Even if the cache leaks a suspense source, this source doesn't stay active forever and the added memory overhead would in theory be minimal.

## Set of changes

- Make `operation.context.suspense` more consistent. It now explicitly determines whether an operation has Suspense-mode enabled
- Move all `useSource` trigger effect to `useQuery` and `useSubscription` effectively (which also fixes the case of `useSubscription` triggering suspense unnecessarily in some cases
- Remove `toSuspenseSource` logic from `Client` and move it to `useQuery` with an added hook-wide cache
- Move `pollInterval` implementation to `executeRequestOperation` in case we find a way to reduce `useQuery`'s bundlesize impact now


JoviDeCroock (approved):

packages/preact-urql/src/hooks/useQuery.ts
```diff
@@ -38,67 +50,120 @@ export type UseQueryResponse<Data = any, Variables = object> = [
...
+
+    // If we haven't got a previous result then start suspending
+    // otherwise issue the last known result immediately
+    if (cache !== undefined) {
+      const signal = [cache] as [T] & { tag: 1 };
+      signal.tag = 1;
+      sink(signal);
+    } else {
+      hasSuspended = true;
+      sink(0 /* End */);
```
kitten, line 81: It really isn't great that we're manually sending source events here in case of breaking changes in Wonka (which are in fact coming) but I haven't really found a better implementation for this yet that's as short. The alternative is to use `Wonka.make` I suppose but then we need to manually pipe all events through.

PR #1131, opened 2020-11-07
(docs) - Update CONTRIBUTING guide with new information

kitten:
This basically adds:

- How our issues work, especially regarding the RFC process
- How PRs are written with respect to changesets and titles
- How the automated releases work

It also removes some duplicate sections and revises some other things I've found


JoviDeCroock (approved):
Nice one! :shipit:

PR #1139, opened 2020-11-09
(react, preact) Add default context to useMutation

Krisztiaan:
## Summary

When writing a `Login` mutation, I could add `additionalTypenames` without much extra code.

## Set of changes

Additional parameter for `useMutation`, that is a `Partial<OperationContext>`, that get's merged (and overwritten) by mutation-time context.

### DIFF
--- a/.changeset/friendly-eyes-sip.md
+++ b/.changeset/friendly-eyes-sip.md
@@ -1,0 +1,6 @@
+---
+'@urql/preact': minor
+'urql': minor
+---
+
+Add optional parameter for `useMutation` to supply a default `OperationContext`
--- a/packages/preact-urql/src/hooks/useMutation.ts
+++ b/packages/preact-urql/src/hooks/useMutation.ts
@@ -32,7 +32,8 @@ export type UseMutationResponse<Data = any, Variables = object> = [
 ];
 
 export function useMutation<Data = any, Variables = object>(
-  query: DocumentNode | TypedDocumentNode<Data, Variables> | string
+  query: DocumentNode | TypedDocumentNode<Data, Variables> | string,
+  defaultContext: Partial<OperationContext> = {}
 ): UseMutationResponse<Data, Variables> {
   const isMounted = useRef(true);
   const client = useClient();
@@ -42,13 +43,13 @@ export function useMutation<Data = any, Variables = object>(
   );
 
   const executeMutation = useCallback(
-    (variables?: Variables, context?: Partial<OperationContext>) => {
+    (variables?: Variables, context: Partial<OperationContext> = {}) => {
       setState({ ...initialState, fetching: true });
 
       return pipe(
         client.executeMutation<Data, Variables>(
           createRequest<Data, Variables>(query, variables),
-          context || {}
+          { ...defaultContext, ...context }
         ),
         toPromise
       ).then(result => {
--- a/packages/react-urql/src/hooks/useMutation.ts
+++ b/packages/react-urql/src/hooks/useMutation.ts
@@ -32,7 +32,8 @@ export type UseMutationResponse<Data = any, Variables = object> = [
 ];
 
 export function useMutation<Data = any, Variables = object>(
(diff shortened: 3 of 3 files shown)


andyrichardson:
Hey @Krisztiaan, thanks for the PR 👋 

Do you have some examples where this functionality would have been useful? Something like a before and after maybe?

Krisztiaan:
If I were to use `login` in multiple places, but still need to invalidate other related typenames, I'd create a hook like this, instead of handling the common effects one-by-one.

```typescript
export default function useLogin() {
  const setToken = useSetToken()
  const loginMutationTuple = useLoginMutation({
    additionalTypenames: ['User', 'Booking'],
  })
  const [loginResult] = loginMutationTuple

  useEffect(() => {
    setToken(loginResult.data?.login?.accessToken)
  }, [loginResult.data, setToken])

  return loginMutationTuple
}
```

kitten:
Mostly my concerns here are with API design. The `useMutation` hook is unique compared to the `useQuery` and `useSubscription`. In terms of "eccentricity" `useQuery` is the most complex and eccentric having to handle suspense, `useSubscription` comes next. Both need to handle mounting with the correct state and starting subscriptions to the `Client` consistently.

`useMutation` is special in that it's more reduced. Compared to the others it's just a lean wrapper that "echoes" the eventual mutation result on the hook's state, which isn't always used. Compared to using the `client.mutation(...).toPromise()` wrapper it doesn't do much more than to also update the hook state.

This may seem confusing since the other hooks provide a much more thorough parameters, and even alias properties to the top level of its parameters, i.e. they accept some context values like `requestPolicy` directly but also accept `context` as a separate parameter.

What concerns me about this API design is that it feels like it may be confused with `useMutation(document, ?variables)` which wouldn't make much sense to use.

Rather we imagined that it's more common to use the hook and create a wrapper callback for `executeMutation` or to use `client.mutation` directly if necessary.

andyrichardson:
> Hey @andyrichardson sure thing,
> 
> If I were to use `login` in multiple places, but still need to invalidate other related typenames, I'd create a hook like this, instead of handling the common effects one-by-one.
> 
> ```ts
> export default function useLogin() {
>   const setToken = useSetToken()
>   const loginMutationTuple = useLoginMutation({
>     additionalTypenames: ['User', 'Booking'],
>   })
>   const [loginResult] = loginMutationTuple
> 
>   useEffect(() => {
>     setToken(loginResult.data?.login?.accessToken)
>   }, [loginResult.data, setToken])
> 
>   return loginMutationTuple
> }
> ```

@Krisztiaan  Thanks for the example - auth is a tricky one!

I think @kitten has made some really good points :+1:

Not too sure why you would want to trigger a login in multiple places within your app but either way, if you do need to hoist state and logic, context might be the solution you're looking for 🤞 

```tsx
// Provider
const AuthProvider = (props) => {
  const [token, setToken] = useState();
  const [loginState, executeMutation] = useLoginMutation();

  const login = useCallback((args) => 
    executeMutation(args, { additionalTypenames: ['User', 'Booking'] })),
    []
  );

  useEffect(() => {
    if (!loginState.data) {
      return;
    }

    setToken(loginState.data.login.accessToken)
  }, [loginState.data])
  
  return <Auth.Provider {...props} value={{ token, setToken, login }} />
};

// Consuming components
const MyComponent = () => {
  const { login } = useAuth();

  // ...
};
```

kitten:
Hiya 👋 sorry, closing this due to inactivity. We do have an RFC process and are also open to discussing changes via that first or in a discussions thread (which does exist first, so that you won't have to propose changes before having a full discussion)

The tl;dr here is that sometimes we'll put a pragmatic API design for bindings before making certain usages slightly shorter ✌️ cheers!

Krisztiaan:
Thanks for all the input guys, I've opted for a slightly modified version of @andyrichardson 's solution. I agree, this may not be worth changing the API for.

PR #1142, opened 2020-11-10
(react/preact) - Fix client-side Suspense loop regression

kitten:
## Summary

This regressed in the last patch, the version before that explicitly fixing it. This patch was aimed at server-side rendering but cleared the local suspense cache too early.

## Set of changes

- Ensure that caching is still disabled on the server-side
- Clear suspense cache in effect rather than render loop

### DIFF
--- a/.changeset/eighty-rivers-sniff.md
+++ b/.changeset/eighty-rivers-sniff.md
@@ -1,0 +1,6 @@
+---
+'@urql/preact': patch
+'urql': patch
+---
+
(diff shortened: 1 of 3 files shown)


JoviDeCroock (commented):
Did some smoke tests in former reproductions, :shipit:

JoviDeCroock (approved):
Did some smoke tests in former reproductions, :shipit:

(comment instead of approve....)

PR #1148, opened 2020-11-12
(docs) - Add note about third argument in readFragment

vitosamson:
## Summary

Adds an example of passing variables in `cache.readFragment`. This was previously undocumented.


kitten (approved):
You're absolutely right in that this should be documented 🙌 I believe I was reluctant to do so since it comes up quite rarely and is a little dangerous since there are no variable definitions to normalise the input, but it's there as an escape hatch and convenience afterall

PR #1151, opened 2020-11-14
(vue-urql) - Refactor usage of ref/unref

LinusBorg:
## Summary

The current implementation of `useQuery` and `useSubscription` uses refs heavily and also uses `unref` in a lot of places, which adds noise to the code. This PR aims to simplify the implementation and to have more readable code.

## Set of changes

1. We can simplify the implementation substantially by using `reactive` on the args object, which will unwrap all nested refs automatically.
2. The conditional ref wrapping for `isPaused` and `scanHandler` can be simplified with `computed` in the former (for casting to a boolean) and a simple `ref()` for the latter (calling `ref()` on a `Ref` simply returns the ref as is).

This is purely an internal refactoring and doesn't have any influence on the API. therefore, I didn't include a changeset. Please let me know if one is required nonetheless and I'll add it.

### DIFF
--- a/packages/vue-urql/src/useQuery.ts
+++ b/packages/vue-urql/src/useQuery.ts
@@ -1,4 +1,4 @@
-import { Ref, unref, ref, isRef, watchEffect } from 'vue';
+import { Ref, ref, watchEffect, reactive, computed } from 'vue';
 import { DocumentNode } from 'graphql';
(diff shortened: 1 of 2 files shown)


kitten (approved):
This is a really neat trick! I feel like I missed something really crucial then 😅

PR #1153, opened 2020-11-14
(vue-urql) - Add Tests

LinusBorg:
## Summary

This PR adds tests for:
- `useMutation`
- `useQuery`
- `useSubscription`

## Set of changes

- The added tests cover the same cases as the svelte implementation.
- I had to add `jest` to the package's tsconfig type option in order to stop TS from complaining, this should not have any impact outside of the vue-urql package.
- This is a draft PR for now as I came across some problems to test `useQuery().then()` behavior, and I'm not certain what the desired behavior here should be. Please see the code comments 

This only adds tests, so I didn't create a changeset. I can provide one if that's still required for this kind of PR.

### DIFF
--- a/packages/vue-urql/src/useQuery.test.ts
+++ b/packages/vue-urql/src/useQuery.test.ts
@@ -1,0 +1,132 @@
+jest.mock('vue', () => {
+  const vue = jest.requireActual('vue');
+
+  return {
+    __esModule: true,
+    ...vue,
+    inject: () => client,
+  };
+});
+import { makeSubject } from 'wonka';
+import { createClient } from '@urql/core';
+import { useQuery } from './useQuery';
+import { nextTick, reactive } from 'vue';
+
+const client = createClient({ url: '/graphql', exchanges: [] });
+
+describe('useQuery', () => {
+  it('runs a query and updates data', async () => {
+    const subject = makeSubject<any>();
+    const executeQuery = jest
+      .spyOn(client, 'executeQuery')
+      .mockImplementation(() => subject.source);
+
+    const _query = useQuery({
+      query: `{ test }`,
+    });
+    const query = reactive(_query);
+
+    expect(executeQuery).toHaveBeenCalledWith(
+      {
+        key: expect.any(Number),
+        query: expect.any(Object),
+        variables: {},
+        context: undefined,
+      },
+      {
+        pollInterval: undefined,
+        requestPolicy: undefined,
+      }
+    );
+
+    expect(query.fetching).toBe(true);
+
+    subject.next({ data: { test: true } });
+
+    expect(query.fetching).toBe(false);
+    expect(query.data).toEqual({ test: true });
+  });
+
+  it('runs Query and awaits results', async () => {
+    const subject = makeSubject<any>();
+    const executeQuery = jest
+      .spyOn(client, 'executeQuery')
+      .mockImplementation(() => subject.source);
+
+    const query = useQuery({
+      query: `{ test }`,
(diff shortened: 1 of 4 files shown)


packages/vue-urql/src/useQuery.test.ts
```diff
@@ -0,0 +1,156 @@
...
+    const executeQuery = jest
+      .spyOn(client, 'executeQuery')
+      .mockImplementation(() => {
+        subject.next({ data: { test: true } });
+        return subject.source;
+      });
+
+    const query = await useQuery({
+      query: `{ test }`,
+      // pause: true,
```
kitten, line 89: Oh that's interesting! I think this is because we can liken this behaviour to suspense in React, so once `useQuery`'s promise resolves and we subscribe again we don't have a result, since no `cacheExchange` is used.

Adding the appropriate exchanges however is not what we want though. While it would fix this, the cache doesn't cache errored results for instance, so instead we'll have to copy some code from `react-urql`'s `useQuery` where the result is cached just for the suspended stream for the next rerender after suspending.

I can take a look at fixing this

LinusBorg: My main problem here was that I wasn't sure what the use case of making it " thenable" is, so I wasn't sure what an appropriate fix would be.

Am I understanding correctly that this is supposed to support Vue's version of `Supsense`?

packages/vue-urql/src/useQuery.test.ts
```diff
@@ -0,0 +1,156 @@
...
+      },
+      {
+        pollInterval: 500,
+        requestPolicy: 'cache-first',
+      }
+    );
+
+    expect(query.fetching).toBe(true);
+
+    subject.next({ data: { test: true } });
```
LinusBorg, line 148: I'm not too sure if this test is adequate for polling. Maybe we can use some timer mocks in the  `executeQuery`'s `mockImplementation` so it really sends a couple of data points after timeouts?

then again, would that really improve the tests? Not sure.

packages/vue-urql/src/useQuery.test.ts
kitten: Yea, I think I also don't quite know enough unfortunately quite yet to fix this. So it is indeed to support Vue's version of suspense, so we'd like to await the first result that `executeQuery` may yield but during normal execution we'd also like future results to still update the component.

I'm not quite clear on how the component lifecycle functions once Vue suspends, so it'd be interesting to know whether the lifecycle aborts when it does.

I think we may get into some trouble here with our `isThenable` flag, but essentially the idea was always to return the promise, resolve it on a result, and otherwise still keep a subscription active to update the result refs passively.

packages/vue-urql/src/useQuery.test.ts
kitten: I think this is adequate, since all it'd have to support is results being actively pushed into the `useQuery` subscription

packages/vue-urql/src/useQuery.test.ts
LinusBorg: Gotcha. I'll think a bit about it as well.

LinusBorg:
@kitten Suggestion: we merge these tests as they are, with the skipped one, just so we have some basic checks in place for the rest of the functionality.

Then we can work on solving the issue with that Promise for Suspense without breaking stuff accidentally.

JoviDeCroock (approved):
This is some awesome work 🎉 I agree @LinusBorg let's merge this as is, having tests in place is already awesome

PR #1155, opened 2020-11-15
(vue-urql) - Fix pausing behavior

LinusBorg:
## Summary

Fixes a bug introduced in #1151, which broke `pause()`/resume()` functionality of `useQuery()`

## Set of changes

I'm sorry about this - in #1151 I turned `isPaused` into a computed ref, which makes it read-only as it's a derived value.

This PR turns it into a writeable `ref()` again.

It doesn't contain any breaking changes, just a fix for a patch release.

### DIFF
--- a/.changeset/flat-rats-fetch.md
+++ b/.changeset/flat-rats-fetch.md
@@ -1,0 +1,5 @@
+---
+'@urql/vue': patch
+---
+
+fix pausing feature of useQuery by turning "isPaused" into a ref again.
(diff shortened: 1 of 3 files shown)


LinusBorg:
I have a few other improvements in mind but we should get some basic tests going with #1153 before we touch anything else.

JoviDeCroock (approved):
This looks really good! Thank you so much for all your efforts already! 👍

PR #1157, opened 2020-11-16
(react) - Fix client-side Suspense update edge-case

### LINKED ISSUE #1150: useQuery not re requesting if variables changes while request is already sent
urql version & exchanges:

**Steps to reproduce**

1. have a query
2. update the query variables while the query is in flight

**Expected behavior**

The result is based on the first variable and not the second one

**Actual behavior**

The result should be a request with the new input variables

**Example code**
```typescript
export const useCurrentUser = () => {
// This return first a undefined, then 100 ms later will send the user
  const firebaseUser = useFirebaseUser(userWithCorrectToken$);

  const [
    { data, error, fetching, extensions, operation },
    retry,
  ] = useQuery({
    query: myQuery,
    variables: {
      firebaseId: firebaseUser?.uid ?? 'anonymous',
    },
    requestPolicy: 'network-only',
  });

  console.log('useCurrentUser', {
    data,
    error,
    fetching,
    extensions,
    operation,
  });

  const user = data?.user?.[0];
  return { error, user, fetching, loading: fetching };
};
```

This error occurs in 1.11 but not in 1.10. This may be related to the change in this pr : https://github.com/FormidableLabs/urql/pull/1104

I will make a sandbox soon, I don't have time right now but will soon !

kitten:
Fix #1150 

## Summary

We encountered another edge-case in client-side React Suspense, which is caused by React handling suspense-on-updates differently than suspense-on-mount. We basically have to be more sensible with state on suspense-on-update because React doesn't clear any state when we suspend during an update.

This can be tested against this reproduction sandbox: https://codesandbox.io/s/ecstatic-ritchie-lzxvl (See CodeSandbox CI output)

## Set of changes

- Use the latest known `input` in `useSource` when subscribing

### DIFF
--- a/.changeset/nasty-kids-bake.md
+++ b/.changeset/nasty-kids-bake.md
@@ -1,0 +1,5 @@
+---
+'urql': patch
+---
+
+Fix in edge-case in client-side React Suspense, where after suspending due to an update a new state value is given to `useSource` in a render update. This was previously then causing us to subscribe to an outdated source in `useEffect` since the updated source would be ignored by the time we reach `useEffect` in `useSource`.
--- a/packages/react-urql/src/hooks/useSource.ts
+++ b/packages/react-urql/src/hooks/useSource.ts
@@ -37,7 +37,16 @@ export function useSource<T, R>(
     ]);
 
     const updateInput = (nextInput: T) => {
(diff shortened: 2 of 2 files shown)


JoviDeCroock (approved):
:shipit:

PR #1159, opened 2020-11-16
(vue) - Fix Vue Suspense by making thenable reactive

kitten:
##  Summary

This should fix the thenable in `useQuery`'s Vue Suspense implementation by making it more reactive.

cc @LinusBorg

### DIFF
--- a/packages/vue-urql/src/useQuery.ts
+++ b/packages/vue-urql/src/useQuery.ts
@@ -1,6 +1,20 @@
-import { Ref, ref, watchEffect, reactive, isRef } from 'vue';
+import { Ref, ref, watch, reactive, isRef } from 'vue';
 import { DocumentNode } from 'graphql';
-import { pipe, take, publish, share, onPush, toPromise, onEnd } from 'wonka';
+
+import {
+  Source,
+  concat,
+  fromValue,
+  map,
+  pipe,
+  take,
+  publish,
+  share,
+  onStart,
+  onPush,
+  toPromise,
+  onEnd,
+} from 'wonka';
 
 import {
   OperationResult,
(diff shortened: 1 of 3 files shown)


JoviDeCroock (approved):
Neat :shipit:

packages/vue-urql/src/useQuery.ts
```diff
@@ -69,80 +82,106 @@ export function useQuery<T = any, V = object>(
...
-      if (!isPaused.value) {
-        executeQuery();
-        onInvalidate(() => {
-          unsubscribe.value();
-        });
+  watch(
+    [source],
+    (_, __, onInvalidate) => {
+      if (!source.value) {
+        return unsubscribe.value();
```
LinusBorg, line 128: Maybe I'm missing something, but the `onInvalidate` callback should already take care of unsubscribing in either case. we can just return here without unsubscribing again.

packages/vue-urql/src/useQuery.ts
```diff
@@ -151,13 +190,24 @@ export function useQuery<T = any, V = object>(
...
+    ...state,
     then(onFulfilled, onRejected) {
-      isThenable = true;
-      return executeQuery()
-        .then(() => result)
-        .then(onFulfilled, onRejected);
+      let result$: Promise<UseQueryState<T, V>>;
+      if (fetching.value && source.value) {
+        result$ = pipe(
+          source.value,
```
LinusBorg, line 199: Disclaimer: This might be my lack of understanding for wonka speaking:

What happens if, while the request of the first source is still running, a side-effect i.e. changes variables and a new source & request is created?

* Would this promise never resolve?
* Would this promise still resolve to the original requests value?
* Would this promise somehow resolve to the second request's value, as we concat sources on line 134?

The third behavior is what I think *should* happen but I can't tell if it does, and the tests don't cover it yet.

LinusBorg (approved):
This is a great start and definitely an improvement. But I think we can do better and refactor those multiple specific `watch` effects into something less verbose.

When I find some time tomorrow I'll take a stab at it, provided this is merged by then.

packages/vue-urql/src/useQuery.ts
kitten: It's the second case right now, but potentially we can now try shoehorning the third case in.

kitten:
Awesome! Generally I made the watchers more verbose as I found it's easier to make it work this way first and I'm sure we can trim it back down later ✌️

Good shout on the promise behaviour. I'm not sure if we do want to support a case where the source is a moving target as that could maybe lead to some gnarly looping behaviour when user code goes wrong, but generally I agree.

We can try to do that, yea

packages/vue-urql/src/useQuery.ts
kitten: this is here because of pausing. `executeQuery` can update `source.value` and the watcher may set it to `undefined` when everything's paused, so that `executeQuery` can explicitly take precedence over the watcher's subscription being paused.

PR #1174, opened 2020-11-19
(graphcache) - Add mergeMode to simplePagination helper

### LINKED ISSUE #1173: RFC: Add mergeMode to simplePagination helper
## Summary

[Relay Pagination](https://formidable.com/open-source/urql/docs/graphcache/computed-queries/#relay-pagination) allows setting `mergeMode` to define how the paginated items are merged. From the doc:

> Example series of requests:
> 
> ```
> first: 1 => node 1, endCursor: a
> first: 1, after: 1 => node 2, endCursor: b
> ...
> last: 1 => node 99, startCursor: c
> last: 1, before: c => node 89, startCursor: d
> ```
> 
> With inwards merging the nodes will be in this order: [1, 2, ..., 89, 99] And with outwards merging: [..., 89, 99, 1, 2, ...]

We should implement a similar feature in [Simple Pagination](https://formidable.com/open-source/urql/docs/graphcache/computed-queries/#simple-pagination).

A use case is for the infinite scroll that needs to load previous items when scrolling up, where the pages with higher `skip` or `offset` value should be inserted to the beginning of the array. Ex: Chat app. In this case, later paginated items should be added to the beginning instead of the end.

## Proposed Solution

Add a new option `mergeMode = "before" | "after"` to define how pages will be merged. If `mergeMode = "after"` (default), (later) pages with higher offset will be added **after** the others. This matches the current behavior so it will not be a breaking change. Otherwise, if `mergeMode = "before"`, (later) pages with higher offset will be added **before** the others.

Example:

*Where `mergeMode = "after"` (default) makes sense:*

```
offset = 0 => 1,2,3
offset = 3 => 4,5,6
```

```
mergeMode = "after" => [1,2,3,  4,5,6]
mergeMode = "before" => [4,5,6,  1,2,3]
```

*Where `mergeMode = "before"` makes sense:*

```
offset = 0 => 4,5,6
offset = 3 => 1,2,3
```

```
mergeMode = "after" => [4,5,6,  1,2,3]
mergeMode = "before" => [1,2,3,  4,5,6]
```

Here is a demo image ([source](https://use-the-index-luke.com/sql/partial-results/fetch-next-page)):
image

hoangvvo:
Resolves #1173 

## Summary

Peek 2020-11-19 01-32

## Set of changes

Adds `mergeMode` to `simplePagination` similar to that of `relayPagination`. Default to `inwards`, so it will not be a breaking change.

### DIFF
--- a/.changeset/little-crabs-sell.md
+++ b/.changeset/little-crabs-sell.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-graphcache': minor
+---
+
+Add mergeMode to simplePagination helper to handle how pages are merged when you paginate forwards and backwards
--- a/.changeset/little-crabs-sell.md
+++ b/.changeset/little-crabs-sell.md
@@ -2,4 +2,4 @@
 '@urql/exchange-graphcache': minor
 ---
 
-Add mergeMode to simplePagination helper to handle how pages are merged when you paginate forwards and backwards
+Add mergeMode = 'before' | 'after' to simplePagination helper to define whether pages are merged before or after when you paginate forwards and backwards
--- a/.changeset/little-crabs-sell.md
+++ b/.changeset/little-crabs-sell.md
@@ -2,4 +2,4 @@
 '@urql/exchange-graphcache': minor
 ---
 
-Add mergeMode = 'before' | 'after' to simplePagination helper to define whether pages are merged before or after when you paginate forwards and backwards
+Add a `mergeMode: 'before' | 'after'` option to the `simplePagination` helper to define whether pages are merged before or after preceding ones when pagination, similar to `relayPagination`'s option
--- a/docs/graphcache/computed-queries.md
+++ b/docs/graphcache/computed-queries.md
@@ -192,26 +192,29 @@ options in here `limitArgument` and `offsetArgument` these will default to `limi
 and `skip` respectively. This way you can use the keywords that you are using in
 your queries.
 
-You can also specify `mergeMode`, which defaults to `inwards` and
-can otherwise be set to `outwards`. This will handle how pages are merged when you paginate
-forwards and backwards at the same time. Outwards pagination assumes that pages
-that come in last should be merged before the first pages, so that the list
-grows outwards in both directions. The default inwards pagination assumes that
-pagination last pages is part of the same list and come after first pages.
-Hence it merges pages so that they converge in the middle.
+You can also specify `mergeMode`, which defaults to `after` and can otherwise
+be set to `before`. This will handle how pages are merged when you paginate
+forwards and backwards at the same time. The `before` mode assumes that pages
+that come in last should be merged _before_ the first pages. The default `after`
+mode assumes that pages that come in last should be merged _after_ the first pages.
 
-Example series of requests:
+Examples series of requests with `limit: 3`:
 
 ```
-limit: 2, skip: 0 => 1, 2
-limit: 2, skip: 1 => 2, 3
-...
-limit: 2, skip: c - 1 => 79, 89
-limit: 2, skip: c => 89, 99
-```
+// Where mergeMode: after makes sense
+skip: 0 => 1, 2, 3
+skip: 3 => 44, 55, 66
 
-With inwards merging the items will be in this order: `[1, 2, 3, ..., 79, 89, 99]`
-And with outwards merging: `[..., 79, 89, 99, 1, 2, 3, ...]`
+mergeMode: after => 1, 2, 3, 44, 55, 66
+mergeMode: before => 44, 55, 66, 1, 2, 3
+
+// Where mergeMode: before makes sense
+skip: 0 => 44, 55, 66
+skip: 3 => 1, 2, 3
+
+mergeMode: after => 44, 55, 66, 1, 2, 3
+mergeMode: before => 1, 2, 3, 44, 55, 66
+```
 
 ### Relay Pagination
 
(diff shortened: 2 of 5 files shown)


exchanges/graphcache/src/extras/simplePagination.ts
```diff
@@ -74,21 +78,20 @@ export const simplePagination = ({
...
+      for (let j = 0; j < links.length; j++) {
+        const link = links[j];
+        if (visited.has(link)) continue;
+        tempResult.push(link);
+        visited.add(link);
+      }
+
+      const isAhead = !prevOffset || currentOffset > prevOffset;
+
+      if (isAhead === (mergeMode === 'inwards')) {
```
JoviDeCroock, line 92: Let's simplify this to:

```suggestion
      if ((!prevOffset || currentOffset > prevOffset) && mergeMode === 'inwards') {
```

I think that will make it a bit easier to ready, we can remove line 90 with this addition

JoviDeCroock (commented):
Looks good! Thank you for taking the time to PR this, mind adding a `changeset` by running `yarn changeset` at the root? This would be a `minor` for `@urql/exchange-graphcache` since we're adding a pagination feature! 👍

hoangvvo:
Thanks for such a quick review! I added documentation for `mergeMode` of Simple Pagination based on that of Relay Pagination. Changeset also added.

kitten:
I believe this is misunderstanding what the `mergeMode` does on Relay pagination.
There it takes care of which order things are merged in, yes, but in relation to how the pagination arguments are used, since `before` and `after` cursors can be used for the same list. This then creates ambiguity because both directions can be "outward" from the same point or "inward" towards a shared centre in the list.

I believe `mergeMode` may not make much sense for `simplePagination` as we're not dealing with two directions of pagination so inwards/outwards aren't descriptive.

Furthermore I'm unsure the explanation in the RFC is accurate and whether a lot of APIs even use this pattern? Do we have some prior art here?

hoangvvo:
@kitten You are right. I am not familiar with Relay and possibly misusing the term `inwards` and `outwards`. Either way, here are the relevant words from relay pagination definition:

> Outwards pagination assumes that pages
> that come in **last** should be merged **before the first pages**, [...] The default inwards pagination assumes that
> pagination **last pages** is part of the same list and come **after first pages**.

This matches the description of what I am trying to achieve. Perhaps,  `mergeMode = "before" | "after"` or `mergeMode = "reverse" | "forward"` might be better. It satisfies use-cases of infinite scroll but instead of scrolling down for older items, we scroll up for them. The GIF above demonstrates that.

kitten:
@hoangvvo i believe would rename it like that 👍 the reason why the description matches is because for this paragraph it describes rather what happens when we start pagination from in reverse while forward pagination has been used and vice-versa. So the behaviour does match but it's for a subtly different reason 😅

hoangvvo:
> @hoangvvo i believe would rename it like that  the reason why the description matches is because for this paragraph it describes rather what happens when we start pagination from in reverse while forward pagination has been used and vice-versa. So the behaviour does match but it's for a subtly different reason

I agree that it is still a somewhat uncommon pattern. I can definitely workaround this by simply reverse the returned GraphQL array appropriately.

```
offset = 0 => 9,8,7
offset = 3 => 6,5,4
```

The current `simplePagination` will merge it into `[9,8,7,6,5,4]` and I would only need to reverse it to achieve the above effect. So it is okay to not putting this into the helper.

Otherwise, how would you rename it? I am steering toward `mergeMode = "before" | "after"` with the default being `after`, where pages with higher offset will be put **after**, and can otherwise be "before" where pages with higher offset will be put **before**. The doc will also be updated to reflect this instead of using the terms `inwards` and `outwards`

hoangvvo:
@kitten I updated the [RFC](https://github.com/FormidableLabs/urql/issues/1173) to clarify the intention and approach

kitten:
Awesome! Yea, I think `'before' | 'after'` is definitely the most intuitive solution :+1:

.changeset/little-crabs-sell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/exchange-graphcache': minor
+---
+
+Add mergeMode = 'before' | 'after' to simplePagination helper to define whether pages are merged before or after when you paginate forwards and backwards
```
kitten, line 5:
```suggestion
Add a `mergeMode: 'before' | 'after'` option to the `simplePagination` helper to define whether pages are merged before or after preceding ones when pagination, similar to `relayPagination`'s option
```

Just suggesting some details here since we can use full markdown in our changelog 😇

docs/graphcache/computed-queries.md
```diff
@@ -192,6 +192,30 @@ options in here `limitArgument` and `offsetArgument` these will default to `limi
 and `skip` respectively. This way you can use the keywords that you are using in
 your queries.
 
+You can also specify `mergeMode`, which defaults to `after` and can otherwise
+be set to `before`. This will handle how pages are merged when you paginate
```
kitten, line 196:
```suggestion
We may also add the `mergeMode` option, which defaults to `'after'` and can otherwise
be set to `'before'`. This will handle in which order pages are merged when paginating
```

We usually use "we" as often as possible when writing (with some exceptions). I realise the Graphcache pages may need some revamping to make sure they're as helpful as can be compared to some others, but let's take that into account here

kitten (approved):
There's one more section in the docs that needs updating, but I can totally do that after this PR myself if you'd like ✌️ https://github.com/FormidableLabs/urql/blob/main/docs/api/graphcache.md#simplepagination

Thanks for adding this!

hoangvvo:
Thanks for the review, I updated the documentation and also added to `graphcache.md`.

PR #1181, opened 2020-11-27
(react/preact) - Force suspense to be disabled for executeQuery calls

### LINKED ISSUE #1180: `rexecuteQuery` escapes Suspense boundary
Hey folks, found a little issue with suspense:

urql version & exchanges:
```
version: urql@1.11.3
exchanges: graphcache + built-ins
```

**Steps to reproduce**

- Go to https://codesandbox.io/s/gifted-bohr-honuf?file=/src/App.tsx
- After data comes in, click on the "Load again please" button above the list

**Expected behavior**

Refetching a query should trigger the suspense boundary

**Actual behavior**

Refetching a query completely bypasses Suspense and leads to an uncaught "exception" (the promise that should be suspended on) 🤷

kitten:
Fix #1180

### DIFF
--- a/packages/preact-urql/src/hooks/useQuery.ts
+++ b/packages/preact-urql/src/hooks/useQuery.ts
@@ -175,7 +175,9 @@ export function useQuery<Data = any, Variables = object>(
 
   // This is the imperative execute function passed to the user
   const executeQuery = useCallback(
-    (opts?: Partial<OperationContext>) => update(makeQuery$(opts)),
+    (opts?: Partial<OperationContext>) => {
+      update(makeQuery$({ suspense: false, ...opts }));
+    },
     [update, makeQuery$]
   );
 
--- a/.changeset/hungry-penguins-teach.md
+++ b/.changeset/hungry-penguins-teach.md
@@ -1,0 +1,6 @@
+---
+'@urql/preact': patch
+'urql': patch
+---
+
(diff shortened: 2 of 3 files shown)


packages/preact-urql/src/hooks/useQuery.ts
```diff
@@ -175,7 +175,9 @@ export function useQuery<Data = any, Variables = object>(
 
   // This is the imperative execute function passed to the user
   const executeQuery = useCallback(
-    (opts?: Partial<OperationContext>) => update(makeQuery$(opts)),
+    (opts?: Partial<OperationContext>) => {
+      update(makeQuery$({ suspense: false, ...opts }));
```
JoviDeCroock, line 179: Nit: Shouldn't we move suspense overrides to the back so users can't accidentally invoke the feature.

```suggestion
      update(makeQuery$({ ...opts, suspense: false }));
```

JoviDeCroock (approved):
Minor nit but look good to me

packages/preact-urql/src/hooks/useQuery.ts
kitten: I believe it should still be possible to force enable it, since there may be legitimate uses to trigger Suspense manually in the render thread of lower components... somehow 🤷

PR #1183, opened 2020-11-29
feat: Expose reponse headers inside OperationResult

juan267:
## Summary

Goal: Expose response `headers` inside `OperationResult`.  

Reason to do the change: 

Sometimes is necessary to capture the response headers after every request and fire some side effects based on them. 

An example of this is an Auth flow where the `access-token` changes after every request. The new token is send back by the server in an `access-token` response header. So is necessary to capture this header and save them so that the next request is sent with the correct headers. https://github.com/lynndylanhurley/devise_token_auth is an example of a popular JWT library following this scheme. 

## Set of changes

Affect packages: Core

Changes: 

- makeResult now includes `response.headers` in it's return value
- OperationResult interface now includes a `headers: Headers` property

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -69,6 +69,8 @@ export interface OperationResult<Data = any, Variables = any> {
   operation: Operation<Data, Variables>;
   /** The data returned from the Graphql server. */
   data?: Data;
+  /** The headers returned from the Graphql server response */
+  headers?: Headers;
   /** Any errors resulting from the operation. */
   error?: CombinedError;
   /** Optional extensions return by the Graphql server. */
--- a/packages/core/src/utils/result.ts
+++ b/packages/core/src/utils/result.ts
@@ -8,6 +8,7 @@ export const makeResult = (
 ): OperationResult => ({
   operation,
   data: result.data,
+  headers: response.headers,
   error: Array.isArray(result.errors)
     ? new CombinedError({
         graphQLErrors: result.errors,


kitten:
Hiya, we don't expose an unlimited amount of result properties on `OperationResult`. In fact, what's there is all that's defined in the spec for GraphQL results (albeit abstracted)

So I don't think we'll want to expose this on there and encourage adding more things to the result that aren't ultimately spec-compliant.

For operations there's the `context` but you'll also notice it's separate. For results there isn't since they closely match the spec.

There's several things that you could do on your end:

- Use `extensions` rather than headers in your API
- Create a custom `fetchExchange` based on `core`'s one and expose the headers
- Use cookies, etc, or wrap around `fetch` itself

Given multiple options are available I don't think this is the right approach to add to the core library 😅 Sorry!

Also please note, we do have an RFC process as part of our issue templates which allows more room to describe proposed changes and discuss them, which can avoid having the discussion "late" for a PR ✌️

juan267:
Hi @kitten thanks, for the quick feedback. Wanting to keep `OperationResult` spec compliant makes sense I will close this pr then. 

>Create a custom fetchExchange based on core's one and expose the headers

Yeah I currently have a working solution with this approach but creating a custom `fetchExchange` is a tedious process because if I wanted to preserve almost all the current `fetchExchange` code and only do this small change, I have to start copying private functions from utils etc.. which is not ideal. Would it be possible to allow users to extend the current fetchExchange like with `fetchOptions` but for `OperationResult`? 

that way the default `fetchExchange` can remain being spec compliant but users can if they want extend the `OperationResult` from it.

> it's also worth noting that an access token that changes for each request and is only valid for one may be a bad fit for GraphQL in general 

Hehe yeah I know the server can batch parallel request so that they all can use the same accessToken 😄.

PR #1187, opened 2020-12-02
(core) - Add built-in gql tag function

kitten:
## Summary

No RFC, since this is a rather small change that naturally emerged after #1186 

We can merge the functionality of `graphql-tag` into a leaner function in `@urql/core`. Our implementation does not check fragment names globally, so that it works well with `@urql/exchange-graphcache` but mostly checks them in the document that's currently generated.

The output is still compatible with `graphql-tag` and works similarly. There's potential for more optimisations, but the most important optimisation is that the function reuses the key-cache of `request.ts` in the same file.

The size increase is negligible and could be gained back with more refactors potentially (0.1kB gzip)


JoviDeCroock (approved):
This is superb

zachasme:
We tried replacing `graphql-tag` with this built-in tag, but as far as I can tell this doesn't do fragment de-duplication, which we rely on.

Do you still recommend using `graphql-tag` when de-duplication is needed, or is there another approach using the built-in tag?

kitten:
@zachasme Just so I don't miss the point here; I suppose we may have forgotten to handle duplicate _identical_ fragments? I'll look into that! That's a good point :+1:

zachasme:
Yes exactly, duplicate identical fragments. I was wondering if it was intentionally left out :)

kitten:
@zachasme well, I hope that with the patch this is going to work just fine now ✌️ I hope that didn't waste any time on your part and will work as expected

zachasme:
Wow that was fast, and it appears to be working--thanks a bunch, @kitten! :+1:

No time wasted, I only did a quick search+replace on our develop branch last week to try it out, and wrote the comment to figure out what was happening.

PR #1189, opened 2020-12-03
(graphcache) - Replace buildClientSchema with lean implementation

kitten:
## Summary

Our largest impact on `@urql/exchange-graphcache` is our usage of `graphql/utilities/buildClientSchema` which uses `GraphQLSchema` and other heavy implementations, which are basically server-side implementations with a lot of invariants. Given that we know that with `@urql/introspection` and other introspection methods the result can be trusted we don't even need any of the strict validation of `buildClientSchema` and what it pulls in.

Furthmore, we don't need any of the `graphql/type/definition` types, e.g. `GraphQLObjectType` and helpers like `isObjectType` if we transform the client schema to a format that is just good enough for our purposes.

This change increases `@urql/exchange-graphcache`'s size slightly from `6.74kB gzip` to `6.93kB gzip`. However, the impact on what's pulled in from `graphql` is significant and we save in `-9.61kB gzip` (up to; lowest, `8.53kB`). This makes a **net impact** of `-9.42kB gzip`

## Set of changes

- Reimplement `buildClientSchema` and replace `GraphQLSchema` types
- Started to drill down `buildClientSchema` to what's necessary, but then transitioned to just using string-based schema structure
- Update `store.ts` and `schemaPredicates.ts` to use new implementation

No unit tests had to be updated, which is encouraging.
The only remaining usage of `graphql` compared to `@urql/core` is `valueFromASTUntyped`

### DIFF
--- a/.changeset/nervous-beds-battle.md
+++ b/.changeset/nervous-beds-battle.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-graphcache': patch
+---
+
+Replace `graphql/utilities/buildClientSchema.mjs` with a custom-tailored, lighter implementation built into `@urql/exchange-graphcache`. This will appear to increase its size by about `0.5kB gzip` but will actually save around `-5.8kB gzip` in any production bundle by using less of `graphql`'s code.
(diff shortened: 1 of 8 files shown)


JoviDeCroock (approved):
Awesome work, wooooooooow

PR #1214, opened 2020-12-10
Add compatibility for Wonka compiled with ReScript >= 8

Kingdutch:
## Summary
It took a bit of investigating but it turns out that when Wonka is compiled with ReScript >= 8 it's variant representation changes. Urql has one location where it manually encodes a variant in `userQuery`, thus this cause problems when Urql expects Wonka to be compiled with ReScript < 8, but that's not the case.

The full investigation story can be found in https://github.com/FormidableLabs/reason-urql/issues/232

I [searched for the usage of `.tag` in the codebase](https://github.com/FormidableLabs/urql/search?p=6&q=.tag) to find where this manual encoding was performed and only found this in `react-urql` and `preact-urql`.

A better change than my fix would be to let Wonka encode the variant (perhaps we can expose a function?) so that it automatically uses whatever encoding ReScript decided. This more properly decouples Urql from ReScript's internal decisions and avoids breakage in the future if they decide to further change this representation.

## Set of changes

Urql's useQuery for React and Preact now encodes it's cache results in a manner compatible with ReScript >= 8 as well as ReScript < 8 to ensure that Urql doesn't break when Wonka has been compiled with newer ReScript versions.

### DIFF
--- a/packages/preact-urql/src/hooks/useQuery.ts
+++ b/packages/preact-urql/src/hooks/useQuery.ts
@@ -74,8 +74,14 @@ function toSuspenseSource<T>(source: Source<T>): Source<T> {
     // If we haven't got a previous result then start suspending
     // otherwise issue the last known result immediately
     if (cache !== undefined) {
-      const signal = [cache] as [T] & { tag: 1 };
+      const signal = [cache] as [T] & { tag: 1, TAG: 1, _0: T };
       signal.tag = 1;
+      // ReScript has new variant encoding from version 8 onwards.
+      // Mirror the variant encoding above to a compatbile version
+      // so urql works with Wonka no matter with which version of
+      // ReScript it was compiled.
+      signal.TAG = signal.tag;
+      signal._0 = signal[0];
       sink(signal);
     } else {
       hasSuspended = true;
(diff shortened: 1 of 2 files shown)


kitten:
First of all; I do have to politely ask you to read the `CONTRIBUTING.md` guide for PRs and documenting changes. Furthermore, we may need an issue for this.

I was aware that this changes some intervals, but using the new version of Wonka is deeply unsettling to me, since, compiled with BS 8 / RS, it isn't compatible with any other version of it.

Now BuckleScript has always been making "curious choices" in terms of transpilation and not using bundles / packages properly which it really should.

This wouldn't be an issue, but were sometimes already encountering people that have duplicate installations of `@urql/core`.
This is less severe since this will most of the times only cause type issues.

With Wonka it could be the same, except that the two different ways of compiling these types are inherently incompatible with one another. Which is suboptimal given that different versions of exchanges could be installed.

So this does fix it, but opens up a new problem that we didn't want to address just yet.

I'll close it for now, since it's unlikely to completely solve this whole problem, but I'm aware that it would _appear_ to fix it temporarily.

That being said, the simplest thing to do on your end is not to use suspense-mode if at all possible with `reason-urql`

kitten:
More info here: https://github.com/FormidableLabs/reason-urql/issues/232#issuecomment-742453847 So an additional issue in this repo may not be needed yet, since we haven't even scheduled a `wonka` upgrade or changes thereof as part of our RFC pipeline.

Issue #1228, opened 2020-12-13
docs: unclear mention of `getIntrospectionQuery` in schema-awareness.md


Kingdutch:
The [documentation mentions](https://github.com/FormidableLabs/urql/blame/main/docs/graphcache/schema-awareness.md#L110) that `getIntrospectionQuery` can take various arguments and is a convenience helper and unneeded in the example. However, the example doesn't provide it with any of the mentioned arguments (instead gives `{ description: false }`). 

The line in question is probably talking about another function. It probably means to say that `minifyIntrospectionQuery` takes many different arguments and that `getIntrospectedSchema` is there for convenience but isn't actually needed?

Given the example it's somewhat unclear what the paragraph is trying to explain.

JoviDeCroock:
Hey,

This isn't really a bug so let's handle this in a discussion next time, you are indeed correct. We should rephrase this to talk about `minify` and that it's an optimization

Kingdutch:
Ah sorry about that, I saw another docs issue so I wasn't sure where it qualified ^^

kitten:
I think that's a typo and it's supposed to say `getIntrospectedSchema`

PR #1229, opened 2020-12-13
(docs) - Change wording in schema-awareness to reflect intent of minify

JoviDeCroock:
This changes the wording of `schema-awareness.md` to reflect what `minifyIntrospectionQuery` can take as an argument and that it's not a requirement to make schema-awarenss work, it is an optimization step.

Closes https://github.com/FormidableLabs/urql/issues/1228

### DIFF
--- a/docs/graphcache/schema-awareness.md
+++ b/docs/graphcache/schema-awareness.md
@@ -107,9 +107,9 @@ fetch('http://localhost:3000/graphql', {
   });
 ```
 
-The `getIntrospectionQuery` doesn't only accept `IntrospectionQuery` JSON data as inputs, but also
+The `minifyIntrospectionQuery` doesn't only accept `IntrospectionQuery` JSON data as inputs, but also
 allows you to pass a JSON string, `GraphQLSchema`, or GraphQL Schema SDL strings. It's a convenience
-helper and not needed in the above example.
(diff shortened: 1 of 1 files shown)


docs/graphcache/schema-awareness.md
```diff
@@ -107,9 +107,9 @@ fetch('http://localhost:3000/graphql', {
   });
 ```
 
-The `getIntrospectionQuery` doesn't only accept `IntrospectionQuery` JSON data as inputs, but also
+The `minifyIntrospectionQuery` doesn't only accept `IntrospectionQuery` JSON data as inputs, but also
```
kitten, line 110: I think this was meant to say `getIntrospectionSchema` as it allows the conversion of multiple inputs to the JSON format that `minifyIntrospectionSchema` expects 🤔

Kingdutch:
With the latest change I'm still a bit puzzled what the convenience method `getIntrospectionSchema` is for, especially since the example shows it being used, but if it's not needed, why use it? 

Perhaps we can extend 
> It's a convenience helper and not needed in the above example.

To something like
> It's a convenience helper [that normalises the type for minifyIntrospectionSchema or something completely different] and not needed in the above example.

to better illustrate why it's not needed in the above example (and allow users to figure out if they do/do not need this in their code).

Issue #1213, opened 2020-12-10
next-urql: Switch from NextPageContext to PartialNextContext breaks usage with NextPage type


impulse:
Nextjs `NextPage` uses `NextPageContext` to type the context. With the recent update in urql to `PartialNextContext`, this breaks passing in a component typed with `NextPage` into `withUrqlClient`.

urql version & exchanges:
```
"@urql/devtools": "^2.0.2",
"@urql/exchange-graphcache": "^3.3.2",
"next": "10.0.3",
"next-urql": "^2.1.1",
"urql": "^1.11.4",
```
**Steps to reproduce**

Use the official Nextjs example <https://nextjs.org/docs/api-reference/data-fetching/getInitialProps#typescript> and pass it in `withUrqlClient`.

```tsx
import { NextPage } from "next";
import { withUrqlClient } from "next-urql";
import { createUrqlClient } from "../lib/createUrqlClient";

interface Props {
  userAgent?: string;
}

const Page: NextPage<Props> = ({ userAgent }) => (
  <main>Your user agent: {userAgent}</main>
);

Page.getInitialProps = async ({ req }) => {
  const userAgent = req ? req.headers["user-agent"] : navigator.userAgent;
  return { userAgent };
};

export default withUrqlClient(createUrqlClient)(Page);
```

**Expected behavior**
No type errors

**Actual behavior**
```
Argument of type 'FunctionComponent<Props> & { getInitialProps?(context: NextPageContext): Props | Promise<Props>; }' is not assignable to parameter of type 'NextComponentType<PartialNextContext, {}, {}>'.
  Type 'FunctionComponent<Props> & { getInitialProps?(context: NextPageContext): Props | Promise<Props>; }' is not assignable to type 'FunctionComponent<WithUrqlProps> & { getInitialProps?(context: PartialNextContext): {} | Promise<{}>; }'.
    Type 'FunctionComponent<Props> & { getInitialProps?(context: NextPageContext): Props | Promise<Props>; }' is not assignable to type '{ getInitialProps?(context: PartialNextContext): {} | Promise<{}>; }'.
      Types of property 'getInitialProps' are incompatible.
        Type '((context: NextPageContext) => Props | Promise<Props>) | undefined' is not assignable to type '((context: PartialNextContext) => {} | Promise<{}>) | undefined'.
          Type '(context: NextPageContext) => Props | Promise<Props>' is not assignable to type '(context: PartialNextContext) => {} | Promise<{}>'.
[...]

PR #1230, opened 2020-12-13
(next) - use the App and PageContext types provided by Next

JoviDeCroock:
## Summary

This uses the built-in types provided by Next to type the context used in our `getInitialProps`.

Fixes: https://github.com/FormidableLabs/urql/issues/1213

Testable at: https://codesandbox.io/s/twilight-rgb-vxuwu?file=/pages/index.tsx

## Set of changes

- Type getInitialProps with `AppContext` and `PageContext`

### DIFF
--- a/packages/next-urql/package.json
+++ b/packages/next-urql/package.json
@@ -42,7 +42,6 @@
     "enzyme": "^3.11.0",
     "enzyme-adapter-react-16": "^1.15.2",
     "graphql": "^15.4.0",
-    "next": "^10.0.3",
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
     "react-is": "^17.0.1"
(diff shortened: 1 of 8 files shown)


packages/next-urql/package.json
```diff
@@ -42,6 +42,7 @@
     "enzyme": "^3.11.0",
     "enzyme-adapter-react-16": "^1.15.2",
     "graphql": "^15.4.0",
+    "next": "^10.0.3",
```
kitten, line 45: I think there's a PR that we could basically revert to get here. However, the original intention was to make sure that we don't need the `next` package or depend on it as it literally comes with hundreds of its own dependencies for builds and etc some conflicting with our own 😅

We could look into why the generic isn't working as intended

JoviDeCroock:
Closing since this doesn't solve the issue, this will most likely relate to using the generic available on `withUrql`

Issue #1272, opened 2021-01-05
Client may reuse results during SSR when typeof window isn't undefined


juanallo:
When updating to the latest packages on URQL I noticed a potential regression in SSR. When rendering a react tree using `react-ssr-prepass` and URQL the same query in 2 different requests gets back the same response and does not trigger a fetch call the second time. 

urql version & exchanges:
    "react": "17.0.1",
    "react-dom": "17.0.1",
    "react-ssr-prepass": "1.2.1",
    "urql": "1.11.4",    
     "@urql/core": "1.16.1",
    "graphql": "15.4.0",

**Steps to reproduce**

1. Open this sandbox project: https://codesandbox.io/s/urql-cross-request-cache-issue-jbfws
2. Run the test cases (more details on the README)

There are 3 test cases:
- The first 2 are exactly the same. Mock fetch, create a client, render an APP that makes a query. 
- The last test case uses 2 different clients in the same test cases and mounts the same App with the different clients, expecting 2 calls to happen

**Expected behavior**

I would expect each client instance to handle their own cache so each SSR should make a call to fetch and render the data. 
- The second test case fails because the fetch call never happens (Probably due to the first test response being cached and served in the second render)

**Actual behavior**

Each individual request calling a different client instance should result in invidual requests to fetch and those should not be cached across instances

PR #1275, opened 2021-01-05
(react/preact) - Fix issue where ssr would leak across renders

JoviDeCroock:
## Summary

When we'd invoke two parallel renders with two different clients  the global in `useQuery` would cause us to share data between the clients.

Resolves https://github.com/FormidableLabs/urql/issues/1272
Sandbox: https://codesandbox.io/s/urql-cross-request-cache-issue-forked-sx6oe?file=/src/useQuery.js

## Set of changes

- register sources on the urql-client

### DIFF
--- a/packages/preact-urql/src/hooks/useQuery.ts
+++ b/packages/preact-urql/src/hooks/useQuery.ts
@@ -90,7 +90,8 @@ function toSuspenseSource<T>(source: Source<T>): Source<T> {
 const isSuspense = (client: Client, context?: Partial<OperationContext>) =>
   client.suspense && (!context || context.suspense !== false);
 
-const sources = new Map<number, Source<OperationResult>>();
+type Sources = Map<number, Source<OperationResult>>
+const sourcesByClient = new WeakMap<Client, Sources>();
 
 export function useQuery<Data = any, Variables = object>(
   args: UseQueryArgs<Variables, Data>
@@ -103,6 +104,9 @@ export function useQuery<Data = any, Variables = object>(
   // Create a new query-source from client.executeQuery
   const makeQuery$ = useCallback(
     (opts?: Partial<OperationContext>) => {
+      let sources = sourcesByClient.get(client);
+      if (!sources) sourcesByClient.set(client, (sources = new Map()))
+
       // Determine whether suspense is enabled for the given operation
       const suspense = isSuspense(client, args.context);
       let source: Source<OperationResult> | void = suspense
@@ -182,6 +186,9 @@ export function useQuery<Data = any, Variables = object>(
   );
 
   useEffect(() => {
+    let sources = sourcesByClient.get(client);
+    if (!sources) sourcesByClient.set(client, (sources = new Map()))
+
     sources.delete(request.key); // Delete any cached suspense source
     if (!isSuspense(client, args.context)) update(query$);
   }, [update, client, query$, request, args.context]);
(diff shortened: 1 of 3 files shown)


tazsingh (approved):
🚀

packages/preact-urql/src/hooks/useQuery.ts
```diff
@@ -90,7 +90,8 @@ function toSuspenseSource<T>(source: Source<T>): Source<T> {
 const isSuspense = (client: Client, context?: Partial<OperationContext>) =>
   client.suspense && (!context || context.suspense !== false);
 
-const sources = new Map<number, Source<OperationResult>>();
+type Sources = Map<number, Source<OperationResult>>;
+const sourcesByClient = new WeakMap<Client, Sources>();
```
JoviDeCroock, line 94: This will possibly heap a lot though so it might be worth to moving this to the `client` itself?

juanallo: Thanks for starting work on this bug so fast!!

My feeling is that putting these caches on the client instance would be the best bet. I saw this line and figure it could be related but I still don't fully understand the context in which this cache is used. Why wouldn't the cache or dedup exchanges take care of this case?

kitten: I suppose this implementation detail may even change as we experiment with more implementations of suspense?

Regarding this PRs relation to #1271, why is this relevant if we don't wish to use this temporary cache on the server-side? Has something gone wrong in the SSR setup where it's accidentally active again?

JoviDeCroock: That is a good point, the error we're seeing is only because of the tests running in a browser environment. This wouldn't happen in a real node environment. 😅

juanallo: The tests I shared are running in a node sandbox.

juanallo: https://codesandbox.io/s/urql-cross-request-cache-issue-jbfws

juanallo: This also was not happening on previous versions. Only when I updated to the latest in a project I got similar tests to fail, so I created a standalone version of that and eventually ported that to codesandbox to be able to quickly share it. If you download the files from the sandbox and you run them in your computer (jest running in node) then you will see the tests failing too. 

I believe the problem is that this cache `const sources = new Map<number, Source<OperationResult>>();` is in the file scope, so once you imported the file it will be cached for as long as the node process lives. So in the case where you have a server running SSR that gets different requests you are likely to see the issues I represented in the test case.

JoviDeCroock: When you run your tests we see `typeof window === 'object'` which means it's being run in a browser context. The sources are unused in a Node context (`typeof window === 'undefined'`)

juan-allo-ron-ck: Ohh so does it mean that because I mocked window.fetch the way I did then I am seeing the issues?

juanallo: Ohh, so it is probably because I mocked window.fetch the way I did that caused it to fail then? Interesting. I will update that on my tests and try it out.

kitten: So I mean, following the principle of least surprise this probably wasn't a great choice of implementations on our part, but it's rather temporary since we'll also need to figure out other suspense edge cases

juanallo: Thanks for responding so quickly on this "bug". Makes me think that moving that cache to the client is the safest too, but I defer that to you :)

JoviDeCroock:
We're going to be handling this in another refactor.

PR #1280, opened 2021-01-06
(docs) - correct next-urql getStaticProps example

### LINKED ISSUE #1279: next-urql - getStaticProps doesn't provide static data to useQuery
Following the example from the [documentation](https://formidable.com/open-source/urql/docs/advanced/server-side-rendering/#using-getstaticprops-or-getserversideprops) which seems to be wrong (tries using the ssrCache exchange without defining it first and a missing comma) I am able to get the query running server side but the client side does an independent query as it doesn't seem to have access to urqlState.
Am I missing something?

urql version & exchanges:

**Steps to reproduce**

https://gist.github.com/pocorschi/0ba2a6c699df917e8a6120acdfb78c30

**Expected behavior**
The query should be executed with a request on server and it should be read from cache on client.
**Actual behavior**
The query gets executed with a request both on server and client side.

pocorschi:
ref: https://codesandbox.io/s/condescending-kapitsa-9p8y7?file=/pages/index.js:548-702

## Summary

The docs referring to usage in getStaticProps was incorrect.
Resolves: #1279

## Set of changes
Fixed example of getStaticProps usage.

### DIFF
--- a/docs/advanced/server-side-rendering.md
+++ b/docs/advanced/server-side-rendering.md
@@ -265,11 +265,6 @@ const TODOS_QUERY = `
   query { todos { id text } }
 `;
 
-const createUrqlClient = (ssr, ctx) => ({
-  url: "your-url"
-  exchanges: [dedupExchange, cacheExchange, ssrCache, fetchExchange]
-});
-
 function Todos() {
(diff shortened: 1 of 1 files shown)


JoviDeCroock (approved):
Nice work, thank you!

PR #1281, opened 2021-01-07
(docs) add a search input to the sidebar

### LINKED ISSUE #1064: docs: Add a search bar to our documentation site
## Summary

The sidebar should have basic search capability. Similar to that available in other Formidable open source, such as Victory https://formidable.com/open-source/victory/docs/. To start off with, the design/functionality should more or less mirror the implementation for Victory. Partly for consistency purposes, and partly because a search located elsewhere will require changes to the mobile design.

The search input should function by searching keywords against page headings and filtering the links in the sidebar. Our `react-static-plugin-md-pages` package has a small hook that can return information and headings to fulfil this requirement: https://github.com/kitten/react-static-plugin-md-pages/blob/0f3cd06da8102ab2038765068b6ec341d2fc118f/src/index.js#L51-L53

### Rough Design
urql sidebar with search

## Set of Changes

- [ ] Create search input in the sidebar
- [ ] Change the displayed pages in the sidebar based on the value of the input

wgolledge:
Resolves #1064 

## Summary

The reasoning behind this change is mostly explained in the above issue. 

We want to add the capability to search the docs for relevant content based on the page title or subheading(s).

## Set of changes

- add `< SidebarSearchInput />`
- filter sidebar headings and subheadings based on input 
- while searching, include matching H3 headers in the sidebar
- highlight matching substrings by wrapping them in `<strong>`
- add `fuse.js` fuzzy-search dep
- increase the sidebar width

searching via the sidebar

## Outstanding points

- Depending on how we like the current design we may want to relocate the search bar to the header, showing relevant matches in a dropdown as opposed to the sidebar

### DIFF
--- a/packages/site/package.json
+++ b/packages/site/package.json
@@ -32,6 +32,7 @@
   "dependencies": {
     "@mdx-js/react": "^1.6.22",
     "formidable-oss-badges": "0.3.5",
+    "fuse.js": "^6.4.6",
     "history": "^4.7.2",
     "path": "^0.12.7",
     "preact": "^10.5.7",
--- a/packages/site/src/components/navigation.js
+++ b/packages/site/src/components/navigation.js
@@ -10,11 +10,11 @@ export const SidebarContainer = styled.div`
   right: 0;
   bottom: 0;
   min-height: 100%;
+  width: ${p => p.theme.layout.sidebar};
 
   @media ${({ theme }) => theme.media.sm} {
     display: block;
     position: relative;
-    width: ${p => p.theme.layout.sidebar};
     margin-left: calc(2 * ${p => p.theme.layout.stripes});
   }
 `;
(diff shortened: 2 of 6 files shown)


kitten:
I think what's left from a functionality point is that it doesn't return H3 header data? I think that's a pretty important point since some items are more descriptively listed under those and all H2 headers are represented in the sidebar already 😅 

On a design note, maybe we can check how this looks when the inner text is aligned with the sidebar text, i.e. with negative horizontal margin. I'd also test out a larger margin on the search itself as it needs to keep its distance from the logo and also has a reduced distance to the first sidebar item compared to the other items' distance to each other.

wgolledge:
Yeah agreed that H3 is fairly important. Do you think we should give any design indication that a H3 has been matched? Otherwise we can just show the parent and if there's no direct match it can be assumed that it must be a child H3 match.

Good points on the design front, will do some more testing!

kitten (approved):

PR #1283, opened 2021-01-08
docs: added 'Updating the cache after a mutation'

bacebu4:
## Summary

After scrolling through all the docs haven't found any mentions about automatic cache updates after a mutation. but I think it's really important to mention what mutation should return specific fields in order to automatic updates come to action.

## Set of changes

Added a brief paragraph about this feature so it's more understandable to newcomers.

### DIFF
--- a/docs/basics/mutations.md
+++ b/docs/basics/mutations.md
@@ -45,6 +45,11 @@ Unlike the `useQuery` hook, the `useMutation` hook doesn't execute automatically
 our example, no mutation will be performed. To execute our mutation we instead have to call the
 execute function — `updateTodo` in our example — which is the second item in the tuple.
 
+### Updating the cache after a mutation
+
+If a mutation updates a single existing entity, `urql` can automatically update that entity’s value in its cache. To do so, the mutation must return the id of the modified entity, along with the values of the fields that were modified. You can clearly see that in the example above.
+However if mutation affects other entities, you may want to [set up manual updates that react to mutations.](https://formidable.com/open-source/urql/docs/graphcache/custom-updates/)
+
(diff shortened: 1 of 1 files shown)


JoviDeCroock:
Hey @bacebu4 

This isn't entirely correct, in the basics section we assume the user to use `document-caching`, this doesn't rely on the `__typename:id` combination that is our normalized cache named graphCache.

The updating mechanism for `document-caching` can be found here https://formidable.com/open-source/urql/docs/basics/document-caching/
The updating mechanism you describe for `normalized caching` can be found here https://formidable.com/open-source/urql/docs/graphcache/custom-updates/

bacebu4:
> Hey @bacebu4 
> 
> 
> 
> This isn't entirely correct, in the basics section we assume the user to use `document-caching`, this doesn't rely on the `__typename:id` combination that is our normalized cache named graphCache.
> 
> 
> 
> The updating mechanism for `document-caching` can be found here https://formidable.com/open-source/urql/docs/basics/document-caching/
> 
> The updating mechanism you describe for `normalized caching` can be found here https://formidable.com/open-source/urql/docs/graphcache/custom-updates/

Gotcha!
Though still in the "Custom updates" section it's not so obvious for me that it can automatically update single entities. Maybe that's because I'm not native speaker...

Thanks for the reply!

JoviDeCroock:
Hmm you are right in that, we might want to consider renaming it to `cache-updates` instead.

kitten:
Closing this since as discussed the change isn't quite in the right place; if you'd like to suggest a change as per the discussed stuff above, let's reopen this ❤️

PR #1307, opened 2021-01-14
(docs) clarify necessity of `__typename` for abstract types in readFr…

Kingdutch:
…agment

## Summary

If you're using a fragment on an abstract type in `readFragment` then you must specify the concrete type of the entity that you're trying to match. Not doing so will cause a cache miss.

## Set of changes

This PR updates the `readFragment` documentation to add this snippet of information.

### DIFF
--- a/docs/api/graphcache.md
+++ b/docs/api/graphcache.md
@@ -315,8 +315,9 @@ cache.readFragment(
 ); // Data or null
 ```
 
-Note that the `__typename` may be left out on the partial entity, since the `__typename` is already
-present on the fragment itself.
+Note that the `__typename` may be left out on the partial entity if the fragment isn't on an
+interface or union type, since in that case the `__typename` is already present on the fragment
+itself.
 
 If any fields on the fragment require variables, you can pass them as the third argument like so:
 


kitten:
I'm kind of concerned that the wording here is cryptic so that no additional information is added. By definition and by what's allowed on the backend, it's already possible to deduct that a type name cannot be a union or interface name. Also if someone knows what "concrete type@ means in this case they're already likely to have known or be able to assume this

Kingdutch:
I am very happy to use different wording but this was the best I could come up with :) It tripped me up somewhat which is why I proposed the change.

> By definition and by what's allowed on the backend, it's already possible to deduct that a type name cannot be a union or interface name.

I'm not sure what you mean with this? A fragment on an abstract type is still a valid fragment. 

The documentation currently suggests that readFragment will work without specifying the typename. However, if you're doing generalised work for an interface (or union) then that's no longer true, even though the fragment itself is valid. You'll receive `null` as your result (since GraphCache has no data stored for the abstract type) which could be hard to track down.

In my current experience there are all sorts of reasons why readFragment could return null (e.g. requesting a non-nullable field that's removed from other queries, thus not in the cache) when you're learning the ropes, so I believe that having this caveat pointed out is useful.

kitten:
I see what you want to clarify here, but I'd opt for phrasing like "if the fragment isn't on an interface name." But my point was that no entity ever has a `__typename` field that is not a type, so phrasing this in a roundabout manner could be more confusing than helpful 😅

Kingdutch:
I've updated the text based on your feedback to read
> Note that the `__typename` may be left out on the partial entity if the fragment isn't on an interface or union type, since in that case the `__typename` is already present on the fragment itself.

I'm not sure if we should make that "in that case the [correct] `__typename`" or "in that case the [concrete] `__typename`" since technically "a type" is always present for the fragement.

kitten (approved):

kitten:
I think it could say "the entity's `__typename`" but I think it's clear enough which one it's referring to.

Issue #1297, opened 2021-01-13
relayPagination incorrectly handles subsequent uncursored queries with larger `last` values


Kingdutch:
When using `relayPagination`, requesting the last item in a list and then subsequently requesting the last N (N>1) items in a list, the last item will be put in the front of the other results, rather than as last item. 

A scenario where this occurred was in rendering a list of conversations which wants to display the last message as a preview. Only upon opening a conversation will the last N messages be loaded for display.

urql version & exchanges: `main` branch, Graphcache

**Steps to reproduce**

0. Given a source of [1, 2] (where 2 is newer than 1 and thus later in the list).
1. Request the last result.
2. Request the last 2 results.

**Expected behavior**
[1, 2]

**Actual behavior**
[2, 1]

**Reproduction test**

I've written a test in `relayPagination.test.ts` that demonstrates the issue.

```typescript
it('handles subsequent queries with larger last values', () => {
  const Pagination = gql`
    query($last: Int!) {
      __typename
      items(last: $last) {
        __typename
        edges {
          __typename
          node {
            __typename
            id
          }
        }
        nodes {
          __typename
          id
        }
        pageInfo {
          __typename
          hasPreviousPage
          startCursor
        }
      }
    }
  `;

  const store = new Store({
    resolvers: {
      Query: {
        items: relayPagination(),
      },
    },
  });

  const pageOne = {
    __typename: 'Query',
    items: {
      __typename: 'ItemsConnection',
      edges: [itemEdge(2)],
      nodes: [itemNode(2)],
      pageInfo: {
        __typename: 'PageInfo',
        hasPreviousPage: true,
        startCursor: '2',
      },
    },
  };

  const pageTwo = {
    __typename: 'Query',
    items: {
      __typename: 'ItemsConnection',
      edges: [itemEdge(1), itemEdge(2)],
      nodes: [itemNode(1), itemNode(2)],
      pageInfo: {
        __typename: 'PageInfo',
        hasPreviousPage: false,
        startCursor: '1',
      },
    },
  };

  write(store, { query: Pagination, variables: { last: 1 } }, pageOne);
  write(store, { query: Pagination, variables: { last: 2 } }, pageTwo);

  const res = query(store, { query: Pagination, variables: { last: 2 } });

  expect(res.partial).toBe(false);
  expect(res.data).toEqual(pageTwo);
});
```
<details><summary>Test output</summary>
<p>

[...]

PR #1311, opened 2021-01-14
(graphcache) - Relay backwards pagination

JoviDeCroock:
Resolves https://github.com/FormidableLabs/urql/issues/1297

## Summary

When we'd have two backwards overlapping pages the results would start mixing up.

## Set of changes

- Reverse the backwards pagination node - edge concatenation

### DIFF
--- a/.changeset/strange-scissors-swim.md
+++ b/.changeset/strange-scissors-swim.md
@@ -2,4 +2,4 @@
 '@urql/exchange-graphcache': patch
 ---
 
-Fix relay-pagination would bug out when requesting overlapping pages through the last parameter.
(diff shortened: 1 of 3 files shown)


.changeset/strange-scissors-swim.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/exchange-graphcache': patch
+---
+
+Fix relay-pagination would bug out when requesting overlapping pages through the last parameter.
```
kitten, line 5:
```suggestion
Fix a Relay Pagination edge case where overlapping ends of pages queried using the `last` argument would be in reverse order.
```

kitten (approved):

PR #1318, opened 2021-01-16
(docs) - Redirects

JoviDeCroock:
This is an attempt at redirects for the pages we've moved around in our docs-split up

### DIFF
--- a/packages/site/static.config.js
+++ b/packages/site/static.config.js
@@ -52,23 +52,23 @@ export default {
     },
     {
       path: '/docs/concepts/core-package',
-      redirect: '/docs/basics/core',
+      redirect: 'docs/basics/core',
     },
     {
       path: '/docs/basics/getting-started',
-      redirect: '/docs/basics',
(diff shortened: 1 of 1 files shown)


wgolledge:
Good idea!

JoviDeCroock:
Currently not working yet it doesn't respect the base-url yet.

kitten:
@JoviDeCroock wondering what this outputs 🤔 if we want the proper redirection codes we may have to look into some changes on the AWS bucket side

JoviDeCroock:
@kitten currently redirects to the absolute url (without normal prefix) wanted to introduce this as an in-between fix so we could introduce the split-up docs already.

kitten:
This kind of seems to work as it uses `http-equiv="refresh"` as expected. The generation of them seems to be extremely flawed though and it outputs: `/open-source/urql//docs/basics/core`, which isn't normalised or correct in this case.

kitten (approved):

PR #1324, opened 2021-01-18
docs(subscriptions): change websocket protocol prefix

impulse:
## Summary
Just tried using GraphQL Subscriptions with `subscriptions-transport-ws`. I can only get it to work when using `ws://`. No response whatsoever when using the `wss://` prefix. Not sure what is causing this because they should specify the same protocol, but I guess it would be a good change either way for consistency with other resources like the Apollo Docs.

## Set of changes

Changed `wss://` to `ws://`


JoviDeCroock:
ws is commonly used for localhost setups for remotes you'll need to use wss (equivalent of https)

impulse:
> ws is commonly used for localhost setups for remotes you'll need to use wss (equivalent of https)

since the doc uri's contain localhost, it would make sense to use `ws://` then 👍🏻

kitten (approved):

PR #1356, opened 2021-02-02
(graphcache) - Handle fields with associated GraphQLError as cache misses and provide errors to updaters

### LINKED ISSUE #1344: Using offlineExchange with storage option using metaData methods swallows errors
I am experiencing an issue when trying to utilise offline caching for gql client. I have project setup to persist cache to a storage solution and include the metaData methods when I do this, queries that fail, the error value is always undefined I am never seeing a `CombinedError` object.

image

If I comment out `readMetadata` inside the function to `makeLocalStorage` which is used inside `offlineExchange` then it works.

I have set up a code sandbox with a working example of the issue. The example is set up to hit an endpoint that doesn't exist to it "fails" always.

https://codesandbox.io/s/urql-bug1-nlzbf?file=/src/graphql/makeLocalStorage.ts

urql version & exchanges:
|package|version|
|:---|:---|
|urql|1.11.6|
|@urql/exchange-graphcache|3.4.0|
|graphql|15.5.0|

**Steps to reproduce**
1. Load code sandbox example
- Observe the error is undefined (prints null to page)
2. Comment out `readMetadata` inside `/src/graphql/makeLocalStorage.ts`
- Observe the error is now a `CombinedError`

I added a clear cache button to page as you need to clear storage if you re-enable `readMetadata` else error is persisted and "appears" to works.

I have also noticed that if readMetadata is enabled the request is always doing a  "cache-only" request even though I set in my query for it to be "cache-and-network"

kitten:
Resolve #1344

## Summary

This change attempts to improve Graphcache's error handling and error awareness. Instead of ignoring the added information that `GraphQLError#path` may provide for individual null'ed out fields, Graphcache now processes these errors and dynamically replaces `null` values (links and records) as it writes them with `undefined` — i.e. uncached / cache misses — when an associated `GraphQLError` has been received for a given field.

## Set of changes

- Add `Context.__internal.path` tracking the current data path via aliases and indices
- Add `Context.__internal.errorMap` tracking the current `CombinedError`'s `GraphQLError`s by their paths
- Add `Context.error` which shows an updater its current field's error if any exists
- Replace `fieldValue` in `write` with `undefined` rather than `null` if the current field has an associated error to it

Size change from `6.87kB gzip` to `7.07kB gzip` (an increase of `0.2kB gzip`)

### DIFF
--- a/exchanges/graphcache/src/operations/write.ts
+++ b/exchanges/graphcache/src/operations/write.ts
@@ -299,20 +299,20 @@ const writeSelection = (
     }
 
     if (isRoot) {
-      // We have to update the context to reflect up-to-date ResolveInfo
-      updateContext(
-        ctx,
-        data,
-        typename,
-        typename,
-        joinKeys(typename, fieldKey),
-        fieldName
-      );
-
       // We run side-effect updates after the default, normalized updates
       // so that the data is already available in-store if necessary
       const updater = ctx.store.updates[typename][fieldName];
       if (updater) {
+        // We have to update the context to reflect up-to-date ResolveInfo
+        updateContext(
+          ctx,
+          data,
+          typename,
+          typename,
+          joinKeys(typename, fieldKey),
+          fieldName
+        );
+
         data[fieldName] = fieldValue;
         updater(data, fieldArgs || {}, ctx.store, ctx);
       }
--- a/exchanges/graphcache/src/operations/write.ts
+++ b/exchanges/graphcache/src/operations/write.ts
@@ -246,9 +246,14 @@ const writeSelection = (
       }
     }
 
-    if (fieldName === '__typename') {
-      continue;
-    } else if (ctx.optimistic && isRoot) {
+    // We simply skip all typenames fields and assume they've already been written above
+    if (fieldName === '__typename') continue;
+
+    // Add the current alias to the walked path before processing the field's value
+    ctx.path.push(fieldAlias);
+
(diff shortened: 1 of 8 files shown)


exchanges/graphcache/src/operations/write.ts
```diff
@@ -246,9 +255,14 @@ const writeSelection = (
...
     }
 
-    if (fieldName === '__typename') {
-      continue;
-    } else if (ctx.optimistic && isRoot) {
+    // We simply skip all typenames fields and assume they've already been written above
+    if (fieldName === '__typename') continue;
+
+    // Add the current alias to the walked path before processing the field's value
+    ctx.__internal.path.push(fieldAlias);
```
wgolledge, line 262: I feel like the pushing / popping to `path` is quite verbose. But it seems fine and I can't think of a better way of doing it off the top of my head

wgolledge (approved):
This is awesome! Definitely a nice upgrade and I'm a big fan of handling errors consistently with `undefined`

exchanges/graphcache/src/operations/write.ts
kitten: Yea, this was the simplest implementation I could come up with. I tried to merge it into the `SelectionIterator` stuff, but then it wouldn't capture the indices, so this is indeed the simplest way to get that path stack in there

JoviDeCroock (approved):

PR #1367, opened 2021-02-05
(graphcache) - Fix uncached GraphQLError fields making it impossible to get first API result

kitten:
This is a follow-up for #1356

## Summary

The #1356 change introduced behaviour that prevents `null` values from fields that have errored and are associated with `GraphQLError` from being cached. This caused a regression where when we "read-back" results from the cache that have just arrived from the API these fields would cause a cache miss:

https://github.com/FormidableLabs/urql/blob/6a687091b67dc3d0ee78353ca79b0aa401ba983c/exchanges/graphcache/src/cacheExchange.ts#L237-L243

We can prevent this by also tracking errored fields in the `query` and ignoring these uncached fields there too by replacing them with `null` while we still have access to the `GraphQLError`s.
This is preferable to other solutions as we've gotten rid of comparing the cached result to the API result in #1196 as we had trouble differentiating between `null` selection sets from previous selections and `null` values from the API result.

No changeset is needed since the PR is unreleased.

## Set of changes

- Add (previously failing) test for retrieving a result with `null`'d fields due to `GraphQLError`s from an API
- Track `ctx.__internal.path` in all environments in `query` too
- Check for errors before aborting a selection in `query` and set the field to `null` if it does have an error

### DIFF
--- a/exchanges/graphcache/src/operations/query.ts
+++ b/exchanges/graphcache/src/operations/query.ts
@@ -1,4 +1,5 @@
 import { FieldNode, DocumentNode, FragmentDefinitionNode } from 'graphql';
+import { CombinedError } from '@urql/core';
 
 import {
   getSelectionSet,
@@ -6,9 +7,6 @@ import {
   SelectionSet,
   getFragmentTypeName,
   getFieldAlias,
-} from '../ast';
-
-import {
   getFragments,
   getMainOperation,
   normalizeVariables,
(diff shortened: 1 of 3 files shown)


exchanges/graphcache/src/operations/query.ts
```diff
@@ -405,6 +405,9 @@ const readSelection = (
       dataFieldValue = null;
     }
 
+    // After processing the field, remove the current alias from the path again
+    ctx.__internal.path.pop();
```
wgolledge, line 409: nice 👍

wgolledge (approved):
This makes sense to me! The `result.error` was probably going to have to be brought in at some point anyway

PR #1374, opened 2021-02-10
(breaking) - Remove pollInterval option from urql

### LINKED ISSUE #531: pollInterval doesn't wait for a response before dispatching next call
I am using `pollInterval` with a short interval (~1s) to get a reasonably current "live view" of some data on the server (pending a better implementation with subscriptions). The other day the server experienced a slowdown in its DB access, causing it to take multiple seconds to respond to requests usually served in <50ms. I then found out that the `pollInterval` option dispatches a new request every interval, instead of waiting between successful requests, which ended up sending lots of requests to the already-under-load server. It may be that the previous request is "cancelled" when the new is dispatched, but the server is already doing the work for it.

Do you think that this is the correct behaviour, or should the interval be from the completion of one request to the start of the next request instead? I didn't look into it in-depth, but I think this behaviour could be achieved by changing `switchMap` into `concatMap` [here](https://github.com/FormidableLabs/urql/blob/master/packages/core/src/client.ts#L274). If this shouldn't be the default behaviour, would it make sense to at least make it an option?

kitten:
Resolves #531

## Summary

This PR removes the `pollInterval` option from `@urql/core` and all bindings. Generally it isn't hard to replace `pollInterval` in user-land with even more correct behaviour, as it can ultimately read the results of the bindings and react to those accordingly.

### React / Preact replacement

This just demonstrates that it's rather simple to replace the implementation in React-land.

```js
const [result, executeQuery] = useQuery(...);

useEffect(() => {
  if (!result.fetching) {
    const id = setTimeout(() => executeQuery({ requestPolicy: 'network-only' }), 5000);
    return () => clearTimeout(id);
  }
}, [result.fetching, executeQuery]);
```

**NOTE re #531**: It's worth noting that this PR doesn't resolve the issue explicitly, however we see it as much easier to implement effects in user-land that call `executeQuery` to refresh the results on an interval.

## Set of changes

- Remove `pollInterval` from `@urql/core`
- Remove `pollInterval` from `useQuery` in `urql`
- Remove `pollInterval` from `useQuery` in `@urql/preact`
- Remove `pollInterval` from `useQuery` in `@urql/vue`

### DIFF
--- a/.changeset/gorgeous-squids-exercise.md
+++ b/.changeset/gorgeous-squids-exercise.md
@@ -2,4 +2,4 @@
 '@urql/core': major
 ---
 
-Remove `pollInterval` feature from `OperationContext`. Instead consider using a source that uses `Wonka.interval` and `Wonka.switchMap` over `client.query()`'s source.
+**Breaking**: Remove `pollInterval` feature from `OperationContext`. Instead consider using a source that uses `Wonka.interval` and `Wonka.switchMap` over `client.query()`'s source.
--- a/.changeset/gorgeous-squids-exercise.md
+++ b/.changeset/gorgeous-squids-exercise.md
@@ -1,0 +1,5 @@
+---
+'@urql/core': major
+---
+
+Remove `pollInterval` feature from `OperationContext`. Instead consider using a source that uses `Wonka.interval` and `Wonka.switchMap` over `client.query()`'s source.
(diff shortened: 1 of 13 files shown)


JoviDeCroock (approved):
A few mentions of `pollInterval` in the vue tests but generally looks awesome!

PR #1375, opened 2021-02-10
(core) - Reemit results as stale if they're refetched in the background

### LINKED ISSUE #1372: RFC: Stale to be "true" if any background fetching is happening
## Summary

Stale to be "true" if any background fetching is happening

## Proposed Solution

Hey!

First of all, thanks for the awesome library. Currently, I lead a team that migrates big application from Apollo to urql, and I have very high hopes that this is a correct decision.

Most of our needs are covered by what urql has to offer at the moment, except for one. Comparing to Apollo, urql provides very limited information about current state of the rendered query (e.g. **networkStatus**). I've carefully read reasoning in this [thread](https://github.com/FormidableLabs/urql/issues/417), and can definitely see the logic in an overall strategy, however few details are messing the whole picture for me right now.

Our app mostly consists of different tables with data. User can view them, and modify / delete any row using dropdown menu. When that happens, we run delete mutation, invalidate deleted entity in the cache using `cache.invalidate` in updater, and wait for query that populates the table to be refetched.

Unfortunately, I have not found a way to us to know if that query is refetched. You may suggest for us to not only invalidate an entity, but also update query result in cache manually, but in our case we can't make that due to side-effects on delete. The easiest & safest way would to just show kind of "Syncing..." status in the table until data is freshly fetched.

Same goes for cases when variables change. Often we have a header in the table with range picker or toggles that modify query variables. Query can take few seconds to resolve after change, and whole this time users has no visual feedback to their actions.

I think it can be handled (probably) by custom hooks, maybe exchanges and stuff like that... But from what I see now, this library aims to become a real competitor for Apollo. However, in my opinion, it will not happen until such fundamental functionality has to be hacked around (sorry if I sound rude).

 
## Requirements

I agree that `fetching` flag should only be responsible for explicitly called action.

So it will be only logical that `stale` flag should indicate _staleness of the data_ - which is by dictionary definition **stale** in case query variables change or data becomes invalidated in the cache manually. That way, it would be explicit & straightforward - you will always know if the query is being refetched if you need that info, if no - you just use `fetching`.

kitten:
Resolve #1372 

## Summary

Usually we have a couple of refetches that are emitted as `cache-and-network`. In these cases the cache exchanges (default or Graphcache) take care of emitting a first result with the `stale` flag set. In other cases we may have a subsequent `network-only` reexecution which happens in the background. In these cases no part of the exchange pipeline or the `Client` took responsibility of emitting a `stale` result.

This PR changes this behaviour to allows the `Client`'s `executeRequestOperation` stream to add-on a single `stale` result onto a prior result if it detects a `network-only` operation.

- Has been tested in `threed-web` against Graphcache's `cache.invalidate` which triggers a `network-only` refetch.
- This has also been fixed in 5c1b1fd to be applied to all operations that are not `cache-only`

## Set of changes

- Add a `switchMap` add-on stream per result that reemits the result as `stale` if it detects a `network-only` operation.
  - Only emit this `stale` add-on result when the previous result wasn't already `stale`
  - Ensure that this won't emit when the cache responds synchronously
  - Apply this to `cache-and-network` and `cache-first` (with cache misses) too
  - Emit this at most once since we expect a result afterwards

### DIFF
--- a/.changeset/strong-emus-confess.md
+++ b/.changeset/strong-emus-confess.md
@@ -1,0 +1,5 @@
+---
+'@urql/core': minor
+---
+
+Reemit an `OperationResult` as `stale: true` if it's being reexecuted as `network-only` operation to give bindings immediate feedback on background refetches.
--- a/packages/core/src/client.test.ts
+++ b/packages/core/src/client.test.ts
@@ -465,4 +465,57 @@ describe('queuing behavior', () => {
     unsubscribe();
     jest.useRealTimers();
   });
+
+  it('does not reemit previous results as stale if it was marked as stale already', async () => {
+    const output: OperationResult[] = [];
+    const exchange: Exchange = () => ops$ => {
+      return pipe(
+        ops$,
+        filter(op => op.kind !== 'teardown'),
+        map(op => ({
+          data: 1,
+          operation: op,
+          stale: true,
+        }))
+      );
+    };
+
+    const client = createClient({
+      url: 'test',
+      exchanges: [exchange],
+    });
+
+    const { unsubscribe } = pipe(
+      client.executeRequestOperation(queryOperation),
+      subscribe(result => {
+        output.push(result);
+      })
+    );
+
+    expect(output.length).toBe(1);
+    expect(output[0]).toHaveProperty('operation.key', queryOperation.key);
+    expect(output[0]).toHaveProperty(
+      'operation.context.requestPolicy',
+      'cache-first'
+    );
+
+    client.reexecuteOperation(
+      makeOperation(queryOperation.kind, queryOperation, {
+        ...queryOperation.context,
+        requestPolicy: 'network-only',
+      })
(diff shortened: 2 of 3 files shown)


kitten:
I've tested this against our `threed-web` example by invalidating a thread when we create a reply for it. Previously this would only refetch the thread query in the background while we now see `stale: true` pop up correctly.

JoviDeCroock (approved):

Coobaha:
@kitten @JoviDeCroock What will be the recommended way of re-executing background queries without marking results as stale? My use case is custom refocus exchange, that compares refetched data with existing and only updates if it was actually changed so that we are not triggering rendering without a need. I don't know if exactly this PR is the reason but it was working before correctly, now first background refetch on focus is marking data as stale (and triggers re-renders), but all the subsequent background refetches are not marking it and views are not re-rendered.

PR #1393, opened 2021-02-15
(docs) - Fix path rewrites for non client routed links

### LINKED ISSUE #1380: Docs links are broken when *open in a new tab*
It looks like client-routing is working properly but the _passed href_ isn't correct.

Links like `https://formidable.com/open-source/urql/docs/api/retry-exchange/`
are being shown as `https://formidable.com/open-source/api/retry-exchange/` (missing `/urql/docs` part) 

_PS: I deleted issue template since it's not about library itself. To reproduce, just open docs and click open in a new tab_

Kingdutch:
## Summary

Hopefully fixes #1380

## Set of changes

Fixed path concatenation in documentation site.

### DIFF
--- a/packages/site/static.config.js
+++ b/packages/site/static.config.js
@@ -52,23 +52,23 @@ export default {
     },
     {
       path: '/docs/concepts/core-package',
-      redirect: 'docs/basics/core',
+      redirect: '/docs/basics/core',
     },
     {
       path: '/docs/basics/getting-started',
-      redirect: 'docs/basics',
+      redirect: '/docs/basics',
     },
     {
       path: '/docs/basics/mutations',
-      redirect: 'docs/basics',
+      redirect: '/docs/basics',
     },
     {
       path: '/docs/basics/queries',
-      redirect: 'docs/basics',
+      redirect: '/docs/basics',
     },
     {
       path: '/docs/basics/document-caching',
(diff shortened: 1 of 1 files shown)


Kingdutch:
I'm not sure why Circel CI is failing to download the hpack package :/ I had hoped this would quickly build a copy of the docs site to see if my train of thought was correct.

JoviDeCroock:
Hmm, when we initially made this the leading / seemed to break url's. Will look into this as well

renatorib:
I don't checked code yet, but I think using `/docs` would fail because it will wrongly go for `formidable.com/docs` instead of `formidable.com/open-source/urql/docs`. 

I will *try* to test somthing tonight.

Kingdutch:
> I don't checked code yet, but I think using `/docs` would fail because it will wrongly go for `formidable.com/docs` instead of `formidable.com[/open-source/urql]/docs`.
> 
> I will _try_ to test somthing tonight.

That was my initial thought, but I wanted to try it anyway for two reasons:

1. It’s the only doc change I see that seems recent enough and related to be a potential cause
2. The generated URLs show a concatenation with some other string that ends before the `/docs` segment. This concatenation fails because there’s a missing / and some included .. so my hypothesis was that adding the / would fix the concatenation and fix the issue. :D 🤔

kitten (approved):

PR #1421, opened 2021-02-26
(docs) - use formideploy

JoviDeCroock:
One issue is that our CircleCI runs on pushes, `staging deploy` should only run when a PR has been opened. This seems easier on github-actions compared to Circle, do we want to switch to actions?

Temp deploy: https://github.com/FormidableLabs/urql/pull/1421/files#diff-e40148fce781f35f317b38a6914e5bb6c37beec42acce64ee30a270f5c53ab09R8 let's do a main merge and then check whether this was correct and do a subsequent merge to remove these lines.

We also have to add a `GITHUB_DEPLOYMENT_TOKEN` to get links into PR's.


packages/site/formideploy.config.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Formideploy configuration overrides.
+ */
+module.exports = {
+  lander: {
+    name: 'urql',
+  },
+  production: {
```
JoviDeCroock, line 8: When we merge, we'll check whether or not this deployed correctly, if it has we can move on to remove this in a follow-up PR.

kitten:
Closing for now until we can rebase this and update it for GH Actions, if it's actually needed yet 😅

ryan-roemer:
@kitten -- Jovi and I were working out some kinks in CircleCI (the last CI provider that I was integrating). But GH Actions are already completely supported by `formideploy`. I'd suggest a new branch, and it should be pretty straightforward. Check out:

- spectacle: https://github.com/FormidableLabs/spectacle/blob/main/.github/workflows/docs.yml
- renature: https://github.com/FormidableLabs/renature/blob/saturn/.github/workflows/ci.yml#L36-L79

And if you want to test it out before going live, the `formideploy.config.js` file here uses a temp testing site/infrastructure.

The only _real_ switch that I think you want to do is having **all** of localdev, staging, and production use the base path `/open-source/urql/` which will hopefully reduce our bug surface of docs issues. (There are ways to configure `formideploy` around this, but I'd strongly recommend having localdev match all other deploy scenarios).

Ping me if you're having any problems getting going, or in a pinch I'm pretty sure I can actually do the work if no one else is available.

Issue #672, opened 2020-03-26
relayPagination and network-only


zenios:
Hi All

I have the following query
```
query FriendsQuery($after: String!) {
  me {
    id
    friends(after: $after) {
      edges {
        node {
          ...FriendshipFragment
        }
      }
      pageInfo {
        endCursor
        hasNextPage
      }
    }
  }
}
```

And the following hook

```

export const useFriends = () => {
  const client = useClient();
  const [{ fetching, error, data }] = useQuery<
    FriendsQueryQuery,
    FriendsQueryQueryVariables
  >({
    query: FRIENDS_QUERY,
    requestPolicy: "network-only",
    variables: {
      after: "",
    },
  });

  const friends =
    data === undefined ? undefined : data?.me?.friends ?? undefined;

  const fetchMore = React.useCallback(
    (cursor: string) => {
      return client
        .query<FriendsQueryQuery, FriendsQueryQueryVariables>(
          FRIENDS_QUERY,
          { after: cursor },
          { requestPolicy: "network-only" }
        )
        .toPromise();
    },
    [client]
  );

  return React.useMemo(
    () => ({
      fetching,
      error,
      friends,
      fetchMore,
    }),
    [fetching, error, friends, fetchMore]
  );
```

Results are merged using relayPagination 
```
resolvers: {
  User: {
    friends: relayPagination()
  }
}
```

I have a component that does an infinite scrolling and shows all my friends. 

When the component is unmounted and afterwards remounted **i would like to fetch again the lists of friends (hence the network-only)**

The problem is that due to the relay pagination the friends variable contains all the friends fetched from the previous mount with pageInfo.hasNextPage === false.

The only request that is requested again is the initial with after === "" 

Thanks

zenios:
I think the problem relies in the pageInfo of relayPagination. It always sets it to the last even though after the remount i have refetched the first page.

I am not sure if its wise to reset pageInfo to the currently fetched page results

JoviDeCroock:
Hey Zenios,

Good question but this isn't really the use case of a resolver, for us it would be impossible (and a waterfall) to refetch it all.

Let me explain, if we would refetch your first page we can't eagerly refetch your second page, we would have to check the pageInfo of your first response to go for your second one and so son.

The better solution for this situation would be to make a custom hook:

```
const useRevalidatingPagination = () => {
  const [list, setList] = useState([])

  const [data] = useQuery(query);

  useEffect(() => {
    if we go from fethcing to not fetching 
[...]

PR #1474, opened 2021-03-21
(graphcache) support relayPagination cache refresh

duytbp:
Related to #672 

## Summary

This is an attempt to allow force refreshing pagination cache inside helper relayPagination.

Basically, with this new option enabled, when we want to invalidate pagination cache, we reload first page with an empty cursor

I think naming of this new option can be improved, but this is the best I can come up now.

## Set of changes
- Add `relayPagination` option `cleanCacheOnEmptyCursors`

### DIFF
--- a/.changeset/shaggy-rats-own.md
+++ b/.changeset/shaggy-rats-own.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-graphcache': minor
+---
+
+Add option `cleanOnEmptyCursors` to helper `relayPagination`. With this enabled, when resolve a page with both before & after cursors empty, cache for all prior & subsequent pages are invalidated, and will not be included in resolving outcome.
--- a/exchanges/graphcache/src/extras/relayPagination.test.ts
+++ b/exchanges/graphcache/src/extras/relayPagination.test.ts
@@ -1319,3 +1319,145 @@ it('handles subsequent queries with larger first values', () => {
   expect(res.partial).toBe(false);
   expect(res.data).toEqual(pageTwo);
 });
+
+it('cleans cache when args.after is empty and cleanOnEmptyCursors=true', () => {
+  const Pagination = gql`
+    query($cursor: String) {
+      __typename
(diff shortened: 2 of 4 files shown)


kitten:
I think it may be necessary for us to have an RFC here and follow the full process, as it exists to discuss these kinds of proposals in detail before implementing them 😅

Specifically, I believe that this approach doesn't fulfil a couple of requirements we'd have in terms of being predictable, intuitive, and follow the principle of least surprise:

- The logic here may wipe the cache as a result of a resolver. This is not expected. A resolver should aim never to modify the cache and doing so may put the storage into a state where it can't respond to the change. This isn't in response to a result so it violates the principle that the same cache data will lead to the same result.
- Concurrent requests may lead to one query deleting data that another was about to use, which may then start delivering two results that aren't consistent with one another.
- This approach may trigger in scenarios where the developer hasn't asked for another request (cache-and-network would do that) which means that any attempt to "passively" access the first page would fail (`client.readQuery`, first mount in React, or `cache-only` policies)
- as per the code, there are some requirements that it doesn't fit, if we go beyond the problems of the approach. It must be used after `compareArgs` is applied to fields, otherwise it applies to all fields, not just the ones that have been filtered for
- It must check for unset cursors, not for empty/trimmed ones as those are set.

But overall, sorry for ripping this apart a little 😅 we've also thought about this before and haven't written an RFC for this approach because it simply doesn't lead to predictable or usable behaviour.

Overall, we actually believe that the pagination helpers don't lead to usable behaviour in most cases. They're great for the simplest cases but beyond them it makes much more sense to handle infinite pagination logic where it applies: in the UI code and lifecycles.

This is easily done for the web with fragments, eg in React, but currently only an issue in React Native. So while we're talking about patterns for those and prepare a docs page, we haven't made a page for it quite yet.

duytbp:
Hey, thanks for the detailed feedback. No hard feelings . 

Let me close PR and wait for release of those mentioned documents.

PR #1502, opened 2021-04-02
(examples) - next.js

JoviDeCroock:
## Summary

Provides a Next.JS example describing the two main avenues you can consider when starting a new application, the functions provided by Next or the prepass method.


examples/with-next/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+## getStaticProps
+
+This requires some manual work, when we look at [`static.js`](./pages/static.js) we can see that we define our own
+`getStaticProps` method, this because these methods are only `user-facing`. When doing a `yarn next build` we'll need to
+ensure that the server we're targetting is running so we can successfully execute the static prerender.
+
+## getServerSideProps
+
+This requires some manual work, when we look at [`static.js`](./pages/server.js) we can see that we define our own
```
jesstelford, line 20:
```suggestion
This requires some manual work, when we look at [`server.js`](./pages/server.js) we can see that we define our own
```

jesstelford (commented):
Great examples!

gszliu:
Are there blockers for this change?

examples/with-next/pages/static.js
```diff
@@ -0,0 +1,55 @@
...
+import {
+  ssrExchange,
+  dedupExchange,
+  cacheExchange,
+  fetchExchange,
+  useQuery,
+  useClient
+} from "urql";
+
+const POKEMONS_QUERY = `
```
kitten, line 11: I think in examples we should use `gql` so that it's as "idiomatic" as possible

examples/with-next/pages/static.js
```diff
@@ -0,0 +1,55 @@
...
+  query {
+    pokemons(limit: 10) {
+      id
+      name
+    }
+  }
+`;
+
+function Static(props) {
+  const client = useClient();
```
kitten, line 21:
```suggestion
```

kitten (approved):

PR #1509, opened 2021-04-03
(core) - Make DocumentNode serialization idempotent

### LINKED ISSUE #1508: Under certain conditions a different key is generated for the same DocumentNode
Under certain conditions (all of which are a bit confusing) a different key is generated (by [`keyDocument()`](https://github.com/FormidableLabs/urql/blob/ed45f960122e5e3848dfbb4382304f774417b5e9/packages/core/src/utils/request.ts#L56-L87)) given the same DocumentNode.

This is noticeable in a next-urql app when we see unexpected cache misses during either final server-side render (`Error: ReactDOMServer does not yet support Suspense.`) or client-side rehydration (since cached data isn't found, React warns about unexpected dom nodes, and data is fetched over network again).

versions are the latest as of writing this:
- urql @ v2.0.2
- next-urql @ v3.0.1

Uses next-urql's default exchanges

**Steps to reproduce**

1. Clone & install the reproduction project: https://github.com/zenflow/urql-documentnode-key-generation-issue
2. Start dev server with `yarn next dev`
3. Open http://localhost:3000 & open Chrome devtools
4. `touch pages/index.tsx`, or make a trivial change to it's contents like `<h1>` to `<h2>`
5. Wait for recompile & then refresh the browser
6. Observe error rendering server-side: `Error: ReactDOMServer does not yet support Suspense.`
7. Refresh browser (or allow it to automatically refresh) until a page loads without the above error
8. Observe console warning & network request to graphql server

**Analysis**

To debug I built `packages/core` with [`hash()`](https://github.com/FormidableLabs/urql/blob/ed45f960122e5e3848dfbb4382304f774417b5e9/packages/core/src/utils/hash.ts#L14) modified to `console.log` it's input (unique graphql query string) & output (hash number). 

I found that, for my single query, either of two different query strings would be hashed: `"query MyQuery { ... }"` (bare) or `"# MyQuery\nquery MyQuery { ... }"` (prefixed), resulting in a different hash number.

See in [`stringifyDocument()`](https://github.com/FormidableLabs/urql/blob/ed45f960122e5e3848dfbb4382304f774417b5e9/packages/core/src/utils/request.ts#L24-L52) where the query string gets prefixed.

I see a problem that `stringifyDocument()` is not idempotent. When first called with a DocumentNode that has no `.loc` property, it adds the `.loc` property and returns only the bare query string. The next time it's called with the exact same DocumentNode object, since the `.loc` property is now present, the result will be prefixed with operation name if available. 

TBH It was pretty confusing, because the code looks ok at first glance - you would think `stringifyDocument()` wouldn't be called twice with the same object anyway, due to short-circuiting in [`keyDocument()`](https://github.com/FormidableLabs/urql/blob/ed45f960122e5e3848dfbb4382304f774417b5e9/packages/core/src/utils/request.ts#L56-L73) when that object has a `.__key` property defined. 
But I looked closer and noticed that this `.__key` property is only added to the document objects that are being kept in the [`docs` Map](https://github.com/FormidableLabs/urql/blob/ed45f960122e5e3848dfbb4382304f774417b5e9/packages/core/src/utils/request.ts#L54). Document objects that never get added to the `docs` Map (because another identical document object was "keyed" first) never have their `.__key` property defined, and are therefore stringified & hashed on each call to `keyDocument()`.

zenflow:
## Summary

Fixes #1508

## Set of changes

This makes DocumentNode serialization idempotent.
In other words, this makes it so we get the same serial representation (and therefore hash key) for a DocumentNode each time it is computed.

### DIFF
--- a/.changeset/fair-beds-happen.md
+++ b/.changeset/fair-beds-happen.md
@@ -2,4 +2,4 @@
 '@urql/core': patch
 ---
 
-Make DocumentNode serialization idempotent, closes #1508
+Fix inconsistency in generating keys for `DocumentNode`s, especially when using GraphQL Code Generator, which could cause SSR serialization to fail.
--- a/.changeset/fair-beds-happen.md
+++ b/.changeset/fair-beds-happen.md
@@ -1,0 +1,5 @@
+---
+'@urql/core': patch
+---
+
+Make DocumentNode serialization idempotent, closes #1508
--- a/exchanges/multipart-fetch/src/__snapshots__/multipartFetchExchange.test.ts.snap
+++ b/exchanges/multipart-fetch/src/__snapshots__/multipartFetchExchange.test.ts.snap
@@ -16,7 +16,7 @@ Object {
     "key": 2,
     "kind": "query",
     "query": Object {
-      "__key": 3044551916,
+      "__key": 3521976120,
       "definitions": Array [
         Object {
           "directives": Array [],
@@ -140,9 +140,10 @@ Object {
       ],
       "kind": "Document",
       "loc": Object {
-        "end": 76,
+        "end": 86,
         "source": Object {
-          "body": "query getUser($name: String) { user(name: $name) { id firstName lastName } }",
+          "body": "# getUser
+query getUser($name: String) { user(name: $name) { id firstName lastName } }",
           "locationOffset": Object {
             "column": 1,
             "line": 1,
@@ -185,7 +186,7 @@ Object {
     "key": 2,
     "kind": "query",
     "query": Object {
-      "__key": 3044551916,
+      "__key": 3521976120,
       "definitions": Array [
         Object {
           "directives": Array [],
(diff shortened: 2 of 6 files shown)


.changeset/fair-beds-happen.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/core': patch
+---
+
+Make DocumentNode serialization idempotent, closes #1508
```
kitten, line 5:
```suggestion
Fix inconsistency in generating keys for `DocumentNode`s, especially when using GraphQL Code Generator, which could cause SSR serialization to fail.
```

Just a minor one here; this goes directly into the changelog, so we make an effort to describe exactly what went on in the change and how it affects users (also bug fixes / critical hot fixes are prefixed by "Fix" so they're highlighted) ✌️

kitten (commented):
This is a really awesome catch! I didn't realise that this would introduce an inconsistency in GraphQL Code Generator output since I didn't consider that `loc` mane left out at times depending on its mode, which is entirely probable in other use cases too. Also nice idea to fix it this way 🔥

The only thing that's left is that some snapshots need updating now, however a `yarn test -u` run should fix that. The changeset also has a suggestion by me to make it a little more descriptive

zenflow:
@kitten Suggestion for changeset applied and Just snapshots updated :)

kitten (approved):
Cheers! I'll have to check next week whether any of the hash changes will adversely affect anything else, but it seems perfectly fine to me ✌️

Issue #1485, opened 2021-03-26
RFC: Add `cache.link` method to write links manually to Graphcache


kitten:
## Summary

Graphcache's public APIs allow a user to dynamically read from the cache and write to it during updates. However, updates — most of the time — have very limited use cases they apply to. Typically it's common to at least use `updateQuery` with an ad-hoc created query, but this is often overkill. Fragments however aren't always desireable since creating them adhoc is just as wordy and the `cache.updateQuery` method is perceived to be more elegant. (which it admittedly is)

If we look at all available methods then it becomes clear that one useful one is missing: https://formidable.com/open-source/urql/docs/api/graphcache/#cache
We've got `cache.resolve` (read w/o a document), `cache.readFragment`, `cache.writeFragment`, `cache.readQuery`, `cache.updateQuery`.

Most of these are useful when updating a lot of lists or records, but they make a more common use-case harder to achieve: updating individual links. While `cache.resolve` is useful to read individual links and records, there isn't an equivalent API to write to the cache.

## Considerations

- Any API that looks like `cache.resolve` but writes to the cache shouldn't take a fixed amount of arguments.
- The API should _not_ write records, as `cache.writeFragment` is much safer for that.
- The API should chain nicely with `cache.resolve`

## Proposed Solution

Add a `cache.link` method that only writes links and updates them. It will not work for records and will throw if it receives an invalid value to write to the cache. When it shouldn't have to check whether the link exists since garbage collection will take care of deleting it if it doesn't apply to the cache.

## Requirements

This API can check for link data as we usually do for resolver results: https://github.com/FormidableLabs/urql/blob/65e15c568c9a0755894dfb02995f0262e1fe75f9/exchanges/graphcache/src/operations/query.ts#L547-L549

Furthermore it should accept:

1. A partial entity or a key (if `null` is passed as the first argument then it's a noop)
2. A field argument name
3. field arguments (or `null`) which is an optional argument
4. The new links value, which may be `null | Data | string | Array<null | string | Data>`

The field arguments may be left out by checking whether `cache.link` was called with either four or three arguments.

PR #1551, opened 2021-04-20
(graphcache) - Add cache.link method for writing links

JoviDeCroock:
## Summary

This allows the user to pass in a rootKey or entity from which we'll derive the rootKey for instance Query or { __typename:'Todo', id: 1 }, secondly we accept a field, this can be for instance author which will be applied to the rootKey. As a thirth argument we accept the entity or entities to be written as keys to the rootKey.field, for example { __typename: 'Author', id: 1 } and as a last argument we accept fieldArguments, these will be applied to the fieldKey and will help with fields that have their own arguments for instance: Todo:1.authors({ first: 10 }).

Resolves: #1485

## Set of changes

- Adds `cache.link` to graphCache to be used in updaters

### DIFF
--- a/exchanges/graphcache/src/store/store.ts
+++ b/exchanges/graphcache/src/store/store.ts
@@ -8,6 +8,7 @@ import {
   DataField,
   Variables,
   FieldArgs,
+  Link,
   Data,
   QueryInput,
   UpdatesConfig,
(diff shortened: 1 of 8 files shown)


exchanges/graphcache/src/store/store.ts
```diff
@@ -209,4 +209,22 @@ export class Store implements Cache {
...
+  link(
+    entity: Data | string,
+    field: string,
+    entities: Data | Data[],
+    fieldArgs?: Variables
+  ): void {
+    const entityKey = this.keyOfEntity(entity);
+    if (!entityKey) return;
+
+    InMemoryData.writeLink(
```
kitten, line 222: We can probably create a variant of `writeField` here; my main concern being that embedded entities should work and that repeatedly nested lists should work too 🤔 https://github.com/FormidableLabs/urql/blob/ef4b5b002d31e17d84fb41e5feb24d2200c9656f/exchanges/graphcache/src/operations/write.ts#L305

We don't want to add `filter(Boolean)` either since `null` in a list is valid, although we may want to check that against the schema. Let's see what the current logic is on that. We may want to write `undefined` for schema awareness of a schema is present.

exchanges/graphcache/src/types.ts
```diff
@@ -128,6 +128,13 @@ export interface Cache {
     data: T,
     variables?: V
   ): void;
+
+  link(
+    entity: Data | string,
+    field: string,
+    entities: Data | Data[],
```
kitten, line 135: I think this current signature is a little confusing. Originally I proposed this because it's more intuitive:

```ts
link(
  entity: string | Data | null,
  fieldName: string,
  fieldArgs: Variables | null,
  link: Link<Data | string>
): void;

link(
  entity: string | Data | null,
  fieldName: string,
  link: Link<Data | string>
): void;
```

The overload helps here in case people use `cache.keyOfField` or have no arguments. It also helps that arguments come after the field name.

The `Link` type can be reused here for nested lists and for allowing `string | Data | null`.

kitten: We also have to check here what happens when scalars are passed in accidentally or what the common case is for partial data, since it becomes more likely for people to chain `cache.link` and write `null` data.

I'm not sure whether in that case we need to make special arrangements to write `undefined` instead of `null`

kitten:
The bundle size is going to change from `7.16kB` (main) to `7.24kB` minzipped. I think it's a little odd that this increase is a little larger in particular, but we could still look into that if we wanted to.

kitten (approved):
@JoviDeCroock Let's do this the opposite way 😆 I'll approve and you can dismiss/merge depending on how it looks

PR #1562, opened 2021-04-24
(graphcache) - Add generics for codegen to store and cacheExchange

JoviDeCroock:
Spinning up a PR so I can give the CSB CI package a whirl in [my experiment repo](https://github.com/JoviDeCroock/graphcache-code-generation)

Currently dealing with some issues in the `code-generation` repo where it's annoyed at `__typename: 'Todo'` vs `__typename: string`, the problem is located [here](https://github.com/FormidableLabs/urql/blob/main/exchanges/graphcache/src/types.ts#L14-L19) where for instance `Todo.author` can't be it's own type as `__typename` is `never`

## Summary

This adds generics for graphCache

## Set of changes

- Generics for graphcache to be used with generated types


andyrichardson:
Typescript doesn't look to enjoy types where property presence is optional, and also the value is optional

```ts
prop?: Type | undefined
```

https://github.com/FormidableLabs/urql/compare/generics-graphcache...generics-graphcache-andy?expand=1

exchanges/graphcache/src/types.ts
```diff
@@ -136,22 +136,21 @@ type ResolverResult =
...
 export interface ResolverConfig {
   [typeName: string]: {
     [fieldName: string]: Resolver;
   };
 }
 
-export type UpdateResolver = (
-  result: Data,
-  args: Variables,
+export type UpdateResolver<Result = Data | Data[], Args = Variables> = (
```
JoviDeCroock, line 151: Not entirely sure about this change but imo it should be possible to return array data from a mutation, no?

andyrichardson: I could have sworn we had a type definition for this (basically JSON) in core - did we revert it?

kitten: It will never return an array as `result` will always be the mutation data itself, never the item or field, since it imitates `parent` from GraphQL resolvers (like the other resolvers)

See: https://github.com/FormidableLabs/urql/blob/f6979d5be57c05a2c5ad911f9d4ca1501b7c688e/exchanges/graphcache/src/operations/write.ts#L292-L293

exchanges/graphcache/src/cacheExchange.ts
```diff
@@ -50,21 +50,19 @@ type OperationMap = Map<number, Operation>;
...
-}
-
-export const cacheExchange = (opts?: CacheExchangeOpts): Exchange => ({
-  forward,
-  client,
-  dispatchDebug,
-}) => {
-  const store = new Store(opts);
+export type CacheExchangeOpts = {
+  updates: Partial<UpdatesConfig>;
```
kitten, line 54: Let's remove `Partial` here and instead make the internal `UpdatesConfig` type partial? We can still "unpartial" it on the store

exchanges/graphcache/src/cacheExchange.ts
```diff
@@ -50,21 +50,19 @@ type OperationMap = Map<number, Operation>;
...
-  storage?: StorageAdapter;
-}
-
-export const cacheExchange = (opts?: CacheExchangeOpts): Exchange => ({
-  forward,
-  client,
-  dispatchDebug,
-}) => {
-  const store = new Store(opts);
+export type CacheExchangeOpts = {
```
kitten, line 53: Is there a point to wrapping this in `Partial` rather than keeping each field as optional like before?

exchanges/graphcache/src/store/store.ts
```diff
@@ -27,26 +26,20 @@ import { keyOfField } from './keys';
...
 import {
-  IntrospectionData,
   SchemaIntrospector,
   buildClientSchema,
   expectValidKeyingConfig,
   expectValidUpdatesConfig,
   expectValidResolversConfig,
   expectValidOptimisticMutationsConfig,
 } from '../ast';
+import { CacheExchangeOpts } from '../cacheExchange';
```
kitten, line 36: Potentially circular; maybe we should move it to `types.ts`

exchanges/graphcache/src/types.ts
```diff
@@ -136,22 +136,21 @@ type ResolverResult =
...
 
 // Cache resolvers are user-defined to overwrite an entity field result
-export type Resolver = (
-  parent: Data,
-  args: Variables,
-  cache: Cache,
-  info: ResolveInfo
-) => ResolverResult;
+export type Resolver<
+  ParentData = Data,
```
kitten, line 140: I'm actually thinking; this may also have to be `DataFields` and we made a mistake here? https://github.com/FormidableLabs/urql/blob/aa5685b390e733fe2a6f46655502ea31df0daded/exchanges/graphcache/src/operations/query.ts#L339

exchanges/graphcache/src/cacheExchange.ts
JoviDeCroock: It's an interface as we're extending from it so that's not possible.

exchanges/graphcache/src/types.ts
```diff
@@ -135,23 +136,31 @@ type ResolverResult =
...
   | undefined;
 
+export type CacheExchangeOpts = Partial<{
+  updates: Partial<UpdatesConfig>;
+  resolvers: ResolverConfig;
+  optimistic: OptimisticMutationConfig;
+  keys: KeyingConfig;
+  schema: IntrospectionData;
+  storage: StorageAdapter;
+}>;
```
kitten, line 146:
```suggestion
export type CacheExchangeOpts = {
  updates?: Partial<UpdatesConfig>;
  resolvers?: ResolverConfig;
  optimistic?: OptimisticMutationConfig;
  keys?: KeyingConfig;
  schema?: IntrospectionData;
  storage?: StorageAdapter;
};
```

.changeset/silent-turtles-check.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/exchange-graphcache': minor
+---
+
+Add generics for the graphCache config
```
kitten, line 5:
```suggestion
Add on a generic to `cacheExchange` and `offlineExchange` for future, experimental type-generation support.
```

kitten (approved):

PR #1564, opened 2021-04-25
(chore) - track size of bundles

JoviDeCroock:
Track bundle size of changes compared to the `main` branch


kitten:
Closing for now :v: I think the main problem right now is that the action writes a comment and not to the checks output, which we could set up as a custom action; but mainly I've also checked and the action double-builds and runs checkouts of the `main` branch, which I find a little unexpected. My hope is that #1568 can supersede this and that we can compare actual artifact outputs between the target branch and PR branch. But for now, I also planned that PR to double as a CSB CI replacement which it can't be right now, because GitHub's artifacts are always zipped on download.

Issue #1552, opened 2021-04-20
Weird naming collision in @urql/exchange-graphcache crashes in react-native


edenstrom:
Hey, 

So we've noticed that urql offline exchange doesn't work in the release builds of our react native app. It works fine in Debug mode (I guess because of the different engines)

Could be something with our setup, but the following is what I've found.

**urql version & exchanges:**
"@urql/core": "^2.0.0",
"@urql/devtools": "^2.0.3",
"@urql/exchange-auth": "^0.1.2",
"@urql/exchange-graphcache": "^4.0.0",
"@urql/exchange-retry": "^0.2.0",
"urql": "^2.0.2",

**Steps to reproduce**

1. Try to use offline exchange in a react native app. 

Below is a code snippet from https://unpkg.com/@urql/exchange-graphcache@4.0.0/dist/urql-exchange-graphcache.js

You can also search for it with `function a(a)` and you'll find it in the function `persistData`.

```js
}, persistData = function() {
  if (currentData.storage) {
    currentOptimistic = !0;
    currentOperation = "read";
    var b = makeDict();
    currentData.persist.forEach((function a(a) {
      var f, c = deserializeKeyInfo(a), d = c.entityKey;
      void 0 !== (f = readLink(d, c = c.fieldKey)) ? b[a] = ":" + core.stringifyVariables(f) : void 0 !== (f = readRecord(d, c)) ? b[a] = core.stringifyVariables(f) : b[a] = void 0;
    }));
    currentOptimistic = !1;
    currentData.storage.writeData(b);
    currentData.persist.clear();
  }
}, hydrateData = function(a, b, c) {
```

2. Look at the line  `currentData.persist.forEach((function a(a) {`. Both the function and the parameter is `a`.
3. The next line is `deserializeKeyInfo(a)` which sends `a` as an argument into the `deserializeKeyInfo`.
4. In debug mode, `a` is the variable `a` (parameter to the function in step 2).
5. In release mode, `a` is the function.
6. In release mode, deserializeKeyInfo throws an error. `indexOf` doesn't really work on a function 👀 
7. This error is hidden for the user. And the storage.writeData that actually triggers the persistance will never run again. This is because the `defer` boolean is never set to `false` again.
8. Changing the `function a` to `function aWhateverElse` makes it work.
8. Is this some bug with the rollup configuration used to build the library?

Sorry for the rambling above. Hopefully, it's actionable.

**Expected behavior**
The offline cache should work

**Actual behavior**
[...]

PR #1570, opened 2021-04-26
(chore) - remove closure-compiler

JoviDeCroock:
Fixes: https://github.com/FormidableLabs/urql/issues/1552

## Summary

The current rollup plugin isn't really maintained and we were seeing frequent issues with metro and bundles compiled by closure

## Set of changes

- remove `rollup-plugin-closure-compiler`

### DIFF
--- a/.changeset/curvy-bobcats-fry.md
+++ b/.changeset/curvy-bobcats-fry.md
@@ -14,7 +14,6 @@
 'next-urql': patch
 '@urql/preact': patch
 'urql': patch
-'urql-docs': patch
 '@urql/storybook-addon': patch
 '@urql/svelte': patch
 '@urql/vue': patch
--- a/.changeset/curvy-bobcats-fry.md
+++ b/.changeset/curvy-bobcats-fry.md
@@ -1,0 +1,23 @@
+---
+'@urql/exchange-auth': patch
+'@urql/exchange-execute': patch
+'@urql/exchange-graphcache': patch
+'@urql/exchange-multipart-fetch': patch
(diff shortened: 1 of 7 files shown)


.changeset/curvy-bobcats-fry.md
```diff
@@ -0,0 +1,22 @@
+---
```
kitten, line 1: This is just personal taste FWIW but I believe we shouldn't forcefully bump all builds, maybe? It'd be cool if there was a way to tell changesets to "ignore" one and only include it when a non-ignored change set was in a release too

scripts/rollup/plugins.js
```diff
@@ -104,10 +103,6 @@ export const makeOutputPlugins = ({ isProduction, extension }) => {
     isProduction && replace({
       'process.env.NODE_ENV': JSON.stringify('production')
     }),
-    !settings.mayReexport && compiler({
-      formatting: 'PRETTY_PRINT',
-      compilation_level: 'SIMPLE_OPTIMIZATIONS'
-    }),
     cleanup({ extension }),
     isProduction ? terserMinified : terserPretty,
```
kitten, line 112: I think it's worth investigating the impact per bundle 🤔

kitten (approved):

kitten:
Removing Closure Compiler (and Babel Plugin Closure Elimination) in this PR. Added a transformation to clean up arrow functions. Instead of having a lot of function expressions, a lot of those are now converted to function declarations which leads to a net negative bundle size on average

PR #1573, opened 2021-04-27
(examples) - react.js

yankovalera:
Provides a React example describing the way to initialise the urlq client and execute a query to fetch data from a GraphQL backend API. Handles the loading state and error messages.


examples/react-urql-query/README.md
```diff
@@ -0,0 +1,13 @@
+# Integrating with React
+
+Integrating urql is as simple as:
+
+1. Install packages 
+
+yarn add urql graphql
+# or
+npm install --save urql graphql
```
JoviDeCroock, line 9:
```suggestion
```sh
yarn add urql graphql
# or
npm install --save urql graphql
```
```

File: examples/react-urql-query/package.json [hunk 2]
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "react-urql-query",
+  "version": "1.0.0",
+  "private": true,
+  "dependencies": {
+    "@testing-library/jest-dom": "^5.12.0",
+    "@testing-library/react": "^11.2.6",
+    "@testing-library/user-event": "^12.8.3",
```
JoviDeCroock, line 8: let's remove these

examples/react-urql-query/src/index.js
```diff
@@ -0,0 +1,17 @@
...
+  <React.StrictMode>
+    <App />
+  </React.StrictMode>,
+  document.getElementById('root')
+);
+
+// If you want to start measuring performance in your app, pass a function
+// to log results (for example: reportWebVitals(console.log))
+// or send to an analytics endpoint. Learn more: https://bit.ly/CRA-vitals
+reportWebVitals();
```
JoviDeCroock, line 17: Let's remove this as well, don't really need this special stuff, that way we scope it exclusively to urql-react

JoviDeCroock (approved):
Few minor nits, looks awesome great work!

.gitignore
```diff
@@ -16,3 +16,4 @@ exchanges/*/LICENSE
 # TODO: Figure out how to remove these:
 tmp/
 dist/
+examples/react-urql-query/public/*
```
kitten, line 19: is it worth generalising this to the following for the other `examples` that are about to be added?

```suggestion
examples/*/public
```

On another small note, I believe you've used `npm` rather than `yarn` to install dependencies. That's not a problem, but what I wanted to point out is that no lockfile was added to the respository because we're ignoring `package-lock.json`. I'd actually avoid adding lockfiles in the future for each example too, because they're really not worth keeping up-to-date as little breakage should occur with the correct dependency ranges for simple example apps, so I'd also add:

```
examples/*/yarn.lock
```

examples/react-urql-query/README.md
```diff
@@ -0,0 +1,15 @@
+# Integrating with React
```
kitten, line 1: I think the only general nit I have is that we can make each example start with `with-`, e.g. `with-next` or `with-react`. I think each name is then as short as what it demonstrates, i.e. `with-react-urql-query` is a little long, because `with-react-urql-mutation` is something that wouldn't have to exist, or if it does, would have to be added to `with-react` as a "general" one.

So basically, we'd be following the least amount of code necessary to demonstrate an integration point.

examples/react-urql-query/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "name": "react-urql-query",
+  "version": "1.0.0",
+  "private": true,
+  "dependencies": {
+    "graphql": "^15.5.0",
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2",
+    "react-scripts": "4.0.3",
+    "urql": "^2.0.2",
+    "web-vitals": "^1.1.1"
```
kitten, line 11: Accidental addition or leftover from some example?

examples/react-urql-query/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "react-urql-query",
+  "version": "1.0.0",
+  "private": true,
+  "dependencies": {
+    "graphql": "^15.5.0",
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2",
+    "react-scripts": "4.0.3",
```
kitten, line 9: nit: Let's get this added to `devDependencies` and keep the range as in-minor 🙌

kitten (approved):
Nice! generally just some nits that we could clean up any other time too, mostly related to the (added) `package.json`  and (changed) `.gitignore` 🙌 ❤️

examples/react-urql-query/README.md
yankovalera: Agreed!

examples/react-urql-query/package.json
yankovalera: Removed!

PR #1576, opened 2021-04-27
(docs) Improve retryIf function in example

yankovalera:
## Summary

In the docs, the existing retryIf condition could fail if error is not defined. That's why I wanted to check error is defined before checking for graphQLErrors or networkError

### DIFF
--- a/docs/advanced/retry-operations.md
+++ b/docs/advanced/retry-operations.md
@@ -79,7 +79,7 @@ const client = createClient({
     cacheExchange,
     retryExchange({
       retryIf: error => {
-        return Boolean(error && (error.graphQLErrors.length > 0 || error.networkError));
+        return !!(error.graphQLErrors.length > 0 || error.networkError);
       },
     }),
     fetchExchange,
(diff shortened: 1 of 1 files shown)


docs/advanced/retry-operations.md
```diff
@@ -79,9 +79,7 @@ const client = createClient({
     cacheExchange,
     retryExchange({
       retryIf: error => {
-        if ((error && error.graphQLErrors.length > 0) || error.networkError) {
-          return true;
-        }
+        return Boolean(error && (error.graphQLErrors.length > 0 || error.networkError));
```
kitten, line 82: The `retryIf` function is guaranteed to have an error passed to it https://github.com/FormidableLabs/urql/blob/ae650d103f7de7c266d8b0c5f9a489991bede1e4/exchanges/retry/src/retryExchange.ts#L110,

the error here is probably a mistake in the braces, i.e.

```suggestion
        return !!error && (error.graphQLErrors.length > 0 || error.networkError));
```

yankovalera: Oh right, in that case I think we can just do

return error.graphQLErrors.length > 0 || error.networkError;

kitten (approved):
Before I forget, feel free to merge yourself once you feel you have the appropriate amount of reviews 🙏

PR #1580, opened 2021-04-27
(docs) Add Vite-specific installation instructions

tmaxmax:
## Summary

When using either this or other GraphQL clients together with Vite it's common to encounter this error:

```
Uncaught Error: Function called outside component initialization
    at get_current_component (index.mjs:649)
    at setContext (index.mjs:679)
    at setClient (context.ts:11)
```

The fix is simple, as seen in the changes, but in spite of that there exist very few resources that provide a concrete solution to it. Many threads are open on both StackOverflow and Github on this issue, so I believe specifying the solution in the docs will help a lot of developers in the long run.

### DIFF
--- a/docs/basics/svelte.md
+++ b/docs/basics/svelte.md
@@ -31,6 +31,17 @@ all `@urql/svelte` packages will define a range of compatible versions of `graph
 for breaking changes in the future however, in which case your package manager may warn you about
 `graphql` being out of the defined peer dependency range.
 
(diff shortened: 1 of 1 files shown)


kitten (approved):
Wow that's a really weird one! Thanks for digging into providing a fix 🙏

Do you know what happens when this optimisation kicks in that causes this to happen? I'm happy to do some research if not, I'm just curious whether it's a bundling issue we can work around

tmaxmax:
I'm not exactly sure. I believe this happens with packages that have multiple imports paths, as they are compressed into one from as far as I can understand, but I couldn't say why this causes that problem. 

Other than that, no problem!

PR #1585, opened 2021-04-28
(examples) Add with-multipart example

yankovalera:
## Summary

Add an example to show how to use the multipartFetchExchange exchange to handle file uploads


JoviDeCroock (approved):
Nice! It's so great to see this coming together, thank you Yanko!

PR #1586, opened 2021-04-28
(examples) Add with-pagination

yankovalera:
## Summary

Add an example for infinite scrolling pagination using the UI


examples/with-pagination/src/pages/PaginatedNpmSearch.js
```diff
@@ -0,0 +1,34 @@
+import React, { useState } from "react";
+
+import SearchResultPage from './components/SearchResultPage';
```
kitten, line 3: This is just a small nit, but I'd just put both components into a single file so that the pattern of pages & container becomes more obvious :v:

kitten (approved):

PR #1587, opened 2021-04-28
fix imports on next examples

PascalPixel:


kitten (approved):
Hm, seems like I'll have to look into activating lint-staged and eslint for these after all 🤣 Cheers!

PR #1591, opened 2021-04-29
(graphcache) - Improve internal typings for lists, scalars, and overall cleanliness

kitten:
This is a supporting PR for #1551 and #901 and some commits were lifted from other branches into this new PR.

## Summary

Some types in Graphcache weren't accurate enough, especially when using them as extensions for generated types, for conveying the intent of some methods on `Store` / `Cache`, and lastly when typing some scalar and link fields.

Mainly interesting is that `FieldArgs` is a new type that was generalisable, because it was used in this form quite frequently. It's simply `undefined | null | Variables`, a common input for a field's arguments.

The other adjustment that's notable is that `NullArray` can now be nested in on itself, because lists in GraphQL can be nested arbitrarily often. (There was actually one case where Graphcache wasn't handling this well in reference counting)

And lastly, the `ScalarObject` type previously forbid the `__typename` field although a scalar should technically be entirely opaque to us. So we've swapped this out with a type that reflects each field's value type as `unknown`

`7.15kB => 7.16kB minzipped`

## Set of changes

- Add `FieldArgs` type to `store.ts`
- Update reference counting to handle nested lists for links
- Add `Entity` type as a generic entity reference (either as a key or partial entity)
- Update `NullArray` to be recursive

### DIFF
--- a/exchanges/graphcache/src/store/data.ts
+++ b/exchanges/graphcache/src/store/data.ts
@@ -299,8 +299,13 @@ const updateRCForLink = (
   if (typeof link === 'string') {
     updateRCForEntity(gc, refCount, link, by);
   } else if (Array.isArray(link)) {
-    for (let i = 0, l = link.length; i < l; i++)
-      updateRCForLink(gc, refCount, link[i], by);
+    for (let i = 0, l = link.length; i < l; i++) {
+      if (Array.isArray(link[i])) {
+        updateRCForLink(gc, refCount, link[i], by);
+      } else if (link[i]) {
+        updateRCForEntity(gc, refCount, link[i] as string, by);
+      }
+    }
   }
 };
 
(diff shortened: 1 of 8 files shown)


exchanges/graphcache/src/store/data.ts
```diff
@@ -300,9 +300,10 @@ const updateRCForLink = (
...
   } else if (Array.isArray(link)) {
     for (let i = 0, l = link.length; i < l; i++) {
-      const entityKey = link[i];
-      if (entityKey) {
-        updateRCForEntity(gc, refCount, entityKey, by);
+      if (Array.isArray(link[i])) {
+        updateRCForLink(gc, refCount, link[i], by);
+      } else if (link[i]) {
+        updateRCForEntity(gc, refCount, link[i] as string, by);
       }
```
kitten, line 307: @JoviDeCroock This is notable, because this was a potential bug. The logic here didn't allow for correct reference counting of entities in nested lists. It potentially was creating bogus keys here.

JoviDeCroock (approved):

PR #1592, opened 2021-04-29
[@urql/vue] Allow client to be passed to useQuery, useMutation and useSubscription

### LINKED ISSUE #1445: [@urql/vue] useQuery/useMutation must be used in setup()
@kitten  Pls look at this:

https://codesandbox.io/s/romantic-zhukovsky-gdmg6?file=/src/components/HelloWorld.vue

When using `useQuery` in an event callback not the `setup()` directly ( Just to demonstrate the issue here, I know there is a `pause` option can do the same thing), it will fail with message
```
No urql Client was provided. Did you forget to install the plugin or call `provideClient` in a parent?
```
But it does provided the client in `App.vue`.

That is because, when you call `useQuery/useMutation`, firstly it will call `useClient()` to get the client that provided in ancestor or root components, by using the vue api `inject(symbol)`.  But when you are not be inside the `setup()` function directly, the `inject` function will fail because there is no vue's `currentInstance` in the callback context.

https://github.com/FormidableLabs/urql/blob/69195d19d15d6f8b16a0e07130f5a4469494e37a/packages/vue-urql/src/useQuery.ts#L89-L93

https://github.com/FormidableLabs/urql/blob/69195d19d15d6f8b16a0e07130f5a4469494e37a/packages/vue-urql/src/useClient.ts#L15-L16

bartenra:
Resolves  #1445

## Summary

With Vue 3 you can use a an `async setup` function as also [described in the docs](https://formidable.com/open-source/urql/docs/basics/vue/#vue-suspense). This allows you to `await` promise-returning functions such as `useQuery`. There is one downside, however. Vue's dependency injection mechanism `inject` / `provide` does not work after an `await` in an `async setup` function.

Because `useQuery`, `useMutation` and `useSubscription` all call `useClient`, which in turn calls `inject`, that means it's not possible to, for example:

  - Have an `async setup` function where you `await` two queries, with the result of one query being used as an input to the second query.
  - Call `useMutation` after having `await`ing an `useQuery` call. This can often make a lot of sense. The whole idea of the Composition API is to [group code for related logic together](https://v3.vuejs.org/guide/composition-api-introduction.html). But not being able to call `useMutation` in a place where it makes sense breaks this idea.

## Set of changes

The solution to the problem above is to allow passing of a `client` to `useQuery`, `useMutation` and `useSubscription`. The default behavior is unchanged so this is a fully backward-compatible change. But users of `async setup` will be able to use `useClient` at the top of their `async setup` and pass that client explicitly to all calls to `useQuery`, `useMutation` and `useSubscription`.

### DIFF
--- a/.changeset/nine-wolves-suffer.md
+++ b/.changeset/nine-wolves-suffer.md
@@ -1,0 +1,5 @@
+---
+'@urql/vue': minor
+---
+
+Allow client to be passed to useQuery, useMutation and useSubscription
--- a/docs/api/vue.md
+++ b/docs/api/vue.md
@@ -16,8 +16,9 @@ Accepts a single required options object as an input with the following properti
 | `requestPolicy` | `?RequestPolicy`         | An optional [request policy](./core.md#requestpolicy) that should be used specifying the cache strategy. |
 | `pause`         | `?boolean`               | A boolean flag instructing [execution to be paused](../basics/vue.md#pausing-usequery).                  |
 | `context`       | `?object`                | Holds the contextual information for the query.                                                          |
+| `client`        | `?Urql.Client`           | Client to use. If not provided will use client provided in parent component with `provideClient()`.      |
 
-Each of these inputs may also be [reactive](https://v3.vuejs.org/api/refs-api.html) (e.g. a `ref`)
+Each of these inputs (except `client`) may also be [reactive](https://v3.vuejs.org/api/refs-api.html) (e.g. a `ref`)
 and are allowed to change over time which will issue a new query.
 
 This function returns an object with the shape of an [`OperationResult`](./core.md#operationresult)
@@ -46,7 +47,7 @@ once a result from the API is available.
 
 ## useMutation
 
-Accepts a single `query` argument of type `string | DocumentNode` and returns a result object with
+Accepts a `query` argument of type `string | DocumentNode` and an optional `client` argument of type `Urql.Client`, returns a result object with
 the shape of an [`OperationResult`](./core.md#operationresult) with an added `fetching` property.
 
 All of the properties on this result object are also marked as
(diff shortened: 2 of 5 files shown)


kitten:
I unfortunately will have to reject this PR for the same reason as we've rejected a similar one for Svelte: https://github.com/FormidableLabs/urql/pull/1527 (Although for SvelteKit support it seems we'll add a `prefetch` method)

The reason is that this isn't an idiomatic use of the composition API. While Vue 3 doesn't seem to document or cater for this edge case that well, the `Client` should always be read from context predictably. Now, I haven't been able to look into this but `vue-apollo` seems to have come up with a system for this that should be more intuitive and work predictably: https://github.com/vuejs/vue-apollo/blob/9f2d0d72601891f101aac8751101f50c1b4fa81f/packages/vue-apollo-composable/src/useQuery.ts#L212

But passing the client manually isn't quite the solution we're looking for as the vision for Vue 3's composition API (at least as far as I know) aligns much to the vision of React Hooks, where — although we are dealing with `async setup()` in this case — the context / provided client should still be shared statically via the use of the hook from within `setup()` rather than manually.

I think in your case, as mentioned in the issue you're referencing, you've chained two awaited `useQuery` calls; I think in that case you could still work around this by using `await Promise.all([...])` instead potentially, which also prevents a waterfall, since the queries would otherwise run sequentially; Idiomatically I would recommend you to simply merge the queries though as in GraphQL it's uncommon to see two queries on the same component / level when their data requirements can be hoisted into one query. But that's on the side, this should at least of course "just work" 😄

So we were anticipating to create an RFC once we find out more about how `vue-apollo` solves this case with client IDs (?) or rather what I assume are shared client references.

I'm very sorry to put this into text only now, now that you've put time into this. No RFC has existed for this just yet because we haven't looked into the aforementioned mechanisms yet, but [we do encourage the creation of RFCs for proposed changes](https://github.com/FormidableLabs/urql/blob/main/CONTRIBUTING.md#how-do-i-propose-changes) (rather than fixes) for any discussion to be started early. I hope we can find out how other libraries deal with this soon however :v:

bartenra:
@kitten Thanks for taking the time to comment. I agree with your comments that it's not very elegant. However, I also think this is a fundamental limitation of the inject / provide mechanism.

Apollo's implementation [does not seem to be working very well](https://github.com/vuejs/vue-apollo/issues/1082). All it seems to me is that it allows you to set a default client and allows passing a client ID to `useQuery`. If you ask me, passing a client is a much better API than than passing an ID (I know how to define a map myself and it can be made type safe).

Looking forward to know if there's more to Apollo's solution.

> await Promise.all([...])

That's not possible when one query is dependent on another query.  I personally didn't run into this particular issue though. Agreed that it's probably not idiomatic GraphQL usage.

bartenra:
Here's the docs for Apollo: https://v4.apollo.vuejs.org/guide-composable/setup.html#_2-connect-apollo-client-to-vue

If you want, I can close this MR and make an RFC out of it. If you or some other core maintainer like to do it yourself at some point that's also OK of course. Let me know, and thanks again for taking the time.

kitten:
@bartenra If you'd like to write an RFC that'd be awesome! We definitely don't want to support multiple clients, but if you have an idea of how `vue-apollo` is solving this, or how we can solve the issue of the provider access in other contexts, then feel free to open an RFC! 🙌 

I suppose the difficulty is that all lifecycle hooks, like `async setup()` are set up so that they're aware of their surrounding context, which is likely reset once `setup()` completes synchronously. Maybe @LinusBorg can also clarify whether there's a recommended solution for this? ❤️

bartenra:
I think the best solution is actually to have users import a `client` and then use `client.query`, `client.mutation` and `client.subscription`.

But the urql plugin for GraphQL Code Generator only generates types for useQuery, useMutation and useSubscription. So I think I just need to make a MR there. :+1:

kitten:
@bartenra carerul with that though; either you use these and implement a full subscription lifecycle, or what's more common, you use the `toPromise()` helper in the UI code which defeats the purpose of a GraphQL client since you'll never get any updates 😅

I was actually thinking; the easiest solution maybe to create a "client handle" an object that we can call the `use*` APIs on and access the client via.

kitten:
There are some open questions still (mainly I'm concerned about the internals of `watchEffect` and how that works) but I believe this at least brings us closer to discussing one potential solution: https://github.com/FormidableLabs/urql/issues/1596

PR #1599, opened 2021-05-02
(vue) - Implement client handles with methods to call hooks

### LINKED ISSUE #1596: RFC: Add client handles to @urql/vue to avoid invalid injections outside of a provided scope
## Summary

Currently, there are two main problems with the `@urql/vue` bindings. The Vue 3 composition API is still new and partly that means that we're getting reports over confusion with some of the API's limitations. Namely, people often call into question _when_ the hook-like functions like `useQuery` are allowed to be called. Typically they'd be called in a lifecycle, like `async setup()` where they can get access to a `Client`. However, the second problem — apart from the confusion — is that this fails if they're not called synchronously. For instance, `useQuery` cannot be called once `async setup()`'s first synchronous call expires, since the context is only set up as a per call global state. When that happens then the access to the `Client` isn't possible. (For instance, when chaining two `useQuery` calls)

This isn't very intuitive and while we could enforce these limitations of the API, some of them are just not very convenient.
While we could allow the client to be passed to the hooks this doesn't yet address the questions below, and it isn't idiomatic as per how the Composition API is supposed to be used, as far as we understand it; It could cause more confusion since it'd also accidentally allow multiple clients to be created and used and the calls to happen in any part of an app, outside of lifecycles.

#### Open Questions

- [x] is `watchEffect` effective and valid outside of synchronous `async setup()` calls?
- [x] what happens when `ref`s and effects are created and not "associated" with the setup lifecycle?

## Proposed Solution

We could abstract the `Client` and allow `client.useQuery` calls and the likes, however, this would confuse users more, likely, since it creates a "double-meaning" of what a `Client` is and what it is used for. Instead we can create **client handles**.

A **client handle** is an abstraction that allows access to the underlying client (like `useClient` today) but is tied to this client and allows calls to the hook-like functions, e.g.:

```js
import { useClientHandle } from '@urql/vue';

export default {
  async setup() {
    const handle = useClientHandle();

    const resultOne = await handle.useQuery(/* ... */);
    const resultTwo = await handle.useQuery(/* ... */);

    return { resultOne, resultTwo };
  }
};
```

This is simply achieved by creating a mapping from the client to a handle. We keep some similar state around for React: https://github.com/FormidableLabs/urql/blob/7921737454b6e274352332767a372af72166c86f/packages/react-urql/src/hooks/cache.ts#L17-L45
This will add `useQuery`, `useMutation`, and `useSubscription` hooks as methods onto a handle object, which could also contain a circular reference to the client, eliminating the need for `useClient`.

**Edit:** Just a clarification I'd like to add; usage of the hooks as they are today won't change. Simply calling `handle.useQuery` rather than `useQuery` will be enough for them to always have access to a client instance. So, the only change is basically when the client is read (i.e. injected) from context.

## Requirements

- Allow a client to be provided (providing a client _should not_ change)
- Implement a `useClientHandle` hook that creates a unique/shared handle per client, that binds the `use*` hooks as methods to the handle.
- Replace all `use*` hooks and only allow their use via the handles
- Have the hooks dynamically add to a list of `watchEffect` stop functions that the handle carries for the current component, which should be called when the component unmounts (See [comment with details](https://github.com/FormidableLabs/urql/issues/1596#issuecomment-830165093))
  - Note on the above: this may have a couple of different methods to be implemented, but it mostly depends also on how often the `setup()` function may be called, i.e. for updates, which may change the state of the effects (automatically invalidate old ones, or we'll have to manually invalidate old ones? Check `watchEffect` implementation for component lifecycles for this)

kitten:
Resolve #1596

## Summary

This adds a `useClientHandle()` function as per the #1596 RFC. The function creates a `handle` instance on which all `use*` hooks can be called as methods. This allows `handle.useMutation` to be called at any time and `handle.useQuery` and `handle.useSubscription` to be chained for more elaborate setups, e.g. for passing an ID from a `await handle.useQuery(...)` call into a `await handle.useSubscription(...)` call.

The general usage may look something like this (this is obviously pseudo code and chaining should be avoided when possible to avoid query waterfalls):

```js
import { ref, computed } from 'vue';
import { gql, useClientHandle } from '@urql/vue';

export default {
  async setup() {
    const handle = useClientHandle();

    const pokemons = await handle.useQuery({
      query: gql`
        {
          pokemons(limit: 10) {
            id
            name
          }
        }
      `
    });

    const index = ref(0);

    const pokemon = await handle.useQuery({
      query: gql`
        query ($id: ID!) {
          pokemon(id: $id) {
            id
            name
          }
        }
      `,
      variables: {
        id: computed(() => pokemons.data.value.pokemons[index.value].id),
      },
    });

    return {
      pokemons: pokemons.data,
      pokemon: pokemon.data,
      nextPokemon() {
        index.value = index.value < (pokemons.data.value.pokemons.length - 1)
          ? index.value + 1
          : 0;
      },
    };
  },
}
```

## Set of changes

- Add a more helpful error message to `useClient()` when it's called outside of `setup()` and other lifecycle hooks
- Add `useClientHandle()` to preserve client instance across asynchronous hook calls
- Add errors to `useClientHandle()` methods in development when they're used after a component has mounted and it's called in another lifecycle hook
- Add stop functions to all hooks so that all `watchEffect`s are still bound to the component's unmount when called via `handle.use*`

### DIFF
--- a/.changeset/beige-candles-kneel.md
+++ b/.changeset/beige-candles-kneel.md
@@ -1,0 +1,5 @@
+---
+'@urql/vue': minor
+---
+
+A `useClientHandle()` function has been added. This creates a `handle` on which all `use*` hooks can be called, like `await handle.useQuery(...)` or `await handle.useSubscription(...)` which is useful for sequentially chaining hook calls in an `async setup()` function or preserve the right instance of a `Client` across lifecycle hooks.
--- a/.changeset/fluffy-ligers-draw.md
+++ b/.changeset/fluffy-ligers-draw.md
@@ -1,0 +1,5 @@
+---
+'@urql/vue': patch
+---
+
+The `useClient()` function will now throw a more helpful error when it's called outside of any lifecycle hooks.
--- a/docs/api/vue.md
+++ b/docs/api/vue.md
@@ -104,14 +104,28 @@ The result furthermore carries several utility methods:
 [Read more about how to use the `useSubscription` API on the "Subscriptions"
 page.](../advanced/subscriptions.md#vue)
 
+## useClientHandle
+
+The `useClientHandle()` function may, like the other `use*` functions, be called either in
+`setup()` or another lifecycle hook, and returns a so called "client handle". Using this `handle` we
+can access the [`Client`](./core.md#client) directly via the `client` property or call the other
+`use*` functions as methods, which will be directly bound to this `client`. This may be useful when
+chaining these methods inside an `async setup()` lifecycle function.
+
+| Method                 | Description                                                                                                               |
+| ---------------------- | ------------------------------------------------------------------------------------------------------------------------- |
+| `client`               | Contains the raw [`Client`](./core.md#client) reference, which allows the `Client` to be used directly.                   |
+| `useQuery(...)`        | Accepts the same arguments as the `useQuery` function, but will always use the `Client` from the handle's context.        |
+| `useMutation(...)`     | Accepts the same arguments as the `useMutation` function, but will always use the `Client` from the handle's context.     |
+| `useSubscription(...)` | Accepts the same arguments as the `useSubscription` function, but will always use the `Client` from the handle's context. |
+
 ## Context API
 
 In Vue the [`Client`](./core.md#client) is provided either to your app or to a parent component of a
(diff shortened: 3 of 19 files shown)


JoviDeCroock (approved):
wooow, this is awesome

PR #1601, opened 2021-05-02
(docs) - Expand willAuthError section with more notes on usage

kitten:


docs/advanced/authentication.md
```diff
@@ -284,6 +284,45 @@ const willAuthError = ({ authState }) => {
...
+          definition.kind === 'OperationDefinition' &&
+          definition.selectionSet.selections.some(node => {
+            // The field name is just an example, since signup may also be an exception
+            return node.kind === 'Field' && node.name.value === 'login';
+          })
+        );
+      })
+    );
+
+    return true;
```
kadikraman, line 313: We'd never actually reach this line. Could be worth assigning the above to a variable, e.g.
```js
const isLoginMutation = ... ;
return isLoginMutation ? false : true;
```

kadikraman (approved):

docs/advanced/authentication.md
kitten, line 313:
```suggestion
```

Issue #1567, opened 2021-04-26
RFC: Support next/link prefetched data in next-urql


yoshiwarab:
## Summary

The `Link` component from next/link has prefetch turned on by default. For pages using static generation, this includes preloaded JSON data. Currently pages wrapped by withUrql ignore this data on client side transitions. I propose that support is added to take advantage of the prefetched SSG data. 

## Proposed Solution

Currently if an ssr exchange already exists, withUrql ignores `pageProps.urqlState` so it doesn't get the prefetched data on a client side transition.  By updating this https://github.com/FormidableLabs/urql/blob/b6890ba1b0c9fa9e99301338555ea10d602cee22/packages/next-urql/src/with-urql-client.ts#L55-L58 to restore data to an existing ssr exchange it would be possible to hydrate the client with the prefetched data.
`} else if (ssr) {
          ssr.restoreData(urqlState)
}`

Also, thank you for the amazing library :)! Hopefully this wouldn't break something else!

kitten:
I think we'll need to test this; Personally, I don't have a huge overview of `next-urql` but @JoviDeCroock may know whether this will "just work". I don't think we've ever tested the prefetch behaviour, so once we see how it works we can probably test it properly. But in theory this does check out for me :+1:

PR #1602, opened 2021-05-03
(next) - allow subsequent statically generated pages to hydrate the ssr-cache

JoviDeCroock:
Resolves https://github.com/FormidableLabs/urql/issues/1567

## Summary

Next will inject data available for a route as JSON, similar to how it does so for the first-page-serve, this is why we have to add a special else branch to account for this scenario.

## Set of changes

- Add check for when a client-side transition introduces new data to hydrate into the SSG page.


examples/with-next/pages/static-2.js
```diff
@@ -0,0 +1,54 @@
+import { initUrqlClient } from "next-urql";
```
JoviDeCroock, line 1: Can remove this page, mainly meant to illustrate/give a testing point for reviewers 😅

kitten: Yea probably worth removing once you're done 😅

packages/next-urql/src/with-urql-client.ts
```diff
@@ -55,6 +55,8 @@ export function withUrqlClient(
         if (!ssr || typeof window === 'undefined') {
           // We want to force the cache to hydrate, we do this by setting the isClient flag to true
           ssr = ssrExchange({ initialState: urqlServerState, isClient: true });
+        } else if (ssr && typeof window !== 'undefined') {
+          ssr.restoreData(urqlServerState);
```
kitten, line 59: That's really a pleasant surprise!

kitten (approved):

PR #1605, opened 2021-05-04
(examples) - With auth

yankovalera:
## Summary

Add an example for auth with refresh tokens using authExchange


examples/with-auth/src/client/index.js
```diff
@@ -0,0 +1,84 @@
...
+
+const didAuthError = ({ error }) => {
+  return error.graphQLErrors.some(e => e.extensions?.code === 'UNAUTHORIZED');
+};
+
+const client = createClient({
+  url: 'https://trygql.dev/graphql/web-collections',
+  exchanges: [
+    dedupExchange,
+    authExchange({ getAuth, addAuthToOperation, didAuthError }),
```
kitten, line 79: Just a quick note that the cache is missing here; I think for the sake of some folks copy pasting I'd add it back

examples/with-auth/README.md
```diff
@@ -0,0 +1,21 @@
+# Integrating with `@urql/exchange-auth`'s authExchange
```
kitten, line 1: Just a small note, I'd call this example `with-refresh-auth` or something of the likes to make sure the name isn't a catch all (ie if someone just adds a token without refresh)

yankovalera: sounds good

kitten (approved):

PR #1607, opened 2021-05-04
Add with-react-native example

yankovalera:
## Summary

Add with-react-native example

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,8 @@ tmp/
 dist/
 examples/*/public
 examples/*/yarn.lock
+examples/*/ios/
+examples/*/android/
+examples/*/.watchmanconfig
+examples/*/metro.config.js
+examples/*/babel.config.js
--- a/examples/with-react-native/package.json
+++ b/examples/with-react-native/package.json
@@ -1,0 +1,27 @@
+{
+  "name": "withreactnative",
+  "version": "0.0.1",
+  "private": true,
+  "scripts": {
+    "android": "react-native run-android",
+    "ios": "react-native run-ios",
+    "start": "react-native start",
+    "test": "jest",
+    "lint": "eslint ."
+  },
+  "dependencies": {
+    "graphql": "^15.5.0",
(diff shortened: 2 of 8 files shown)


examples/with-react/src/pages/PokemonList.js
```diff
@@ -16,19 +17,9 @@ const PokemonList = () => {
...
-
-      {data && (
-        <ul>
-          {data.pokemons.map((pokemon) => (
-            <li key={pokemon.id}>{pokemon.name}</li>
-          ))}
-        </ul>
-      )}
-    </div>
+    <View>
```
JoviDeCroock, line 20: I think this isn't intended?

examples/with-react-native/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "react-native": "0.64.0",
+    "urql": "^2.0.2"
+  },
+  "devDependencies": {
+    "@babel/core": "^7.12.9",
+    "@babel/runtime": "^7.12.5",
+    "babel-jest": "^26.6.3",
+    "metro-react-native-babel-preset": "^0.64.0"
+  },
+  "jest": {
```
JoviDeCroock, line 24: We can remove jest

examples/with-react/src/pages/PokemonList.js
yankovalera: oops

JoviDeCroock (approved):

.gitignore
```diff
@@ -17,4 +17,9 @@ exchanges/*/LICENSE
 tmp/
 dist/
 examples/*/public
-examples/*/yarn.lock
\ No newline at end of file
+examples/*/yarn.lock
+examples/*/ios/
```
kitten, line 21: Since these are one offs I'd maybe just add them to the React Native example & create a new gitignore there 🤔

yankovalera: I'd prefer them here in case we add more RN examples later, which could be the case for e.g. handling auth, etc.

PR #1625, opened 2021-05-09
(docs) - use formideploy and change base-path

JoviDeCroock:
## Summary

With this PR we'll use `formideploy` to have consistent deployments as well as change the base-path to always be `/open-source/urql` just like itis on the live-site

## Set of changes

- use formideploy to deploy staging/production
- use `/open-source/urql` as basepath


.github/workflows/site.yml
```diff
@@ -62,8 +62,9 @@ jobs:
       - name: Deploy Production Site
         working-directory: ./packages/site
         if: github.ref == 'refs/heads/main'
-        run: node scripts/deploy/aws.js
+        run: yarn run deploy:prod
```
kitten, line 65: Let's just call `yarn run formideploy deploy --production` (and the likes) here, since we'd never run this manually, I suppose

packages/site/formideploy.config.js
```diff
@@ -0,0 +1,13 @@
...
+module.exports = {
+  lander: {
+    name: 'urql',
+  },
+  production: {
+    // TODO: temp retarget when we have verified this working
+    domain: 'tmp-experiment-02.formidable.com',
+    bucket: 'tmp-experiment-02.formidable.com',
+  },
+};
```
kitten, line 13: Why is this a random `.config.js` file? Would this be more suited in a different location? And I think it should be safe to use the prod bucket here, assuming it's the same logic as before, since the scripts seem almost identical :+1:

packages/site/package.json
```diff
@@ -10,10 +10,11 @@
     "clean": "rimraf dist",
     "prepublishOnly": "run-s clean build",
     "stage:build": "react-static build --staging",
-    "stage:deploy": "node scripts/deploy/surge.js",
     "prod:build": "react-static build",
-    "prod:deploy": "node scripts/deploy/aws.js",
-    "prod:serve": "serve -l 3000"
+    "prod:serve": "serve -l 3000",
+    "serve": "formideploy serve",
```
kitten, line 15: Is it possible for us then to get rid of `serve` here?

yarn.lock
```diff
@@ -12711,11 +12894,16 @@ react-inlinesvg@^1.2.0:
     exenv "^1.2.2"
     react-from-dom "^0.3.0"
 
-"react-is@^16.12.0 || ^17.0.0", react-is@^16.13.1, react-is@^16.6.0, react-is@^16.7.0, react-is@^16.8.1, react-is@^16.8.6, react-is@^17.0.1, react-is@^17.0.2:
+"react-is@^16.12.0 || ^17.0.0", react-is@^17.0.1, react-is@^17.0.2:
   version "17.0.2"
   resolved "https://registry.yarnpkg.com/react-is/-/react-is-17.0.2.tgz#e691d4a8e9c789365655539ab372762b0efb54f0"
   integrity sha512-w2GsyukL62IJnlaff/nRegPQR94C/XXamvMWmSHRJ4y7Ts/4ocGRmTHvOs8PSE6pB3dWOrD/nueuU5sduBsQ4w==
 
+react-is@^16.13.1, react-is@^16.6.0, react-is@^16.7.0, react-is@^16.8.1, react-is@^16.8.6:
```
kitten, line 12902: Seems like this is missing a `yarn-deduplicate` run and `yarn` also wasn't run in the root (since we have some resolutions?)

.github/workflows/site.yml
JoviDeCroock: Good psot, done

packages/site/package.json
JoviDeCroock: Just remove it?

kitten (approved):

PR #1626, opened 2021-05-09
(core) - Fix concurrent operations not respecting request policies

### LINKED ISSUE #1621: Re-executing query doesn't work
Hi there. I'm trying to upgrade a project from `urql@1.11.6` to `urql@2.0.3`, and I'm running into an issue when trying to re-execute a query using the `'network-only'` request policy. I've created a minimal reproduction here https://github.com/bausmeier/urql-repro, and the request in question is [in the `AccountInfo` component](https://github.com/bausmeier/urql-repro/blob/b329302bc47af6dbecd1fc0222528b505785035c/src/AccountInfo.jsx#L20).

Effectively what I've got is two components which render alongside each other:
1. A panel which shows some details of the viewer's account.
2. A login form which on success [stores an access token](https://github.com/bausmeier/urql-repro/blob/b329302bc47af6dbecd1fc0222528b505785035c/src/LoginForm.jsx#L11) so that it can be added to the `Authorization` header of subsequent requests [in `fetchOptions`](https://github.com/bausmeier/urql-repro/blob/b329302bc47af6dbecd1fc0222528b505785035c/src/UrqlProvider.jsx#L9-L19).

The rough reproduction looks like this:

2021-05-08-193100_grim

The `viewer` query which I'm executing will return `null` if there is no `Authorization` header, and an object containing an id if there is an `Authorization` header.

What I expect to happen, and what _does_ happen with `urql@1.11.6`, is that when the "Refresh" button is pressed, a GraphQL network request is made to get the viewer. If an access token has been set by "logging in" then it should be included in the request, and the account info should display the viewer's id.

What actually happens with `urql@2.0.3` is that after the initial request is made on page load, pressing the "Refresh" button doesn't fire off any more GraphQL requests at all.

I may be doing something completely stupid, but this used to work with the older version of urql, and a similar approach works with Apollo, so I'm a bit perplexed as to why it's not working with the latest urql. I'm happy to provide more info if needed.

kitten:
Resolves #1621

## Summary

We discovered a slight regression in `@urql/core@2.1.0` that caused reexecuting an operation to not be possible under the usual circumstances anymore. The reason for this was that the `replayOnStart` utility was effective but we'd share the entire source too eagerly. Since this source was shared, no new operation would be created for future concurrent runs which makes changing the `requestPolicy` of an operation, e.g. for the `executeQuery` function in bindings, impossible.

Instead, I've now moved the implementation back to `Client` and cleaned it up a little, which should make it easier for us to predict how it'll behave. Sources aren't completely shared anymore and instead (which was the same in a prior implementation) we only share the result source and store that internally. The `OperationResult` stream and `onStart` behaviour isn't shared any longer.

### DIFF
--- a/.changeset/new-dingos-clean.md
+++ b/.changeset/new-dingos-clean.md
@@ -1,0 +1,5 @@
+---
+'@urql/core': patch
+---
+
+Fix a regression in `@urql/core@2.1.1` that prevented concurrent operations from being dispatched with differing request policies, which for instance prevented the explicit `executeQuery` calls on bindings to fail.
--- a/packages/core/src/client.test.ts
+++ b/packages/core/src/client.test.ts
@@ -565,6 +565,63 @@ describe('shared sources behavior', () => {
     expect(resultTwo).toHaveBeenCalledTimes(1);
   });
 
+  it('dispatches the correct request policy on subsequent sources', async () => {
+    const exchange: Exchange = () => ops$ => {
+      let i = 0;
+      return pipe(
+        ops$,
+        map(op => ({
+          data: ++i,
+          operation: op,
+        })),
+        delay(1)
+      );
+    };
+
+    const client = createClient({
+      url: 'test',
+      exchanges: [exchange],
+    });
+
+    const resultOne = jest.fn();
+    const resultTwo = jest.fn();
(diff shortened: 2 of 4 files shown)


JoviDeCroock (approved):
Awesome, that was quick!

PR #1669, opened 2021-05-25
(docs) - Polled queries are no longer supported

bartenra:
I saw that polling is supported in the docs, but couldn't find anything when searching for it. Then I found out the feature was actually [removed](https://github.com/FormidableLabs/urql/pull/1374).

I think the docs should be updated to reflect this change.

In case you want to include snippets in the individual docs of how to do polling, the Vue equivalent of @kitten's React snippet would be:

```js
let { data, fetching, executeQuery } = useQuery(...);

const id = setInterval(() => {
  if (!fetching)
    executeQuery({ requestPolicy: 'network-only' });
}, 1000);

onBeforeUnmount(() => {
  clearInterval(id);
});
```

I am happy to add the React and Vue snippets for polling, however, someone else would have to chip in for svelte since I'm not familiar with it.

### DIFF
--- a/docs/comparison.md
+++ b/docs/comparison.md
@@ -43,7 +43,7 @@ All features are marked to indicate the following:
 | Devtools                                   | ✅                                  | ✅                                            | ✅                             |
 | Subscriptions                              | ✅                                  | ✅                                            | ✅                             |
 | Client-side Rehydration                    | ✅                                  | ✅                                            | ✅                             |
-| Polled Queries                             | ✅                                  | ✅                                            | ✅                             |
+| Polled Queries                             | 🛑                                  | ✅                                            | ✅                             |
 | Lazy Queries                               | ✅                                  | ✅                                            | ✅                             |
 | Stale while Revalidate / Cache and Network | ✅                                  | ✅                                            | ✅                             |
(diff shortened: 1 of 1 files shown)


docs/comparison.md
```diff
@@ -43,7 +43,7 @@ All features are marked to indicate the following:
 | Devtools                                   | ✅                                  | ✅                                            | ✅                             |
 | Subscriptions                              | ✅                                  | ✅                                            | ✅                             |
 | Client-side Rehydration                    | ✅                                  | ✅                                            | ✅                             |
-| Polled Queries                             | ✅                                  | ✅                                            | ✅                             |
+| Polled Queries                             | 🛑                                  | ✅                                            | ✅                             |
```
kitten, line 46: Cheers! 🙌 The only thing I'd propose is to use "🔶" because we removed the option since it requires very little code to implement in userland, i.e. "🔶" standing for "Supported and documented, but requires custom user-code to implement."

I believe what we need to do though is to make sure to readd it to the docs with our suggested code, which we haven't done yet because it'd be grouped with docs on the "Request Policy" and "Refocus" exchanges, which we in turn wanted to merge into a single one first

```suggestion
| Polled Queries                             | 🔶                                  | ✅                                            | ✅                             |
```

PR #1681, opened 2021-05-31
(graphcache) extend Primitive type to include Date

eschaefer:
## Summary

Return type for local resolvers could include a `Date`, which is even in the [documentation](https://formidable.com/open-source/urql/docs/graphcache/local-resolvers/) for local resolvers. This PR extends the `Primitive` type so that this is supported.

## Set of changes

Graphcache's `Primitive` type now includes `Date` in the union.

### DIFF
--- a/.changeset/green-carpets-enjoy.md
+++ b/.changeset/green-carpets-enjoy.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-graphcache': patch
+---
+
+Graphcache Primitive type now includes Date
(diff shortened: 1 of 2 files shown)


kitten:
`Date` should already be caught by `ScalarObject` which as been adjusted to account for constructors. That's because a date isn't a primitive but an object so it doesn't fall into the same type for us.

That said, you won't get a date object from a GraphQL API directly, so there's some behaviour here that's only desirable on the outputs hence we're adding the differentiation since these types are partially also used internally

eschaefer:
@kitten I just noticed that `ScalarObject` was updated in the latest version. Thanks for the heads-up, I was still one one patch version behind.

kitten:
@eschaefer oh, that makes a lot of sense! sorry fromy end, I didn't write the down that it was a rather new addition. I must've had my head in the clouds 😄

PR #1715, opened 2021-06-09
(next) - Prevent restoring SSR data when client has been reset

### LINKED ISSUE #1714: (next-urql) - doing resetUrqlClient doesn't refetch the graphql queries and uses previous client or previous ssr cache
```js
import { withUrqlClient, SSRExchange, NextUrqlContext } from 'next-urql'
import { dedupExchange, cacheExchange, fetchExchange, ClientOptions, Client } from 'urql'

withUrqlClient(
    (_ssrExchange: SSRExchange, ctx?: NextUrqlContext): ClientOptions => {
      const user = ctx?.getUserData ? ctx.getUserData() : getUserData()
      const exchanges = [dedupExchange, cacheExchange, _ssrExchange, fetchExchange]

      if (process.env.NODE_ENV !== 'production' && process.browser) {
        exchanges.unshift(require('@urql/devtools').devtoolsExchange)
      }

      return {
        url: "https://graphql.fauna.com/graphql",//meteorBackend.graphql,
        suspense: true,
        exchanges,
        fetchOptions: {
          headers: {
            authorization: "Basic Zm5BRFFVdWNRb0FDQ1VpZDAxeXVIdWt2SnptaVY4STI4a2R6Y0p2UDo=",//user?.token
          },
          credentials: 'same-origin',
        },
      }
    },
    { ssr: true }
  )(UrqlWrapper(Component)
```

**urql version & exchanges:** 
urql: 2.0.3
next-urql: 3.1.0
**exchanges:** 
production: [dedupExchange, cacheExchange, _ssrExchange, fetchExchange]
development: [devtoolsExchange,dedupExchange, cacheExchange, _ssrExchange, fetchExchange]

Screenshot 2021-06-09 at 7 15 20 PM

Reset calls the resetUrqlClient.

Biboswan:
Resolves #1714 

This aims to fix two things
1. When we first visit a page and later do `resetUrqlClient`. We do end up using a new client because `urqlClient` prop was set to null. But due to `ssr.restoreData(urqlServerState)` Results are fetched from old ssr cache and network isn't called.
2. When we navigate to a new page (client side rendered) `urqlClient` prop is not null and that is the client we use. Now if you call `resetUrqlClient`, we don't get a `new client` because of existing `urqlClient` prop value. `if (urqlClient) return  urqlClient`

### DIFF
--- a/packages/next-urql/src/with-urql-client.ts
+++ b/packages/next-urql/src/with-urql-client.ts
@@ -1,4 +1,4 @@
-import React, { createElement, useState } from 'react';
+import { createElement, useCallback, useReducer, useMemo } from 'react';
 import ssrPrepass from 'react-ssr-prepass';
 
 import {
@@ -42,20 +42,18 @@ export function withUrqlClient(
       urqlState,
       ...rest
     }: WithUrqlProps) => {
-      // eslint-disable-next-line react-hooks/rules-of-hooks
-      const forceUpdate = useState(0);
+      const [version, forceUpdate] = useReducer(prev => prev + 1, 0);
       const urqlServerState = (pageProps && pageProps.urqlState) || urqlState;
 
-      // eslint-disable-next-line react-hooks/rules-of-hooks
-      const client = React.useMemo(() => {
-        if (urqlClient && forceUpdate[0] === 0) {
+      const client = useMemo(() => {
+        if (urqlClient && !version) {
           return urqlClient;
         }
 
         if (!ssr || typeof window === 'undefined') {
           // We want to force the cache to hydrate, we do this by setting the isClient flag to true
           ssr = ssrExchange({ initialState: urqlServerState, isClient: true });
-        } else if (forceUpdate[0] === 0) {
+        } else if (!version) {
           ssr.restoreData(urqlServerState);
         }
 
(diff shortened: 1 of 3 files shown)


packages/next-urql/src/with-urql-client.ts
```diff
@@ -48,14 +48,14 @@ export function withUrqlClient(
...
-        if (urqlClient) {
+        if (urqlClient && forceUpdate[0] === 0) {
           return urqlClient;
         }
 
         if (!ssr || typeof window === 'undefined') {
           // We want to force the cache to hydrate, we do this by setting the isClient flag to true
           ssr = ssrExchange({ initialState: urqlServerState, isClient: true });
-        } else if (ssr && typeof window !== 'undefined') {
+        } else if (forceUpdate[0] === 0) {
```
Biboswan, line 58: `ssr && typeof window !== 'undefined` is removed because that condition is automatically implied in the else if block due to demorgan's law

JoviDeCroock:
I've been a bit on and off about this particular issue as it introduces a very undeterministic scenario, when we have different SSG'ed pages on our web-application when reset is called these won't be able to restore anymore. Afaik server-side injecting authorized state is not really advised so logging off and on by resetting the urql-client generally won't pull data from `__NEXT_DATA__` if this is done. I guess when we talk about pure SSR applications without a very strict CDN there will be cases where users store authorized state in this injected part, I would personally advise clearing the authorized state from the dom-node that injects this rather than fully clearing the client.

kitten (approved):

PR #1720, opened 2021-06-10
feat: refactor fetch — get ready for defer/stream

maraisr:
This PR aims to simplify the `fetchSource` so that we get access to the  raw `response` — in that we can later introduce defer/stream pipes.

### DIFF
--- a/exchanges/multipart-fetch/src/multipartFetchExchange.test.ts
+++ b/exchanges/multipart-fetch/src/multipartFetchExchange.test.ts
@@ -153,7 +153,7 @@ describe('on error', () => {
     expect(data).toMatchSnapshot();
   });
 
-  it.skip('ignores the error when a result is available', async () => {
+  it('ignores the error when a result is available', async () => {
     fetch.mockResolvedValue({
       status: 400,
       json: jest.fn().mockResolvedValue(response),
--- a/exchanges/multipart-fetch/src/multipartFetchExchange.test.ts
+++ b/exchanges/multipart-fetch/src/multipartFetchExchange.test.ts
@@ -153,7 +153,7 @@ describe('on error', () => {
     expect(data).toMatchSnapshot();
   });
 
-  it('ignores the error when a result is available', async () => {
+  it.skip('ignores the error when a result is available', async () => {
     fetch.mockResolvedValue({
       status: 400,
       json: jest.fn().mockResolvedValue(response),
--- a/packages/core/src/internal/fetchSource.ts
+++ b/packages/core/src/internal/fetchSource.ts
@@ -14,7 +14,8 @@ export const makeFetchSource = (
     const fetcher = operation.context.fetch || fetch;
 
     let ended = false;
-    let catch_response: Response;
+    let statusNotOk = false;
+    let response: Response;
 
     Promise.resolve()
       .then(() => {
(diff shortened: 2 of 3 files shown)


exchanges/multipart-fetch/src/multipartFetchExchange.test.ts
```diff
@@ -153,7 +153,7 @@ describe('on error', () => {
     expect(data).toMatchSnapshot();
   });
 
-  it('ignores the error when a result is available', async () => {
+  it.skip('ignores the error when a result is available', async () => {
```
maraisr, line 156: @JoviDeCroock @kitten — can you please assist me. Was struggling to understand the point of this test? If there was an error (400), then we should error and not return data?

Or is the idea to always return the data — even if there was an error?

packages/core/src/internal/fetchSource.ts
```diff
@@ -57,21 +24,44 @@ export const makeFetchSource = (
...
-          ended = true;
-          if (result) next(result);
-          complete();
-        }
-      });
+      // now we handle the maybe response
+      .then(async (response: Response | undefined) => {
+        if (!response || ended) return;
+
+        const networkResult = await response.json();
```
maraisr, line 45: @kitten  this is the line where we'd do the `meros` checks with the mime. (which is [built into meros](https://github.com/maraisr/meros/blob/main/src/browser.ts#L27) anyway)

packages/core/src/internal/fetchSource.ts
```diff
@@ -57,21 +24,44 @@ export const makeFetchSource = (
...
+
+        const result = makeResult(operation, networkResult, response);
+        if (result) next(result);
+      })
+      .catch((error: Error) => {
+        if (error.name === 'AbortError') return;
+
+        next(makeErrorResult(operation, error, catch_response));
+      })
+      .finally(() => complete());
```
maraisr, line 60: `.finally` is correctly polyfilled yeah?

exchanges/multipart-fetch/src/multipartFetchExchange.test.ts
JoviDeCroock: We allow for errors and results to both be in the response, think about a sub-resolver failing, that sub-resolver will insert an error message into the [`CombinedError`](https://formidable.com/open-source/urql/docs/basics/errors/) but when that sub-resolver is optional we can still return a valid result to the user.

maraisr: Should be ready again for you @JoviDeCroock 🤘🏻

packages/core/src/internal/fetchSource.ts
```diff
@@ -57,21 +25,41 @@ export const makeFetchSource = (
...
-        if (!ended) {
-          ended = true;
-          if (result) next(result);
-          complete();
-        }
-      });
+      .then(async (res: Response | undefined) => {
+        if (!res || ended) return;
+
+        let result = await res.json();
```
maraisr, line 39: defer/stream will slice into here — we get access to raw response, and next's

packages/core/src/internal/fetchSource.ts
```diff
@@ -57,21 +25,41 @@ export const makeFetchSource = (
...
+        });
       })
-      .then((result: OperationResult | undefined) => {
-        if (!ended) {
-          ended = true;
-          if (result) next(result);
-          complete();
-        }
-      });
+      .then(async (res: Response | undefined) => {
```
kitten, line 36: I'd avoid mixing `async/await` and `then` chaining in the same blocks in a file. We do have a Babel plugin that converts these to regular `then` chains, but I'd prefer to make due without `async/await` where possible or alternatively check transpilation output. I think at least in this case we can definitely avoid this

packages/core/src/internal/fetchSource.ts
kitten: It's in fact not, and dynamic transpilation of methods is always a hairy situation, so I'd opt for not doing this here, personally. This will likely also become clearer as we add the actual multipart logic into this file, but generally, I think it isn't a bad thing to make sure that whatever uses `makeResult` also calls `next(result)` _and_ `complete()` as needed, since in multipart responses we can clearly tell when we're done.

For errors we can also always call it eagerly as we know we're done anyway

packages/core/src/internal/fetchSource.ts
maraisr: its just a tad tricky to handle that "early return" in this scenario without it. `.then` chains will continue to execute. If the response is ended, we need to "break out" of the .then chain. Think the only way is to "throw".

maraisr:
PR is ready for you again @kitten — I think i can make this work w/o async generators or async/await.

kitten:
would it be easier here to directly jump to the implementation of multipart responses?

maraisr:
@kitten  — ya probably. Already started, and its already deviated a tad from this. 

I will close this PR, and get cracking on the real thing.

PR #1743, opened 2021-06-25
(core) - Add ability to customize logs in debugExchange

Mookiies:
## Summary
Adds the ability for customizable logs in the `debugExchange`  which makes it more useful for a few reasons:
- Can have multiple debug exchanges since you can adjust the exchange to label log output depending which order its in
- Can change console severity  (useful for react-native since `warn`/`error`'s can show up on screen)
- Can drill down to only log pertinent information about an operation or result

This `debugExchange` has been especially useful for me since the library ships minified which makes stepping through library code with a debugger less viable.

## Set of changes
- Adds optional configuration options  `onIncoming` and `onCompleted` to the debugExchange. Both are functions intented to be used for custom log output.
- Updates documentation.
- As a consequence of having configuation options in the `debugExchange` api changes 

Before:
```
createClient({
   ...
   debugExchange,
   ...
})
```

After
```
createClient({
   ...
   debugExchange({}),
   ...
})
```
Not sure if this is considered a breaking change.

### DIFF
--- a/.changeset/wicked-queens-applaud.md
+++ b/.changeset/wicked-queens-applaud.md
@@ -1,0 +1,5 @@
+---
+'@urql/core': minor
+---
+
+Adds ability to customize logs in the debugExchange
--- a/.changeset/wicked-queens-applaud.md
+++ b/.changeset/wicked-queens-applaud.md
@@ -1,5 +1,5 @@
 ---
-'@urql/core': minor
+'@urql/core': major
 ---
 
 Adds ability to customize logs in the debugExchange
--- a/docs/advanced/debugging.md
+++ b/docs/advanced/debugging.md
@@ -138,3 +138,8 @@ custom exchanges:
 - ❌ **Don't send warnings in debug events**: Informing your user about warnings isn't effective
   when the event isn't seen. You should still rely on `console.warn` so all users see your important
   warnings.
+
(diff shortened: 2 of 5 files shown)


kitten:
Definitely a very logical addition to the internals 👍 I'll defer to @JoviDeCroock for a second opinion on this.

One thing that's already obvious is that it's a break change,waning that we may want to delay landing this for a more opportune moment. Since it has no bundle size impact I also have no concerns regarding any specific code decisions, but it's implemented in an obvious way since its not too large of a change

.changeset/wicked-queens-applaud.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/core': minor
```
JoviDeCroock, line 2:
```suggestion
'@urql/core': major
```

JoviDeCroock (approved):
Looks like a good addition to me!

moosefaceee:
:( need this

kitten:
Sorry for the delay here! We couldn't find a way to fit this in without breaking changes, and the breaking change didn't quite sit right with us, as the usage of the `debugExchange` is indeed very basic, and allowing it to be customised doesn't immediately add the value we were seeking.

It's been superseded by #2846

PR #1744, opened 2021-06-26
(exchange-persisted-fetch) - Make indirect eval call to optimize for minifiers

### LINKED ISSUE #875: Avoid eval in persisted-fetch exchange
```
"@urql/exchange-graphcache": "3.0.1",
"@urql/exchange-persisted-fetch": "1.0.1",
"@urql/svelte": "0.3.0",
```

I'm trying for the first time https://github.com/FormidableLabs/urql/tree/master/exchanges/persisted-fetch exchange.

In my https://github.com/sveltejs/template project Rollup warns with:

image

Is there a way to avoid using `eval` here: https://github.com/FormidableLabs/urql/blob/master/exchanges/persisted-fetch/src/sha256.ts#L34?

This is stopping all of our customers' CI pipelines.

Thanks for your amazing work!

nderscore:
## Summary

The `@urql/exchange-persisted-fetch` package is not currently being minified to the fullest extent and even throws warnings in some minifiers/bundlers, such as esbuild, because it contains a direct call to `eval()`.

Every minifier (including terser, which is being used to generate minified bundles of urql libraries) avoids mangling names when there is a direct `eval` call in scope. Because a direct call to eval gives the code being evaluated access to the scope from which it was called, there's no way for the minifier to know what side-effects that evaluted code has, so it avoids making optimizations.

By changing it to an indirect eval, we get smaller outputs, even though more boilerplate was added to the code:

* `urql-exchange-persisted-fetch.min.js` shrinks from `3299b` to <strike>`2802b`</strike> `2792b`
* `urql-exchange-persisted-fetch.min.mjs` shrinks from `3281b` to <strike>`2828b`</strike> `2818b`

This PR also resolves #875 which I believe was erroneously closed.

## Set of changes

* Replaced direct `eval()` call in `exchanges/persisted-fetch/src/sha256.ts` with an indirect call, guarded against errors with a try/catch.

## Side-note

This is probably not the _ideal_ fix. I question whether an `eval` here is even appropriate as a means of avoiding including `crypto` in a client-side bundle.

It would probably be better to generate separate builds for node and browser environments, where a direct call to `require('crypto')` gets tree-shaken out of the `browser` build.

### DIFF
--- a/exchanges/persisted-fetch/src/sha256.ts
+++ b/exchanges/persisted-fetch/src/sha256.ts
@@ -30,18 +30,14 @@ const sha256Browser = (bytes: Uint8Array): Promise<Uint8Array> => {
   });
 };
 
-const nodeCrypto =
-  typeof window === 'undefined'
-    ? (() => {
-        try {
-          // Call eval indirectly to guarantee no side-effects in module scope
-          // (optimization for minifiers)
-          return (0, eval)("r => r('crypto')")(require);
-        } catch (e) {
-          return null;
-        }
-      })()
-    : null;
+let nodeCrypto;
+if (typeof window === 'undefined') {
+  try {
+    // Indirect eval/require to guarantee no side-effects in module scope
+    // (optimization for minifiers)
+    nodeCrypto = new Function('require', 'return require("crypto")')(require);
+  } catch (e) {}
+}
 
 export const hash = async (query: string): Promise<string> => {
   if (
--- a/exchanges/persisted-fetch/src/sha256.ts
+++ b/exchanges/persisted-fetch/src/sha256.ts
@@ -31,7 +31,17 @@ const sha256Browser = (bytes: Uint8Array): Promise<Uint8Array> => {
 };
 
 const nodeCrypto =
-  typeof window === 'undefined' ? eval("require('crypto')") : null;
+  typeof window === 'undefined'
+    ? (() => {
+        try {
+          // Call eval indirectly to guarantee no side-effects in module scope
+          // (optimization for minifiers)
+          return (0, eval)("r => r('crypto')")(require);
+        } catch (e) {
+          return null;
+        }
+      })()
+    : null;
 
 export const hash = async (query: string): Promise<string> => {
   if (
--- a/.changeset/silent-carpets-judge.md
+++ b/.changeset/silent-carpets-judge.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-persisted-fetch': patch
+---
+
+Optimize for minification by avoiding direct eval call.


exchanges/persisted-fetch/src/sha256.ts
```diff
@@ -31,7 +31,17 @@ const sha256Browser = (bytes: Uint8Array): Promise<Uint8Array> => {
 };
 
 const nodeCrypto =
-  typeof window === 'undefined' ? eval("require('crypto')") : null;
+  typeof window === 'undefined'
+    ? (() => {
+        try {
+          // Call eval indirectly to guarantee no side-effects in module scope
+          // (optimization for minifiers)
+          return (0, eval)("r => r('crypto')")(require);
```
nderscore, line 39: I originally wanted to put this here:
```js
(0, eval)("require('crypto')");
```
But, apparently global scope doesn't have `require`. It only exists in module scope, so passing a reference is needed.

kitten: I think we'll want to test this with a multitude of bundlers and options today, namely Rollup and some other solutions on top.

While I'm pretty confident that Webpack won't mind this, I do think we'll want to play this one safely

kitten:
The choice of not providing separate bundles is explicitly based on the currently fractioned ecosystem around `package.json:exports`. While we could provide separate bundles, it's questionable whether all environments currently would handle them correctly, and this often introduces an additional requirement on users to know what could go wrong, when things do go wrong.

That out of the way, I'll do a couple of spot checks next week to see whether this is working as intended ✌️

exchanges/persisted-fetch/src/sha256.ts
kitten: Ok, haven't quite gotten to testing this yet, but I'll aready note down that this will require transpilation, which may break in non obvious ways, so a simpler written form may be: 

```suggestion
return (new Function('require', 'return require("crypto")'))(require);
```

nderscore: I just tested this version out:
```typescript
let nodeCrypto;
if (typeof window === 'undefined') {
  try {
    // Indirect eval/require to guarantee no side-effects in module scope
    // (optimization for minifiers)
    nodeCrypto = new Function('require', 'return require("crypto")')(require);
  } catch (e) {}
}
```

And confirmed that it accomplishes the same thing. The function constructor doesn't inherit local scope.

This brings our sizes down to:

* `urql-exchange-persisted-fetch.min.js` -> `2783b`
* `urql-exchange-persisted-fetch.min.mjs` -> `2809b`

Small, but still a gain 💪 (and it looks nicer than the evaled-IIFE inside an IIFE inside a ternary)

I haven't tested it with `rollup` (which was the source of the original issue raised a few months ago), but this clears out a warning I was seeing in `esbuild`

Updated here: b13b3633

kitten (approved):

kitten:
@nderscore Can't find any problems with this, so let's ship it 💪 Thanks for taking a look at this!

kitten:
shipped in `@urql/exchange-persisted-fetch@1.3.2`

PR #1753, opened 2021-06-30
update vue.md to include a issue about async/await

viztor:

### DIFF
--- a/docs/basics/vue.md
+++ b/docs/basics/vue.md
@@ -405,6 +405,9 @@ when `pause` is `true`, which would usually stop all automatic queries.
 
 ### Vue Suspense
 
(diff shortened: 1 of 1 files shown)


kitten:
I know this issue is taking us a bit, but `@urql/vue` is pre-stable and at `v0.4.1` so we're actually expecting some regressions here and there while we settle on an implementation. So I'm actually hoping we get to resolve this issue this week instead, and typically we don't add "current" notifications into the docs, i.e. all documentation should assume that things work and the issues should help otherwise. I do tend to make some exceptions in some cases, but I'd say this issue is generally easily identifiable as a bug in `@urql/vue`, so as long as the issue is bumped we're good for now

PR #1796, opened 2021-07-12
(auth) - Allow mutate from the getAuth to accept a TypedDocumentNode

younesmln:
First of all thanks a lot for this awesome GraphQL client 

## Summary

`useMutation` gives up the posiblity to infer Data/Variables if we pass a typed-document as you can see here [useMutation.ts#L35](https://github.com/FormidableLabs/urql/blob/main/packages/react-urql/src/hooks/useMutation.ts#L35), unlike mutate from auth-exchange which only accepts `DocumentNode` and not `TypedDocumentNode`

### auth-exchange:
- allow `mutate` exposed by the `getAuth` to accept `TypedDocumentNode` to make type inference possible

### DIFF
--- a/.changeset/weak-boxes-speak.md
+++ b/.changeset/weak-boxes-speak.md
@@ -2,4 +2,4 @@
 '@urql/exchange-auth': patch
 ---
 
-allow mutate to infer mutation Data and Variables by using TypedDocumentNode
+Allow `mutate` to infer the result's type when a `TypedDocumentNode` is passed via the usual generics, like `client.mutation` for instance.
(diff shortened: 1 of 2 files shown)


.changeset/weak-boxes-speak.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/exchange-auth': patch
+---
+
+allow mutate to infer mutation Data and Variables by using TypedDocumentNode
```
kitten, line 5:
```suggestion
Allow `mutate` to infer the result's type when a `TypedDocumentNode` is passed via the usual generics, like `client.mutation` for instance.
```

kitten (commented):
That's a nice catch! 🙌 We must've missed this one all the way back when we switched everything over. I only have a small change request for the changeset, to make it a little clearer

JoviDeCroock (approved):
Awesome catch, thank you!

PR #1800, opened 2021-07-15
Added a section about setting up custom headers

sunpietro:
Added a section about setting up custom headers in useQuery/useMutation hooks.

## Summary

I was looking for information on how to set up custom headers when using `useQuery` and `useMutation` hooks in React applications. I was not able to find it in the existing docs.

## Set of changes

- [x] updated the React usage docs.

### DIFF
--- a/docs/api/urql.md
+++ b/docs/api/urql.md
@@ -153,7 +153,10 @@ Sometimes, it's required to send custom headers with your query/mutation. In suc
 import React from 'react';
 
 const SomeComponent = (): JSX.Element => {
-  const [{ data, error, fetching }] = useQuery({ query: YourQuery, context: { fetchOptions: { headers: { 'x-custom-header': 'some-value' } } });
+  const context = React.useMemo(() => ({
+    fetchOptions: { headers: { 'x-custom-header': 'some-value' } }
+  }), [])
+  const [{ data, error, fetching }] = useQuery({ query: YourQuery, context });
   
   return <div>{fetching ? 'Loading ...' : 'Data loaded'}</div>;
 }
--- a/docs/api/urql.md
+++ b/docs/api/urql.md
@@ -149,7 +149,7 @@ above, and should thus be preferred over manually using `useContext` with `urql`
 
 Sometimes, it's required to send custom headers with your query/mutation. In such case you should send them via `context` property of `useQuery`/`useMutation` params object:
 
-```
+```tsx
 import React from 'react';
 
 const SomeComponent = (): JSX.Element => {
--- a/docs/api/urql.md
+++ b/docs/api/urql.md
@@ -144,3 +144,36 @@ const useClient = () => React.useContext(Context);
 
 However, this hook is also responsible for outputting the default client warning that's mentioned
 above, and should thus be preferred over manually using `useContext` with `urql`'s `Context`.
+
+### Setting up custom headers for a query/mutations
+
+Sometimes, it's required to send custom headers with your query/mutation. In such case you should send them via `context` property of `useQuery`/`useMutation` params object:
+
+```
+import React from 'react';
+
+const SomeComponent = (): JSX.Element => {
+  const [{ data, error, fetching }] = useQuery({ query: YourQuery, context: { fetchOptions: { headers: { 'x-custom-header': 'some-value' } } });
+  
+  return <div>{fetching ? 'Loading ...' : 'Data loaded'}</div>;
+}
+```
+
+However you have to make an additional addition to such, as passing the `context` object will cause an infinite loop of re-rendering. In such case, you should memoize the `context` value in order to avoid re-renderings.
+
+```
+import React, { useMemo } from 'react';
+
+const SomeComponent = (): JSX.Element => {
+  const context = useMemo(() => {
+    return {
+      fetchOptions: { headers: { 'x-custom-header': 'some-value' } }
+    };
+  }, []);
+  const [{ data, error, fetching }] = useQuery({ query: YourQuery, context} });
+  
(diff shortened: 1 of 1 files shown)


docs/api/urql.md
```diff
@@ -144,3 +144,36 @@ const useClient = () => React.useContext(Context);
...
+
+### Setting up custom headers for a query/mutations
+
+Sometimes, it's required to send custom headers with your query/mutation. In such case you should send them via `context` property of `useQuery`/`useMutation` params object:
+
+```
+import React from 'react';
+
+const SomeComponent = (): JSX.Element => {
+  const [{ data, error, fetching }] = useQuery({ query: YourQuery, context: { fetchOptions: { headers: { 'x-custom-header': 'some-value' } } });
```
JoviDeCroock, line 156: We need to memoize this as we can't predict when the user wants this to cause a refetch.

```suggestion
  const context = React.useMemo(() => ({
    fetchOptions: { headers: { 'x-custom-header': 'some-value' } }
  }), [])
  const [{ data, error, fetching }] = useQuery({ query: YourQuery, context });
```

docs/api/urql.md
```diff
@@ -144,3 +144,36 @@ const useClient = () => React.useContext(Context);
 
 However, this hook is also responsible for outputting the default client warning that's mentioned
 above, and should thus be preferred over manually using `useContext` with `urql`'s `Context`.
+
+### Setting up custom headers for a query/mutations
+
+Sometimes, it's required to send custom headers with your query/mutation. In such case you should send them via `context` property of `useQuery`/`useMutation` params object:
+
+```
```
JoviDeCroock, line 152:
```suggestion
```tsx
```

docs/api/urql.md
```diff
@@ -144,3 +144,36 @@ const useClient = () => React.useContext(Context);
 
 However, this hook is also responsible for outputting the default client warning that's mentioned
 above, and should thus be preferred over manually using `useContext` with `urql`'s `Context`.
+
+### Setting up custom headers for a query/mutations
+
+Sometimes, it's required to send custom headers with your query/mutation. In such case you should send them via `context` property of `useQuery`/`useMutation` params object:
```
JoviDeCroock, line 150:
```suggestion
Sometimes, it's required to send custom headers with your query. In such case you can leverage the `context` property of your `useQuery` call:
```

docs/api/urql.md
```diff
@@ -144,3 +144,36 @@ const useClient = () => React.useContext(Context);
...
+const SomeComponent = (): JSX.Element => {
+  const [{ data, error, fetching }] = useQuery({ query: YourQuery, context: { fetchOptions: { headers: { 'x-custom-header': 'some-value' } } });
+  
+  return <div>{fetching ? 'Loading ...' : 'Data loaded'}</div>;
+}
+```
+
+However you have to make an additional addition to such, as passing the `context` object will cause an infinite loop of re-rendering. In such case, you should memoize the `context` value in order to avoid re-renderings.
+
+```
```
JoviDeCroock, line 164:
```suggestion
```tsx
```

docs/api/urql.md
```diff
@@ -144,3 +144,36 @@ const useClient = () => React.useContext(Context);
...
+import React from 'react';
+
+const SomeComponent = (): JSX.Element => {
+  const [{ data, error, fetching }] = useQuery({ query: YourQuery, context: { fetchOptions: { headers: { 'x-custom-header': 'some-value' } } });
+  
+  return <div>{fetching ? 'Loading ...' : 'Data loaded'}</div>;
+}
+```
+
+However you have to make an additional addition to such, as passing the `context` object will cause an infinite loop of re-rendering. In such case, you should memoize the `context` value in order to avoid re-renderings.
```
JoviDeCroock, line 162: I think we can lose this section in exchange for the above as the above will always cause infinite loops without the `useMemo`

kitten:
On a conceptual structure of the hooks, I'm not sure whether it belongs there in the first place. Obviously, there isn't really a more suitable place, but the language and structure of what's being added here doesn't adhere to how we write API docs. 

The documentation has two general parts: Instructive and Descriptive parts. The instructive parts is generaly the "Advanced" and "Basics" sections and each will answer "How do I do X?" for a certain topic, the descriptive parts are the "Architecture" and "API" docs, where we answer "What is X?", in other words. So, the API docs are demonstrating available APIs without delivering usage information. If they did we either run into duplication or excessive amounts of information. Because the API docs demonstrate _any_ possible API option, function, methods, etc, it would get overloaded if we also added "How to do X?" into each API there.

In this case, I find the addition a bit random (sorry, for a lack of a better word 😅) because `context` is not bindings-specific, and `context.fetchOptions` is even less so. I also don't find this particularly instructive if added in a different place, especially since it takes a generic API and arbitrarily documents it for one of the bindings but not the others.

I think it may make more sense to add a _note_ to the API docs for `urql` that says that `context` must be memoised using e.g. `useMemo`. The `fetchOptions` are documented in various other places however: https://formidable.com/open-source/urql/docs/basics/react-preact/#setting-up-the-client (for each bindings)

We should extend each _bindings_ introduction's "Setting up the client" section to possibly link to [the `context`'s API docs](https://formidable.com/open-source/urql/docs/api/core/#operationcontext) and add a hint that the context can also be altered selectively, since that's currently only briefly touched on in each bindings: https://formidable.com/open-source/urql/docs/basics/react-preact/#reexecuting-queries

I think for that, we should find an additional example in which it'd make sense to alter `context` for queries. I don't think altering the `url` or `fetchOptions` is a best practice, so that should be avoided. We definitely do not see this as a "happy path", so maybe we could use `preferGetMethod` as an example or demonstrate how `requestPolicy` is often a shorthand to the `context` property 🤔

PR #1804, opened 2021-07-15
Implemented a global context for useMutation hook

sunpietro:
## Summary

When using `useQuery` it is possible to pass a context in `useQuery` call. While with `useMutation` the context can be passed at some point later, when invoking `executeMutation` function. With this PR there's a possibility to pass a kind of "global context" to `useMutation` hook that helps to avoid adding the contextual information with every call of `executeMutation` function.

## Set of changes

- [x] added a second param of `globalContext`, type `Partial<OperationContext>`,
- [x] updated tests

### DIFF
--- a/.changeset/wise-days-doubt.md
+++ b/.changeset/wise-days-doubt.md
@@ -1,0 +1,8 @@
+---
+'@urql/preact': minor
+'urql': minor
+'@urql/vue': minor
+'urql-docs': minor
+---
+
+Added a second param to useMutation hook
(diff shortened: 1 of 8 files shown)


kitten:
On a side note, this doesn't merge the context, so it's in more than one way distinct from other behaviour we already adopted. A similar proposal was rejected
 https://github.com/FormidableLabs/urql/pull/1139

Like in the last case however, I'll leave this, please don't open PRs for minor API changes especially, before submitting an RFC to discuss the change 💕😅 https://github.com/FormidableLabs/urql/blob/main/CONTRIBUTING.md#how-do-i-propose-changes

PR #1810, opened 2021-07-16
(vue) - unwrap the proxy variables before we send it to the urql-client

JoviDeCroock:
Relates to https://github.com/FormidableLabs/urql/discussions/1809

## Summary

When we use `variables` in vue we'll pass it in as a `ref` which in `vue` is a proxy or proxy-like object, this means if we don't unwrap it we'll send this through the exchanges-chain which in this case caused issues in IndexedDB.

I'm wondering if there's more places where we could mutate/.... a variable that goes into the exchange-pipeline.

## Set of changes

- unwrap `variables` proxy-ref before sending it through to the client

### DIFF
--- a/packages/vue-urql/src/utils.ts
+++ b/packages/vue-urql/src/utils.ts
@@ -1,0 +1,9 @@
+import { Ref } from 'vue';
+
+export function unwrapPossibleProxy<V>(
+  possibleProxy: V | Ref<V> | undefined
+): V {
+  return possibleProxy && typeof possibleProxy === 'object'
+    ? JSON.parse(JSON.stringify(possibleProxy))
+    : possibleProxy;
+}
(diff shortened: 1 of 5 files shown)


packages/vue-urql/src/utils.ts
```diff
@@ -0,0 +1,9 @@
+import { Ref } from 'vue';
+
+export function unwrapPossibleProxy<V>(
+  possibleProxy: V | Ref<V> | undefined
+): V {
+  return possibleProxy && typeof possibleProxy === 'object'
+    ? JSON.parse(JSON.stringify(possibleProxy))
+    : possibleProxy;
```
kitten, line 8: could this be `isRef(value) ? value.value : value`?

kitten (approved):

PR #1821, opened 2021-07-27
Update subscriptions.md

andyrichardson:
## Summary

Minor tweaks to the subscriptions docs.

## Set of changes

 - Change precedence of subscription methods
 - Add deprecation notice to `subscriptions-transport-ws`
 - Trim code examples down
 - Decrease verbosity of `graphql-ws` instructions (maybe too much information IMHO)

### DIFF
--- a/docs/advanced/subscriptions.md
+++ b/docs/advanced/subscriptions.md
@@ -40,41 +40,9 @@ When we define this function it must return an "Observable-like" object, which n
 [Observable spec](https://github.com/tc39/proposal-observable), which comes down to having an
 object with a `.subscribe()` method accepting an observer.
 
-## Setting up `subscriptions-transport-ws`
-
-If your GraphQL API is using [the Apollo Server](https://www.apollographql.com/docs/apollo-server/),
-you'll be able to use [Apollo's `subscriptions-transport-ws`
-package](https://github.com/apollographql/subscriptions-transport-ws).
-
-```js
-import { Client, defaultExchanges, subscriptionExchange } from 'urql';
-import { SubscriptionClient } from 'subscriptions-transport-ws';
-
-const subscriptionClient = new SubscriptionClient('ws://localhost/graphql', { reconnect: true });
-
-const client = new Client({
-  url: '/graphql',
-  exchanges: [
-    ...defaultExchanges,
-    subscriptionExchange({
-      forwardSubscription(operation) {
-        return subscriptionClient.request(operation);
-      },
-    }),
-  ],
-});
-```
-
-In this example, we're creating a `SubscriptionClient`, are passing in a URL and some parameters,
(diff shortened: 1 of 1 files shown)


JoviDeCroock (approved):
LGTM

docs/advanced/subscriptions.md
```diff
@@ -40,74 +40,64 @@ When we define this function it must return an "Observable-like" object, which n
...
-[Read more about `subscription-transport-ws` on its README.](https://github.com/apollographql/subscriptions-transport-ws/blob/master/README.md)
+## Setting up `subscriptions-transport-ws`
 
-## Setting up `graphql-ws`
+For backends supporting `subscriptions-transport-ws`, [Apollo's `subscriptions-transport-ws`
+package](https://github.com/apollographql/subscriptions-transport-ws) can be used.
 
-If your GraphQL API server is using [graphql-ws](https://github.com/enisdenjo/graphql-ws),
-you'll be able to use it here too!
+> Note: `subscriptions-transport-ws` is deprecated, consider using `graphql-ws` (above)
```
kitten, line 77:
```suggestion
> The `subscriptions-transport-ws` package isn't actively maintained. If your API supports the new protocol or you can swap the package out, consider using [`graphql-ws`](#setting-up-graphql-ws) instead.
```

It never has been put into official deprecation or archival, so I'd love to keep this generic, especially since the GraphQL over HTTP spec extension moving to standardise the protocol anyway

docs/advanced/subscriptions.md
```diff
@@ -40,74 +40,64 @@ When we define this function it must return an "Observable-like" object, which n
...
-      },
+      forwardSubscription: (operation) => subscriptionClient.request(operation)
     }),
   ],
 });
 ```
 
-We create a WebSocket client with the necessary options and use the `subscribe` method from it to
-create a Subscription Observable, which we return to the `subscriptionExchange` inside `forwardSubscription`.
+In this example, we're creating a `SubscriptionClient`, are passing in a URL and some parameters,
```
kitten, line 96: I think it's inconsistent to keep this and delete the other description. I'm actually not sure whether to keep both or delete both, but better to be consistent 🤷 I'll leave it up to you 😸

docs/advanced/subscriptions.md
andyrichardson: Can do that if you want - I was basing that deprecation status on what is in the [README](https://github.com/apollographql/subscriptions-transport-ws)

> The subscriptions-transport-ws library is not being actively maintained. It is recommended that you use the graphql-ws library instead.

docs/advanced/subscriptions.md
andyrichardson: Sorry yeah that was a whimsical change - I'll add back in the old one for now to keep this scoped :+1:

PR #1826, opened 2021-07-29
Update README.md

frontshift:
## Summary

Updated readme to reflect change in `NextUrlqClientConfig` 

## Set of changes
renamed `neverSuspend` flag to `suspense`

### DIFF
--- a/packages/next-urql/README.md
+++ b/packages/next-urql/README.md
@@ -252,7 +252,7 @@ component has children using `getInitialProps` but that component itself is not 
 `withUrqlClient`, this means `withUrqlClient(() => clientOptiosn, { ssr:true })`.
 This measure is available so we can support `getStaticProps`, ...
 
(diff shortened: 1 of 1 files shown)


kitten:
These aren't the `Client`'s options, but the `withUrqlClient` options, so this instruction is still completely correct. The HOC does have a `neverSuspend` option that can be set to `true`, although I'm actually unsure whether it can be made obsolete and `getClientConfig`'s config can be read instead 🤔 cc @JoviDeCroock

JoviDeCroock:
Hmm, I think we could actually do that 😅 the complexity when we originally added this feature is because we don't read `suspense: true` when using our `prepass` approach. The problem presents itself when we use `getServerSideProps` (no prepass) and a query is uncovered by `getServerSideProps` which was sometimes intended 😅 it couldn't throw.

https://github.com/FormidableLabs/urql/pull/923#issue-457698628

frontshift:
Ok thanks, I just realised that you can pass the HOC options as second argument.... all good :)

PR #1828, opened 2021-07-30
(auth) - expose AuthConfig type

arempe93:
## Summary

Exposes the `AuthConfig<T>` type in the auth exchange package. This would be useful for doing something like:

```ts
const baseAuthConfig: AuthConfig<MyAuthContext> = { ... }
```

Currently `AuthConfig` cannot properly be derived at all because it is generic.

```ts
type AuthConfig = Parameters<typeof authExchange>[0] // => AuthConfig<unknown>
```

## Set of changes

 - re-export `AuthConfig<T>` from `authExchange.ts`

### DIFF
--- a/.changeset/spicy-suits-tease.md
+++ b/.changeset/spicy-suits-tease.md
@@ -2,4 +2,4 @@
 '@urql/exchange-auth': patch
 ---
 
-Expose `AuthContext` type, by [@arempe93](https://github.com/arempe93) (See [#1828](https://github.com/FormidableLabs/urql/pull/1828))
+Expose `AuthContext` type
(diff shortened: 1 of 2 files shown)


JoviDeCroock (approved):
Good spot!

.changeset/spicy-suits-tease.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/exchange-auth': patch
+---
+
+Expose `AuthContext` type, by [@arempe93](https://github.com/arempe93) (See [#1828](https://github.com/FormidableLabs/urql/pull/1828))
```
JoviDeCroock, line 5:
```suggestion
Expose `AuthContext` type
```

The name gets automatically appended when changeset picks these up

Issue #1827, opened 2021-07-29
Error (10) writeOptimistic when accidentally passing a query to useMutation hook


jakobo:
Running a mutation via `useMutation` that contains a `query` operation throws Error (10) `writeOptimistic(...) was called with an operation that isn't a mutation`

<blockquote>This should never happen, please open an issue if it does. This occurs when writeOptimistic attempts to write an optimistic result for a query or subscription, instead of a mutation.</blockquote>

I should also say this is 100% user error, as I tried to mutate with a query operation. I just thought urql would catch it.

urql version & exchanges:
```
"@urql/core": "^2.1.4",
"@urql/devtools": "^2.0.3",
"@urql/exchange-auth": "^0.1.3",
"@urql/exchange-graphcache": "^4.1.4",
"urql": "^2.0.4",
```

**Steps to reproduce**
Query that causes issue:
```js
const UPDATE_CONTACT_DETAILS = gql`
  query UpdateContactDetails(
    $id: uuid!
    $firstName: String!
    $fullName: String!
  ) {
    update_contacts(
      where: { id: { _eq: $id } }
      _set: { first_name: $firstName, full_name: $fullName }
    ) {
      returning {
        id
        updated_at
        first_name
        full_name
      }
    }
  }
`;
```

Execution in hook:
```js
const [, updateInfo] = useMutation(UPDATE_CONTACT_DETAILS); // I actually expected it to throw right here
// ... later, possibly in a callback
updateInfo({...})
```

**Expected behavior**
Honestly, I expected urql to throw when I accidentally passed a query operation into the `useMutation` hook instead of waiting until execution the query came back before throwing Error (10).

**Actual behavior**
Error (10), in which urql did exactly what it was asked to b/c I fat-fingered "query" instead of "mutation" in my gql.

PR #1829, opened 2021-07-30
(core) - error for invalid operation passed to query/mutation/subscription

JoviDeCroock:
Fixes https://github.com/FormidableLabs/urql/issues/1827

## Summary

Errors out when an invalid operation-node is passed into our client-methods

## Set of changes

 - throw an error when passing in the wrong operation-type to the urql-client

### DIFF
--- a/.changeset/weak-doors-cheer.md
+++ b/.changeset/weak-doors-cheer.md
@@ -1,0 +1,5 @@
+---
+'@urql/core': minor
+---
+
+Warn for invalid operation passed to query/subscription/mutation
--- a/.changeset/weak-doors-cheer.md
+++ b/.changeset/weak-doors-cheer.md
@@ -1,5 +1,5 @@
 ---
-'@urql/core': minor
+'@urql/core': patch
 ---
 
 Warn for invalid operation passed to query/subscription/mutation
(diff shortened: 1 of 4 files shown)


packages/core/src/client.ts
```diff
@@ -289,6 +289,25 @@ export const Client: new (opts: ClientOptions) => Client = function Client(
...
 
     createRequestOperation(kind, request, opts) {
+      if (
+        kind !== 'teardown' &&
+        !request.query.definitions.some(
+          x => x.kind === Kind.OPERATION_DEFINITION && x.operation === kind
+        ) &&
+        process.env.NODE_ENV !== 'production'
+      ) {
+        for (let i = 0; i < request.query.definitions.length; i++) {
```
JoviDeCroock, line 299: `.find` wasn't allowed with the `ES5` lint-rule 😅

kitten: The `some` combined with the extra loop seems a bit... insufficient even if it's just for development. So we could probably add an operation type helper here that's similar to the name one: https://github.com/FormidableLabs/urql/blob/30447ec77a8c44e0585db5af4febf58e1c253f35/packages/core/src/utils/request.ts#L93

kitten (approved):

.changeset/weak-doors-cheer.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/core': minor
```
kitten, line 2: I'd say this is a patch actually, since nothing will have broken in expected usage

PR #1835, opened 2021-08-01
Make "async fetchOptionsExchange" example copy-able

44px:
## Summary

The current example perfectly suits some use-cases but can't be used "as is". If you copy it into TS project, you'll see a few errors. Fixing these errors requires some understanding of exchange internals and wonka API.  

I see two possibilities how to make it easier to get started with the `urql` here:
1) Make it able to copy and paste the code example
2) Add `fetchOptionsExhange` to the core (as @RodolfoSilva originally suggested [here](https://github.com/FormidableLabs/urql/issues/234#issuecomment-602305153))

I implemented first as a simple solution, but maybe second would be even a better option?

## Set of changes

Changed the current example a bit by adding the missing imports and types.

### DIFF
--- a/docs/common-questions.md
+++ b/docs/common-questions.md
@@ -11,25 +11,34 @@ If you need `async fetchOptions` you can add an exchange that looks like this:
 
 ```js
 import { makeOperation } from '@urql/core';
+import { Exchange, Operation } from 'urql';
+import { pipe, mergeMap, map, fromPromise } from 'wonka';
 
-export const fetchOptionsExchange = (fn: any): Exchange => ({ forward }) => ops$ => {
-  return pipe(
-    ops$,
-    mergeMap((operation: Operation) => {
-      const result = fn(operation.context.fetchOptions);
-      return pipe(
-        typeof result.then === 'function' ? fromPromise(result) : fromValue(result),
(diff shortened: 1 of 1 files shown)


docs/common-questions.md
```diff
@@ -11,25 +11,34 @@ If you need `async fetchOptions` you can add an exchange that looks like this:
...
-            fetchOptions,
-          });
-        })
-      );
-    }),
-    forward
-  );
-};
+export const fetchOptionsExchange =
+  (fn: (fetchOptions: RequestInit) => Promise<RequestInit>): Exchange =>
```
JoviDeCroock, line 18: to avoid issues we initially made this `(fetchOptions: RequestInit) => Promise<RequestInit> | RequestInit` and had a ternary that did this in the stream:

```
const result = fn(operation.context.fetchOptions);
return pipe(
  typeof result.then === 'function' ? fromPromise(result) : fromValue(result),
);
```

This mainly to avoid introducing a footgun 😅

44px: Yep, sounds good. Fixed! Thanks for the tip 👍

JoviDeCroock (approved):
Awesome thank you!!!!

PR #1837, opened 2021-08-03
expose ExecuteExchangeArgs

taneba:
## Summary

This PR exposes ExecuteExchangeArgs from execute-exchange which would be useful when you wrap executeExchange in your function and pass the configuration args accordingly (e.g. testing).

## Set of changes
* export ExecuteExchangeArgs

### DIFF
--- a/exchanges/execute/src/execute.ts
+++ b/exchanges/execute/src/execute.ts
@@ -24,7 +24,7 @@ import {
   getOperationName,
 } from '@urql/core';
 
-interface ExecuteExchangeArgs {
+export interface ExecuteExchangeArgs {
   schema: GraphQLSchema;
   rootValue?: any;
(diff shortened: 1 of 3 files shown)


exchanges/execute/src/execute.ts
```diff
@@ -24,7 +24,7 @@ import {
   getOperationName,
 } from '@urql/core';
 
-interface ExecuteExchangeArgs {
+export interface ExecuteExchangeArgs {
```
JoviDeCroock, line 27: This will also need exporting from `./src/index.ts`

taneba: oops I messed up... updated!

JoviDeCroock (approved):
Awesome, thank you for your contribution!!! 🎉

PR #1849, opened 2021-08-11
(auth) - Fix willAuthError causing duplicate operations

### LINKED ISSUE #1825: willAuthError query executes twice
When a willAuthError is present in auth-exchange it results to queries being executed twice.

I saw that there was a PR that fixed the same issue but even that does not seem to fix the error.

If willAuthError is removed and we rely only on didAuthError then everything is working as expected.

Sorry that i am not provided a reproducible case but i think the issue is present in also a very simple auth-exchange scenarios

yankovalera:
Resolve #1825

### DIFF
--- a/.changeset/orange-pets-wave.md
+++ b/.changeset/orange-pets-wave.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-auth': patch
+---
+
+Fixes an issue with willAuthError causing duplicated queries
--- a/.changeset/orange-pets-wave.md
+++ b/.changeset/orange-pets-wave.md
@@ -2,4 +2,4 @@
 '@urql/exchange-auth': patch
 ---
 
-Fixes an issue with willAuthError causing duplicated queries
+Fix willAuthError causing duplicate operations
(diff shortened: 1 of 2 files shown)


.changeset/orange-pets-wave.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/exchange-auth': patch
+---
+
+Fixes an issue with willAuthError causing duplicated queries
```
kitten, line 5: Fix `willAuthError` causing duplicate operations

exchanges/auth/src/authExchange.ts
```diff
@@ -103,28 +103,32 @@ export function authExchange<T>({
...
 
-      let pendingPromise: Promise<any> | void = Promise.resolve()
+      let authPromise: Promise<any> | void = Promise.resolve()
         .then(() => getAuth({ authState, mutate }))
         .then(updateAuthState);
 
-      const refreshAuth = (operation: Operation): Promise<any> => {
+      const refreshAuth = (
+        operation: Operation,
+        addAuthAttempt = true
```
JoviDeCroock, line 117: We might want to invert this to `skipAddAuthAttempt` this because the polyfill used for `argument defaulting` is extremely expensive and will cause a big bump in the bundle size of this exchange 😅

yankovalera: reviewed

JoviDeCroock (approved):

kitten (approved):

PR #1850, opened 2021-08-11
(core) - fix single-source behavior in integrations

### LINKED ISSUE #1841: When two identical queries are executed and one is paused then both are aborted and return no data
urql version & exchanges: urql 2.0.4, cache exchange 4.2.1

**Steps to reproduce**

 - Have two identical queries
 - Both not paused (so they start executing)
 - Then pause one of the two
 - The fetch request (if it was sent) is aborted, both queries return no data (even the query that was not paused).

https://codesandbox.io/s/nifty-feather-e3esz?file=/pages/index.js

The page shows the result from the two queries. I'd expect the first line to read `{"fetching":false,"stale":false}` (because the query is paused), but the second line to dump the data into the page, since the query in the second component is not paused.

**Expected behavior**

 - Fetch request should not be aborted, any other parallel and non-paused query should continue executing as usual.
 - The paused request may or may not return data (since the data will be fetched by the other request and stored in the cache).

**Actual behavior**

The second query never resolves, even though it is not paused.

JoviDeCroock:
## Summary

The [single-source behavior previously](https://github.com/FormidableLabs/urql/pull/1515) wasn't effective for implementations like React, where the issue presents itself when the state of an operation is first polled. This led to the operation being torn down erroneously.

Fix #1841

## Set of changes

- We now ensure that operations started at the same time still use a shared single-source.

### DIFF
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -230,9 +230,6 @@ export const Client: new (opts: ClientOptions) => Client = function Client(
       onPush(result => {
         replays.set(operation.key, result);
       }),
-      onStart(() => {
-        active.set(operation.key, source);
-      }),
       onEnd(() => {
         // Delete the active operation handle
         replays.delete(operation.key);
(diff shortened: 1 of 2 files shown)


packages/core/src/client.ts
```diff
@@ -230,9 +230,6 @@ export const Client: new (opts: ClientOptions) => Client = function Client(
       onPush(result => {
         replays.set(operation.key, result);
       }),
-      onStart(() => {
-        active.set(operation.key, source);
-      }),
```
kitten, line 235: Context on this: `onStart` only triggers _after_ the source has started, so if someone subscribes and gets a synchronous result, that result would arrive _before_ `onStart`'s callback is called. This is fixed by moving this to `make` whose inner function is called immediately.

kitten (approved):

PR #1854, opened 2021-08-11
(experimental) - Add full support for multipart/mixed defer and stream directive responses

kitten:
Add **experimental** support for `@defer` and `@stream` responses for GraphQL. This implements the ["GraphQL Defer and Stream Directives"](https://github.com/graphql/graphql-spec/blob/4fd39e0/rfcs/DeferStream.md) and ["Incremental Delivery over HTTP"](https://github.com/graphql/graphql-over-http/blob/290b0e2/rfcs/IncrementalDelivery.md) specifications. If a GraphQL API supports `multipart/mixed` responses for deferred and streamed delivery of GraphQL results, `@urql/core` (and all its derived fetch implementations) will attempt to stream results. This is _only supported_ on browsers [supporting streamed fetch responses](https://developer.mozilla.org/en-US/docs/Web/API/Response/body), which excludes IE11.

The implementation of streamed multipart responses is derived from [`meros` by `@maraisr`](https://github.com/maraisr/meros), and is subject to change if the RFCs end up changing.

For more information, read: https://graphql.org/blog/2020-12-08-improving-latency-with-defer-and-stream-directives

On fresh builds with a quick gzip check across all minified files (`$ cat packages/core/dist/*.min.mjs | gzip-size`) we see an increase from `5.32kB` minzipped to `6.02kB` minzipped, meaning a delta of approximately `700B`.
On Graphcache (`$ cat exchanges/graphcache/dist/*.min.mjs | gzip-size`) we see a change from `9.38kB` to `9.55kB`, a delta of `170B`.
Therefore we can estimate that the total impact of this feature is limited to `<1kB` minzipped.


packages/core/src/internal/fetchSource.ts
```diff
@@ -1,68 +1,165 @@
...
-      return makeResult(operation, result, response);
-    })
-    .catch((error: Error) => {
-      if (error.name !== 'AbortError') {
-        return makeErrorResult(
-          operation,
-          statusNotOk ? new Error(response.statusText) : error,
-          response
-        );
+      buffer += chunk;
```
maraisr, line 69: can try mention this code was largely copied from my work (https://github.com/maraisr/meros/blob/d914f7461e2f3d0e3c2064c91ac95c28ef1e79d4/src/lib/browser.ts#L34)

packages/core/src/internal/fetchSource.ts
```diff
@@ -1,68 +1,165 @@
...
+          }
+
+          // hit a tail boundary, break
+          if (
+            next.slice(0, 2) === '--' ||
+            (payload && payload.hasNext === false)
+          ) {
+            if (nextResult) onResult(nextResult);
+            return;
+          }
```
maraisr, line 98: https://github.com/maraisr/meros/blob/d914f7461e2f3d0e3c2064c91ac95c28ef1e79d4/src/lib/browser.ts#L68-L69

👀

packages/core/src/internal/fetchSource.ts
kitten: yea, of course, we're just internally discussing what an attribution comment should look like

maraisr: no need — this is OSS. You don't HAVE to do anything 😅 just a little upset my lib wasnt good enough to have just been used. Ill be watching this very closely to improve my things <3

kitten: Oh, either way, attribution fully necessary as you drive the RFC. It'll also make it easier for external contributors to spot divergences for us and compare them. I'd love to keep it aligned going forward, but this is the easiest approach, so far ✨

One part of the reason to diverge is to allow isomorphic use-cases. I've seen the `package.json` switch and generally have made some bad experiences with those. That's why the persisted fetch exchange doesn't use it for Node/Browser and instead uses an eval-require. 😅 I've realised that there are so many commonalities in both files that we can attempt to skip the differences.

The other part is IE11 support. We'd like to be able to "partially support" it at least. Meaning that we won't fiddle around with XMLHttpRequest (especially since that deprecates the configurable `fetch` and is hence even more confusing) but because we'd have to transpile the generator and `for await ... of` code, which wouldn't yield compact code.

Speaking of compact, we're just generally aiming to shave off bytes and keep things simple. This approach doesn't change our stack a lot and we have some control to golf some bytes and reduce the bundle size impact. While this point is admittedly not that important in general, we do try to keep things small, especially when adding new features

maraisr: Think the idea was — and still in flux. Node is kept in buffer as opposed to operating on strings. Re the async iterables — would it makes sense for me to expose a potentially `meros/legacy` or something to that effect, where its implemented as a callback? For me its more about make sure that my other 3.5M downloaded users benefit from patches and bug fixes that may result from findings discovered here. 

Also re the isomorphic nature, honestly can say this isnt an issue. unpkg correctly handles it, esbuild and webpack also. But for the unknown browserify or whatever bundler, i could forgo performance here and expose a `meros/isomorphic` sub-module that merges the 2 behaviors.

Ultimately feel we should still try and work together here, rather than go down this forked path. 

I'm open for amends/changes/bends/alignment to suffice urql's needs.

kitten: I mean, don't get me wrong, I think `meros/legacy` would be neat and there are compromises to be had, but even all things considered currently we keep the dependencies we take on at practically zero either way; the big exception was `@graphql-typed-document-node/core` since TypeScript needs this to mostly be a deduplicated, unique package. So if we hypothetically said that everything was as we needed it in `meros` (which for the most part, they are of course) we'd still be reluctant to take on an extra dependency.

That's not a direct attack at other maintainers, it's mostly for practical reasons. For the one part, it does allow us to have full control over our build output and that allows us to code golf and uniformly keep sizes small and the experience coherent. One big point we picked up from the ecosystem, especially in the early days of GraphQL, is that any given library picks up tens of dependencies to achieve one thing. And while that doesn't add up to any difference in the quality of what we're building, it's important to us to outwardly show what we stand for with stats.

Apart from that however, we believe in alignment through specifications, throughout the GraphQL ecosystem. We're big fans of the GraphQL over HTTP spec-extension in general, and weren't happy how we had to approach `@urql/exchange-persisted-fetch` and `@urql/exchange-multipart-fetch` with a vague, coincidental convention.
This then obviously leaves us with questions of what happens when a spec does emerge and we're faced with either maintaining an implementation or integrating an external library.

With `graphql-ws`, this choice was rather easy. Since the beginning, there wasn't a spec to align but there was a de-facto practical conventional package in `subscriptions-transport-ws`. We decided to align loosely on that, later seeing that other vendors like AppSync were following even different practices, which then worked out. Now that a spec is emerging this will likely not change, for the simple reason that most people have to opt into subscriptions either way, and this isn't a super common choice we see.

With this decision it's a little different. It's impossible for us to _not_ integrate defer/stream closely as it influences our core implementation. On the other hand, as you could see, we also didn't want to issue a breaking change that replaces our pluggable `fetch` option, so we're stuck with making some assumptions here. This includes that we may see some breakage from users that are using `unfetch` and are trying this out, which means we'd like to control in the future how we surface this (similar to the risk of IE11), and only as a last resort consider deprecating the `fetch` option.

It's also possible that we'd like to diverge a little from `meros` in general (outside of spec concerns) every now and then. By having the code ourselves, not only do we get benefits in our output and little to no overhead in integrating a library and duplicating some code, when something goes wrong (say, with React Native, or a new bundler) we can act fast and still contribute changes back.

So, it's not a matter of not contributing back or not working together, which we still will and want to do, it's just that we don't see collaboration as effective in shared code at all times, and compromising on a couple of things that we see as necessary. In fact, I believe we'd likely work together more if we're not stuck putting all our eggs in one basket, which leads to a situation where all roads lead to ~rome~ one library, but rather to shared issues and spec changes 😄

JoviDeCroock (approved):

kitten:
We've released initial prerelease versions:

| Package | Version |
| -------- | -------- |
| `@urql/core@defer` | `2.3.0-defer.0` |
| `@urql/exchange-graphcache@defer` | `2.3.0-defer.0` |
| `@urql/exchange-execute@defer` | `2.3.0-defer.0` |

This will automatically enable this PR's patch. It may be necessary to create a resolution in Yarn, e.g. `"resolutions": { "@urql/core": "2.3.0-defer.0" }`, for the core package to remain deduplicated and on the new version.

PR #1859, opened 2021-08-16
(graphcache) - Increase structural (deep) equality in normalized query results

### LINKED ISSUE #1860: RFC: Graphcache's query results to maintain structural equality
## Summary

We previously received numerous questions and requests regarding structural / deep-referential equality in Graphcache.

Currently, when a query is read from the normalised cache, Graphcache creates fresh objects and results each time. This means that even if we queried the cache twice while no data has changed, it would still create fresh objects all the way down. Graphcache is fast and can often read normalised data thousands of times each second. Even complex queries (many fragments and long lists) can often be read and written several hundreds of times per second. This typically puts us well into a small fraction of the typical "frame budget" of a web app (at 60FPS we'd just want to take up a fraction of 16.6ms!)

However, the same can't be said about rendering in apps. Often, rendering and updating the DOM and views may take exponentially longer for complex queries than querying this data from the cache. The typical wisdom is to avoid as much work as possible. This is why we often advised to write apps without any updates in mind, then check which part of an app is becoming too expensive to render, and adding strategic `React.memo` HOCs in React with `react-fast-compare` deep comparisons. This is **tedious** though and often still not optimal.

As, in the Web world, people are by now used to treating data as immutable and taking (deep) referential equality for granted, it's easier for us to implement it in Graphcache, rather than relying on users to hunt for excessive updates early on. While this isn't even recommended in the React community (the mantra being "let apps re-render, optimise later"), it's common for excessive updates to become confusing.

## Proposed Solution

Introduce a (weak) guarantee of structural / deep-equality. Given a previous result that Graphcache has generated from its normalised cache (or has just written to the cache when an API result comes in), use this previous result and reuse unchanged objects/data in the next result it generates.

This may be done by A) tracking which objects Graphcache has created, B) checking referential equality and only using the "fresh" objects when they've diverged from the previous one. Since we're already querying recursively this is "free".

**⚠️ Limitation:** This solution focuses on update performance in UI frameworks at minimal cost to Graphcache itself. This means that if an entity is queried with the same selection set in two _different_ queries, they won't be referentially equal!

kitten:
Resolve #1860

## Summary

This change enables Graphcache to compare new query results (from its normalized cache) with previous query results. While we previously had some remaining logic to reuse field values for mutations, we now extended the concept to queries as well. Graphcache will query each field and check whether it has changed, only creating new objects when a selection set has actually changed. This creates structural / deep referential equality in more parts of the query as it changes and potentially helps UI frameworks (e.g. React with `useMemo` and `React.memo`) to avoid more work.

In testing, I've noticed that currently (this isn't super optimized yet) this sacrifices about -10% of performance compared to the `main` branch. However, given that rendering may be much more expensive than querying our normalized cache we currently even deem this worth it. For typical queries we'd expect to see performance drop from an order of 20K operations per second to 18K, which means this won't be noticeable in typical use, but help development a lot.

I've checked the minzipped change of sizes (`$ cat exchanges/graphcache/dist/*.min.mjs | gzip-size`) and am seeing a change from `9.38kB` to `9.61kB`, a delta of `230B`.

**⚠️ Limitation:** We're still focusing on UI update performance here, and not strict structural equality. This means that two _different_ queries requesting the same entity with the same selection set will not be referentially equal.

## Set of changes

- Store references to previously queried results in `cacheExchange`
- Pass previous references to `query` operations
- Refactor `query` operations (e.g. `readSelection`) to reuse references as much as possible


kitten:
We've released initial prerelease versions:

| Package | Version |
| -------- | -------- |
| `@urql/exchange-graphcache@deepeql` | `4.3.0-deepeql.0` |

This will automatically enable this PR's patch.

JoviDeCroock (approved):

PR #1870, opened 2021-08-18
(core) - mark query.__key as non-enumerable

JoviDeCroock:
## Summary

When a plugin manipulates `operation.query` with for instance removing a directive, changing the selectionSet, ... we would restore the original query with subsequent invocations of `formatDocument`. By marking `query.__key` as non-enumerable we prevent this, this does assume that the consuming exchange uses `visit` or any other cloning traversal mechanism to manipulate the query.

## Set of changes

- mark `query.__key` as non-enumerable

### DIFF
--- a/.changeset/rude-pumpkins-yawn.md
+++ b/.changeset/rude-pumpkins-yawn.md
@@ -1,0 +1,5 @@
+---
+'@urql/core': patch
+---
+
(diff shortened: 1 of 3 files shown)


kitten (approved):
A small test would make me unbelievably happy ❤️ Also, 😂
> rude pumpkins yawn

yankovalera (approved):

Issue #964, opened 2020-09-07
RFC: Increase feature-set of the populateExchange for progressive Graphcache adoption


kitten:
## Summary

The `populateExchange` is an effective tool to progressively start using Graphcache more effectively. The current problem with adopting Graphcache is that the user has to at some point make the switch from the Document Cache. The interesting properties of the `@populate` directive and Graphcache combined mean however that the `populateExchange` can be used to progressively use Graphcache like the document cache.

In such a scenario the goal would be to allow the `populateExchange` to trigger refetches like the Document Cache does _without_ the user having to write updaters for Graphcache that invalidate entities.

```graphql
query TodoList {
  todos(first: 10) {
    id
    title
  }
}

mutation AddTodo {
  addTodo(title: "Test") {
    todo @populate
  }
}
```

Given this mutation to add a `Todo` it may be common to define a `cacheExchange.updates.Mutation.addTodo` function that adds the new `Todo` item to several lists. In this example the mutation returns a `Todo` and `@populate` adds the appropriate fields.

If the mutation also returns a `viewer` field then `@populate` may be used on that!

```graphql
mutation {
  addTodo(title: "Test") {
    viewer @populate
  }
}
```

This mutation may now update `Query` which `viewer: Query!` links to by adding the appropriate fields.
**Currently the `populateExchange` does not support field arguments, but if we add those then this use-case is covered!** The mutation effectively would be allowed to update any field related to `Query`.

In cases where _no `viewer` field exists on the schema_, we could allow a novel **use of the `@populate` directive on the mutation operation itself!**

```graphql
mutation @populate {
  addTodo(title: "Test")
}
```

We could allow the `populateExchange` to send an _entirely_ new query operation in this case which is generated from fields that the `populateExchange` knows are affected by `addTodo` (i.e. each path from `Query` to `Todo`)

Generally it'd be useful for `@populate` to only add **affected fields** when necessary, meaning when any types under `@populate` loop back to `Query`, only fields on `Query` and below should be included until the path reaches the types affected by the mutation. Once these types are reached, the usual `@populate` logic applies.

Once we allow this, all data may be updated automatically. We can then **let the user transition to more Graphcache-based usage** by letting them move the `@populate` directive to lower fields.

## Proposed Changes

- Allow `@populate` to add and track fields that have arguments, e.g. `timestamp(format: UTC)` or `todos(first: 10)`
[...]

PR #1872, opened 2021-08-18
(populate) - refactor and increase reliability of query-reading

yankovalera:
## Summary

(relates to #964) Increase feature-set of the populateExchange. Address "classic" mode. As discussed with @JoviDeCroock we might address other cases in other PRs.

"Classic" mode for populate exchange:

` mutation { addTodo @populate }`

We find out what type is used inside of the [[Mutation]] and look in our data-structure what fields we've used for this type.

### DIFF
--- a/exchanges/populate/src/populateExchange.ts
+++ b/exchanges/populate/src/populateExchange.ts
@@ -289,9 +289,15 @@ export const extractSelectionsFromQuery = (
       if (node.kind === Kind.FRAGMENT_DEFINITION) {
         extractedFragments.push(node);
       } else if (node.kind === Kind.FIELD && node.selectionSet) {
-        const type = unwrapType(
-          resolveFields(schema, visits)[node.name.value].type
-        );
+        const resolvedFields = resolveFields(schema, visits);
+
+        const fieldType = resolvedFields[node.name.value];
+
+        const type = fieldType && unwrapType(fieldType.type);
+
+        if (!type) {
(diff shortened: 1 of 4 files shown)


exchanges/populate/src/populateExchange.ts
```diff
@@ -84,6 +183,8 @@ export const populateExchange = ({
     }
 
     parsedOperations.add(key);
+    currentVariables = variables || {};
+    readFromQuery(query);
 
     const [extractedFragments, newFragments] = extractSelectionsFromQuery(
```
JoviDeCroock, line 189: we should remove this line as

> Fragments won't be tracked anymore. It's assumed that only missing fields from a given selection set are added and that it doesn't matter how they're added, since the API does not care.

exchanges/populate/src/populateExchange.ts
```diff
@@ -284,35 +416,58 @@ export const addFragmentsToQuery = (
         }
 
         const newSelections = possibleTypes.reduce((p, possibleType) => {
-          const typeFrags = activeTypeFragments[possibleType.name];
+          let typeFrags = activeTypeFragments[possibleType.name];
```
JoviDeCroock, line 419: so here we should look at the collectedFieldsByType rather than the fragments we've collected

kitten:
Superseded by #2897. `Co-Author` tag should potentially be added there

PR #1880, opened 2021-08-22
(core) - update toPromise to exclude `hasNext` results

JoviDeCroock:
## Summary

This change ensures that when we call `toPromise()` on a query we wont serve an incomplete result, the user will expect to receive a non-stale full-result when using `toPromise()`

## Set of changes

- Account for `hasNext` in `toPromise`

### DIFF
--- a/.changeset/thick-seals-wait.md
+++ b/.changeset/thick-seals-wait.md
@@ -1,0 +1,7 @@
+---
+"@urql/core": patch
+---
+
+Fix: update toPromise to exclude `hasNext` results. This change ensures that
+when we call toPromise() on a query we wont serve an incomplete result, the
+user will expect to receive a non-stale full-result when using toPromise()
--- a/packages/core/src/utils/streamUtils.ts
+++ b/packages/core/src/utils/streamUtils.ts
@@ -7,7 +7,7 @@ export function withPromise<T extends OperationResult>(
   (source$ as PromisifiedSource<T>).toPromise = () => {
     return pipe(
       source$,
-      filter(result => !result.stale),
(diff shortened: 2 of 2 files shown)


wgolledge (commented):
I need to read up a bit on the _multipart/mixed defer and stream directive responses_ work, but this makes sense to me.

Seeing as we already track `stale`, would it make sense to `stale=true` when `hasNext` instead?

kitten:
@wgolledge I was extremely unsure about that one. Mainly, because the fields that are deferred will initially be `undefined` rather than `null`, so clearly identifiable and distinct from stale results and partial results. We did also get feedback that at times it's hard to handle partial results. But with `useFragment` maybe beoming a thing, this may not be that severe 🤔

The other consideration here are `@live` queries for which `hasNext` may be `true` indefinitely. So both are really good points

wgolledge:
Cheers for the detailed explanation @kitten! Sounds like it's helpful to keep track of both. The other slight query I had is if there's ever a scenario when we wouldn't want to exclude the operation even if it `hasNext`, though I'm guessing not

JoviDeCroock:
> Cheers for the detailed explanation @kitten! Sounds like it's helpful to keep track of both. The other slight query I had is if there's ever a scenario when we wouldn't want to exclude the operation even if it `hasNext`, though I'm guessing not

not in the case of `toPromise`, the reason for this being that if a consumer of urql does `client.query().toPromise()` they expect a full-response rather than something being streamed in, else they'd have to use `subscribe()` to get to a full result

kitten (approved):

PR #1881, opened 2021-08-23
(retry) - Add retryWith option to update operations

kitten:


wgolledge (approved):
nice 👍

exchanges/retry/src/retryExchange.ts
```diff
@@ -117,7 +120,10 @@ export const retryExchange = ({
         if (!maxNumberAttemptsExceeded) {
           // Send failed responses to be retried by calling next on the retry$ subject
           // Exclude operations that have been retried more than the specified max
-          nextRetryOperation(res.operation);
+          nextRetryOperation(
+            retryWith ? retryWith(res.error, res.operation) : res.operation
```
JoviDeCroock, line 124: Do we expect this to not retry if nothing is returned? Replacing retryIf?

JoviDeCroock (commented):
Let's add a test and a docs entry 😀

exchanges/retry/src/retryExchange.ts
kitten: I did consider this and may replace this. Mainly, I noticed as well that we would want to defer `retryWhen` potentially to the time it's retried which I think is more important, since it may be used to alter request-time context values

JoviDeCroock (approved):

exchanges/retry/src/retryExchange.test.ts
```diff
@@ -227,3 +228,105 @@ it(`should still retry if retryIf undefined but there is a networkError`, () =>
   expect(response).toHaveBeenCalledTimes(mockOptions.maxNumberAttempts);
   expect(result).toHaveBeenCalledTimes(1);
 });
+
+it('should allow retryWhen to return falsy value and act as replacement of retryIf', () => {
```
wgolledge, line 232: Should this be `retryWith` instead of `retryWhen`

wgolledge (commented):
Looks good! Like Jovi mentioned should we also add an entry to the [Retrying Operations](https://github.com/FormidableLabs/urql/blob/main/docs/advanced/retry-operations.md#reacting-to-different-errors) and [Retry Exchange](https://github.com/FormidableLabs/urql/blob/main/docs/api/retry-exchange.md) docs?

exchanges/retry/src/retryExchange.test.ts
kitten: damn it 😆 well, it's just a test

PR #1885, opened 2021-08-24
(graphcache) - Fix previous result's null values spilling into next result

### LINKED ISSUE #1878: Graphcache not triggering rerender when a field goes from null to non-null
urql version & exchanges: 2.0.4 (@urql/exchange-graphcache: 4.3.0)

**Steps to reproduce**

Imagine something simple like this : 

```graphql
  type Meal {
    id: ID!
    name: String!
  }

  type Entry {
    id: ID!
    name: String!
    meal: Meal
  }

  type EntryPage {
    total: Int!
    items: [Entry!]!
  }

  type Query {
    entries(): EntryPage!
  }

  type Mutation {
    editEntry(
      id: ID!
      name: String!
      meal: ID
    ): Entry!
  }
```

**Expected behavior**

When `editEntry` returns on the frontend, the data of `entries` should be updated via graphcache magic.

**Actual behavior**

It does not trigger an update when editing an `Entry` which had a `meal` field equal to `null` and now has that same field equal to some `Meal`. Every other edit triggers an update (`Meal` to different `Meal`, `Meal` to `null`, change on field `name`, etc.).

kitten:
Fix #1878

## Summary

Fix previous results' `null` values spilling into the next result that Graphcache issues, which may prevent updates from being issued until the query is reexecuted. This was affecting any `null` links on data, and any queries that were issued before non-optimistic mutations.

## Set of changes

- Separate input from output data in `operations/query.ts`
- Fix typo in owned-data check for `resolveLink`
- Fix skipped layers being disregarded if last result was falsy

### DIFF
--- a/exchanges/graphcache/src/operations/query.ts
+++ b/exchanges/graphcache/src/operations/query.ts
@@ -66,7 +66,7 @@ export const query = (
   error?: CombinedError | undefined,
   key?: number
 ): QueryResult => {
-  initDataState('read', store.data, (data && key) || null);
+  initDataState('read', store.data, key);
   const result = read(store, request, data, error);
   clearDataState();
   return result;
--- a/exchanges/graphcache/src/operations/query.ts
+++ b/exchanges/graphcache/src/operations/query.ts
@@ -121,21 +121,23 @@ const readRoot = (
   ctx: Context,
   entityKey: string,
   select: SelectionSet,
-  data: Data
+  input: Data
 ): Data => {
-  const typename = ctx.store.rootNames[entityKey] ? entityKey : data.__typename;
+  const typename = ctx.store.rootNames[entityKey]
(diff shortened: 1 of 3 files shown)


exchanges/graphcache/src/operations/query.ts
```diff
@@ -152,14 +154,14 @@ const readRoot = (
     }
 
     // Check for any referential changes in the field's value
-    hasChanged = hasChanged || dataFieldValue !== output[fieldAlias];
+    hasChanged = hasChanged || dataFieldValue !== input[fieldAlias];
```
JoviDeCroock, line 157:
```suggestion
    hasChanged = hasChanged || dataFieldValue !== fieldValue
```

JoviDeCroock (approved):

Issue #1888, opened 2021-08-25
If staleWhileRevalidate is set to true, it crashs Next.js and the page keeps loading forever


wfelipe99:
I'm a beginner so maybe I won't express myself correctly, but I have provided steps to reproduce locally and a codesandbox.

It took me hours to identify it. I was thinking that my project was the problem, but even in a fresh Next.js project the result is the same. If `staleWhileRevalidate: true` is set in withUrqlClient() HOC, it will crash Next.js. No error occurs in terminal, browser or console, but the page keeps loading forever. If I kill the command `npm run dev`, it's not enough because the node's process keeps running and the port continues to be used. So, to start the app again, it's necessary to use another port or to kill the node process in terminal.

In the fresh install, I've used the exact same code from here https://formidable.com/open-source/urql/docs/advanced/server-side-rendering/#using-getstaticprops-or-getserversideprops
I'm using an online graphql API for testing: https://countries.trevorblades.com/
(I'll provide the code I'm trying to run)

In codesanbox, I observed either one of these:

1 - "Refused connection" with a broken icon;
2 - "Error 502 - Bad gateway";
3 - It compiles, but doesn't show the page.

urql version & exchanges:

```JSON
"dependencies": {
    "graphql": "^15.5.1",
    "next": "11.1.0",
    "next-urql": "^3.2.0",
    "react": "17.0.2",
    "react-dom": "17.0.2",
    "react-is": "^17.0.2",
    "urql": "^2.0.4"
  },
```

Exchanges: [dedupExchange, cacheExchange, ssrCache, fetchExchange]

**Steps to reproduce**

1. Create a fresh Next.js project using `npx create-next-app --typescript`
2. `npm install --save next-urql react-is urql graphql`
3. In index.tsx, put this code:
<details><summary>Code</summary>
<p>

```typescript
import { withUrqlClient, initUrqlClient } from "next-urql";
import {
  ssrExchange,
  dedupExchange,
  cacheExchange,
  fetchExchange,
  useQuery
} from "urql";

const COUNTRIES_QUERY = `
  query { countries { name } }
`;

function Countries() {
  const [res] = useQuery({ query: COUNTRIES_QUERY });
  return (
    <div>
      {res.data.countries.map((country, index) => (
        <>
            <div key={index}>{country.name}</div>
        </>
      ))}
    </div>
  );
}

export async function getStaticProps(ctx) {
  // Either I add this staleWhileRevalidate here or not
[...]

PR #1895, opened 2021-08-27
(next) - fix issue with next where staleWhileRevalidate would loop

JoviDeCroock:
## Summary

Relates to https://github.com/FormidableLabs/urql/issues/1888

When next resolves `getStaticProps` we rely on the `props.urqlState` to be hydrated during the `renderToString` this ensures that we actually use the data from this call on the server. The issue presents itself when we are in a `renderToString` pass and the exchange opts to mark the result as `stale` and perform a `reexecute`.

## Set of changes

- avoid data being `reexecuted` due to `staleWhileRevalidate` on the server

### DIFF
--- a/.changeset/six-walls-jump.md
+++ b/.changeset/six-walls-jump.md
@@ -1,0 +1,5 @@
+---
+'next-urql': patch
+---
+
+Fix issue where the `renderToString` pass would keep looping due to reexecuting operations on the server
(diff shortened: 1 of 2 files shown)


SteveMelons:
It seems like this issue was not resolved. I updated to v3.2.1. When setting staleWhileRevalidate to true, the page freezes on load and becomes unresponsive while using up 100% CPU.

PR #1919, opened 2021-09-02
Use uncompiled files in a Svelte project

frederikhors:
## Summary

In a Svelte project I can use the `*.svelte` and `*.ts` uncompiled files directly in my project and compile them with my own bundler.

## Set of changes

1. Add `"svelte": "src/index.ts"` line to `package.json`;
1. fix `import type` when needed (for compliance with default SvelteKit tsconfig);
1. **what I don't know is how to export `src` dir in the final bundle, can you please enlight me?**

### DIFF
--- a/packages/svelte-urql/package.json
+++ b/packages/svelte-urql/package.json
@@ -21,6 +21,7 @@
   ],
   "main": "dist/urql-svelte",
   "module": "dist/urql-svelte.mjs",
+  "svelte": "src/index.ts",
   "types": "dist/types/index.d.ts",
   "source": "src/index.ts",
   "exports": {
--- a/packages/svelte-urql/src/context.ts
+++ b/packages/svelte-urql/src/context.ts
@@ -1,5 +1,8 @@
+import type { ClientOptions } from '@urql/core';
+
 import { setContext, getContext } from 'svelte';
-import { Client, ClientOptions } from '@urql/core';
+import { Client } from '@urql/core';
+
 import { _contextKey } from './internal';
 
 export const getClient = (): Client => getContext(_contextKey);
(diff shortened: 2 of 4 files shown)


kitten:
I'm not quite sure why'd you ever want to do this though? Can you explain what your motivation here is? We have quite a long/sophisticated transpilation/compilation pipeline that also gives us a lot of control and guarantees over the code output, so we actually prefer to not expose any source files.

frederikhors:
I was trying some modifications and realized that I didn't have the source and that Svelte was already using the compiled one. In the Svelte world it is common to let Svelte compile (and optimize) various aspects of the code.

Why not at least for this package? In the end it is only relative to Svelte.

kitten:
So, in other words then though, there's no point and only drawbacks to this potentially, I suppose... 😅

frederikhors:
I think an advantage could be that (even in the future) the Svelte compiler could optimize and change that code as it sees fit (from a Suspense perspective and more that is coming to Svelte).

Why drawbacks?

kitten:
> could optimize and change that code as it sees fit (from a Suspense perspective and more that is coming to Svelte).

There's nothing to optimise though, there are no `.svelte` files in `@urql/svelte`. It's a simple API to access stores and interface with the `Client` and the Svelte compiler isn't magic and doesn't even operate on these files 😅  In other words, what Svelte is compiling to (ignoring component lifecycle transpilations) is already present in the stores we've implemented, and it's built to interface with Svelte components. If we'd hypothetically resolve source files in this library then all we'd do is skip our pipeline and go through a simple transpilation/minification pipeline that's separate from ours, but not better.

frederikhors:
You're right.

PR #1949, opened 2021-09-12
(graphcache) offline storage for React Native

kadikraman:
## Summary

The default storage included in Graphcache isn't compatible with React Native. This adds a React Native-compatible version of the default storage that relies on [AsyncStorage](https://react-native-async-storage.github.io/async-storage/) for persisting the data, and [NetInfo](https://github.com/react-native-netinfo/react-native-netinfo) to determine when the app is offline.

## Set of changes

- new package `@urql/storage-rn`
- readme instructions
- note about the RN storage in the docs site


packages/graphcache-rn-async-storage/src/makeAsyncStorage.ts
```diff
@@ -0,0 +1,106 @@
...
+        if (persistedMetadata) {
+          return JSON.parse(persistedMetadata);
+        }
+      } catch (_err) {}
+
+      return [];
+    },
+
+    onOnline: cb => {
+      NetInfo.addEventListener(({ isConnected }) => {
```
JoviDeCroock, line 99: We'll probably want to add a variable that catches a way to unsubscribe here so if this function gets called twice we can unsubscribe the previous and use the new callback. This is a memory leak we recently fixed in the web one

packages/graphcache-rn-async-storage/src/makeAsyncStorage.ts
```diff
@@ -0,0 +1,106 @@
...
+     * The rest of the batches should be merged into a cache in chrono order.
+     * Also hydrate todayBatch if we find it.
+     */
+    readData: async () => {
+      const cache = {};
+
+      try {
+        const persistedDayStamps = (await AsyncStorage.getAllKeys())
+          .reduce<number[]>((filtered, curr) => {
+            const ts = curr.match(prefixCheck)?.[1];
```
JoviDeCroock, line 35: I don't think we currently support `?`

kadikraman: Yeah, looks like tsc is happy with it, the build command not so much. Removed ✅

packages/graphcache-rn-async-storage/src/makeAsyncStorage.ts
kadikraman: Oops! Good spot. Updated ✅

JoviDeCroock (approved):

Mookiies:
Awesome that this is getting added, looking forward to swapping out my implementation 😄 

Any chance that this can be updated to also include a `clear` method like the `defaultStorage` for web?

packages/graphcache-rn-async-storage/README.md
```diff
@@ -0,0 +1,47 @@
+# @urql/graphcache-rn-async-storage
```
kitten, line 1: The only thing I'm a little hesitant about is the name, because it may end up being a little arbitrary 🤔 `@urql/core` or `@urql/*` to `@urql/exchange-graphcache` to `@urql/graphcache-*-storage` seems a bit of a leap

packages/graphcache-rn-async-storage/src/makeAsyncStorage.ts
```diff
@@ -0,0 +1,114 @@
...
+
+            if (ts) {
+              filtered.push(Number(ts));
+            }
+            return filtered;
+          }, [])
+          .sort();
+
+        // eslint-disable-next-line es5/no-for-of
+        for (const dayStamp of persistedDayStamps) {
```
kitten, line 48: `for-of` may or may not be transpiled here, but I'd just axe it. Even if it isn't transpiling and works correctly, it may still be a bit of an inconsistency compared to the rest of the codebase across `urql`

packages/graphcache-rn-async-storage/src/makeAsyncStorage.ts
```diff
@@ -0,0 +1,114 @@
...
+          // Discard
+          if (todayDayStamp - dayStamp > maxAge) {
+            await AsyncStorage.removeItem(`${dataKey}_${dayStamp}`);
+          }
+          // Parse batch and merge in.
+          else {
```
kitten, line 54: nit: hard to read, since the comment is superficial (like "Discard") and this one separates the `if` from `else`

packages/graphcache-rn-async-storage/src/makeAsyncStorage.ts
```diff
@@ -0,0 +1,114 @@
...
+              if (data) {
+                const parsedData = JSON.parse(data);
+                Object.assign(cache, parsedData);
+
+                // We found today's batch, let's hydrate that while we're at it.
+                if (dayStamp === todayDayStamp) {
+                  Object.assign(todayBatch, parsedData);
+                }
+              }
+            } catch (_err) {}
```
kitten, line 66: The await & try-catch combos read a bit unfortunate here and I'm not sure whether they're transpiled nicely. I'd opt for avoiding it or extracting it into a separate function. But overall, unwrapping into regular chained promises ends up much nicely in terms of transpiled code and readability in try-catch cases.

packages/graphcache-rn-async-storage/src/makeAsyncStorage.ts
```diff
@@ -0,0 +1,114 @@
...
+
+      return cache;
+    },
+
+    writeData: async delta => {
+      try {
+        Object.assign(todayBatch, delta);
+        await AsyncStorage.setItem(
+          `${dataKey}_${todayDayStamp}`,
+          JSON.stringify(todayBatch)
```
kitten, line 79: I think all these `JSON.stringify` calls hide a very inconvenient bug 😅 The `SerializedEntries` contain `undefined` values which must be serialised or treated in a special manner: https://github.com/FormidableLabs/urql/blob/63eec7eb8b6efeb49c0bb27249dec58d12121d2a/exchanges/graphcache/src/types.ts#L208-L210

In the original implementation, we didn't actually "merge" them to overwrite old values, as we're dealing with batches that are deserialized in order. But the important point is that a key may be set to `undefined` to denote that it has been deleted or garbage collected in Graphcache. It's possible that failing this may lead to some issues. But I'm not entirely sure. https://github.com/FormidableLabs/urql/blob/63eec7eb8b6efeb49c0bb27249dec58d12121d2a/exchanges/graphcache/src/store/data.ts#L623

This is how we're solving it there, I believe: https://github.com/FormidableLabs/urql/blob/63eec7eb8b6efeb49c0bb27249dec58d12121d2a/exchanges/graphcache/src/default-storage/index.ts#L66-L67

packages/graphcache-rn-async-storage/README.md
kadikraman: How about `@urql/rn-storage` or `@urql/storage-rn` ?

packages/graphcache-rn-async-storage/src/makeAsyncStorage.ts
kadikraman: Actually I think this raises an underlying question of "why are we doing so many async calls anyway". I've updated the logic to store all the data against a single key instead.

packages/graphcache-rn-async-storage/src/makeAsyncStorage.ts
kadikraman: I've changed this logic completely so that else is now gone

packages/graphcache-rn-async-storage/src/makeAsyncStorage.ts
kadikraman: I pulled out the async storage functions into utils, but the but had to [revert the change](https://github.com/FormidableLabs/urql/pull/1949/commits/0496e8f3c0263c74873b8d72deb535fd61f468a1) because the it caused the build command fail with:
```
Cannot read property 'constantViolations' of undefined
```
Looks like we can't define any async functions, e.g. adding
```js
async function anything() {
  return null;
}
```
will fail with the same error. I could write them using promises instead?

packages/graphcache-rn-async-storage/src/makeAsyncStorage.ts
kadikraman: This should be addressed with this commit: https://github.com/FormidableLabs/urql/pull/1949/commits/0f4696c262321795ead32fa4a86e2cb4677fde12

packages/graphcache-rn-async-storage/README.md
kitten: Yea! I don't really mind any name 😆 I just think we may be able to get away with something more catchy, so `storage-rn` makes complete sense.

kadikraman: Ace! I've changed it to `@urql/storage-rn` - that seems to follow the naming scheme a lot better.

kitten (approved):

PR #1982, opened 2021-09-22
(react) - fix issue with paused queries not executing correctly

### LINKED ISSUE #1978: `executeQuery` does not work on the first attempt when `pause: true`
I have the relevant code set up more or less as follows
```
function MyComponent() {
  const [email, setEmail] = useState('');
  const validationToken = useSelector(state => state.tempValidationToken);

  const [res, executeQuery] = useQuery({ query: MyQuery, variables: {email, validationToken}, pause: true })

  // To debug the behaviour of the useQuery hook
  useEffect(() => {
    console.log('Query res', res);
  }, [res]);

  return (
    ...
    <TextInput value={email} onChangeText={setEmail} />
    ...
    ...
    <Button title="..." onPress={executeQuery} />
  )
}
```
The issue is that when the `<Button />` is pressed after typing the email in the `<TextInput />` it should call executeQuery thereby causing the `useEffect` hook to get triggered (because `res.fetching` should now be `true`) and log the current value of `res`. But that does not take place. However pressing the `<Button />` any subsequent number of times after that does in fact correctly call `executeQuery`.

Another observation - If I just hard code the values of `variables` in `useQuery` (in my case something like `{email: 'test@example.com', validationToken: 'test-validation-token'}`) the `executeQuery` works perfectly fine in the first and subsequent attempts.

urql version & exchanges:
```
urql: "2.0.4"
```

**Steps to reproduce**
I have been trying urql on react-native but I have also created a codesandbox in react and was able to reproduce the same issue. Here is the link to the sandbox - https://codesandbox.io/s/bold-feynman-nq83d?file=/src/App.js

Here is a link to a previous github issue that might be related to this - https://github.com/FormidableLabs/urql/issues/432

**Expected behavior**
`executeQuery` should run perfectly in the first attempt itself

**Actual behavior**
`executeQuery` does not run on the first attempt but works for subsequent attempts

JoviDeCroock:
## Summary

Fix issue where a paused query would not behave correctly when calling `executeQuery`, this scenario occured when the query has variables, there would be cases where on the first call it would think that the dependencies had changed (previous request vs current request) which made the source reset to null.

Fixes #1978 

## Set of changes

- update the state to reflect latest deps when calling `executeQuery`

### DIFF
--- a/.changeset/smart-bags-confess.md
+++ b/.changeset/smart-bags-confess.md
@@ -1,0 +1,5 @@
+---
+'urql': patch
+---
+
+Fix issue where a paused query would not behave correctly when calling `executeQuery`, this scenario occured when the query has variables, there would be cases where on the first call it would think that the dependencies had changed (previous request vs current request) which made the source reset to null
--- a/packages/react-urql/src/hooks/useQuery.ts
+++ b/packages/react-urql/src/hooks/useQuery.ts
@@ -207,7 +207,7 @@ export function useQuery<Data = any, Variables = object>(
               })
             )
           : client.executeQuery(request, context);
-        return [source, state[1], state[2]];
+        return [source, state[1], deps];
       });
     },
     [


kitten (approved):

Miraj98:
Urql really needs to have a separate `useLazyQuery` hook. I know `pause: true` achieves that but it's not good api design and is frustrating to use. The perfect solution is having a separate hook for this which returns `executeLazyQuery`. We can have this function accept the query variables directly instead of passing them through the hook. This will also be more efficient because `useQuery` runs a few checks every time the inputs to the hook change. Hence a separate lazy hook will have a simpler implementation and will be more performant.

I know there have been previous discussions around adding `useLazyQuery` but there seems to be a lack of intent to add that api. Mostly the reason given is `pause: true` works just fine. But honestly it's not a great way to implement it given a better approach is possible.

rodperottoni:
I updated URQL to 2.0.5 in a project I'm currently working and this issue is still occurring. I'm using `pause` and `variables` as stated in the description. Calls to `executeQuery` are not doing anything :/

PR #2061, opened 2021-11-05
(execute) - Subscription support

Tigge:
* add support for subscriptions in execute exchange

## Summary

The execute exchange is excellent for testing and various other uses cases. This PR adds support for subscriptions. In `graphql-js` v16, this will be even easier to implement since `execute` will support subscriptions out of the box.

These changes are non-breaking, but let me know if this requires an RFC.

Example of a subscription that generates numbers. This can be added to e.g. `rootValues`:

```js
const subscription = async function* numbers(...args) {
  let number = 0
  while (true) {
    await sleep(1000)
    yield { number: number++ }
  }
}
```

## Set of changes

- Added subscription support
  - Added new `subscribeFieldResolver` argument to the exchange
  - Changed `execute` argument to object
  - When an operation has ended, call `done` on the generator.
- Updated tests

### DIFF
--- a/exchanges/execute/src/execute.test.ts
+++ b/exchanges/execute/src/execute.test.ts
@@ -6,12 +6,13 @@ jest.mock('graphql', () => {
     ...graphql,
     print: jest.fn(() => '{ placeholder }'),
     execute: jest.fn(() => ({ key: 'value' })),
+    subscribe: jest.fn(),
   };
 });
 
 import { fetchExchange } from 'urql';
 import { executeExchange } from './execute';
-import { execute, print } from 'graphql';
+import { execute, print, subscribe } from 'graphql';
 import {
   pipe,
   fromValue,
@@ -21,7 +22,7 @@ import {
   empty,
   Source,
 } from 'wonka';
-import { queryOperation } from '@urql/core/test-utils';
+import { queryOperation, subscriptionOperation } from '@urql/core/test-utils';
 import {
   makeErrorResult,
   makeOperation,
@@ -38,7 +39,10 @@ const exchangeArgs = {
   client: {},
 } as any;
 
-const expectedOperationName = getOperationName(queryOperation.query);
+const expectedQueryOperationName = getOperationName(queryOperation.query);
+const expectedSubscribeOperationName = getOperationName(
+  subscriptionOperation.query
+);
 
 const fetchMock = (global as any).fetch as jest.Mock;
 const mockHttpResponseData = { key: 'value' };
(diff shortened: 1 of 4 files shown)


exchanges/execute/src/execute.test.ts
```diff
@@ -65,16 +65,17 @@ describe('on operation', () => {
     );
```
JoviDeCroock, line 66: Let's add a test for the functionality

Tigge: I've added a test to make sure `subscribe` is being called. Further tests are blocked by the eslint `es5/no-generators` rule, since I would probably need some sort of AsyncIterator to test this further:

```patch
diff --git a/exchanges/execute/src/execute.test.ts b/exchanges/execute/src/execute.test.ts
index f73d8737..092d2061 100644
--- a/exchanges/execute/src/execute.test.ts
+++ b/exchanges/execute/src/execute.test.ts
@@ -51,6 +51,11 @@ beforeEach(() => {
   jest.clearAllMocks();
   mocked(print).mockImplementation(a => a as any);
   mocked(execute).mockResolvedValue({ data: mockHttpResponseData });
+  mocked(subscribe).mockImplementation(async function* b() {
+    yield { data: { key: 'value1' } };
+    yield { data: { key: 'value2' } };
+    yield { data: { key: 'value3' } };
+  });
 });
 
 afterEach(() => {
@@ -133,6 +138,19 @@ describe('on operation', () => {
     });
   });
 
+  it('should return data from subscribe', async () => {
+    const context = 'USER_ID=123';
+
+    const responseFromExecuteExchange = await pipe(
+      fromValue(subscriptionOperation),
+      executeExchange({ schema, context })(exchangeArgs),
+      take(3),
+      toPromise
+    );
+
+    expect(responseFromExecuteExchange.data).toEqual({ key: 'value3' });
+  });
+
   it('should return the same data as the fetch exchange', async () => {
     const context = 'USER_ID=123';
 
diff --git a/exchanges/execute/src/execute.ts b/exchanges/execute/src/execute.ts
index 688fdf5e..29efc6ef 100644
--- a/exchanges/execute/src/execute.ts
```

kitten: @Tigge You could build an async iterator manually, but that's obviously a bit of a waste of time 😄 So, feel free to disable the rule for tests by adding an exception here: https://github.com/FormidableLabs/urql/blob/c2b9eb5afc9088bda6668f80fff74677d4dbf95b/scripts/eslint/common.js#L64

JoviDeCroock (approved):
This is awesome! I love the simplicity of the implementation, nice work 🎉

exchanges/execute/src/execute.test.ts
Tigge: Thanks, more tests are now added. Or at least one more.

PR #2076, opened 2021-11-11
(graphcache) - expose a way to interact with the client outside of operations

JoviDeCroock:
## Summary

This is very much a draft and experiment whether this is a good idea.

We've repeatedly seen the need for people to update data coming from external sources (socket.io,...) outside of GraphQL so with this PR we want to enable this once and for all. Our stance towards this has been hesitant because of the unintended side-effects that might be created by doing this so this should be considered an escape hatch.

An example would be:

```js
import { useClient, gql } from 'urql';

const TodoFragment = gql`
    fragment todoFields on Todo {
      id
      text
      complete
      __typename
    }
`;

export const Component = () => {
  const client = useClient();
  useEffect(() => {
    myTodoSocket.subscribe((updatedTodo) => {
      cache.updateFragment(TodoFragment, { ...updatedTodo, __typename: 'Todo' }, { id: updatedTodo.id })
    })
  }, [])
}
```

TODO:

- [ ] test whether operations are correctly invoked after updating a query
- [ ] find a way to correctly include `core.d.ts` so client is extended with the cache helper

## Set of changes

- expose `cache` on the client interface
- add helpers so the cache can track deps and data state while updating data

### DIFF
--- a/exchanges/graphcache/src/store/data.ts
+++ b/exchanges/graphcache/src/store/data.ts
@@ -192,17 +192,7 @@ export const noopDataState = (
   clearDataState();
 };
 
-export const getCurrentOperation = (): OperationType => {
-  invariant(
-    currentOperation !== null,
-    'Invalid Cache call: The cache may only be accessed or mutated during' +
-      'operations like write or query, or as part of its resolvers, updaters, ' +
-      'or optimistic configs.',
-    2
-  );
-
-  return currentOperation;
-};
+export const getCurrentOperation = (): OperationType | null => currentOperation;
 
 /** As we're writing, we keep around all the records and links we've read or have written to */
 export const getCurrentDependencies = (): Dependencies => {
@@ -414,8 +404,8 @@ export const gc = () => {
 };
 
 const updateDependencies = (entityKey: string, fieldKey?: string) => {
-  if (fieldKey !== '__typename') {
-    if (entityKey !== currentData!.queryRootKey) {
+  if (fieldKey !== '__typename' && currentData) {
+    if (entityKey !== currentData.queryRootKey) {
       currentDependencies![entityKey] = true;
     } else if (fieldKey !== undefined) {
       currentDependencies![joinKeys(entityKey, fieldKey)] = true;
--- a/exchanges/graphcache/src/store/store.ts
+++ b/exchanges/graphcache/src/store/store.ts
@@ -1,5 +1,12 @@
 import { DocumentNode } from 'graphql';
-import { TypedDocumentNode, formatDocument, createRequest } from '@urql/core';
+import {
+  TypedDocumentNode,
+  formatDocument,
+  createRequest,
+  Client,
+  RequestPolicy,
+  Operation,
+} from '@urql/core';
 
 import {
   Cache,
@@ -16,8 +23,10 @@ import {
   KeyingConfig,
   Entity,
   CacheExchangeOpts,
+  OperationType,
 } from '../types';
 
+import { toRequestPolicy } from '../helpers/operation';
 import { invariant } from '../helpers/help';
 import { contextRef, ensureLink } from '../operations/shared';
 import { read, readFragment } from '../operations/query';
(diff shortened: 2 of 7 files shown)


exchanges/graphcache/src/store/data.ts
```diff
@@ -192,17 +192,7 @@ export const noopDataState = (
...
-    currentOperation !== null,
-    'Invalid Cache call: The cache may only be accessed or mutated during' +
-      'operations like write or query, or as part of its resolvers, updaters, ' +
-      'or optimistic configs.',
-    2
-  );
-
-  return currentOperation;
-};
+export const getCurrentOperation = (): OperationType | null => currentOperation;
```
kitten, line 195: Nice catch!

exchanges/graphcache/src/store/data.ts
```diff
@@ -414,8 +404,8 @@ export const gc = () => {
 };
 
 const updateDependencies = (entityKey: string, fieldKey?: string) => {
-  if (fieldKey !== '__typename') {
-    if (entityKey !== currentData!.queryRootKey) {
+  if (fieldKey !== '__typename' && currentData) {
```
kitten, line 407: I think there may be some cases in which this could happen, right? I'm thinking this may be so that `cacheExchange().store.resolve` still works, right? 🤔 Is there any way in which we could ensure that an operation is still available for these cases? (just very loose thoughts)

exchanges/graphcache/src/store/store.ts
```diff
@@ -125,13 +148,63 @@ export class Store<
...
+              this.client!.reexecuteOperation(toRequestPolicy(op, policy));
+            }
+          });
+        }
+        clearDataState();
+      };
+    }
+
+    return null;
+  }
```
kitten, line 191: This is an interesting one! So, we basically now have a system whereby `store` holds a lot more power over the reexecution behaviour. My gut feeling is that it should be possible for us to get rid of `executePendingOperations` and `collectPendingOperations` entirely somehow, but I think I'll need a closer look at the whole files to see whether that's indeed the case 🤔

kitten (approved):

exchanges/graphcache/src/store/data.ts
JoviDeCroock: Hmm, I think this is needed so we don't crash if someone manages to call this without an initialised data state 😅 all though that should not be possible anymore. That would mean we start marking `currentData` as always present which feels a hard one from a typing perspective as it relies on us calling `initDataState`

exchanges/graphcache/src/store/store.ts
JoviDeCroock: I honestly would prefer that as well, this logic is quite shared currently with the main exchange code.

mwangaben:
This is useful

nil1511:
any updates on this? 🙊

paulrostorp:
@JoviDeCroock could this be merged ?

kitten:
@paulrostorp Last where we left it I had some concerns with how far-reaching this change is. Sorry for not posting it here! Generally, if we merged this as-is, it's a huge departure from imposing one limitation which used to be fundamental: Cache Updates are isolated to the Normalised Cache itself.

This may not seem that big, but prevents a lot of the nuances of _succeeding_ at making normalised caches reliable in GraphQL. We've basically found that if all effects are described by mutations and subscriptions (refetches aside) then it's hard to argue that updates should be possible from the outside.
Generally, what I've seen is that this leads to small exceptions, i.e. small updates being added for UI interactions that don't quite go to a mutation, or aren't quite represented by the GraphQL schema, which can get out of hand rather quickly.

That doesn't mean we won't introduce something like this. We actually had an idea to provide an API whereby "passive" mock subscriptions could be triggered instead. In other words, an API that lets you tell the cache "pretend as if this mutation/subscription came in, and update using it". In all likelihood this is going to fix the original issue here, which was around external event sources, — websockets outside of GraphQL for instance.

That said though, I believe the issue we had with those was that it isn't obvious what the API would look like, and the code to trigger a faux mutation/subscription response _and_ write an updater in Graphcache would be rather excessive.

Another alternative could be for us to somehow encourage having the update logic still be in Graphcache, e.g. by having a "triggers" object in there that can be called from the outside. This was an attractive alternative, since it's like "manual triggered updaters." Yet another alternative was to simply expose less of the `cache` methods to the outside.

However, this still requires us to consider some potential impacts on our "Deterministic Cache Updates" guarantees: https://formidable.com/open-source/urql/docs/graphcache/normalized-caching/#deterministic-cache-updates
Basically, if a synchronous update can be triggered at any time, then it may impact optimistic and non-optimistic cache states equally. In other words, it may not be safe at all times to make a change

So there basically are a couple of hurdles here to making this happen 😅 And they're similar to what we've got for subscriptions in general. However, I think we also got side-tracked by how "unnatural" this feels with our current "dependencies" system, which is the system that determines which operations have updated in the cache.

Once these three questions are solved (1. API design, 2. Deterministic Guarantees, 3. Dependencies System) this is basically a go for us, imho.

JoviDeCroock:
I will close this for the time being 😅

paulrostorp:
@JoviDeCroock Thank you for the detailed answer. I appreciate you taking the responsible approach 🙂 

Unfortunately for me, it means I cannot use this great library 😞

JoviDeCroock:
@paulrostorp mind explaining your use case, it might help us get insight

paulrostorp:
@JoviDeCroock I think has been mentioned before: to update cache from push events outside of subscriptions (in my case WS and https://firebase.google.com/docs/cloud-messaging)

PR #2090, opened 2021-11-19
(examples) - Added missing useState and useEffect import.

WildEgo:
## Summary

I decided to try and test the auth exchange package's example and noticed it wasn't compiling so I added the missing imports (useState and useEffect).


JoviDeCroock (approved):
Thank you so much for correcting that!

Issue #2118, opened 2021-12-01
RFC: Vue 2 support with vue-demi inclusion


dargmuesli:
## Summary

I'm coming from Vue 2 with no realistic way to upgrade to Vue 3 just yet. I still want to start using urql, but urql-vue is currently only supported for Vue 3.

## Proposed Solution

By using [vue-demi](https://github.com/vueuse/vue-demi) it seems to be possible to make urql support Vue 2 too fairly simple.

Is that an option to you? I'd be highly pleased!

kitten:
Closing this for now ✌️
Bot because vue-demi wasn't quite a good fit and this is the second attempt 😅 And, we're probably better off focusing on Vue 3 fixes and features

Issue #1327, opened 2021-01-19
RFC: Vue.js Options API support


franciscolourenco:
## Summary

Some people would prefer to keep using the Options API instead of the alternative Composition API.

The [Options API](https://v3.vuejs.org/api/options-api.html), as opposed to the Composition API, is the original way of creating Vue apps and components which can be found all over the [docs](https://v3.vuejs.org/guide/introduction.html#handling-user-input), in which an options object is passed to the Vue constructor, or exported in the case of [Single File Components](https://v3.vuejs.org/guide/single-file-component.html#introduction).

## Proposed Solution

A way to use urql using the declarative Options API in addition to the existing Composition API. The API could look something like the [vue-apollo API](https://apollo.vuejs.org/guide/apollo/#usage-in-vue-components)

JoviDeCroock:
Could you elaborate on what the options api is, how an implementation would look like and so on? This isn't really an RFC rather than a request.

franciscolourenco:
It was not my ambition to create a RFC, just a feature request. How this was the closest template which fitted the feature request case.
![image](https://user-images.githubusercontent.com/208149/105076974-1a69e180-5a8c-11eb-866b-aecbf778c285.png)

With that out of the way, the [Options API](https://v3.vuejs.org/api/options-api.html) as opposed to the Composition API, is the original way of creating Vue apps and components which can be found all over the [docs](https://v3.vuejs.org/guide/introduction.html#handling-user-input), in which an options object is passed to the Vue constructor, or exported in the case of [Single File Components](https://v3.vuejs.org/guide/single-file-component.html#introduction).

franciscolourenco:
vue-apollo is an example of what supporting the Options API could look like https://apollo.vuejs.org/guide/apollo/queries.html#query-with-parameters

kitten:
I'm wondering however, would it make sense to still provide an Options API if Vue 3 shows the Composition API front and centre. Our ambition was to support mostly Vue 3and this seemed to be the more idiomatic approach of the two at least there.
cc @LinusBorg: Do you have any insights on whether it'd be necessary or make sense to support both?

franciscolourenco:
My understanding is that:

1. The majority of the docs is written with the Options API
1. There is a section in the docs for the Composition API
1. There is no intention to deprecate the Options API
1. The Composition API is not touted as the successor to the options API

LinusBorg:
[...]

PR #2119, opened 2021-12-02
feat(vue-2): use vue-demi

dargmuesli:
## Summary

Implements vue-demi as proposed in #2118.

## Set of changes

- replaces import from vue with imports from vue-demi
- splits test calls for v2 and v3
- uses app.mixin for v2 installation (is that correct?)

## Draft

Is draft because I cannot figure out why tests fail for v2... looks like some async issues. Maybe you can help @victorgarciaesgi because of your experiences with vue-chart-3? :pray:

### DIFF
--- a/packages/vue-urql/package.json
+++ b/packages/vue-urql/package.json
@@ -39,7 +39,9 @@
     "dist/"
   ],
   "scripts": {
-    "test": "jest",
+    "test": "yarn test:2 && yarn test:3",
+    "test:2": "vue-demi-switch 2 vue2 && jest",
+    "test:3": "vue-demi-switch 3 && jest",
     "clean": "rimraf dist",
     "check": "tsc --noEmit",
     "lint": "eslint --ext=js,jsx,ts,tsx .",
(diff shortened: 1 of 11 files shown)


victorgarciaesgi:
Hi @dargmuesli ! I had a lot of problems too with testing with `vue-demi`. I could only test on Vue 3, and it was complicated to install each dependencies (like `@vue/test-utils` and `@vue/test-utils@next`. 
My solution was to maintain the main branch for Vue 3, and a legacy branch for Vue 2. That way I could split tests and dependencies and avoid conflicts

dargmuesli:
Thanks a lot for your feedback, @victorgarciaesgi! Highly appreciated.

kitten:
Sorry for the lack of responses 😅 We did have an eye on this and were hoping for this to be ... magically resolved more or less 😆 But in a monorepo like this the method in which `vue-demi` works didn't quite sit well with us. We actually did find and explore it all the way in the beginning of writing the bindings. However, that didn't quite work out for us as we quickly realised we'd run into problems, exactly like the ones with testing. But I do think the list goes on a little. Oh well 😅 

But basically, we were pretty quiet around issues like these (and potential support for Vue 2) because of this: https://blog.vuejs.org/posts/vue-3-as-the-new-default.html#potential-required-actions
We were holding out (for longer than expected) until Vue 3 became the new stable release.

So, I am thinking, in theory, given that Vue 2 support can be "hacked" into `@urql/vue` — if I'm not mistaken — with some bundling and aliasing tricks, I'd currently weigh on the side of cancelling #2118. Of course, I'm very thankful though of the progress and attempt here (albeit, I'm still not too convinced of libraries that modify module source in monorepos using some CLI commands 😅 (sorry... but of course not substantial to you personally)), so I'll leave it up more to you rather than to @JoviDeCroock and myself :v:

This of course doesn't quite eliminate the need for #1327, since afaik Vue 3 will continue supporting the Options API. But, as may be obvious, we don't spend a lot of our time with Vue, and have limited time to maintain and work on new urql changes, apart from planning ahead, hence some of the Vue/Svelte related changes have stalled recently.

dargmuesli:
Thank you for the insight!
Hmm, I'm a little disappointed by Vue's / Nuxt's current state of helpfulness regarding things like this issue. Everything is kinda there, but not really. I had a look in READMEs, documentations and other projects' sources, but then testing is such an issue.
I understand if you'd like to close this. I mean we can reopen this once someone comes around the corner with better ideas than me, but I'd have to dig really deep to make progress, which I - and surely almost everyone else - can't really afford. I really tried, but seems like bad DX on the frameworks' end. I'd be happy to be convinced of the opposite. Until then :rocket:

PR #2123, opened 2021-12-04
(persisted-fetch) - Fix Crypto API support for Web Workers and Node ESM mode

### LINKED ISSUE #2077: (persisted-fetch) - The Node Crypto API is not available.
Using nextjs 12, when trying to use Automatic Persisted Queries using Node `lts/fermium`, I got this error:
`[@urql/exchange-persisted-fetch]: The Node Crypto API is not available.`

In order to make it work, I copied the [hash.js file](https://github.com/FormidableLabs/urql/blob/00ad6421b8f08ce41892b14c91e12464f59d4202/exchanges/persisted-fetch/src/sha256.ts#L42) locally and replaced the `nodeCrypto = new Function('require', 'return require("crypto")')(require);`by `nodeCrypto = require('crypto')`

```
export const client = createClient({
  url: `${process.env.BACKEND_URL}/graphql`,
  preferGetMethod: true,
  exchanges: [
    dedupExchange,
    cacheExchange({
      updates: {...},
      optimistic: {...},
    }),
    persistedFetchExchange({
      preferGetForPersistedQueries: true,
    }),
    fetchExchange,
  ],
  fetchOptions: () => {
    const token =
      typeof window !== 'undefined' ? localStorage.getItem('jwt') : undefined
    return {
      headers: { authorization: token ? `Bearer ${token}` : '' },
    }
  },
})
```

urql version & exchanges:

urql: ^2.0.5
urql/exchange-persisted-fetch: ^1.3.2

1. 2.

**Expected behavior**
No error message

**Actual behavior**
Error message

kitten:
Resolves #2077 

## Summary

Add support for Web Workers via the `self` global and fix support in Node ESM by trying a dynamic import when `require` fails. This also fixes up a mistake in IE11 support, where `msCrypto` was treated as `msCrypto.subtle` accidentally.

## Set of changes

- Check for `self` and `window` global
- Fix `msCrypto.subtle` usage
- Add dynamic `import("crypto")` fallback in require check

### DIFF
--- a/exchanges/persisted-fetch/src/sha256.ts
+++ b/exchanges/persisted-fetch/src/sha256.ts
@@ -35,7 +35,7 @@ if (typeof window === 'undefined') {
   try {
     // Indirect eval/require to guarantee no side-effects in module scope
     // (optimization for minifiers)
-    nodeCrypto = new Function('require', 'return require("crypto")')(require);
+    nodeCrypto = new Function('return require("crypto")')();
   } catch (e) {}
 }
 
--- a/exchanges/persisted-fetch/src/sha256.ts
+++ b/exchanges/persisted-fetch/src/sha256.ts
@@ -35,11 +35,12 @@ if (typeof window === 'undefined' && !webCrypto) {
   // Indirect eval'd require/import to guarantee no side-effects in module scope
   // (optimization for minifiers)
   try {
-    nodeCrypto = Promise.resolve(new Function('return require("crypto")')());
+    nodeCrypto = Promise.resolve(
+      new Function('require', 'return require("crypto")')(require)
+    );
   } catch (_error) {
-    // @ts-expect-error
     try {
-      nodeCrypto = import('crypto');
+      nodeCrypto = new Function('return import("crypto")')();
     } catch (_error) {}
   }
 }
--- a/exchanges/persisted-fetch/src/sha256.ts
+++ b/exchanges/persisted-fetch/src/sha256.ts
@@ -1,15 +1,15 @@
-const jsCrypto =
-  typeof window !== 'undefined'
-    ? window.crypto || (window as any).msCrypto
-    : undefined;
-const cryptoSubtle =
-  jsCrypto && (jsCrypto.subtle || (jsCrypto as any).webkitSubtle);
-const isIE = !!(jsCrypto && (window as any).msCrypto);
+const globals: { crypto?: Crypto; msCrypto?: Crypto } = (typeof window !==
+'undefined'
+  ? window
+  : typeof self !== 'undefined'
+  ? self
+  : null) as any;
+const webCrypto = globals && (globals.crypto || globals.msCrypto);
 
 const sha256Browser = (bytes: Uint8Array): Promise<Uint8Array> => {
-  const hash = cryptoSubtle!.digest({ name: 'SHA-256' }, bytes);
+  const hash = webCrypto!.subtle.digest({ name: 'SHA-256' }, bytes);
   return new Promise((resolve, reject) => {
-    if (isIE) {
+    if (globals.msCrypto) {
       // IE11
       (hash as any).oncomplete = function onComplete(event: any) {
         resolve(new Uint8Array(event.target.result));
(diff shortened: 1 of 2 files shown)


exchanges/persisted-fetch/src/sha256.ts
```diff
@@ -30,61 +30,56 @@ const sha256Browser = (bytes: Uint8Array): Promise<Uint8Array> => {
...
+let nodeCrypto: Promise<typeof import('crypto')> | void;
+if (typeof window === 'undefined' && !webCrypto) {
+  // Indirect eval'd require/import to guarantee no side-effects in module scope
+  // (optimization for minifiers)
   try {
-    // Indirect eval/require to guarantee no side-effects in module scope
-    // (optimization for minifiers)
-    nodeCrypto = new Function('require', 'return require("crypto")')(require);
-  } catch (e) {}
+    nodeCrypto = Promise.resolve(new Function('return require("crypto")')());
```
ryan-roemer, line 38: Have you actually seen this work from a file?

If I go in the node repl, `nodeCrypto = new Function('return require("crypto")'` works. But from a file just like above I never get this to actually successfully do the `require`.

exchanges/persisted-fetch/src/sha256.ts
```diff
@@ -30,61 +30,56 @@ const sha256Browser = (bytes: Uint8Array): Promise<Uint8Array> => {
...
+  } catch (_error) {
+    // @ts-expect-error
+    try {
+      nodeCrypto = import('crypto');
```
ryan-roemer, line 42: Are bundlers still going to try and add `crypto` to the bundle with this present (even deeply inside try/catches)?

(I should mention that haven't actually experimented with the build end of things and inspected any bundles. 😂 )

exchanges/persisted-fetch/src/sha256.ts
kitten: Oh, yea, looks like in some environments `require` isn't available in the global scope. I was doubting this for a bit, hence the change 🤔

exchanges/persisted-fetch/src/sha256.ts
kitten: Actually, let's make that one indirect too, just in case. I didn't test this thoroughly, so who knows what happens in other bundlers 🤔 I can totally see this causing problems when it isn't treated as a Node builtin.

kitten:
Adjusted as per above comments, and tested this in a quick ESM/CJS file. So basically just sanity checking this 👀
Screenshot 2021-12-04 at 18 02 19

ryan-roemer (approved):
I pasted the new implementation into my experiment setup and tried a lot of node versions and CJS / ESM modes and everything works fine. Nice work! 🎉 

Haven't seen what this translates to in bundlers and such, so don't know if we've addressed all the complaining about `require`, etc.

JoviDeCroock (approved):
Awesome, haven't seen much dynamic imports in libs yet, happy to see this becoming a thing!

kitten:
Actually, I hope it doesn't become a thing and ESM `package.json:exports` actually becomes better and adopted by all bundlers with a deeper spec or de facto convention 😆

ryan-roemer:
We're in an awkward time. 😂 

For things like dynamic `import()`, Node 12 is EOL'ed 2022-04-30, but the actual ESM / dynamic import support only shows up unflagged in `v12.17.0 `. So an ambiguity for library maintainers is "what does mainstream Node support mean?"

And yeah, `package.json:exports` is exceedingly challenging in that it's got just a ton of not-really-intuitive options and is just nowhere near the [browser import maps](https://github.com/WICG/import-maps) spec, which I would have assumed was a target for figuring how to reason about bare specifiers, etc.

Not sure if y'all caught it, but I posted a short ESM basics and footguns a little while back: https://formidable.com/blog/2021/node-esm-and-exports/ (Phil and I were originally planning on doing this together, as he was the very first of us to really feel the Node ESM pains 😭 )

kitten:
I mean, if we do think this point a little further though 😅 the problem with `package.json:exports` is not only its flaws but where it's lacking, misunderstood, and/or not even implemented. Additionally, Deno is a huge curve ball here too together with newer bundlers.

PR #2124, opened 2021-12-04
(retry) - Fix delay misqueue in React Native

### LINKED ISSUE #1932: Retry-exchange Operations never resolving in react-native
urql version & exchanges:
```
"urql": "^2.0.3",
"@urql/exchange-retry": "^0.3.0",
"react-native": "~0.63.4"
"expo": "~41.0.1"
```

**Steps to reproduce**
In react-native app:
1. Setup queries to fail and be retried with `retryExhcnage`
2. Fire off query

**Expected behavior**
Query retried `maxNumberAttempts` times and then resolves 

**Actual behavior**
- Query never resolves (error never received by hook or manual caller)
- Retry query never fired (iOS + sometimes android)

Demo:
https://snack.expo.dev/@git/github.com/Mookiies/request-policy-demo-native
change expo to `41.0.0` in bottom bar
Screen Shot 2021-09-07 at 5 07 13 PM

In this demo the `toggle todos` button will cause text to render when an error comes back from a `useQuery` and showing loading until then. It is common for it to get stuck in loading and never receive an error. This happens basically every time ios, sometimes android, never web (all same expo snack code). 

Commenting out the retry exchange causes this problem to go away. When running locally with debugExchange you can see that operations stop passing between fetch and retry exchanges.

kitten:
Resolve #1932 

**NOTE:** This is simply an _attempt_ at resolving the issue above. I'm not sure if this is actually the cause so it'll need to be tested.

## Summary

Our current theory is that because the `delay` operator separately applies a timeout (e.g. `setTimeout`) to each signal, that signals in the same tick (in this case `Push` then `End`) are sometimes flipped. Usually this isn't an issue but with only one value to push the reordering can cause the signal to be swallowed entirely.

To mitigate this I'm attempting to simply validate this by swapping out `delay` with `debounce`, since the latter applies timeouts differently but otherwise behaves identically (in theory) for single values.

## Set of changes

- Replace `delay` with `debounce` in `retryExchange` retry logic

### DIFF
--- a/.changeset/four-moose-scream.md
+++ b/.changeset/four-moose-scream.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-retry': patch
+---
+
+Fix operations sometimes not being executed after a retry is supposed to be triggered, due to a `setTimeout` reordering issue when the timer isn't as predictable as it should be.
--- a/exchanges/retry/src/retryExchange.ts
+++ b/exchanges/retry/src/retryExchange.ts
@@ -5,7 +5,7 @@ import {
   merge,
   filter,
   fromValue,
-  delay,
+  debounce,
   mergeMap,
   takeUntil,
 } from 'wonka';
@@ -96,7 +96,7 @@ export const retryExchange = ({
               retryCount,
             })
           ),
-          delay(delayAmount),
+          debounce(() => delayAmount),
           // Stop retry if a teardown comes in
           takeUntil(teardown$)
         );


JoviDeCroock (approved):
Judging from the [reproduction](https://snack.expo.dev/8CM1c1za4) this patch works correctly, when I use an iOS device with the normal `retryExchange` "toggle todos" will load forever and with this version it actually resolves after a few retries

yankovalera:
I can confirm this works with https://trygql.formidable.dev/graphql/intermittent-colors API and the [with-retry](https://github.com/FormidableLabs/urql/tree/main/examples/with-retry) code example

Issue #2007, opened 2021-10-05
Relay pagination in cache exchange breaks querying totalCount


Tam:
urql version & exchanges:
```
"@urql/exchange-auth": "^0.1.6",
"@urql/exchange-graphcache": "^4.3.3",
"@urql/exchange-multipart-fetch": "^0.1.13",
"@urql/exchange-retry": "^0.2.1",
```

**Steps to reproduce**

1. Setup your client with the following cache exchange:
```js
cacheExchange({
    resolvers: {
        Query: {
            users: relayPagination(),
        },
    },
}),
```

2. Try to query data from the `users` query (or your relay paginated query) that _doesn't_ include paginated data.
```graphql
query GetTotalUsers {
  users {
    totalCount
  }
}
```

**Expected behavior**
Query should return object with total count:
```js
{ users: { totalCount: 124 } }
```

**Actual behavior**
Query returns `null`.

Tam:
Specifying the `first` param works as a work-around to get the response.

```graphql
query GetTotalUsers {
  users (first: 1) {
    totalCount
  }
}
```

I should note that the query to the server is always successful, it's URQL parsing/caching the data that causes the issue. It's also fine if you remove the `relayPagination()` from the exchange.

LeBenLeBen:
FYI this issue can be reproduced with `simplePagination` as well.

JoviDeCroock:
After testing and talking this over a bit through #2137 we came to the conclusion that this is actually not spec-compliant as it doesn't include paginated arguments. This means that using `users (first: 1)` is actually the correct approach.

[Spec](https://relay.dev/graphql/connections.htm#sec-Arguments)

> A field that returns a Connection Type must include forward pagination arguments, backward pagination arguments, or both.

This means that passing `users (first: 0)` is also a possibility

JoviDeCroock:
Hey all,

After researching this a bit in #2137 we concluded that this is actually against the [`relay-spec`](https://relay.dev/graphql/connections.htm#sec-Arguments)

> A field that returns a Connection Type must include forward pagination arguments, backward pagination arguments, or both.

Which means we always need a arguments to support paginating, so in essence `first: 1` working is the right thing here. In general the `relay-spec` specifies that we _can_ support arbitrary arguments on the connection which will work for arguments as you alluded to _but_ without arguments we'd have to pick these from a random-page which isn't guaranteed to be correct.

kitten:
Closing this for now, due to a lack of responses.
[...]

PR #2137, opened 2021-12-09
(graphcache) - fix returning totalCount in relay pagination

JoviDeCroock:
Resolves https://github.com/FormidableLabs/urql/issues/2007

## Summary

In the `relay-pagination` we can have "edges", "nodes" and "pageInfo" but along side that we can also have a `totalCount` we should support and return this.

## Set of changes

- return `totalCount`

### DIFF
--- a/exchanges/graphcache/src/extras/relayPagination.ts
+++ b/exchanges/graphcache/src/extras/relayPagination.ts
@@ -20,6 +20,7 @@ interface Page {
   edges: NullArray<string>;
   nodes: NullArray<string>;
   pageInfo: PageInfo;
+  totalCount?: string;
 }
 
 const defaultPageInfo: PageInfo = {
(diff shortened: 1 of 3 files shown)


exchanges/graphcache/src/extras/relayPagination.ts
```diff
@@ -263,7 +271,8 @@ export const relayPagination = (
     const hasCurrentPage = !!ensureKey(
       cache.resolve(entityKey, fieldName, fieldArgs)
     );
-    if (!hasCurrentPage) {
+
+    if (!hasCurrentPage && (fieldArgs.first || fieldArgs.last)) {
```
JoviDeCroock, line 275: If we aren't asking for any pages we can just return the totalCount, alternatively we could check whether there's a selectionSet containing nodes/edges/...

kitten:
I think this field is non-standard and hence puts us in an awkward position where we have to fall back to supporting non-cumulative fields from the most recent connection entity https://relay.dev/graphql/connections.htm#sec-Connection-Types.Fields

So, `totalCount` isn't standard basically, but I'm also not confident it's the most common name for it, which is unlike `nodes`

> Connection types must have fields named edges and pageInfo. They may have additional fields related to the connection, as the schema designer sees fit.

JoviDeCroock:
Hmm, I've seen a few implement this but I suppose we can tackle this as a derivation from the selectionSet? WDYT?

JoviDeCroock:
Closing due to https://github.com/FormidableLabs/urql/issues/2007#issuecomment-989981090

PR #2164, opened 2021-12-25
(react) - leverage the new `use-sync-external-store` package

JoviDeCroock:
## Summary

Leverage the new `use-sync-external-store` for `useQuery` to make the implementation more in line with React. Our current implementation works already with React Suspense, ... but having an implementation that coexists with React will be better for future changes.

Todos:

- [x] typing issues
- [x] prepassing currently doesn't work due to us setting the initial source too late

[Sandbox demo](https://codesandbox.io/s/inspiring-antonelli-im7k0?file=/src/lib/useQuery.ts)

## Set of changes

- use `synx-external-store` for `useQuery` and `useSubscription`

### DIFF
--- a/.changeset/wicked-terms-sneeze.md
+++ b/.changeset/wicked-terms-sneeze.md
@@ -1,0 +1,5 @@
+---
+'urql': minor
+---
+
+Leverage the nex `use-sync-external-store` for `useQuery` to make the implementation more in line with React. Our current implementation works already with React Suspense, ... but having an implementation that coexists with React will be better for future changes
--- a/.changeset/wicked-terms-sneeze.md
+++ b/.changeset/wicked-terms-sneeze.md
@@ -2,4 +2,4 @@
 'urql': minor
 ---
 
-Leverage the new `use-sync-external-store` for `useQuery` to make the implementation more in line with React. Our current implementation works already with React Suspense, ... but having an implementation that coexists with React will be better for future changes
+Leverage the new `use-sync-external-store` package and `useSyncExternalStore` hook in `useQuery` implementation to bring the state synchronisation in React in line with React v18. While the current implementation works already with React Suspense and React Concurrent this will reduce the maintenance burden of our implementation and ensure certain guarantees so that React doesn't break us.
--- a/.changeset/wicked-terms-sneeze.md
+++ b/.changeset/wicked-terms-sneeze.md
@@ -2,4 +2,4 @@
 'urql': minor
 ---
 
-Leverage the nex `use-sync-external-store` for `useQuery` to make the implementation more in line with React. Our current implementation works already with React Suspense, ... but having an implementation that coexists with React will be better for future changes
+Leverage the new `use-sync-external-store` for `useQuery` to make the implementation more in line with React. Our current implementation works already with React Suspense, ... but having an implementation that coexists with React will be better for future changes
--- a/packages/react-urql/package.json
+++ b/packages/react-urql/package.json
@@ -61,7 +61,7 @@
   },
   "dependencies": {
     "@urql/core": "^2.3.6",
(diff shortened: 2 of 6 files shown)


.changeset/wicked-terms-sneeze.md
```diff
@@ -0,0 +1,5 @@
+---
+'urql': minor
+---
+
+Leverage the new `use-sync-external-store` for `useQuery` to make the implementation more in line with React. Our current implementation works already with React Suspense, ... but having an implementation that coexists with React will be better for future changes
```
kitten, line 5:
```suggestion
Leverage the new `use-sync-external-store` package and `useSyncExternalStore` hook in `useQuery` implementation to bring the state synchronisation in React in line with React v18. While the current implementation works already with React Suspense and React Concurrent this will reduce the maintenance burden of our implementation and ensure certain guarantees so that React doesn't break us.
```

packages/react-urql/package.json
```diff
@@ -60,6 +61,7 @@
   },
   "dependencies": {
     "@urql/core": "^2.3.6",
+    "use-sync-external-store": "^1.0.0-rc.0",
```
kitten, line 64: I'm wondering; when is this releasable given that this package is still tagged as an `rc`? Would we want to define `^1.0.0-rc.0 || ^1.0.0`?

packages/react-urql/src/hooks/useQuery.ts
```diff
@@ -1,9 +1,8 @@
 /* eslint-disable react-hooks/exhaustive-deps */
-
 import { DocumentNode } from 'graphql';
-import { Source, pipe, subscribe, onEnd, onPush, takeWhile } from 'wonka';
-import { useState, useEffect, useCallback, useMemo } from 'react';
-
+import { Source, pipe, subscribe, onPush, takeWhile } from 'wonka';
+import { useCallback, useMemo, useState } from 'react';
+import { useSyncExternalStore } from 'use-sync-external-store/shim';
```
kitten, line 5: We may have to do some build tests on some bundlers to _ehem_ double check their work here

packages/react-urql/src/hooks/useQuery.ts
```diff
@@ -114,112 +105,96 @@ export function useQuery<Data = any, Variables = object>(
...
-          : client.executeQuery(request, context);
-        return [source, state[1], deps];
       });
+
+      const source = suspense
+        ? pipe(
+            fetchSource,
+            onPush(result => {
+              cache.set(request.key, result);
+            })
```
kitten, line 150: Could this be moved to the two separate `subscribe` calls?

packages/react-urql/src/hooks/useQuery.ts
```diff
@@ -41,70 +40,62 @@ export type UseQueryResponse<Data = any, Variables = object> = [
...
-let currentInit = false;
-
 export function useQuery<Data = any, Variables = object>(
   args: UseQueryArgs<Variables, Data>
 ): UseQueryResponse<Data, Variables> {
   const client = useClient();
-  const cache = getCacheForClient(client);
-  const suspense = isSuspense(client, args.context);
   const request = useRequest<Data, Variables>(args.query, args.variables);
+  const suspense = isSuspense(client, args.context);
```
kitten, line 54: We can potentially lift a limitation here for suspense. Previously it wasn't possible to disable `suspense` using `executeQuery`; however, we can circumvent this by deriving this in `getSnapshot` and keeping an extra suspense flag around in `executeQuery` + `meta`

kitten (commented):
I think the only remaining task is that we may have to pair and look at `useSubscription` to finish this off 🙌

kitten (approved):

PR #2179, opened 2022-01-06
Support multiple API clients

wucdbm:
## Summary

I have two GraphQL APIs in my app
I need to be able to provide myself two urql instances
As a consequence, I can't use useQuery as it injects the first (or last?!) provideClient client arg

Followup of https://github.com/FormidableLabs/urql/pull/2173

## Set of changes

I need to be able to create my own useThisApiQuery and useAnotherApiQuery functions with the respective client, and having this function exposed is a way for me to stich things together

A better solution may be to improve the provideClient and useClient functions with an optional argument that defaults to the currently used `'$urql'`, then improve the useQuery function with a second optional argument to select the correct client to run the query against
In this particular case, it would be trivial to reuse the functions provided by this library by doing so

```
provideClient(client, '$myApiIdentifier');
```

```
export function useMyApiQuery<T = any, V = object>(args: UseQueryArgs<T, V>) {
    return useQuery(args, '$myApiIdentifier');
}
```

To be honest, I may very well need to execute queries and it seems that the client's `executeQuery` function isn't very friendly towards executing a string-query, as I'd need to call `createRequest`, etc, so I'd still expose `callUseQuery` to the public API of the library. Even with my second idea, the `useQuery` wrapper function will not help because if I want to call the API in my app bootstrap files (in order to start some requests very early on), it will try to inject the client, as opposed to directly using one. In that case, we could make the parameter accept both an identifier or a client, but that will unnecessarily increase the complexity, and IMO that's undesirable when one could simply expose `callUseQuery`

As for my second suggestion, that'd make for much cleaner code in my app

PR only includes exposing `callUseQuery` as that'll do for me at this point and I'd like to know your thoughts on my second suggestion.
And before you flame me on how to do GQL, the frontend app will connect to two completely distinct GQL APIs

### DIFF
--- a/.changeset/bright-hornets-remember.md
+++ b/.changeset/bright-hornets-remember.md
@@ -5,3 +5,47 @@
 Exposed callUseQuery to allow executing queries in app setup functions.
 Add clientId parameter to useClient and provideClient, as well as useQuery to support multi-client setups (Note: You must provide app.provide('\$myUrql', myUrqlClient) yourself).
 Then it becomes trivial to write code like const useMyApiQuery = (args) => useQuery(args, '$myUrql') and provide this function across your app to execute queries against a concrete API Client
+
+For example:
+
+```ts
+export function createApp() {
+  const app = createApp();
+
+  const myApiExchange = createSSRExchange({
+    isClient: !import.meta.env.SSR,
+    initialState: initialState?.myApi,
+    staleWhileRevalidate: false,
+  });
+  const myApi = new Client({
+    url: 'https://my-api.com',
+    exchanges: [dedupExchange, cacheExchange, myApiExchange, fetchExchange],
+  });
+  app.provide('$myApi', myApi);
+
+  const anotherApiExchange = createSSRExchange({
+    isClient: !import.meta.env.SSR,
+    initialState: initialState?.anotherApi,
+    staleWhileRevalidate: false,
+  });
+  const anotherApi = new Client({
+    url: 'https://another-api.com',
+    exchanges: [dedupExchange, cacheExchange, anotherApiExchange, fetchExchange],
+  });
+  app.provide('$anotherApi', anotherApi);
+
+  return app;
+}
+
+export function useMyApiQuery<T = any, V = object>(
+  args: UseQueryArgs<T, V>
+): UseQueryResponse<T, V> {
+  return useQuery(args, '$myApi');
+}
+
+export function useAnotherApiQuery<T = any, V = object>(
+  args: UseQueryArgs<T, V>
+): UseQueryResponse<T, V> {
+  return useQuery(args, '$anotherApi');
+}
+```
--- a/.changeset/bright-hornets-remember.md
+++ b/.changeset/bright-hornets-remember.md
@@ -1,0 +1,7 @@
+---
+'@urql/vue': patch
+---
+
+Exposed callUseQuery to allow executing queries in app setup functions.
+Add clientId parameter to useClient and provideClient, as well as useQuery to support multi-client setups (Note: You must provide app.provide('\$myUrql', myUrqlClient) yourself).
+Then it becomes trivial to write code like const useMyApiQuery = (args) => useQuery(args, '$myUrql') and provide this function across your app to execute queries against a concrete API Client
--- a/.changeset/bright-hornets-remember.md
+++ b/.changeset/bright-hornets-remember.md
@@ -34,6 +34,19 @@ export function createApp() {
   });
   app.provide('$anotherApi', anotherApi);
 
+  const apiResult = callUseQuery(
+    {
+      query: 'some query',
+    },
+    myApi
+  );
+
+  apiResult.then(() => {
+    // well, now we can execute requests right here before components are rendered
+    // and at the same time leverage the goodies in callUseQuery that
+    // and we don't need to parse a string query ourselves, and then call client.executeQuery
+  });
+
   return app;
 }
 
--- a/packages/vue-urql/src/index.ts
+++ b/packages/vue-urql/src/index.ts
@@ -5,6 +5,7 @@ export { install, provideClient } from './useClient';
 
 export {
   useQuery,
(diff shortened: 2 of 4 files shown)


grigoryosifov:
Hi @kitten, could you please elaborate on why you closed the PR? This is a huge show stopper for our team as well, and the only alternative would be to fork the entire repo with the proposed changes.

kitten:
Sorry for not replying that quickly, including on the other PR.
Typically, we ask people to open issues and discuss problems they're having before submitting changes, to save them time on code that wouldn't get merged later on ([Contributing Guide](https://github.com/FormidableLabs/urql/blob/main/CONTRIBUTING.md))

In this case, I'm not inclined to go ahead with this change because it's come up repeatedly in multiple duplicate issues under a name like "switching `client` in hook" (in the React sense), i.e. allowing a `Client` to be passed in to have multiple clients.

We wouldn't really like to use multiple clients or recommend multiple clients, and our API is built around having a client for one app (or in certain cases per request on the server-side, or per subtree in apps in rare occasions)

Hence, there are issues that describe exchanges and extensions around creating "splits" that send queries to different APIs.

Now, the reason why in general this isn't an approach I'd recommend, from a general GraphQL perspective, is that schemas are built to work for apps; Obviously, there'll be times when there are multiple external schemas (rather "random" schemas so to say), and it'll be tempting to just call them, like REST APIs. But I'd say that it's important to highlight that that isn't using GraphQL as a GraphQL schema-defined API, but rather as a generic API resource

grigoryosifov:
Thank you for the explanation, and yes, references would be helpful!

However how you suggest to work around having a single app (say entire complex website front-end with SSR) , working with 2 completely different GraphQL resources, as the second one is coming form a 3rd party SaaS and is only used in one part of the app? We prefer not to proxy the 3rd party GraphQL as it would add additional latency, dependency, and maintenance cost.

wucdbm:
All due respect, and thank you for creating and maintaining such an awesome library, which otherwise does absolutelty everything we needed and will save us months of reinventing the wheel, but this defies all principles of clean code and software architecture.

I don't want to get into religious arguments, but really, if I were to go down that rabbit hole and create an exchange (which to my understanding is merely a fancy name for a middleware), and if I were to create another such middleware for something else, then it too would need to know which "client" (read: url or service) I'm currently calling in order to restrict it to only be active for one of the resources I'm consuming. I guess you realize where this is going once the project becomes larger.

These are two distinct APIs and they have nothing to do with one another; Proxying the second resource through the first one is also a huge no-go due to having another set of network latency between the two services. A CORS call straight to the second API (since it's public and there's nothing to hide/restrict) will be much quicker than proxying it through the main service API.

Really, besides the core functionality of urql, what I have a problem with is basically the `callUseQuery` function not being exposed -- it's these helpers that really `@urql/vue` anything special, without them, we'd be better off copy/pasting them from this codebae and depending on `@urql/core` instead.

Another interesting question is, how do you propose we make async API calls before any component has been rendered, and data used for internal app doings by leveraging the goodies in these helpers, such as not having to stitch together calls to `createRequest`?

At this point, I'll copy/paste the vue lib and expose `callUseQuery`, probably also `callUseMutation` and `callUseSubscription`, then publish a ripoff on NPM.

Rather unfortunate, but if I have a gun and the will to shoot myself in the foot, there's little that cap stop me. I don't understand where the need for such restrictions comes from. Very little other OS libraries and frameworks will try to restrict you from stitching together what you need. If you know what you're doing and that's how you want things, why not?!

kitten:
> Another interesting question is, how do you propose we make async API calls before any component has been rendered, and data used for internal app

This sounds a bit like a one-off query to the API then? For that, we'd use the core API methods on the `Client`

- https://github.com/FormidableLabs/urql/blob/0b1c3d792b4bd2ca8699f77e06d05aca026926eb/packages/core/src/client.ts#L366-L374
- https://formidable.com/open-source/urql/docs/basics/core/#one-off-queries-and-mutations
- https://formidable.com/open-source/urql/docs/api/core/#clientquery

> Very little other OS libraries and frameworks will try to restrict you from stitching together what you need. If you know what you're doing and that's how you want things, why not?!

I'd disagree with that; the basis of many libraries is to restrict surface/user-level APIs. We're also not preventing you from doing this. You can bind to any `Client` in `urql` and implement a hook like `@urql/vue`'s `useQuery` bindings or `callUseQuery` binding. We just want to explicitly make sure that this _isn't easy_ because otherwise we'd simply be encouraging a pattern that we think isn't productive.

In your case you have stated that you could make a schema that suits your app, which involves hosting it. That's basically the point. And I get your hesitation in investing that time. But there are many ways around this:

- You can stitch on the client-side and use `@urql/exchange-execute`
- You could combine that with something like [GraphQL Mesh](https://www.graphql-mesh.com/) or a simple [GraphQL Tools](https://www.graphql-tools.com/docs/schema-stitching/stitch-combining-schemas) utility
- You could use an exchange to switch, as posted in the issue

So I totally get you — you're going for a quick and dirty solution; The reason I'm "branding" it as such and the reason for resistance (as seen in other issues and discussions) is that it's a UI-driven switch based on that. You could even still do that. You can override the `url` parameter of the `OperationContext` in a wrapper around `useQuery`: https://formidable.com/open-source/urql/docs/api/core/#operationcontext

The problem I'm basically showing you is, what you've decided is the quickest and most natural seeming solution for you here is the line we've set where we think discouraging these patterns isn't sufficiently guaranteed anymore. So I'm basically saying with an exchange or with `OperationContext` there are two more ways to shoot yourself in the foot 😅 And on top of that, there are also more proper solutions and ways to make this safer/cleaner.
But this specific proposal is one we wouldn't accept — not because it adds a third way to shoot yourself in the foot — but because once it's added to the public API it a) needs to remain stable, and b) implies that it's an intended mechanism of `@urql/vue` (and hence `urql` in general) and due to how it offers itself up exponentially may _cause_ people to mess themselves up, even if they didn't explicitly accept it as such, or understood the consequences like you do

Issue #2191, opened 2022-01-10
Cache update errors are caught


bkiac:
I had an error in my update resolver logic which was really hard to find because the error was caught by the cache exchange like nothing happened. I'm not sure if this is intentional or not.

urql version & exchanges: 

<table>
<tr>
	<td> urql
	<td> 2.0.6
<tr>
	<td> @urql/exchange-graphcache
	<td> 4.3.6
</table>

**Steps to reproduce**

```ts
const cache: GraphCacheConfig = cacheExchange({
	schema,
	updates: {
		Mutation: {
			doSomething: () => {
				throw new Error("this is caught")
			},
		},
	},
})
```

**Expected behavior**
Errors are propagated/re-thrown

**Actual behavior**
Errors are caught inside the cache exchange

JoviDeCroock:
Nothing in urql is catching these we are throwing them for the user, here is a [simple reproduction](https://codesandbox.io/s/interesting-wing-kduoz?file=/src/index.js) highlighting this case. When you press a todo it will throw an error.

PR #2210, opened 2022-01-21
(core) - fix incremental fetch not bubbling up JS-errors

JoviDeCroock:
## Summary

When we introduced incremental fetch we made sure that the backwards exchanges flow could be triggered multiple times for an `OperationResult` this made it so that the `fetchSource` would actually have the backwards execution of the exchanges like GraphCache writing data, ... inside of the `try catch` which means that if one of these throws we end up in the catch and handle this as if we encountered a `GraphQL` error.

fixes https://github.com/FormidableLabs/urql/issues/2191

## Set of changes

- filter out `Syntax`, `Type` and normal `Error` scenario's coming from the reverse-exchange run

### DIFF
--- a/packages/core/src/internal/fetchSource.ts
+++ b/packages/core/src/internal/fetchSource.ts
@@ -146,10 +146,7 @@ export const makeFetchSource = (
       })
       .then(complete)
       .catch((error: Error) => {
-        if (
-          error.name === 'SyntaxError' ||
-          error.name === 'TypeError'
-        ) {
+        if (error.name === 'SyntaxError' || error.name === 'TypeError') {
           const e = new Error(error.message);
           e.stack =
             e.stack!.split('\n').slice(0, 2).join('\n') + '\n' + error.stack;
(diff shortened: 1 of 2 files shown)


packages/core/src/internal/fetchSource.ts
```diff
@@ -146,6 +146,14 @@ export const makeFetchSource = (
       })
       .then(complete)
       .catch((error: Error) => {
+        if (error.name === 'SyntaxError' || error.name === 'TypeError') {
```
JoviDeCroock, line 149: This might be a bit too limiting 😅

kitten: Should we limit this to development?

JoviDeCroock: Well my thinking was that people using for instance Sentry in production would want to see these errors surface as well

kitten: oh yea good point

kitten (approved):

PR #2222, opened 2022-01-28
(storage-rn) - Fix issue where cache data is not cleared correctly

benmechen:
## Summary

When clearing the React Native Storage (for example, on logout), the AsyncStorage keys are cleared correctly. However, @urql/storage-rn caches these keys in local memory. This isn't cleared correctly, which means that unless you recreate the entire Urql client the old data persists in local memory even after calling `clear()`.

#### `storage-rn/makeAsnyStorage.ts`
* Set the `allData` variable to an empty object when `clear()` is called

### DIFF
--- a/.changeset/moody-eels-cough.md
+++ b/.changeset/moody-eels-cough.md
@@ -1,0 +1,5 @@
+---
+'@urql/storage-rn': patch
+---
+
+Fix issue where the in-memory cache would not be cleared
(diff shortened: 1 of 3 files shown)


kitten (approved):
Nice catch!

benmechen:
Thanks! Changeset added, sorry forgot to do that before pushing.

JoviDeCroock (approved):

PR #2227, opened 2022-01-31
(react) - Update initial state to not require abort on mount

### LINKED ISSUE #2226: (react) - Fetching initial value is false on first mount
After upgrading to version 2.1.1 from version 2.0.6 I had problems with the initial value of the "fetching" variable.
After calling the query I get that the data and error values ​​are undefined and the fetching is false.
After that fetching changes to true.
The options are only variables, no pause enabled.

In version 2.0.6 the initial value of fetching after the call to useQuery was true.

2.0.6 version, first value was true, after get data get false.
image

2.1.1 version, first value was false, after get true and finally get data and get false.
image

"urql" >= "2.1.0",
"@urql/devtools": "2.0.3",
"@urql/exchange-auth": "0.1.7",
"@urql/exchange-graphcache": "4.3.6",
"@urql/exchange-multipart-fetch": "0.1.14",

**Steps to reproduce**

useQuery with default options and network-only mode.

**Expected behavior**
Initial value of fetching should be true.
Demo: https://codesandbox.io/s/urql-fetching-2-0-6-jcdjb

**Actual behavior**
Initial value of fetching is false.
Demo: https://codesandbox.io/s/urql-fetching-2-1-1-oe9e6

kitten:
Resolve #2226

## Summary

Previously, the `useSyncExternalStore` reimplementation regressed
on us constructing the correct initial state on mount, and instead
relied on an abort on mount, which is confusing to users.
Instead, we can simply revert to what we've done prior to the
change and ensure that the initial state is accurate.

This fix **does not** change any of the current behaviour, but simply avoids the confusing state transition on mount.

## Set of changes

- Update initial state in `useQuery` implementation

### DIFF
--- a/packages/react-urql/src/hooks/useQuery.ts
+++ b/packages/react-urql/src/hooks/useQuery.ts
@@ -53,17 +53,30 @@ export function useQuery<Data = any, Variables = object>(
   const request = useRequest<Data, Variables>(args.query, args.variables);
   const cache = getCacheForClient(client);
 
+  const currDeps: unknown[] = [
+    client,
+    request,
+    args.pause,
+    args.requestPolicy,
+    args.context,
+  ];
+
   const [meta, setMeta] = useState<{
     source: Source<OperationResult<Data, Variables>> | null;
     prevValue: UseQueryState<Data, Variables>;
-    deps: Array<any>;
+    deps: unknown[];
     suspense: boolean;
-  }>({
-    source: null,
+  }>(() => ({
+    source: args.pause
+      ? null
+      : client.executeQuery(request, {
+          requestPolicy: args.requestPolicy,
+          ...args.context,
+        }),
     prevValue: notFetching,
-    deps: [],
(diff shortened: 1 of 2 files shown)


packages/react-urql/src/hooks/useQuery.ts
```diff
@@ -53,17 +53,30 @@ export function useQuery<Data = any, Variables = object>(
...
+
   const [meta, setMeta] = useState<{
     source: Source<OperationResult<Data, Variables>> | null;
     prevValue: UseQueryState<Data, Variables>;
-    deps: Array<any>;
+    deps: unknown[];
     suspense: boolean;
-  }>({
-    source: null,
+  }>(() => ({
```
kitten, line 69: This isn't strictly speaking necessary, but let's avoid calling `client.executeQuery` on every render 😆

JoviDeCroock (approved):

Issue #2309, opened 2022-02-26
Notice: Recent issues with urql@^2.1 React Bindings


kitten:
I'll keep this one brief, the tl;dr is, if you're using the React bindings, please either upgrade to `urql@2.2.0` or revert to `urql@2.0.6`. Ensure that `@urql/core` remains up-to-date for the latest patches on `2.4.3`.

## What happened?

In a recent version, we've decided to implement `use-sync-external-store` and use the `useSyncExternalStore` hook to align with what may become a future norm in React 18.

This wasn't strictly speaking necessary. A lot of this goes into React's tearing research. However, with each hook being rather independent with a GraphQL client, it was already a small concern that hooks being out-of-sync with updates could be of concern to us.

This didn't pan out. We forgot to port over some of the implementation details we needed to preserve. We encountered or re-encountered issues that we need workarounds for. etc etc.

Additionally, while we thought that with our same support of Concurrent, Suspense, and other "modes of operation" in React, like with SSR and `react-ssr-prepass` we could reduce our maintained code, this has proven to be false. The resulting code remained approximately the same.

In summary, any benefit we thought we'd get, we didn't. Meanwhile, we were observing regressions in our implementation. 

## How did we respond?

To address an influx of issues that people encountered, we reverted the PR that implemented this change, #2227, and are back on our old implementation of hooks like `useQuery`.

This is an easy compromise for us to make. As users, you'll notice no improvement in the (broken) `2.1.3` release. In fact, the intention was for both implementations to work identically.

Hence, we decided to revert and `2.2.0` is the new safe version.

## Why did this happen in the first place?

For two reasons.

We thought that `useSyncExternalStore` may actually reduce the amount of issues we'll encounter in React 18. While this wasn't the case (and we would not be alone with keeping a store-like implementation that isn't using it), we still maintained that it would reduce what we'd have to maintain. This leads into the other reason.

Testing. The testing frameworks we use and the tests we maintain are old. They kept us from regressing on basic API behaviour. However, as we encountered more edge cases relating to Strict Mode, Suspense, and Concurrent Mode, we found that they're impossible to test for in the test setup we have. In fact, many of these issues couldn't even happen in `preact/compat` for example, which shows that they only happen in certain conditions and special modes in React.

A lot of issues that we've regressed on and that are associated with edge cases are: only observable in certain conditions, only trigger warnings, or are easily spotted with specific, custom-made test apps.

[...]

PR #2341, opened 2022-03-11
chore(react): create a proof of concept suspense e2e test suite

JoviDeCroock:
## Summary

We have seen several issues relating to our inability to test `Suspense` #2309 being one of those examples, to test this reliably we'd need some kind of e2e test-suite. This takes a stab at leveraging `cypess` component testing to use a real world example of our pokemon API. This serves as an evaluation PR we should finetune this so we can use a similar test-suite more broadly and not only for React.

## Set of changes

- add cypress component testing

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -1,0 +1,7 @@
+{
+  "video": false,
+  "component": {
+    "componentFolder": "./packages/react-urql",
+    "testFiles": "**/e2e-tests/*spec.tsx"
+  }
+}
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -47,6 +47,13 @@ jobs:
         run: yarn run lint
       - name: Unit Tests
         run: yarn run test --maxWorkers=2
+      - name: Build
+        run: yarn run build
+      - name: e2e tests 🧪
+        uses: cypress-io/github-action@v2
+        with:
+          command: yarn cypress run-ct
+          working-directory: packages/react-urql
```
JoviDeCroock, line 56: Could be moved to its own file but imo we have some breathing space in terms of execution time, when we add more integrations we should for parallization

PR #2370, opened 2022-03-29
feat(svelte): v2 — The return of the functional APIs!

### LINKED ISSUE #2369: RFC: Update @urql/svelte
## Summary

Per [discussion](https://github.com/FormidableLabs/urql/discussions/2113#discussioncomment-2411632), there are several limitations with the current @urql/svelte approach. 

> As you point out, the dependence on `context` causes a number of problems:
> 
> * have to put all the logic into a component, or [go boom](https://github.com/FormidableLabs/urql/discussions/2110). (but as you point out, we may want the logic to be shared between multiple components)
> * that issue means [quite a bit of duplicate code](https://github.com/FormidableLabs/urql/discussions/2201)
> * it also [seems confusing](https://github.com/FormidableLabs/urql/discussions/2310) when svelte makes imports easy
> * it looks more like `react` and not so much `svelte` ... which i think can be [hard to diagnose](https://github.com/FormidableLabs/urql/discussions/1988)
> * [can't create tests](https://github.com/FormidableLabs/urql/discussions/2343)
> * I wonder if these are slowing progress for [graphql-code-generator operation store plugin](https://github.com/dotansimha/graphql-code-generator/pull/6074)?
> 
> this issue seems to be at the core of many of the recent [urql discussions re: svelte](https://github.com/FormidableLabs/urql/discussions?discussions_q=svelte+) it almost seems like it could be easier to just skip the svelte bindings and use urql/core (or at least a documentation issue with this)...

## Proposed Solution

**Big picture**: 
1. develop an approach that doesn't rely on svelte context. 
2. simplify if possible, so that it is easier to maintain

I am submitting a PR.  This is a breaking change; it uses almost none of the existing code.

I built new tests, my code is commented, and I would appreciate feedback.  I imagine this issue can be used for comment on the overall approach (dropping the use of svelte context), and specific implementation issues on the PR. I haven't committed any changes to docs yet.  Happy to update docs also if it looks like we're on track.

Although a substantial change, I believe it becomes more maintainable and simplified.  there appears to be support for this:

@kitten 
> I'm not happy at all with our current API in @urql/svelte

@JoviDeCroock 
> Yes would be more than open to a PR! Really happy to hear someone wants to take this on ❤️

jonathanstanley:
## Summary

This replaces the current @urql/svelte with an approach that does not rely on svelte context.

resolves #2369

## Set of changes

1. New interface `UrqlStore` which is a svelte store containing an `OperationResult` and `fetching` status
2. New calls: `queryStore`, `mutationStore`, and `subscriptionStore` which all return `UrqlStore`
3. New tests
4. `queryStore` and `subscriptionStore` output also provide `pause()`, `unpause()` and `isPaused$` so that developers can manage pausing.
5. Svelte store variables all follow naming convention to end with `$` (ex: `let myStore$ = queryStore(...`)


frederikhors:
Amazingly! I can't wait to try it!

packages/svelte-urql/src/index.ts
```diff
@@ -1,4 +1,5 @@
 export * from '@urql/core';
-export * from './operationStore';
-export * from './context';
-export * from './operations';
+export * from './queryStore';
+export * from './mutationStore';
+export * from './subscriptionStore';
+export * from './common';
```
JoviDeCroock, line 5: why are we exporting everything form common, mainly asking because it contains a few methods that aren't useful to the client

packages/svelte-urql/src/queryStore.ts
```diff
@@ -0,0 +1,85 @@
...
+  /** an [Urql client](https://formidable.com/open-source/urql/docs/api/core/#client)  */
+  client: Client;
+  /** a graphql tag */
+  query: DocumentNode | TypedDocumentNode<Data, Variables> | string;
+  /** the variables to be used in the fetch operation */
+  variables: Variables;
+  /** Urql fetching options */
+  context?: Partial<OperationContext>;
+  /** the initial result value (prior to fetch) */
+  placeholderData?: Partial<OperationResult<Data, Variables>>;
```
JoviDeCroock, line 28: What's the reason behind this, in other clients we use the synchronous result that comes back from the client and/or the previous result in case of an update which is mirrored from the same props

packages/svelte-urql/src/subscriptionStore.test.ts
```diff
@@ -0,0 +1,29 @@
+import { createClient } from '@urql/core';
+import { get } from 'svelte/store';
+import { subscriptionStore } from './subscriptionStore';
+
+describe('mutationStore', () => {
```
JoviDeCroock, line 5:
```suggestion
describe('subscriptionStore', () => {
```

packages/svelte-urql/src/subscriptionStore.ts
```diff
@@ -0,0 +1,59 @@
...
+  /** an [Urql client](https://formidable.com/open-source/urql/docs/api/core/#client)  */
+  client: Client;
+  /** a graphql tag */
+  subscription: DocumentNode | TypedDocumentNode<Data, Variables> | string;
+  /** the variables to be used in the fetch operation */
+  variables?: Variables;
+  /** Urql fetching options */
+  context?: Partial<OperationContext>;
+  /** the initial result value (prior to fetch) */
+  placeholderData?: Partial<AnnotatedOperationResult<Data, Variables>>;
```
JoviDeCroock, line 23: same question here, why are we indicating initial data for a subscription?

JoviDeCroock (commented):
This already looks great, had a few questions about the implementation

packages/svelte-urql/src/queryStore.ts
jonathanstanley: placeholderData can be used when there is no synchronous data yet.  if there is any synchronous data/cache, the placeholder data wouldn't be used.

packages/svelte-urql/src/index.ts
jonathanstanley: I could strip out the functions exports.  IMO it would be good to provide them in case others want to build a custom tailored version.  I don't feel strongly either way.

packages/svelte-urql/src/queryStore.ts
```diff
@@ -0,0 +1,85 @@
...
+  fetchProcess,
+  fromStore,
+} from './common';
+
+/**
+ * Creates a Svelte store for an [Urql query](https://formidable.com/open-source/urql/docs/api/core/#clientexecutequery) using [Wonka](https://wonka.kitten.sh/)
+ */
+export function queryStore<Data, Variables extends object = {}>(props: {
+  /** an [Urql client](https://formidable.com/open-source/urql/docs/api/core/#client)  */
+  client: Client;
```
JoviDeCroock, line 20: Why are we passing in a client manually rather than using it from context? Is this an ssr thing or?

packages/svelte-urql/src/queryStore.ts
JoviDeCroock: I personally wouldn't stray too much from our other inputs, I also see little use for `placeHolderData` as it needs a lot of special knowledge, ...

Generally. I also think that we miss the `requestPolicy` convenience method input.

I don't know whether we want to react to the input of `isPaused` to change (if that's possible in Svelte) as in all the others we react to input.pause

frederikhors:
The "placeholder data" thing is borrowed from react-query / svelte-query. I think there is nothing similar to this in urql. Am I right, @jonathanstanley?

packages/svelte-urql/src/queryStore.ts
jonathanstanley: 1. Agreed; I will drop `placeholderData` to keep in line with other bindings.  If someone wants that functionality, they could recreate it with a derived svelte store... so it isn't needed to be part of urql.
2. I'm not sure I understand the `requestPolicy` input.  A user can set the `requestPolicy` in the client, or can change it individually with the `context` partial.  But the bindings also accept another `requestPolicy` input?  If the user provides `context` and separate `requestPolicy`, it looks like [`context` will win](https://github.com/FormidableLabs/urql/blob/f86763cc565f0c36b948ce9017da0c7c2c302d58/packages/react-urql/src/hooks/useQuery.ts#L60-L63)?  Let me know if I have that right and I can add it.
3. Would you clarify your comment on `isPaused`?

packages/svelte-urql/src/subscriptionStore.ts
jonathanstanley: dropped https://github.com/FormidableLabs/urql/pull/2370/commits/df05cd86b6313a41a9eea74df950f3d0b699b8e4

packages/svelte-urql/src/queryStore.ts
jonathanstanley: this is a big reason for the update (RFC https://github.com/FormidableLabs/urql/issues/2369)

In react, this "manual" pattern would be annoying.  In Svelte, I believe this is the preferable approach.

You may be looking for a more thorough answer... and I will be putting together some examples/docs for this to show three patterns:

1. simple case can just use the an `UrqlStore` directly
2. component-driven approach can use svelte context
3. separate-logic approach can use stores and doesn't need svelte context.
4. combine any of the three

packages/svelte-urql/src/common.ts
```diff
@@ -0,0 +1,119 @@
...
+    // react to response
+    pipe(
+      // fetch
+      executeOperation,
+
+      // build annotated result from base (set fetching:false)
+      map(response => ({ ...baseResult, ...response })),
+
+      // mark the fetch as complete
+      onPush(() => complete && complete())
```
jonathanstanley, line 87: Wonka is not very clear to me, so I would especially appreciate any feedback here.

for example: if the client has a cached copy, i assume it would be emitted from the executeOperation? 

if yes: can anyone suggest a way to specify fetching ONLY during a network request?
if no: how does cache emit?

packages/svelte-urql/src/subscriptionStore.ts
```diff
@@ -0,0 +1,57 @@
...
+  UrqlStore,
+  UrqlStoreArgs,
+} from './common';
+import { defaultBaseResult, fetchProcess } from './common';
+
+/**
+ * Create a Svelte store for an [Urql subscription](https://formidable.com/open-source/urql/docs/api/core/#clientexecutemutation) using [Wonka](https://wonka.kitten.sh/)
+ */
+export function subscriptionStore<Data, Variables extends object = {}>(
+  args: UrqlStoreArgs<Data, Variables>
```
jonathanstanley, line 16: I noticed that react binding allows pausing the subscription. https://github.com/FormidableLabs/urql/blob/f86763cc565f0c36b948ce9017da0c7c2c302d58/packages/react-urql/src/hooks/useSubscription.ts#L21

I have not implemented that yet.

packages/svelte-urql/src/common.ts
jonathanstanley: [a comment](https://github.com/FormidableLabs/urql/issues/2371#issuecomment-1085515914) that appears to affirm this implementation:

> as long as the exchange chain is busy we are in theory fetching and that is the only thing a binding should care about.

based on that, it appears we shouldn't try to parse whether there is a network request or not.  fetching is simply a status of whether we're waiting on a response from the client/exchange.

packages/svelte-urql/src/queryStore.ts
jonathanstanley: @JoviDeCroock  I read your comment on `isPaused` and I think I'm understanding now. You would prefer the **input** variable doesn't change from `pause` to `isPaused`?  I went ahead and changed it (ex: https://github.com/FormidableLabs/urql/pull/2370/files#diff-2492165b6a8d5566e8c48fd4a46213f895cc13e6ab6cb8c5c5fbef4904f9a06f)

packages/svelte-urql/src/queryStore.ts
kitten: sorry for chiming in late, I'm still reading this as I find time :v:
I still find it odd to ignore Svelte's context entirely. I get the gap here and the gap context creates. However, I'd find it good if we had a way to continue providing a common context key that's in a way "reserved" for us, since this isn't identity based.

So, I suppose we could:
- Keep `getContextClient` & `setContextClient` intact (with an appropriate rename, not necessarily this one)
- Optionally and additionally: continue supporting context when no `client` is passed with new primitives
- or: export a shared context key that we recommend people to use when providing the `client` via Svelte's context

This way we don't require this but we also don't start ignoring that it's an option for people to use

packages/svelte-urql/src/common.ts
kitten: That's right, yes. Results basically have a `stale` flag that indicate background fetches / requests. These aren't necessarily tied to an HTTP request happening, and that request isn't assumed to be happening right now. In other words, we assume that there's no need for the UI to know why it's waiting for a result because in most cases we assume it's better to deliver results as soon as possible by default.

That plays into how Wonka's sources work and why they're built synchronous and with predictable ordering in mind. We basically try to impose these requirements on bindings for ourselves:
- When a query is initiated, we execute an operation
- If we get a result immediately then we must never see `fetching: true` as a result (not even as a temporary intermediary step)
- If we don't get a result immediately then we must see `fetching: true` first
- If the operation is changed, e.g. variables change, if the framework for these bindings allow for this, we attempt to keep the previous result unchanged and simply set `fetching: true` again

packages/svelte-urql/src/common.ts
```diff
@@ -0,0 +1,140 @@
...
+import type {
+  OperationResult,
+  RequestPolicy,
+  OperationContext,
+  Client,
+} from '@urql/core';
+import { pipe, fromValue, concat, map, makeSubject, onPush } from 'wonka';
+import type { Source } from 'wonka';
+import type { DocumentNode } from 'graphql';
+import type { TypedDocumentNode } from '@graphql-typed-document-node/core';
```
kitten, line 11:
```suggestion
import type {
  OperationResult,
  RequestPolicy,
  OperationContext,
  Client,
  TypedDocumentNode,
} from '@urql/core';
import { pipe, fromValue, concat, map, makeSubject, onPush } from 'wonka';
import type { Source } from 'wonka';
import type { DocumentNode } from 'graphql';
```

This is just a small note. We reexport `TypedDocumentNode` from `@urql/core` and use it from there. If there's a case where we'll switch to the main `graphql` package's generics (if it'll add them) or we switch away from this shared package, it's useful for that to be independent of the bindings.

packages/svelte-urql/src/queryStore.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    // don't continue if paused
+    filter(isPaused => !isPaused),
+
+    // now we want to fetch the return type, so we must concatMap
+    concatMap(() =>
+      fetchProcess(
+        args.client.executeQuery<Data, Variables>(request, context),
+        baseResult,
+        () => (isComplete = true)
```
kitten, line 65: This unfortunately isn't possible 😅 
If it was then we'd be happy to just let everyone use promises in Svelte. However, the reason why we expose sources and not promises or promise-likes is that a result is never complete. A query can always receive updates. Basically, be it with a normalised cache or without, a query may receive multiple updates while it's active, i.e. while it has subscribers ([There's more details on the architecture page](https://formidable.com/open-source/urql/docs/architecture/#requests-and-operations-on-the-client))

This kind of answered another question, I had. At least, I think it does 😅 Which was: How do ensure there's no memory leak when we never unsubscribe (which is what I saw in the `subscriptionStore` file)
The unsubscribe functionality never seems to be passed through from stores to internal subscriptions :v:

kitten:
So, just to make sure I get this right and provide some overarching commentary here 🤔

We're basically looking to combine the older API we had here with a way to input the `Client` as a store, right?
https://github.com/FormidableLabs/urql/blob/7696323f2cf5f38659bea26526a83c64acb40aac/packages/svelte-urql/src/operations/query.ts

I think that raises a couple of questions we could answer before continuing:

- Would we want any input to be a store optionally, meaning, would we want to accept, say, `variables: V | Readable<V>` (similar to Vue's refs basically)
- Would we want the `client` to either be a value, a readable store, or derived from context? (to provide more options)
- I haven't read Svelte's implementation in quite a few versions, so looking at our old API for Svelte, does Svelte allow for synchronous updates on readables while only rerendering once with the latest value? I think it used to. But it's certainly important that if `fetching: true` is synchronously replaced with a (non-fetching) result from cache that we don't render or expose the loading state
- I also think the old implementation I linked wasn't actually checking that the underlying request (i.e. query + variables combination via the `key` we create) didn't change. We could try to preserve this behaviour in Svelte but that'd only apply if we accept `Readable` inputs of course 

Hope that makes sense so far and isn't too much to handle for us ✌️ I'm happy to help though!

jonathanstanley:
> We're basically looking to combine the older API we had here with a way to input the `Client` as a store, right? 

I don't think so.  My focus was to enable urql without the reliance on svelte `context` (which is more of a react-style approach and [mandatory context creates limitations](https://github.com/FormidableLabs/urql/issues/2369#issue-1185054359) in svelte). 

> * Would we want any input to be a store optionally, meaning, would we want to accept, say, `variables: V | Readable<V>` (similar to Vue's refs basically)
> * Would we want the `client` to either be a value, a readable store, or derived from context? (to provide more options)

I would likely advise against any store/readable as input. It could lead users into a constrained approach rather than just setting the logic in their own code however they like.

> * I haven't read Svelte's implementation in quite a few versions, so looking at our old API for Svelte, does Svelte allow for synchronous updates on readables while only rerendering once with the latest value? I think it used to. But it's certainly important that if `fetching: true` is synchronously replaced with a (non-fetching) result from cache that we don't render or expose the loading state

as I understand it, yes [svelte stores/readables are synchronous](https://svelte.dev/docs#component-format-script-4-prefix-stores-with-$-to-access-their-values-store-contract). it doesn't have the issue of [react's delayed state update](https://reactjs.org/docs/faq-state.html#why-is-setstate-giving-me-the-wrong-value).

packages/svelte-urql/src/common.ts
jonathanstanley: thanks!

packages/svelte-urql/src/queryStore.ts
jonathanstanley: > How do ensure there's no memory leak when we never unsubscribe

Thank you for this great point about memory leak.  i will update to ensure unsubscribe is called when the store has no more subscribers.

> A query can always receive updates

Another good point: even though graphql fetch will only return a single result, `executeQuery` could deliver a series of changes (from cache/fetch/exchange)? I will fix this.

i can set each to unsubscribe when the store is no longer in use (also to ensure no memory leak).

packages/svelte-urql/src/common.ts
jonathanstanley: thank you!

packages/svelte-urql/src/queryStore.ts
jonathanstanley: i really appreciate the feedback.

i can see instances where context could be a convenience.  i can also see why it might be unnerving to remove it (especially coming from a React perspective).  i can update it to keep context if you like, but every time i think about it with svelte, I keep coming back to the belief that it is better without context. quotes from a great article on svelte context:

> svelte context is commonly misunderstood [and people] end up reaching for it in many cases where other approaches are simpler.[instead] define a store in another Javascript file, and just import it everywhere that it’s needed.
https://imfeld.dev/writing/svelte_context

as the author points out, it is easy in svelte to use stores instead of context, so the benefit of using context is rather limited.

it might also be thought of as an [inheritance vs composition issue](https://en.wikipedia.org/wiki/Composition_over_inheritance).  svelte context behaves like inheritance and can easily become strange / messy. by removing the context approach, users are spared from finding themselves in an inheritance trap, and the urql binding is easier to learn, understand, and maintain.

if you still think this is needed, i should be able to bring it back in as you propose.  another idea is to see if we release v2 alpha and see if users want to get context back?

kitten:
> I don't think so. My focus was to enable urql without the reliance on svelte context (which is more of a react-style approach and https://github.com/FormidableLabs/urql/issues/2369#issue-1185054359 in svelte).

The point isn't the context though 😅 I actually believe that the old implementation may 1:1 be very close to what's desired here with the added change for where the `client` comes from, hence the note. Where the client comes from is a minor detail, I think, rather than a big change, implementation-wise, hence me linking it :v:

> as I understand it, yes [svelte stores/readables are synchronous](https://svelte.dev/docs#component-format-script-4-prefix-stores-with-$-to-access-their-values-store-contract). it doesn't have the issue of [react's delayed state update](https://reactjs.org/docs/faq-state.html#why-is-setstate-giving-me-the-wrong-value).

Yes, the delayed updates aren't however what I'm after. The question is what happens, say, when a store delivers to results synchronously. If Svelte updates a component twice in reaction we'd have to ensure that the initial signal from the store isn't stale state that's immediately replaced, i.e. `next({ fetching: true }); next({ fetching: false /* ... */ })` triggering two subsequent updates.

I can double check whether it does so next week or whether we're fine on that end though 🤔

packages/svelte-urql/src/index.ts
kitten: I'd prefer to keep them unexported. They really don't belong to our API surface and all other packages in `@urql/*` basically ensure that their API surface is as minimal as possible to keep them stable. This of course hasn't applied to Svelte as much as this is basically the third iteration, but seen from another perspective, `createPausable`, `fetchProcess`, `fromStore`, and `defaultBaseResult` are all not part of our public API or part of urql's logic. So they're utilities that'd not have much use outside of the core bindings.

There are two sides to this:
1. We'd like to hear from people when they run into limitations, and if they realise they'll have to copy `@urql/svelte` code and modify it, we'll hear from them and be able to adapt
2. These utilities are all not part of what we maintain and if we change behaviour we'd like to ensure that we don't have to continue maintaining them when other implementation details change

packages/svelte-urql/src/queryStore.ts
kitten: I think it's not too relevant whether we have `pause: boolean` or `isPaused: boolean`, to keep it brief. Since `fetching` and `pause` are responsibilities and owned by bindings, there's no need to keep their names stable :v: 

In fact, in the `@urql/vue` bindings we ran into the same problem and hence we adapted the names to: `isPaused, pause(), unpause()`: https://github.com/FormidableLabs/urql/blob/823d47b5a831d72b6798b4a504c5f52c9045396a/packages/vue-urql/src/useQuery.ts#L121-L136

packages/svelte-urql/src/queryStore.ts
kitten: >  define a store in another Javascript file, and just import it everywhere that it’s needed.

I can certainly agree with that and see how that saves a lot of trouble in Svelte. I'm not opposed to not having it be the default. However, I do think the point stands that it's wise to keep the `contextKey` reserved to us for future use. Hence, we could think of placing stores in that context, or otherwise create readables from the context value. It doesn't really matter, but I think we should continue to reserve that key at the very least since it's a constant literal 🤔

packages/svelte-urql/src/queryStore.ts
jonathanstanley: fixed in https://github.com/FormidableLabs/urql/pull/2370/commits/e2f837b3d98115e1b6f8932732eff23bd40ea322

packages/svelte-urql/src/queryStore.ts
jonathanstanley: resolved in https://github.com/FormidableLabs/urql/pull/2370/commits/b0a6f286abc8718755f5270f095164c888875332

packages/svelte-urql/src/index.ts
jonathanstanley: resolved in https://github.com/FormidableLabs/urql/pull/2370/commits/ac8e9ed0d4bf56d6eac252487b3d7214d7959d1c

packages/svelte-urql/src/queryStore.ts
jonathanstanley: I added an update for this.  is [my change here](packages/svelte-urql/src/context.ts) what you're recommending? (also added to exports)

kitten:
Sorry about the delay, but I didn't forget about this ✌️
is it alright if I come back to this PR this week? I really want to get this in shape; I also may have some suggestions but could also just push them to a separate branch to streamline it, if that's alright?

frederikhors:
@jonathanstanley it would be amazing to use https://github.com/omer-g/persistent-svelte-store too, what do you think? Also fo cross tabs sync.

frederikhors:
Or this: https://github.com/macfja/svelte-persistent-store.

frederikhors:
@jonathanstanley does it have tabs sync?

EskelCz:
Just a quick question if I may. Will this enable delayed queries? (wrapped in a function, triggered on interaction etc)
Demo https://stackblitz.com/edit/sveltekit-urql-issue-demo
Thanks

frederikhors:
> > Just a quick question if I may. Will this enable delayed queries? (wrapped in a function, triggered on interaction etc) Demo https://stackblitz.com/edit/sveltekit-urql-issue-demo Thanks
> 
> Yes. You would [initialize the query as paused](https://github.com/FormidableLabs/urql/blob/3a162559405d5ae798ce8ea357f7844a70d656bd/packages/svelte-urql/src/queryStore.ts#L24) and then unpause whenever you like.

I do not think @EskelCz is referring to `pause`.

EskelCz:
Sorry I don't want to derail your work here, so I've made a new thread for the issue/use case: [https://github.com/FormidableLabs/urql/issues/2449](https://github.com/FormidableLabs/urql/issues/2449)

packages/svelte-urql/src/common.ts
```diff
@@ -0,0 +1,136 @@
...
+  const urqlSource = pipe(
+    client.executeRequestOperation(baseResult.operation),
+
+    // build annotated result from base (set fetching:false)
+    map(response => ({ ...baseResult, ...response })),
+    share
+  );
+
+  // concat combines sources into a sequential source
+  const source = pipe(concat([baseSource, urqlSource]), share);
```
jonathanstanley, line 104: I'm struggling with the unsubscribe.

when a query is run, it doesn't seem to emit a completion signal.  ex: if i do something like this in `queryStore.ts`:

```ts
pipe(
fetchProcess(...)
toPromise
)
```

the promise will never resolve.  i tried looking at the react bindings, but i'm not seeing a good way to teardown when urql is finished.
Screen Shot 2022-05-18 at 9 26 32 PM

kitten: Sources from operations never complete automatically unless an exchange or other process in urql cancels them forcefully. Instead, once you subscribe, you'll indefinitely get updated results from the `Client`, which is how caching and cache updates work. Hence, `toPromise` is a tricky one. It'll wait until the _last_ result, however, there is no last result to wait for. So if you're using the raw sources you'll have to do `take(1)`.

I'll resolve this btw since it's unrelated to the current PR :v:

kitten (approved):

frederikhors:
Glad this is merged! But without docs on how to use it it is difficult for me to use it (and test on Svelte).

@jonathanstanley can you increase docs?

DoisKoh:
I can't get subscriptions to work though (Svelte/SvelteKit) - the data is being received but the Svelte store is not updated - can we have an example for subscriptions?

rogueyoshi:
Seconded on examples.

francois-blanchard:
@jonathanstanley or @kitten could you give us a `mutationStore` usage example, please?
I tried to call it when I click on a button but I've faced the `Uncaught Error: Function called outside component initialization` error.
A code example could be very useful 🙏

EskelCz:
@francois-blanchard Take a look at this: https://stackblitz.com/edit/sveltekit-urql-working-demo
Just replace the query with a mutation.

francois-blanchard:
@EskelCz thank you.
The client needs to be initialized outside the event handler.
```
const client = getContextClient();
```

francois-blanchard:
Since the `reexecute` method has been removed, what do you recommend to re-execute a query? @EskelCz

rouzwelt:
I would also really like to know what is your recommendation for reexecute() function, since it is gone, Im having a bit of hard time to update the codes.

qbunt:
@kitten is there any sort of documentation around what `operationStore` has turned into? We have an existing sveltekit project with loads of operationStore, but it'd be great to get something in the docs for existing users. Transitioning to v3 is going to be a fair bit of work it seems

frederikhors:
> @kitten is there any sort of documentation around what `operationStore` has turned into? We have an existing sveltekit project with loads of operationStore, but it'd be great to get something in the docs for existing users. Transitioning to v3 is going to be a fair bit of work it seems

Yes please. I need to switch to V3 too soon. Please.

JoviDeCroock:
What are you missing compared to the docs that are there atm 😅 https://formidable.com/open-source/urql/docs/basics/svelte/#queries `operationStore` combined with `query` all turned into one API making it easier to use etc

qbunt:
@JoviDeCroock any mention of what `operationStore` has turned into would be a great start. Docs are for communication, I landed here because I am not following the progress of this project all the time or how my larger codebase has to change to upgrade. Docs around migration from v2 just don't exist.

nmd111956-dot (commented):
Open

PR #2382, opened 2022-04-05
feat: use "application/graphql+json" content-type header

kettanaito:
## Summary

The recept [GraphQL over HTTP specification](https://github.com/graphql/graphql-over-http/blob/92b57a9179834318b6f15e1d23afc49368dd5e3c/spec/GraphQLOverHTTP.md#content-types) encourages GraphQL requests to have a `Content-Type: application/graphql+json` header. This way GraphQL requests can finally be reliably distinguished from other HTTP requests ([see the issue](https://github.com/graphql/graphql-spec/issues/938)).

## Set of changes

- `urql` now sets `Content-Type: application/graphql+json` request header on the issued requests.
- Adjusts tests accordingly.

### DIFF
--- a/docs/graphcache/schema-awareness.md
+++ b/docs/graphcache/schema-awareness.md
@@ -19,9 +19,9 @@ on `cacheExchange` allows us to pass an introspected schema to Graphcache:
 ```js
 const introspectedSchema = {
   __schema: {
-    queryType: { name: 'Query', },
-    mutationType: { name: 'Mutation', },
-    subscriptionType: { name: 'Subscription', },
+    queryType: { name: 'Query' },
+    mutationType: { name: 'Mutation' },
+    subscriptionType: { name: 'Subscription' },
   },
 };
 
@@ -97,7 +97,7 @@ import * as fs from 'fs';
 
 fetch('http://localhost:3000/graphql', {
   method: 'POST',
-  headers: { 'Content-Type': 'application/json' },
+  headers: { 'Content-Type': 'application/graphql+json' },
   body: JSON.stringify({
     variables: {},
     query: getIntrospectionQuery({ descriptions: false }),
@@ -161,7 +161,7 @@ import { getIntrospectedSchema, minifyIntrospectionQuery } from '@urql/introspec
 
 fetch('http://localhost:3000/graphql', {
   method: 'POST',
-  headers: { 'Content-Type': 'application/json' },
+  headers: { 'Content-Type': 'application/graphql+json' },
   body: JSON.stringify({
     variables: {},
     query: getIntrospectionQuery({ descriptions: false }),
--- a/exchanges/graphcache/default-storage/package.json
+++ b/exchanges/graphcache/default-storage/package.json
@@ -16,7 +16,7 @@
     "./package.json": "./package.json"
   },
   "dependencies": {
-    "@urql/core": ">=2.3.3",
+    "@urql/core": ">=2.3.6",
     "wonka": "^4.0.14"
   }
 }
--- a/exchanges/multipart-fetch/src/multipartFetchExchange.ts
+++ b/exchanges/multipart-fetch/src/multipartFetchExchange.ts
@@ -36,7 +36,10 @@ export const multipartFetchExchange: Exchange = ({
       if (files.size) {
         url = makeFetchURL(operation);
         fetchOptions = makeFetchOptions(operation);
-        if (fetchOptions.headers!['content-type'] === 'application/json') {
+        if (
+          fetchOptions.headers!['content-type'] === 'application/json' ||
+          fetchOptions.headers!['content-type'] === 'application/graphql+json'
+        ) {
           delete fetchOptions.headers!['content-type'];
         }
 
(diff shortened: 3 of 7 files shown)


kettanaito:
I will create changesets once this proposal is approved. 

I'm in no position to judge whether this is going to be a breaking change or not. Media type subsets are backwards-comptabile to their parent types, so `application/graphql+json` should be compatible with `application/json`. I leave the proper version bump to the maintainer's best judgement. 

Please let me know if you see this change as a part of URQL. I will be opening similar pull requests in other GraphQL clients to make the ecosystem better. Thank you.

docs/graphcache/schema-awareness.md
```diff
@@ -19,9 +19,9 @@ on `cacheExchange` allows us to pass an introspected schema to Graphcache:
 ```js
 const introspectedSchema = {
   __schema: {
-    queryType: { name: 'Query', },
-    mutationType: { name: 'Mutation', },
-    subscriptionType: { name: 'Subscription', },
+    queryType: { name: 'Query' },
+    mutationType: { name: 'Mutation' },
+    subscriptionType: { name: 'Subscription' },
```
kettanaito, line 24: Yet again a misconfigured formatting :(

exchanges/graphcache/default-storage/package.json
```diff
@@ -16,7 +16,7 @@
     "./package.json": "./package.json"
   },
   "dependencies": {
-    "@urql/core": ">=2.3.3",
+    "@urql/core": ">=2.3.6",
```
kettanaito, line 19: Not intentional: running `yarn` bumps this automatically.

exchanges/multipart-fetch/src/multipartFetchExchange.ts
```diff
@@ -36,7 +36,10 @@ export const multipartFetchExchange: Exchange = ({
       if (files.size) {
         url = makeFetchURL(operation);
         fetchOptions = makeFetchOptions(operation);
-        if (fetchOptions.headers!['content-type'] === 'application/json') {
+        if (
+          fetchOptions.headers!['content-type'] === 'application/json' ||
```
kettanaito, line 40: We should still keep `application/json` content type header check for compatibility.

JoviDeCroock:
However much I do agree that we need to adhere to the spec I _think_ this could be a breaking change for some server implementations 😅  we'd have to closely check whether some server implementations break on this and ensure people can easily override this so in this major version we could choose for this to be opt-in and allow folks to override the header themselves rather than defaulting to the new

kettanaito:
Sounds sensible, @JoviDeCroock. Let's outline the verification strategy to ensure existing servers can handle `application/graphql+json` requests correctly. 

1. They treat such requests as `application/json`, implying any JSON request body parsing middleware work.
2. They still treat such requests as GraphQL requests, implying they send back data/errors.

Are there any other points we have to validate?

JoviDeCroock:
That does sound like the most important points to me yes, Will do some checks later today by creating a few popular servers in codesandbox

kettanaito:
We can add the `application/graphql+json` content type support to the server implementations first (like Apollo Server and express-graphql). Then we can release this change. What do you think about this?

From the server standpoint, they remain in an ambiguous context, so it won't be a breaking change—only checking for one more potential content-type request header value. The bottom line is that if this lands, it must work with the existing servers.

kettanaito:
Let me know if you need any help, we can split the workload here, if necessary. I'm always happy to help!

JoviDeCroock:
@kettanaito oh yeah feel free to PR 😅 was just checking it out and most of these seem related to `body-parser` needing to be extended

kitten:
I think the general gist here is that we'll have to defer this change and close the PR unfortunately until the ecosystem shifts. Sending `Content-Type: application/json` is still the accepted default until all major GraphQL API frameworks adopt the GraphQL over HTTP spec fully. Until then we'll have to keep the default as it is, and since we allow this to be changed that should still cater to all users.

We simply don't have enough driving force to be the first to make this change. It's also notable that there's no added assumptions attached to `application/graphql+json` that'd block any behaviour — at least for now :v:

kettanaito:
@kitten, I understand. It's sad we cannot have a proper request differentiation for that long in GraphQL land. Let's what for the better times (tm) then.

kitten:
cc @JoviDeCroock: I haven't closed this yet because of another thing I did mention to you.

As per the spec, there's actually another note:

> [...] If there is no `Accept` header in the request, the response MUST be serialized as JSON and MUST include a  `Content-Type: application/graphql+json` header.
>
> If another content type is preferable to a client, it MAY include an `Accept` HTTP header listing other acceptable content types in order of preference. In this case a client SHOULD include `application/graphql+json` in the list, according to their preferred priority.
>
> The server MUST respect the given `Accept` header and attempt to encode the response in the first supported content type listed. [...]

This links into our multipart responses though, but that spec does _not_ list any compatibility note.
However, we can therefore assume that any GraphQL response _may_ return this content mode:
- [graphql-helix](https://github.com/contra/graphql-helix/blob/1bbc7146e953556498180c53aa1638dd25072d43/packages/core/lib/send-result/node-http.ts#L37-L42)
- [hotchocolate](https://github.com/ChilliCream/hotchocolate/blob/d412413d5f93f74e897c1253e4786f61fc59f773/src/StrawberryShake/Client/src/Transport.Http/ResponseEnumerator.cs#L60-L66)
- [express-graphql](https://github.com/robrichard/express-graphql/blob/ce8429e5c15172b394e65d5a27491611b5fb354e/src/index.ts#L463-L465)

This means that the correct `Accept` header must be:

```
Accept: application/graphql+json, application/json
```

kitten:
Superseded by #2457

kettanaito:
@kitten, thank you for your work on this! 🎉

PR #2389, opened 2022-04-11
add v-bind:key directive to basic vue docs

Cerchie:
## Summary

Following along the urql docs for Vue on vue version ^3.2.13, I expected to see my data surface to the user interface after inserting my open endpoint and updating the logic, but I ran into:

```
10:7  error  Elements in iteration expect to have 'v-bind:key' directives  vue/require-v-for-key
```

## Set of changes

I added the 'v-bind:key' directive example to the three `v-for` loops in the code examples so that users will anticipate having to provide a key. This PR only consists of changes to the copy in `/docs/basics/vue.md`. Ran `yarn test` on my branch to be sure everything else was looking good. 

I need help with the changeset. When I run `yarn changeset`, type `urql-docs` and enter, nothing happens.

### DIFF
--- a/docs/basics/vue.md
+++ b/docs/basics/vue.md
@@ -125,7 +125,7 @@ todo items. Let's dive right into it!
   </div>
   <div v-else>
     <ul v-if="data">
-      <li v-for="todo in data.todos">{{ todo.title }}</li>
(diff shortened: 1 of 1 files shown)


JoviDeCroock (approved):
Awesome, thank you!

PR #2410, opened 2022-04-20
fix(vue): prevent executeQuery from taking a fetching or prior result

### LINKED ISSUE #2409: executeQuery returns old data and ignores fetched data
For some reason when I'm executing query that was already called before in other part of app, it returns previous data instead. It behaves like cache (that I disabled by network policy in client) but it's not because fetch is actually executed in devtools network tab. I'm receiving fresh data, but `executeQuery` returns old one anyway o.O If I reexecute it again and again it always returns data from previous one.

In real world scenario I'm fetching user data on init. If api does not return data (token expired) user need's to log in. After that I'm fetching user again on same query by using `executeQuery` again. Now API returns correct user, but urql still sees that response as null even that it's not. I'm debugging this now by returning random number in endpoint.

I noticed also:
- In real app I have this query extracted in composable, but it does not matter. Same behavior if I just duplicate queries in two components.
- If I change ANYTHING in one of queries like field or name - it works fine.

urql version & exchanges: @urql/core@2.4.4 @urql/vue@0.6.1

**Steps to reproduce**

1. App.vue

```
<template>
  <router-view />
</template>

<script setup lang="ts">
import { gql, useQuery } from '@urql/vue'

useQuery({
  query: gql`
    query GetLoggedUser {
      me {
        id
        randomNumber
      }
    }
  `,
}).then(({ data }) => {
  console.log(data.value.me)
})
</script>
```
2. SomePage.vue

```
<template>
  <button @click="demo">
    Demo
  </button>
</template>

<script setup lang="ts">
import { gql, useQuery } from '@urql/vue'

const getLoggedUserQuery = useQuery({
  query: gql`
    query GetLoggedUser {
      me {
        id
        randomNumber
      }
    }
  `,
  pause: true,
})

const demo = () => {
  getLoggedUserQuery.executeQuery()
    .then(({ data }) => {
      console.log(data.value.me)
    })
}
</script>
```

**Expected behavior**

In console I should have two random numbers fetched from API.

**Actual behavior**

In console I have two same numbers (from first execution of query).

Screenshot from 2022-04-19 21-39-55

Screenshot from 2022-04-19 21-40-37

JoviDeCroock:
Fixes #2409 

## Summary

Due to the promise already being executed we would be returning old data to the user, this was enforced by us unsubscribing immediately in `take(1)`. Now we wait for the first non fetching and stale result.

## Set of changes

- wait for the first non-stale result

### DIFF
--- a/packages/vue-urql/src/useQuery.ts
+++ b/packages/vue-urql/src/useQuery.ts
@@ -128,7 +128,6 @@ export function callUseQuery<T = any, V = object>(
 
       return {
         ...response,
-        // @ts-ignore
         then(onFulfilled, onRejected) {
           const promise = new Promise(resolve => {
             const sub = pipe(
@@ -145,10 +144,10 @@ export function callUseQuery<T = any, V = object>(
               })
             );
           });
-          return promise.then(res => {
-            // @ts-ignore
-            onFulfilled(res);
-          }, onRejected);
+          return promise.then(
+            x => onFulfilled!(x as UseQueryState<T, V>),
+            onRejected
+          );
         },
       };
     },
@@ -199,7 +198,6 @@ export function callUseQuery<T = any, V = object>(
 
   const response: UseQueryResponse<T, V> = {
     ...state,
-    // @ts-ignore
     then(onFulfilled, onRejected) {
       const promise = new Promise(resolve => {
         if (!source.value) return resolve(state);
@@ -217,10 +215,11 @@ export function callUseQuery<T = any, V = object>(
           })
         );
       });
-      return promise.then(res => {
-        // @ts-ignore
-        onFulfilled(res);
-      }, onRejected);
+
+      return promise.then(
+        x => onFulfilled!(x as UseQueryState<T, V>),
+        onRejected
+      );
     },
   };
 
--- a/packages/vue-urql/src/useQuery.ts
+++ b/packages/vue-urql/src/useQuery.ts
@@ -129,25 +129,24 @@ export function callUseQuery<T = any, V = object>(
       return {
         ...response,
         then(onFulfilled, onRejected) {
-          const promise = new Promise(resolve => {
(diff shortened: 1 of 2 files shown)


packages/vue-urql/src/useQuery.ts
```diff
@@ -120,13 +120,36 @@ export function callUseQuery<T = any, V = object>(
...
+                  if (sub && sub.unsubscribe) sub.unsubscribe();
+                  resolve(result);
+                }
+              })
+            );
+          });
+          return promise.then(
+            x => onFulfilled!(x as UseQueryState<T, V>),
+            onRejected
+          );
```
kitten, line 150:
```suggestion
          return new Promise<UseQueryState<T, V>>(resolve => {
            let hasResult = false;
            const sub = pipe(
              s,
              map(() => state),
              subscribe(result => {
                if (
                  !result.fetching.value &&
                  !result.stale.value
                ) {
                  if (sub) sub.unsubscribe();
                  hasResult = true;
                  resolve(result);
                }
              })
            );
            if (hasResult) sub.unsubscribe();
          }).then(onFulfilled, onRejected);
```

packages/vue-urql/src/useQuery.ts
```diff
@@ -120,13 +120,35 @@ export function callUseQuery<T = any, V = object>(
...
+              subscribe(result => {
+                if (
+                  !result.fetching.value &&
+                  !result.stale.value
+                ) {
+                  if (sub) sub.unsubscribe();
+                  hasResult = true;
+                  resolve(result);
+                }
+              })
```
kitten, line 146:
```suggestion
              subscribe(()=> {
                if (
                  !state.fetching.value &&
                  !state.stale.value
                ) {
                  if (sub) sub.unsubscribe();
                  hasResult = true;
                  resolve(state);
                }
              })
```

kitten (approved):

PR #2419, opened 2022-04-23
fix(graphcache): Shift layers with future updates and improve layer behaviour

kitten:
## Summary

This pull request introduces several small tweaks to Graphcache's data layering.

Usually we reserve layers per query operation, create optimistic layers for optimistic mutations, and write to these layers as responses come in in the order they've been reserved in. If they haven't been reserved they'll be created on demand. This means that layers are essentially just queues and we `unshift` to this queue whenever a layer is created. Queueing in order isn't always desireable however.

As we've found out, when a layer expects future updates (`hasNext: true`, as it's been added with deferred results, via `@defer`) it's added to the queue and multiple queries that are started while this layer is still waiting for updates will layer on top of it. We've deemed this acceptable. After all, it's very unlikely that any query will go on to conflict with another query that's been deferred, since this would imply a deferred vs non-deferred overlap. However, we knew that this behaviour could eventually still cause trouble due to subscriptions (which theoretically may never stop updating) and the updates to queries they may trigger.

A subscription may cause updates at any time, and could hence cause refetches for queries at any time. This can cause problems, because when two subsequent subscription events cause a query to reexecute each time concurrently, then these the queued up query is in a layer lower than the subscription and hence its result has no chance to ever update the cache, if its data overlaps with the subscription's update.

It turns out that there's a very elegant solution for the above problems. **Layer Shifting**.

Whenever a subscription or query result with `hasNext: true` comes in, we proactively call `reserveLayer` with this truthy flag. This now causes the layer to be shifted ahead in the queue until it's past all layers that are still empty. This means it _loses_ precedence in the layering, i.e. becomes a lower layer. This is desireable however in this case.
Any layer that expects update should be the highest layer with data. However, it should also be the lower than all empty layers, so that layers for which we're still awaiting data can still override its update.

```
previously: [ x, 0, 1, 0 ]
new behaviour: [ 0, x, 1, 0 ]
```

The above is an illustration of this. In this case, our layers are ordered from latest & highest precedence (left) to oldest & lowest precedence (right). In this example `0` stands for empty layers (still fetching), while `1` stands for written layers (not fetching). The `x` stands for a layer that will be written using `hasNext: true`.
Instead of us adding this `x` layer to a place where it'll always override the first `0` layer, it's instead shifted behind it.

## Set of changes

- Treat subscription results as if they carried `hasNext: true`
- Ensure `query` (i.e. `read`) operations don't cause layers to be created
- Shift `hasNext: true` layers behind currently empty layers
- Avoid adding a layer when only one layer has been reserved (performance optimisation)

### DIFF
--- a/.changeset/pink-turkeys-live.md
+++ b/.changeset/pink-turkeys-live.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-graphcache': patch
+---
+
+Fix issues with continuously updating operations (i.e. subscriptions and `hasNext: true` queries) by shifting their layers in Graphcache behind those that are still awaiting a result. This causes continuous updates to not overwrite one-off query results while still keeping continuously updating operations at the highest possible layer.
--- a/.changeset/pink-turkeys-live.md
+++ b/.changeset/pink-turkeys-live.md
@@ -1,5 +1,5 @@
 ---
-'@urql/exchange-graphcache': patch
+'@urql/exchange-graphcache': minor
 ---
 
 Fix issues with continuously updating operations (i.e. subscriptions and `hasNext: true` queries) by shifting their layers in Graphcache behind those that are still awaiting a result. This causes continuous updates to not overwrite one-off query results while still keeping continuously updating operations at the highest possible layer.
--- a/exchanges/graphcache/src/store/data.ts
+++ b/exchanges/graphcache/src/store/data.ts
@@ -111,7 +111,7 @@ export const initDataState = (
     // We don't create new layers for read operations and instead simply
     // apply the currently available layer, if any
     currentOptimisticKey = layerKey;
-  } else if (isOptimistic || data.optimisticOrder.length > 0) {
+  } else if (isOptimistic || data.optimisticOrder.length > 1) {
     // If this operation isn't optimistic and we see it for the first time,
     // then it must've been optimistic in the past, so we can proactively
     // clear the optimistic data before writing
@@ -131,6 +131,8 @@ export const initDataState = (
   } else {
     // Otherwise we don't create an optimistic layer and clear the
     // operation's one if it already exists
+    // We also do this when only one layer exists to avoid having to squash
+    // any layers at the end of writing this layer
     currentOptimisticKey = null;
     deleteLayer(data, layerKey);
   }
--- a/exchanges/graphcache/src/store/data.ts
+++ b/exchanges/graphcache/src/store/data.ts
@@ -513,27 +513,37 @@ export const reserveLayer = (
   layerKey: number,
   hasNext?: boolean
 ) => {
-  const index = data.optimisticOrder.indexOf(layerKey);
-  if (index === -1) {
-    // The new layer needs to be reserved in front of all other commutative
-    // keys but after all non-commutative keys (which are added by `forceUpdate`)
-    data.optimisticOrder.unshift(layerKey);
-  } else if (!data.commutativeKeys.has(layerKey)) {
-    // Protect optimistic layers from being turned into non-optimistic layers
-    // while preserving optimistic data
-    clearLayer(data, layerKey);
-    // If the layer was an optimistic layer prior to this call, it'll be converted
-    // to a new non-optimistic layer and shifted ahead
-    data.optimisticOrder.splice(index, 1);
-    data.optimisticOrder.unshift(layerKey);
-  }
-
   if (hasNext) {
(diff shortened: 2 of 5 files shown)


.changeset/pink-turkeys-live.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/exchange-graphcache': patch
+---
+
+Fix issues with continuously updating operations (i.e. subscriptions and `hasNext: true` queries) by shifting their layers in Graphcache behind those that are still awaiting a result. This causes continuous updates to not overwrite one-off query results while still keeping continuously updating operations at the highest possible layer.
```
kitten, line 5: I'm a little unsure of whether to mark this as a fix. Technically, I'd say this is a feature, since previously we knew the lack of this could cause issues but underestimated the nature of concurrent subscription updates.

frederikhors:
How can I try this?

exchanges/graphcache/src/store/data.ts
```diff
@@ -509,27 +515,37 @@ export const reserveLayer = (
...
   if (hasNext) {
     data.deferredKeys.add(layerKey);
   } else {
     data.deferredKeys.delete(layerKey);
   }
 
+  let index = data.optimisticOrder.indexOf(layerKey);
+  if (index > -1) {
+    if (hasNext || !data.commutativeKeys.has(layerKey)) {
+      data.optimisticOrder.splice(index, 1);
```
kitten, line 527: It's worth noting that this can cause issues. When a deferred query reorders itself in the layers it may "lift up" old state.
I'd still say this is preferable since overlapping queries are generally discouraged.
However, it'd be cool if we figured out how to implement "sparse layers", i.e. not writing anything in layers that's identical to the lower layers. This may however introduce a significant slowdown to our writes. I haven't measured its impact yet however.

kitten:
@frederikhors I tried it on your sandbox already and it does fix your particular problem of concurrent subscription and query updates. However, it'll only be released after a review.

frederikhors:
Ok, but usually you have beta package, where can I found it for this?

frederikhors:
I found it. It works amazingly well.

@kitten I don't know how to tell you, how to represent HOW MUCH I LOVE YOU. 😄😄😄

❤️❤️❤️

.changeset/pink-turkeys-live.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/exchange-graphcache': patch
```
JoviDeCroock, line 2:
```suggestion
'@urql/exchange-graphcache': minor
```

I'd opt for this as a feature as technically this isn't a bug. We have seen this request once before

exchanges/graphcache/src/store/data.ts
JoviDeCroock: Hmm, I wonder if this could be based on selectionSets

JoviDeCroock (approved):

tgriesser:
This change seems to have caused issues with how we're expecting graphcache to work with subscriptions. Still trying to debug exactly what's happening & fully wrap my head around the change here.

Since the query is what ultimately triggers the subscription, we actually want the subscription to be the source of truth and be applied after the query, as it always has the latest/most relevant data.

Was wondering if you might be able to better explain the situations you'd anticipate this would cause issues, since it seems to have broken a lot of the assumptions we had around how this should work.

frederikhors:
@tgriesser can you create a reproduction? I did a reproduction for a problem I had that is fixed now with this PR. Before this PR the update order of layers was random. Sorry if I explained it so badly, but this is really a needed improvement.

kitten:
@frederikhors No, the order of layers was not random before this change. Instead, subscriptions created a layer on top every time while queries were on reserved layers in order of when they were triggered. Now, any realtime events skip ahead (i.e. are lower in priority) past any pending queries. This can indeed cause timing issues when one transport is slower than another, i.e. when the websockets are faster and simultaneous to a query.
However, typically I wouldn't expect this to cause an issue since it's very unlikely for a query to be completing or updating with _new_ data while a subscription is updating with a new event. In other words, what's the missing piece here is why the order would be so sensitive and unstable when the query isn't realtime 🤔 

But indeed, I'd love to hear about this in a separate thread, maybe in a discussion to have it kept separate :v:

tgriesser:
Opened as https://github.com/FormidableLabs/urql/discussions/2440, sorry for the long read!

PR #2431, opened 2022-05-03
fix: ignore empty relay pagination pages

### LINKED ISSUE #2430: RFC: preserving startCursor / endCursor in cases of empty pagination results
## Summary

When using relay style cursor pagination to fetch the latest result

```
connectionField(first: 10)
```

We get a result that looks like this:

```
pageInfo: { startCursor: 'ABC', endCursor: 'DEF' }
nodes: [{ ... }, { ... }, { ... }, ...]
```

If we want to poll for items earlier newer than `startCursor`, we might want to query for something like:

```
connectionField(last: 10, before $startCursor)
```

The issue is that when there is an empty result set: 

```
pageInfo: { startCursor: null, endCursor: null }
nodes: []
```

The `relayPagination` will then merge in & return the `null` cursor values for this latest result set. Ideally we'd like to keep the latest known `startCursor` so we can continue to re-issue the query for items before the `startCursor`.

## Proposed Solution

Add a new option, maybe named something like `skipEmpty` or `excludeEmpty`, which would skip over results that don't have any nodes/edges while [looping through the pages](https://github.com/FormidableLabs/urql/blob/e1278c91b8bb601e57835179c202ac4526c6f1bf/exchanges/graphcache/src/extras/relayPagination.ts#L207-L217) in the `relayPagination` helper, so as to not clobber the cursors for the existing paginated list.

```ts
if (params.excludeEmpty && (page.edges.length === 0 && page.nodes.length === 0 && typename)) {
  continue
}
```

tgriesser:
Resolves #2430 

## Summary

Ignores empty pages in the `relayPagination` helper when there is at least one result, so as to preserve the start & end cursor when an empty result is returned with null cursor values.

## Set of changes

- Ignore empty pages when we already have at least one page in the cursor
- Adds test for the intended behavior

### DIFF
--- a/.changeset/fluffy-toys-live.md
+++ b/.changeset/fluffy-toys-live.md
@@ -1,0 +1,5 @@
+---
+"@urql/exchange-graphcache": patch
+---
+
+Fix ignore empty relay edges when using the `relayPagination` resolver
--- a/exchanges/graphcache/src/extras/relayPagination.test.ts
+++ b/exchanges/graphcache/src/extras/relayPagination.test.ts
@@ -1321,7 +1321,7 @@ it('handles subsequent queries with larger first values', () => {
 });
 
 it('ignores empty pages when paginating', () => {
-  const PaginatingForward = gql`
+  const PaginationForward = gql`
     query($first: Int!, $after: String) {
       __typename
       items(first: $first, after: $after) {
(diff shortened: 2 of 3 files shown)


JoviDeCroock (approved):
Awesome!

kitten (approved):

PR #2435, opened 2022-05-08
fix(execute-exchange): prune undefined variables

fathyb:
## Summary

This allows for directives to correctly validate with variables using default values.
```graphql
query($count: Boolean = false) {
  users {
    # Fails if $count is not provided:
    # Argument "if" of non-null type "Boolean!" must not be null.
    count @include(if: $count)
    list { id name }
  }
}
```

## Set of changes

- `exchanges/execute` (`@urql/execute-exchange`): 
	- `src/execute.ts`/`executeExchange()`: filter `undefined` variables before passing them to `graphql.execute()`

### DIFF
--- a/exchanges/execute/src/execute.ts
+++ b/exchanges/execute/src/execute.ts
@@ -142,16 +142,15 @@ export const executeExchange = ({
         const contextValue =
           typeof context === 'function' ? context(operation) : context;
 
-        const variableValues = operation.variables
-          ? { ...operation.variables }
-          : operation.variables;
-
-        if (variableValues) {
-          Object.keys(variableValues).forEach(key => {
-            if (typeof variableValues[key] === 'undefined') {
-              delete variableValues[key];
+        // Filter undefined values from variables before calling execute()
+        // to support default values within directives.
+        const variableValues = Object.create(null);
+        if (operation.variables) {
+          for (const key in operation.variables) {
+            if (operation.variables[key] !== undefined) {
+              variableValues[key] = operation.variables[key];
             }
-          });
+          }
         }
 
         return pipe(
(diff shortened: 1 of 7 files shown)


exchanges/execute/src/execute.ts
```diff
@@ -141,13 +141,26 @@ export const executeExchange = ({
...
+          ? { ...operation.variables }
+          : operation.variables;
+
+        if (variableValues) {
+          Object.keys(variableValues).forEach(key => {
+            if (typeof variableValues[key] === 'undefined') {
+              delete variableValues[key];
+            }
+          });
+        }
```
kitten, line 155: Can you provide a test for this if it's not too much trouble that demonstrates this issue?
Also, on a side note we can make this fix a lot less wasteful by doing:

```suggestion
        const variableValues = Object.create(null)
        if (operation.variables) {
          for (const key in operation.variables) {
            if (operation.variables[key] !== undefined)
              variableValues[key] = operation.variables[key];
          });
        }
```

Generally, this is the style we prefer to avoid unnecessary closures, so we make it a habit throughout the codebase :v:

fathyb: Thank you for the feedback! I've pushed a commit addressing both points.

Few notes:
- I needed a new query fixture, I wasn't sure what was more appropriate: update the existing ones, or create a new one. I went with updating an existing one, which required updating a few snapshots.
- Ideally, we should call `graphql.execute` with an executable schema linked to mocked resolvers. It looked like it would've required a considerable amount of code, so I went with simple assertions in order to keep the PR lean.

JoviDeCroock (approved):
Nice work

kitten (approved):
Thank you so much ❤️🙌

PR #2485, opened 2022-06-08
fix(build): Fix Node.js ESM/cjs-module-lexer compatibility

### LINKED ISSUE #2484: (react-urql): Import error in Node.js ESM mode
urql version: `2.2.1`.

**Steps to reproduce**

See https://github.com/brillout/urql-vite3.

**Expected behavior**

No import error.

**Actual behavior**

Import error. Note that `useQuery` can be imported while `gql` cannot. It seems to be a problem on `urql`'s side.

kitten:
Resolve #2484

## Summary

This fixes Node.js ESM compatibility of `urql`'s `@urql/core` re-exports when the CommonJS module is used in ESM mode.

In Node.js' ESM mode, any CommonJS module will be analyzed with `cjs-module-lexer`, which then ensures that the module actually contains the exports that an ESM module requests. When a re-export is used, a heuristic is applied by the module which is prone to failure. Prior to this change, `urql`'s reexport of `@urql/core` wasn't detected correctly.

I changed Rollup's CommonJS output until this was resolved and added a plugin to ensure future compliance.

## Set of changes

- Upgrade Rollup
- Adjust Rollup's CommonJS output
- Add plugin to check Rollup output with `cjs-module-lexer`


scripts/rollup/cjs-check-plugin.js
```diff
@@ -0,0 +1,37 @@
+import { parse } from 'cjs-module-lexer';
```
JoviDeCroock, line 1: Loving the check-plugin!

PR #2504, opened 2022-06-15
feat: Remove IE11 support

kitten:
## Summary

This PR aims to comprehensively deprecate the family of `urql` packages' support of Internet Explorer 11, which has now reached its end of life officially.

Our main focus is that no upgrade is required. We're currently at a stable state and we do not expect that users who need to continue supporting IE11 for a little while longer will need to upgrade immediately.

As part of this change we want to remove code that was added or written in a such a way to enable IE11 support, for compacter alternatives, update our build scripts and tooling, and perform a general cleanup.

## Set of changes

- Update `buble` transpilation and Terser output config
- Remove unused babel transformations
- Fix `cjs-module-lexer` plugin and improve output
- Remove IE11 code from `@urql/exchange-persisted-fetch`


JoviDeCroock (approved):

ryan-roemer (approved):
👋

kitten:
cc @JoviDeCroock: As discussed, this now uses `wonka@^6.0.0` which is so far 100% backwards compatible but ported to TypeScript to make external contributions easier and further reduce its bundle size impact.

PR #2534, opened 2022-07-07
feat: allow providing a custom isSubscriptionOperation function

n1ru4l:
## Summary

I have a live query implementation that can use the existing `subscriptionExchange`, but it would require customizing the `isSubscriptionOperation` function to the following:

```ts
import { isLiveQueryOperationDefinitionNode } from "@n1ru4l/graphql-live-query"

const isSubscriptionOperation = ({ query, variables }) => {
    const definition = getMainDefinition(query);

    const isSubscription =  definition.kind === 'OperationDefinition' &&  definition.operation === 'subscription'
    const isLiveQuery =  isLiveQueryOperationDefinitionNode(definition, variables)
    return isSubscription || isLiveQuery
  }
```

However, currently, this is not possible, requiring me to copy the subscription exchange code. Other solutions like relay and apollo easily allow customizing this.

## Set of changes

Adds a new parameter to `subscriptionExchange`, that allows providing a custom `isSubscriptionOperation` implementation.

### DIFF
--- a/packages/core/src/exchanges/subscription.ts
+++ b/packages/core/src/exchanges/subscription.ts
@@ -58,20 +58,10 @@ export interface SubscriptionExchangeOpts {
   isSubscriptionOperation?: (operation: Operation) => boolean;
 }
 
-const makeIsSubscriptionOperation = (enableAllOperations?: boolean) => (
-  operation: Operation
-): boolean => {
-  const { kind } = operation;
-  return (
-    kind === 'subscription' ||
-    (!!enableAllOperations && (kind === 'query' || kind === 'mutation'))
-  );
-};
-
 export const subscriptionExchange = ({
   forwardSubscription,
   enableAllOperations,
-  isSubscriptionOperation = makeIsSubscriptionOperation(enableAllOperations),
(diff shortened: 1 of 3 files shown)


packages/core/src/exchanges/subscription.ts
```diff
@@ -104,14 +116,6 @@ export const subscriptionExchange = ({
     });
   };
 
-  const isSubscriptionOperation = (operation: Operation): boolean => {
```
JoviDeCroock, line 107: NIT: We can save a few bytes by leaving this in and doing `const isSubscriptionOperation = isSubscriptionOperation 
|| (operation: Operation): boolean => {`

n1ru4l: I hope [9fc6ede](https://github.com/FormidableLabs/urql/pull/2534/commits/9fc6eded11d38afc32ee6e1147f550dd36b77bcf) is what you meant 😁

Alternatively, I also considered using default parameter assignment, but I assumed you don't wanna use this as it needs to be transpiled for IE? https://caniuse.com/?search=default%20parameters

PR #2537, opened 2022-07-08
fix(next-urql): avoid a redundant copy of the urql-client

### LINKED ISSUE #2524: (next-urql) - Different pages reuse the same URQL Client instance, even if they're not intended to.
urql version & exchanges:
- "@urql/core" "^2.4.4"
- "next-urql": "^3.3.2"

**Steps to reproduce**

1. Create two pages with two intended different resulting ClientOptions
2. Navigate from one to another 

**Expected behavior**
A different instance of URQL Client on the new page is created. 

**Actual behavior**
Due to this [bit of code](https://github.com/FormidableLabs/urql/blob/413a1c1ff66afb3ac168c5d7a8ca60542a19c703/packages/next-urql/src/init-urql-client.ts#L15-L17) the client-side instance gets arbitrarily reused. 

Even though there's [resetClient](https://github.com/FormidableLabs/urql/blob/413a1c1ff66afb3ac168c5d7a8ca60542a19c703/packages/next-urql/src/init-urql-client.ts#L5-L7), it gets exposed only [after react rendered](https://github.com/FormidableLabs/urql/blob/413a1c1ff66afb3ac168c5d7a8ca60542a19c703/packages/next-urql/src/with-urql-client.ts#L91) and after the render' requests are done, from `getInitialProps` there's no way to control whether we want or not to reuse the instance, even if we return a new `urqlClient`, because of [this](https://github.com/FormidableLabs/urql/blob/413a1c1ff66afb3ac168c5d7a8ca60542a19c703/packages/next-urql/src/with-urql-client.ts#L55-L59).

JoviDeCroock:
Resolves #2524

## Summary

In `next-urql` we have three flows, two of which are on the server and one of which is on the client.

### Server

First and foremost we have the classic `getInitialProps`, the automated way of handling data, this is the one that runs two passes on the server with `react-ssr-prepass`. It's our way of passing the `urql-client` from `getInitialProps` to be used for `renderToString`, code located [here](https://github.com/FormidableLabs/urql/blob/main/packages/next-urql/src/with-urql-client.ts#L118-L179) this first pass builds up the cache so we can use it during the `renderToString` pass. After the state is JSON.stringified and the client is ignored thanks to a `toJSON: () => null`.

Secondly we have the `getXProps` methods, [which we documented is a self-serve solution](https://formidable.com/open-source/urql/docs/advanced/server-side-rendering/#using-getstaticprops-or-getserversideprops) as we can't easily intervene in these functions.

Both of the above don't rely on the global variable set in `initUrqlClient` as we always [reset it](https://github.com/FormidableLabs/urql/blob/main/packages/next-urql/src/init-urql-client.ts#L17) when reaching that code-path.

### Client

This is where stuff gets a bit more complicated and I can understand the reason why we went with the approach of sharing a client but reevaluating it, we should get rid off it so people won't fall into this as the issue outlines, we aren't offering a choice. I would advise people that want this behavior to wrap `_app` in `withUrqlClient` and people who don't want this behavior to do it on a page-basis.

## Set of changes

- remove global client var

### DIFF
--- a/.changeset/long-tomatoes-joke.md
+++ b/.changeset/long-tomatoes-joke.md
@@ -1,0 +1,5 @@
+---
+'next-urql': patch
+---
+
+Stop keeping a global copy of the client in our `init-urql-client` method
--- a/.changeset/long-tomatoes-joke.md
+++ b/.changeset/long-tomatoes-joke.md
@@ -1,5 +1,6 @@
 ---
-'next-urql': patch
+'next-urql': major
 ---
 
-Stop keeping a global copy of the client in our `init-urql-client` method
+Rmove global client var from our `init-urql-client` method, this impacts applications that rely on
+the client being shared across wrapped `pages/` a migration path is to wrap `_app` instead.
--- a/packages/next-urql/src/init-urql-client.ts
+++ b/packages/next-urql/src/init-urql-client.ts
@@ -1,15 +1,10 @@
 import { createClient, Client, ClientOptions } from 'urql';
 
-let urqlClient: Client | null = null;
-
-export function resetClient() {
-  urqlClient = null;
-}
-
 export function initUrqlClient(
   clientOptions: ClientOptions,
   canEnableSuspense: boolean
 ): Client | null {
+  let urqlClient: Client | null = null;
   // Create a new Client for every server-side rendered request.
   // This ensures we reset the state for each rendered page.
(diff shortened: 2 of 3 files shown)


JoviDeCroock:
As @kitten pointed out in #2524 it's intended behavior, we can reopen this when we want to do a major and not support this anymore.

PR #2538, opened 2022-07-08
fix(graphcache-relayPagination): always set prev and next page in relaypagination

### LINKED ISSUE #2521: Relay pagination omits pageInfo fields
For some reason, the `relayPagination` implementation doesn't copy the response `pageInfo` but instead picks only specific fields. 

This behavior doesn't seem correct as it does not conform to the [Relay spec](https://relay.dev/graphql/connections.htm#sec-undefined.PageInfo.Fields). The spec says that the server MAY return `hasPreviousPage: true` for the forward pagination and `hasNextPage: true` for the backward one. However, the urql implementation assumes that `hasPreviousPage` for the forward pagination and `hasNextPage,` for the backward one, respectively, response fields can be ignored.

I'm happy to contribute a fix for this.

urql version & exchanges:

- urql@^2.2.2
- @urql/exchange-graphcache@^4.4.3

**Expected behavior**

`relayPagination()` returns all `pageInfo` fields regardless of to query variables.

**Actual behavior**

`relayPagination()` returns only:

-  `endCursor` and `hasNextPage` when query vars have `after`
-  `startCursor` and `hasPreviousPage` when query vars have `before`

JoviDeCroock:
Resolves #2521 

## Summary

We want our cursors to be defined in both directions as we are delivering a result so we can enable our user to paginate in both directions

## Set of changes

- add endCursor/startCursor and hasPrevious/hasNextPage to after/before paths

### DIFF
--- a/.changeset/gentle-pens-listen.md
+++ b/.changeset/gentle-pens-listen.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-graphcache': patch
+---
+
+Always set previous and next page in relayPagination
--- a/exchanges/graphcache/src/extras/relayPagination.test.ts
+++ b/exchanges/graphcache/src/extras/relayPagination.test.ts
@@ -1478,3 +1478,85 @@ it('allows for an empty page when this is the only result', () => {
   expect(res.partial).toBe(false);
   expect(res.data).toEqual(pageOne);
 });
+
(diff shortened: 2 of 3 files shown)


BahaaZidan:
Can I ask why was this closed ? I'm running into this issue and would love for this to move forward.

PR #2540, opened 2022-07-11
docs: clear up section about Next APIs

oreqizer:
## Summary

I noticed docs about SSR with `Next` are a little bit outdated, and missing
some potentially juicy approaches using direct schema execution.

## Set of changes

- adjust docs to assume `{ ssr: false}` is the default
- add an explicit section for `{ ssr: true }`
- move note about `_app` static optimisation to section with `{ ssr: true }`
- add section about server-only code and direct schema execution when using `getServerSideProps` and `getStaticProps`


JoviDeCroock (approved):
Awesome, thank you so much, sorry that it took so long to review this

JoviDeCroock:
Might need to do another push for the CI to run

PR #2551, opened 2022-07-21
chore(build): remove modular graphql

### LINKED ISSUE #2539: @urql/exchange-execute issues when using with 📦s that import `graphql` from root
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `@urql/exchange-execute@1.2.3` for the project I'm working on.

I have a project that uses `@graphql-tools/schema` to build schema. The project imports `graphql` from the root, as:

```ts
import { GraphQLEnumType /* ... */ } from 'graphql';
```

Version `v16.x` of `graphql`, however, has no `exports` field in their `pancake.json`, only `module` and `main`:

```json
{
  // ...
  "main": "index",
  "module": "index.mjs",
  // ...
}
```

This does not work for packages without `"type": "module"`, such as mine. Now this creates problems, because non-module apps doing root imports from `graphql` then import CJS modules implicitly. However, the `@urql/exchange-execute` package does explicit `.mjs` imports.

This causes two distinct versions of the GraphQL executor to be imported, causing an error:

```
 'Error: Cannot use GraphQLSchema 
 ...
 from another module or realm.\n' +
        '\n' +
        'Ensure that there is only one instance of "graphql" in the node_modules\n' +
        'directory. If different versions of "graphql" are the dependencies of other\n' +
        'relied on modules, use "resolutions" to ensure only one version is installed.\n' +
        '\n' +
        'https://yarnpkg.com/en/docs/selective-version-resolutions\n' +
        '\n' +
        'Duplicate "graphql" modules cannot be used at the same time since different\n' +
        'versions may have different capabilities and behavior. The data from one\n' +
        'version used in the function from another could produce confusing and\n' +
        'spurious results.'
    }

```

This is an issue on `graphql` package's side by not having an `"exports"` field in their `pancake.json`.

While I know this is not an issue in your package, I just want to post it here so y'all are aware of it and maybe note it as a temporary workaround.

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/@urql/exchange-execute/dist/urql-exchange-execute.mjs b/node_modules/@urql/exchange-execute/dist/urql-exchange-execute.mjs
index 240b63a..2705e00 100644
--- a/node_modules/@urql/exchange-execute/dist/urql-exchange-execute.mjs
+++ b/node_modules/@urql/exchange-execute/dist/urql-exchange-execute.mjs
@@ -1,8 +1,8 @@
 import { share as e, mergeMap as r, filter as n, takeUntil as t, make as o, merge as i } from "wonka";
 
-import { subscribe as u } from "graphql/subscription/index.mjs";
+import { subscribe as u } from "graphql/subscription/index.js";
 
-import { execute as a } from "graphql/execution/execute.mjs";
+import { execute as a } from "graphql/execution/execute.js";
 
 import { getOperationName as c, makeResult as f, makeErrorResult as l, mergeResultPatch as v } from "@urql/core";
 
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

JoviDeCroock:
Resolves #2539

## Summary

It seems like importing from `graphql` makes us run into issues, especially in node-environments as imports from the root and imports from the specific packages make us run into the infamous multiple GraphQL instances error. Until the main package moves to full ESM compat with export maps, ... this will be the best fix to support previous versions of GraphQL and current ones

## Set of changes

- remove the modular babel plugin

### DIFF
--- a/scripts/rollup/cleanup-plugin.js
+++ b/scripts/rollup/cleanup-plugin.js
@@ -1,6 +1,5 @@
 import { transformSync as transform } from '@babel/core';
 import { createFilter } from '@rollup/pluginutils';
-import babelPluginModularGraphQL from 'babel-plugin-modular-graphql';
 
 function removeEmptyImports({ types: t }) {
   return {
@@ -37,10 +36,7 @@ function cleanup(opts = {}) {
       }
 
       return transform(code, {
-        plugins: [
-          !opts.maintainImports && [babelPluginModularGraphQL, { extension: opts.extension }],
-          removeEmptyImports
-        ].filter(Boolean),
+        plugins: [removeEmptyImports],
         babelrc: false
       });
     }
--- a/.changeset/beige-chefs-add.md
+++ b/.changeset/beige-chefs-add.md
@@ -1,0 +1,10 @@
+---
+'@urql/exchange-auth': minor
+'@urql/exchange-execute': minor
+'@urql/exchange-graphcache': minor
+'@urql/exchange-persisted-fetch': minor
+'@urql/exchange-populate': minor
+'@urql/core': minor
+---
+
+Remove the `babel-plugin-modular-graphql` helper, this because the graphql package hasn't converted to ESM yet which gives issues in node environments
--- a/package.json
+++ b/package.json
@@ -80,7 +80,6 @@
     "@types/jest": "^26.0.23",
     "@typescript-eslint/eslint-plugin": "^4.22.0",
     "@typescript-eslint/parser": "^4.22.0",
-    "babel-plugin-modular-graphql": "^1.1.0",
     "babel-plugin-transform-async-to-promises": "^0.8.15",
     "cjs-module-lexer": "^1.2.2",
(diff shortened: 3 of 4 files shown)


scripts/rollup/cleanup-plugin.js
```diff
@@ -37,10 +36,7 @@ function cleanup(opts = {}) {
       }
 
       return transform(code, {
-        plugins: [
-          !opts.maintainImports && [babelPluginModularGraphQL, { extension: opts.extension }],
-          removeEmptyImports
-        ].filter(Boolean),
+        plugins: [removeEmptyImports],
```
kitten, line 39: Should we still keep the plugin around for CommonJS bundles?

JoviDeCroock: Yes, rollup code splitting still seems to have some empty stuff :(

kitten (approved):

PR #2570, opened 2022-07-28
fix: typo where there was a conditional else block that would never b…

JaredTokuz:
…e used

## Summary

I want to fix a typo in the documentation 

## Set of changes

I want to remove a conditional else block in the willAuthError. Based on the logic it will never be used.

### DIFF
--- a/docs/advanced/authentication.md
+++ b/docs/advanced/authentication.md
@@ -309,8 +309,6 @@ const willAuthError = ({ operation, authState }) => {
         );
       })
     );
-  } else if (false /* JWT is expired */) {
-    return true;
   }
 
   return false;


docs/advanced/authentication.md
```diff
@@ -309,8 +309,6 @@ const willAuthError = ({ operation, authState }) => {
         );
       })
     );
-  } else if (false /* JWT is expired */) {
```
JoviDeCroock, line 312: judging from the comment it would be better to do something like `else if (isJwtExpired(authState))`

JaredTokuz: ohh ya, my mistake. I think both the original snippet and your change makes sense.

JoviDeCroock:
Closing as stale, feel free to reopen if you think my comment is word addressing

PR #2571, opened 2022-07-28
changed to use setInterval in test

jatokz:
## Summary

I want to fix a typo in documentation

## Set of changes

The test seems to be checking on an interval for if the observable pipe subscription is mutating on an interval. The code used setTimeout and I changed to setInterval otherwise the test would never reach done()

### DIFF
--- a/docs/advanced/testing.md
+++ b/docs/advanced/testing.md
@@ -243,7 +243,7 @@ it('should update the list', done => {
     </Provider>
   );
 
-  setTimeout(() => {
+  setInterval(() => {
     expect(wrapper.find('.list').children()).toHaveLength(index + 1); // See how many items are in the list
     index++;
     if (index === 2) done();


docs/advanced/testing.md
```diff
@@ -243,7 +243,7 @@ it('should update the list', done => {
     </Provider>
   );
 
-  setTimeout(() => {
+  setInterval(() => {
```
JoviDeCroock, line 246: I don't feel comfortable doing `setInterval` in our tests without showing a proper cleanup 😅 we could change this test to instead assert that after 200ms we have the old length of children + an extra one due to the subscription firing.

JaredTokuz: I see what you mean. If you still wanted to show testing the interval. Would there be anything wrong when testing with the cleanup like below?

`
 const interval = setInterval(() => {
    expect(wrapper.find('.list').children()).toHaveLength(index + 1); // See how many items are in the list
    index++;
    if (index === 2) {
      clearInterval(interval);
      done();
    }
  }, 200);
`

JoviDeCroock: Well if the test fails you risk never calling `clearInterval`

JaredTokuz: Oops. How about adding a before and after hook in the snippet like this?

```
  before(function() { let interval; });
  after(function() { clearInterval(interval) });
```

If doing "setTimeout", the snippet would get shorter doing something like this I think

```
  setTimeout(() => {
    expect(wrapper.find('.list').children()).toHaveLength(index + 1); // See how many items are in the list
    done();
  }, 200);
```

PR #2574, opened 2022-07-28
feat(next-urql): send App props for pre-pass

### LINKED ISSUE #2573: RFC: `next-urql` – send App props for pre-pass
## Summary

Using the `next-urql` library with `withUrqlClient` is used with `ssr: true`, all the Next App tree is re-rendered. When we use `withUrqlClient` in the `App` everything works fine. But when we use `withUrqlClient` in pages and have React Context providers in the custom `Aapp` that receive props from `App.getInitialProps`. In the pre-pass step, this values (it only fetches the `Page.getInitialProps`) and no App props from   `App.getInitialProps`. This can be an issue if, for example, we use a Feature Flag React context, the behavior in the pre-pass step will be different.

## Proposed Solution

What I want to propose is a callback inside `WithUrqlClientOptions` where we could send this app props from the Next context, something like:
```ts
export interface WithUrqlClientOptions {
  ssr?: boolean;
  neverSuspend?: boolean;
  staleWhileRevalidate?: boolean;
  getAppProps?: (ctx?: NextPageContext) => any;
}
```

And the only change would be in `with-urql-client.ts`:
```ts
       const appProps = isApp ? {} : options?.getAppProps?.(ctx) ?? {};
       // L: 163 
       const appTreeProps = isApp ? props : { pageProps: props, ...appProps };
```

With this, we are able to send the app props, and keep using the context in the pre-pass step.

tomvardasca:
Resolves #2573

## Summary

Using the next-urql library with withUrqlClient is used with ssr: true, all the Next App tree is re-rendered. When we use withUrqlClient in the `App` everything works fine. But when we use withUrqlClient in pages and have React Context providers in _app that receive props from App.getInitialProps. In the pre-pass step, this values (it only fetches the Page.getInitialProps) and no App props from App.getInitialProps. This can be an issue if, for example, we use a Feature Flag React context, the behavior in the pre-pass step will be different.

## Set of changes

Added a new option in `WithUrqlClientOptions`, `getAppProps` where you can send the props for the App. This way the App is rendered in prepass with the same context as without it.

### DIFF
--- a/packages/next-urql/README.md
+++ b/packages/next-urql/README.md
@@ -146,6 +146,8 @@ The second argument for `withUrqlClient` is an options object, this contains one
 `withUrqlClient` that the wrapped component does not use `getInitialProps` but the children of this wrapped component do. This opts you into
 `ssr` for these children.
 
+If you use custom `App.getInitialProps` and use this App props in the App and have `ssr === true` you you can pass the props through `generateAppProps` so the prepass step has all the information needed to render.
+
 ### Different Client configurations on the client and the server
 
 There are use cases where you may need different configurations for your `urql` Client on the client-side and the server-side; for example, you may want to interact with one GraphQL endpoint on the server-side and another on the client-side. `next-urql` supports this as of v0.3.0. We recommend using `typeof window === 'undefined'` or a `process.browser` check.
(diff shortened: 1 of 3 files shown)


nruhe:
I believe we're running into a similar issue when trying to use withUrqlClient on _app.js alongside [appWithTranslations](appWithTranslation) from https://github.com/i18next/next-i18next

PR #2593, opened 2022-08-08
docs: remove double negation

dargmuesli:
## Summary

Replace `isn't dissimilar from` with `is similar to`.
... or was there a reason for the double negation? I stumbled across this for so many times now, that I had to make this PR 😅

### DIFF
--- a/docs/basics/svelte.md
+++ b/docs/basics/svelte.md
@@ -378,7 +378,7 @@ There are some more tricks we can use with `queryStore`.
 
 ## Mutations
 
-The `mutationStore` function isn't dissimilar from the `queryStore` function but is triggered manually and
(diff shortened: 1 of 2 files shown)


.changeset/empty-waves-look.md
```diff
@@ -0,0 +1,5 @@
+---
+'urql-docs': patch
+---
+
+docs: remove double negation
```
JoviDeCroock, line 5: You don't need to add a changeset for docs as we can't publish these

JoviDeCroock (approved):
Good spot, thank you!

.changeset/empty-waves-look.md
dargmuesli: Removed :)

PR #2602, opened 2022-08-11
docs: refactored examples that use the logical `!`

MoustaphaDev:
## Summary

Using the bang operator (!) on non-boolean values can lead to unexpected results. For example, in the case of [this example](../blob/main/docs/basics/react-preact.md#pausing-usequery), using the `pause` option inside the `useQuery` hook, one might set the value of the `from` variable to be `0`. In that case the checks would be inaccurate since `!0 === true`. However, we only expect to pause the query only when `from` or `limit` is nullish. We can replace that condition with a more robust one.

Yes, it's an edge case, but new developers reading this documentation might not be aware of the presence of this bug and might have some unexpected results using this example in their code.

## Set of changes

Replaced `!from || !limit` with the more accurate `from === undefined || from === null || limit === undefined || limit === null` in the `useQuery` hook/composable on [React/Preact Basics](../blob/main/docs/basics/react-preact.md#pausing-usequery) and [Vue basics](../blob/main/docs/basics/vue.md#pausing-usequery).
An intermediate variable `shouldPause` got used to avoid inlining that long condition into the `pause` option.

### DIFF
--- a/docs/basics/vue.md
+++ b/docs/basics/vue.md
@@ -269,6 +269,8 @@ import { useQuery } from '@urql/vue';
 export default {
   props: ['from', 'limit'],
   setup({ from, limit }) {
+    const shouldPause = computed(() => from === undefined || from === null ||
+                                      limit === undefined || limit === null);
     return useQuery({
       query: `
         query ($from: Int!, $limit: Int!) {
@@ -279,7 +281,7 @@ export default {
         }
       `,
       variables: { from, limit },
-      pause: computed(() => !from.value || !limit.value)
+      pause: shouldPause
     });
   }
 };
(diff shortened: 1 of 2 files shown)


docs/basics/vue.md
```diff
@@ -269,6 +269,8 @@ import { useQuery } from '@urql/vue';
 export default {
   props: ['from', 'limit'],
   setup({ from, limit }) {
+    const shouldPause = computed(() => from === undefined || from === null ||
+                                      limit === undefined || limit === null);
```
JoviDeCroock, line 273:
```suggestion
    const shouldPause = computed(() => from == null || limit == null);
```

Is a bit more terse 😅

MoustaphaDev: I admit 😂
Sacrificing the edge case for a more terse code sounds good to me

kitten (approved):

docs/basics/vue.md
MoustaphaDev: Actually your suggestion only applied to the basics/vue.md example. In basics/react-preact.md there's a similar example but it still has my initial suggestion. Should I open a new PR to fix that?

PR #2605, opened 2022-08-13
(docs) - Fix relayPagination example

y15e:
## Summary

Fix cursor value of relayPagination example

## Set of changes

Change `after` value from `1` to `a`


kitten (approved):
Nice catch! 🙌

JoviDeCroock (approved):

PR #2607, opened 2022-08-14
fix(core): Implement stricter variable typings on generic

### LINKED ISSUE #2606: RFC: Fix for useQuery variables being optional
In my codebase this feels like a bug - but I'm posting it as an RFC since I might be missing some perspective on this. An example of what I expect to give a type error:
```GraphQL
query GetPage($pageNumber: Int!) {
  page(pageNumber: $pageNumber){
    title
    body
  }
}
```

Which doesn't give a type error for this:
```TypeScript
const [{ data }] = useGetPage() // <-- obviously missing variables with pageNumber
```

When using useQuery the variables parameter is always optional - even if variables are required by the query type. The issue is with the `UseQueryArgs` interface, which currently looks like this:
```TypeScript
export interface UseQueryArgs<Variables = object, Data = any> {
    query: string | DocumentNode | TypedDocumentNode<Data, Variables>;
    variables?: Variables; // <--- here is the issue
    requestPolicy?: RequestPolicy;
    context?: Partial<OperationContext>;
    pause?: boolean;
}
```

I did some experimenting and I think something like this could be better (works as expected when using useQuery, although it does cause some type compatibility issues with [@graphql-codegen/typescript-urql](https://npmjs.com/package/@graphql-codegen/typescript-urql)):

```TypeScript
export interface UseQueryArgsWithVariables<Variables = object, Data = any> {
    query: string | DocumentNode | TypedDocumentNode<Data, Variables>;
    variables: Variables;
    requestPolicy?: RequestPolicy;
    context?: Partial<OperationContext>;
    pause?: boolean;
}

export interface UseQueryArgsWithoutVariables<Data = any> {
    query: string | DocumentNode | TypedDocumentNode<Data, Variables>;
    requestPolicy?: RequestPolicy;
    context?: Partial<OperationContext>;
    pause?: boolean;
}

export type UseQueryArgs<Variables = object, Data = any> = keyof Variables extends never ? UseQueryArgsWithoutVariables<Data> : UseQueryArgsWithVariables<Variables, Data>
```

I tried with a similar approach for useMutation which has the same issue but wasn't able to get it 100%.

kitten:
Resolve #2606

## Summary

This attempts to introduce a stricter type for variables preventing it from being an optional property or argument (i.e. `variables?: Variables`) in all parts of the codebase. To do this, we also need to make its type a little more strict with `AnyVariables`, which explicitly defaults the generic to a type of `void | undefined | { ... }`. In case of a generic that's passed in, this also becomes the extension type.

This is currently a draft because some methods on `Client` and some utility functions, like `createRequest`, are affected by this change. We have to check that these behave as expected (as per the change and RFC) and that we don't break them if that isn't necessary

## Set of changes

- Add `AnyVariables` and make generics `Variables extends AnyVariables = AnyVariables`
- Update core typings where `variables` were optional
- Update all typings across bindings (wherever `variables?:` was used, a special type must be implemented)

### DIFF
--- a/packages/vue-urql/src/useQuery.ts
+++ b/packages/vue-urql/src/useQuery.ts
@@ -8,6 +8,7 @@ import { Source, pipe, subscribe, onEnd } from 'wonka';
 
 import {
   Client,
+  AnyVariables,
   OperationResult,
   TypedDocumentNode,
   CombinedError,
@@ -23,19 +24,25 @@ import { unwrapPossibleProxy } from './utils';
 
 type MaybeRef<T> = T | Ref<T>;
 
-export interface UseQueryArgs<T = any, V = object> {
+export type UseQueryArgs<T = any, V extends AnyVariables = AnyVariables> = {
   query: MaybeRef<TypedDocumentNode<T, V> | DocumentNode | string>;
-  variables?: MaybeRef<V>;
   requestPolicy?: MaybeRef<RequestPolicy>;
   context?: MaybeRef<Partial<OperationContext>>;
   pause?: MaybeRef<boolean>;
-}
+} & (V extends void
+  ? {
+      variables?: MaybeRef<V>;
+    }
+  : {
+      variables: MaybeRef<V>;
+    });
 
-export type QueryPartialState<T = any, V = object> = Partial<
-  OperationResult<T, V>
-> & { fetching?: boolean };
+export type QueryPartialState<
+  T = any,
+  V extends AnyVariables = AnyVariables
+> = Partial<OperationResult<T, V>> & { fetching?: boolean };
 
-export interface UseQueryState<T = any, V = object> {
+export interface UseQueryState<T = any, V extends AnyVariables = AnyVariables> {
   fetching: Ref<boolean>;
   stale: Ref<boolean>;
   data: Ref<T | undefined>;
@@ -55,13 +62,13 @@ const watchOptions = {
   flush: 'pre' as const,
 };
 
-export function useQuery<T = any, V = object>(
+export function useQuery<T = any, V extends AnyVariables = AnyVariables>(
   args: UseQueryArgs<T, V>
 ): UseQueryResponse<T, V> {
   return callUseQuery(args);
 }
 
-export function callUseQuery<T = any, V = object>(
+export function callUseQuery<T = any, V extends AnyVariables = AnyVariables>(
   _args: UseQueryArgs<T, V>,
   client: Ref<Client> = useClient(),
   stops: WatchStopHandle[] = []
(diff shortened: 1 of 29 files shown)


JoviDeCroock (approved):

packages/vue-urql/src/useQuery.ts
```diff
@@ -23,19 +24,25 @@ import { unwrapPossibleProxy } from './utils';
...
-}
-
-export type QueryPartialState<T = any, V = object> = Partial<
-  OperationResult<T, V>
-> & { fetching?: boolean };
-
-export interface UseQueryState<T = any, V = object> {
+} & (V extends void
+  ? {
+      variables?: MaybeRef<V>;
```
JoviDeCroock, line 34:
```suggestion
      variables?: MaybeRef<{ [K in keyof V]: MaybeRef<V[K]> }>;
```

packages/vue-urql/src/useQuery.ts
```diff
@@ -23,19 +24,25 @@ import { unwrapPossibleProxy } from './utils';
...
+    }
+  : {
+      variables: MaybeRef<V>;
```
JoviDeCroock, line 37:
```suggestion
      variables: MaybeRef<{ [K in keyof V]: MaybeRef<V[K]> }>;;
```

packages/vue-urql/src/useSubscription.ts
```diff
@@ -21,17 +22,29 @@ import { unwrapPossibleProxy } from './utils';
...
+  V extends AnyVariables = AnyVariables
+> = {
   query: MaybeRef<TypedDocumentNode<T, V> | DocumentNode | string>;
-  variables?: MaybeRef<V>;
   pause?: MaybeRef<boolean>;
   context?: MaybeRef<Partial<OperationContext>>;
-}
+} & (V extends void
+  ? {
+      variables?: MaybeRef<V>;
```
JoviDeCroock, line 34:
```suggestion
      variables?: MaybeRef<{ [K in keyof V]: MaybeRef<V[K]> }>;
```

packages/vue-urql/src/useSubscription.ts
```diff
@@ -21,17 +22,29 @@ import { unwrapPossibleProxy } from './utils';
...
+    }
+  : {
+      variables: MaybeRef<V>;
```
JoviDeCroock, line 37:
```suggestion
      variables: MaybeRef<{ [K in keyof V]: MaybeRef<V[K]> }>;
```

JoviDeCroock (commented):
These come from #2608

crmitchelmore:
@kitten is the expected pattern to pass empty variables for cases when there are no variables?

e.g. `client.query(SomeDocument)`

becomes:

`client.query(SomeDocument, {})`

Issue #2566, opened 2022-07-27
(graphcache)- readFragment targets the first fragment of a DocumentNode


narkeeso:
### Describe the bug

When using a fragment with nested fragments readFragment will return null if your nested fragment is ordered a certain way.

Non-working Example:

```
// This doesn't work, observe that the nested Note fragment is defined first before using it
gql`
  fragment Note on note {
    id
  } 

  fragment User on user {
    id
    notes {
      ...Note
    }
  }
`
```

Working Examples (workarounds):

```
// Moving the Note fragment to the bottom strangely works, it's probably being hoisted?
gql`
  fragment User on user {
    id
    notes {
      ...Note
    }

  fragment Note on note {
    id
  } 
`

// or don't use nested fragments
gql`
  fragment User on user {
    id
    notes {
      id
    }
  }
`
```

I've attached a CodeSandbox with the reproduction. Uncomment out the various fragments at the top of the sandbox to observe how readFragment works with various fragment configurations. The result is printed to console.

### Reproduction

https://codesandbox.io/s/urql-nested-fragments-bug-7ez7tr?file=/src/App.tsx

### Urql version

2.2.3

### Validations

- [X] I can confirm that this is a bug report, and not a feature request, RFC, question, or discussion, for which [GitHub Discussions](https://github.com/FormidableLabs/urql/discussions) should be used
- [X] Read the [docs](https://formidable.com/open-source/urql/docs/).
- [X] Follow our [Code of Conduct](https://github.com/FormidableLabs/urql/blob/main/CODE_OF_CONDUCT.md)

JoviDeCroock:
This kind off feels like intended behavior, we take the first fragment we can find because it's a heuristic to see what base-type you are looking for. What assumption would you expect here? Us iterating over all fragments to find the one who doesn't have any spreads, this could also be an unused one in that case though, we could ask to specify the type the user is looking for but this might give issues with interfaces/unions 😅

narkeeso:
The behavior feels inconsistent in my opinion and slightly unintuitive. When using it in a query it works as expected. Declaring the fragment before usage feels like the correct thing to do.

If this is how readFragment is intending to behave we should probably add a note in the docs at least, I lost a good chunk of time figuring this out. Although I do believe it should support the originally intended method of declaring your fragments first.

JoviDeCroock:
[...]

PR #2609, opened 2022-08-15
feat(graphcache): support passing in name for fragments

JoviDeCroock:
Resolves https://github.com/FormidableLabs/urql/issues/2566

## Summary

When re-using document-nodes that contain queries/multiple fragments we might need to target a fragment rather than going for the first one. This PR introduces a new argument named `fragmentName` which allows people to target a specific `FragmentDefinitionNode` in a `DocumentNode` by name.

## Set of changes

- add `fragmentName` argument to write and read operations

### DIFF
--- a/.changeset/hungry-swans-tell.md
+++ b/.changeset/hungry-swans-tell.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-graphcache': minor
+---
+
+Allow passing in `fragmentName` for `write` and `read` operations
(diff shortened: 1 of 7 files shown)


kitten (approved):

narkeeso:
I just noticed this, thank you for this feature. We have a lot of fragments that will benefit from this. I think it also helps people understand how the fragment is targeted. Thank you 🙏

PR #2643, opened 2022-08-25
fix: #2642 root element have no __typename field

### LINKED ISSUE #2642: urql 3.0 - regression on maskTypename
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `@urql/core@3.0.1` for the project I'm working on.

With urql@3.0, even if `maskTypename` option is set, `__typename` fields are not removed from response.

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/@urql/core/dist/urql-core.js b/node_modules/@urql/core/dist/urql-core.js
index 0724265..eef71f3 100644
--- a/node_modules/@urql/core/dist/urql-core.js
+++ b/node_modules/@urql/core/dist/urql-core.js
@@ -74,11 +74,12 @@ var formatDocument = node => {
 };
 
 var maskTypename = data => {
+ 
   if (!data || typeof data !== 'object') {
     return data;
   } else if (Array.isArray(data)) {
     return data.map(maskTypename);
-  } else if (data && typeof data === 'object' && '__typename' in data) {
+  } else if (data && typeof data === 'object') {
     var acc = {};
 
     for (var key in data) {
@@ -791,6 +792,7 @@ var Client = function Client(opts) {
       }))(wonka.take(1)(wonka.filter(op => op.kind === 'query' && op.key === operation.key && op.context.requestPolicy !== 'cache-only')(operations$)))]);
     })( // End the results stream when an active teardown event is sent
     wonka.takeUntil(wonka.filter(op => op.kind === 'teardown' && op.key === operation.key)(operations$))(result$)))));
+ 
     return source;
   };
 
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

mlecoq:
Resolves #2642

### DIFF
--- a/.changeset/many-worms-bow.md
+++ b/.changeset/many-worms-bow.md
@@ -1,0 +1,5 @@
+---
+'@urql/core': minor
+---
+
+Fix case where `maskTypename` would not traverse down when the root query-field does not contain a `__typename`
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -196,7 +196,7 @@ export const Client: new (opts: ClientOptions) => Client = function Client(
     if (opts.maskTypename) {
       result$ = pipe(
         result$,
-        map(res => ({ ...res, data: maskTypename(res.data) }))
+        map(res => ({ ...res, data: maskTypename(res.data, true) }))
       );
     }
 
--- a/packages/core/src/utils/maskTypename.test.ts
+++ b/packages/core/src/utils/maskTypename.test.ts
@@ -66,3 +66,20 @@ it('strips typename from nested objects with arrays', () => {
     },
   });
 });
+
+it('strips typename in nested object from root', () => {
+  expect(maskTypename({ root: { __typename: 'Todo', id: 1 } }, true)).toEqual({
+    root: {
+      id: 1,
+    },
+  });
+});
+
+it('doesn’t strip typename in sub-object when there is no __typename field', () => {
+  expect(maskTypename({ subObject: { __typename: 'Todo', id: 1 } })).toEqual({
+    subObject: {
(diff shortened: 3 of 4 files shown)


.changeset/many-worms-bow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/core': minor
+---
+
+fix maskTypename
```
JoviDeCroock, line 5:
```suggestion
Fix case where `maskTypename` would not traverse down when the root query-field does not contain a `__typename`
```

JoviDeCroock (commented):
Mind adding a test for this?

.changeset/many-worms-bow.md
mlecoq: Thanks for your feedback ! 

I have added a unit test and fixed the case with Date objects

mlecoq:
> Mind adding a test for this?

Hi, does it look good to you after last changes ?

kitten:
Hiya 👋 Sorry about that regression. That was my bad

Unfortunately, the fix here you've got now has a small problem as well.
You've got the `Date` instance check there, however, exchanges may add any kind of objects into the result, making the recursion now potentially dangerous and destroying instances or even recursing infinitely if instances with recursive properties are added to a result 😅

What I forgot when refactoring is that we skip adding a `__typename` field to the root selection set. However, what we can do about this is skip the `__typename` check for the first run of this function, which we could do with an additional argument for instance

mlecoq:
Ok I'll change it

JoviDeCroock (approved):

PR #2667, opened 2022-09-03
fix(react): set `fetching` to the lack or presence of `source`

### LINKED ISSUE #2648: fetching is not set to true for subscriptions on variable change
### Describe the bug

I'm using a subscription which looks like this:

```
const [result] = useExampleSubscription({
  variables: { variableA: valueA, variableB: valueB },
});
```
`useExampleSubscription` is a generated hook by graphql-code-generator, which calls `useSubscription` under the hood:
```
export function useExampleSubscription<TData = ExampleSubscription>(options: Omit<Urql.UseSubscriptionArgs<ExampleSubscriptionVariables>, 'query'> = {}, handler?: Urql.SubscriptionHandler<ExampleSubscription, TData>) {
  return Urql.useSubscription<ExampleSubscription, TData, ExampleSubscriptionVariables>({ query: ExampleDocument, ...options }, handler);
};
```
The graphql server that I'm using is Hasura.

Then I'm changing `valueB` to a new value, the old subscription closes (as expected), a new one opens and I would expect `result.fetching` to be temporarily `true` in the new subscription result until there is new data from the API.
But instead, `result.fetching` of the new subscription result is `false` the first time it is returned (together with the new data). Is this expected ? If yes, how can I know that a subscription is on fetching state after a variable change ?

In the docs it is mentioned: `Whenever the variables (or the query) option on the useQuery hook changes fetching will switch to true, and a new request will be sent to our API, unless a result has already been cached previously.`

I can verify that this works for `useQuery` (using the respective generated hook by graphql-code-generator), but it does not seem to work for `useSubscription`.

I hope I'm not missing something. Thank you.

P.S. Please let me know if I should submit a reproduction url, in case the above description is not sufficient.
P.S.2 This works correctly in case the subscription is first paused, and then resumed at the same time `valueB` is changed. It also works correctly on first render.

### Reproduction

""

### Urql version

urql: 3.0.1
graphql-codegen/typescript-urql: 3.6.4

### Validations

JoviDeCroock:
Resolves #2648 

## Summary

When we changes `variables` `onEnd` seems to fire later than us establishing the new connection, this means we set `fetching` to a truthy value and then switch it off as we assume that we aren't fetching anymore.

Now when a subscription ends we check whether there is a new source already and base our fetching state off that, [sandbox](https://codesandbox.io/s/kind-currying-8v8yvw?file=/src/components/Messages.js)

CC @elias-pap

## Set of changes

- rely on `source` to flip `fetching` state

### DIFF
--- a/.changeset/tame-crews-grab.md
+++ b/.changeset/tame-crews-grab.md
@@ -1,0 +1,5 @@
+---
+"urql": patch
+---
+
+Fix `fetching` going to `false` after changing variables in a subscription
--- a/packages/react-urql/src/hooks/useSubscription.ts
+++ b/packages/react-urql/src/hooks/useSubscription.ts
@@ -114,7 +114,7 @@ export function useSubscription<
       return pipe(
         state[0],
         onEnd(() => {
-          updateResult({ fetching: false });
+          updateResult({ fetching: !!source });
         }),
         subscribe(updateResult)
       ).unsubscribe;


kitten (approved):

elias-pap:
This change might have introduced a different bug. I use wonka's `fromPromise` to mock `executeSubscription` on testing when creating a mock urql client, like this:

```
executeSubscription: ({ query, variables }) =>
      executeMockOperation({ schemaWithMocks, query, variables }),

const executeMockOperation = ({ schemaWithMocks, query, variables }) => {
  const result = graphql({
    schema: schemaWithMocks,
    source: query.loc.source.body,
    variableValues: variables,
  });
  return fromPromise(result);
};
```

Before the fix (3.0.2) `fetching` is eventually set to `false` when the data comes (expected). After the fix (3.0.3) `fetching` is set to `true` and never turns `false` again. I had to implement a custom source with `makeSubject` to work around this.

PR #2692, opened 2022-09-21
feat(core): Allow GET URL limit to be ignored using force option

### LINKED ISSUE #2680: RFC: Configurable url-size limit
## Summary

https://github.com/FormidableLabs/urql/pull/2384#issue-1193259509

its a hardcoded limitation

In our case we don't need to limit the url size because its an internal graphql request. 
With the limitation its not possible to force GET requests.

## Proposed Solution

Please accept a new parameter (eg "urlSizeLimit")

kitten:
Resolve #2680

## Summary

This updates `preferGetMethod` to allow for one more mode, which ignores the URL size limit of 2048 characters.

The option now, apart from boolean flags, also accepts `preferGetMethod: 'within-url-limit'` and `preferGetMethod: 'force'`. This is the proposed method to allow fetch exchanges to bypass the defined limit and always send a GET request for queries.

The default remains unchanged at `preferGetMethod: true | 'within-url-limit'`

## Set of changes

- Update types to include new options for `preferGetMethod`
- Handle `preferGetMethod: 'force'`


packages/core/src/internal/fetchOptions.ts
```diff
@@ -40,7 +40,7 @@ export const makeFetchURL = (
     search.set('extensions', stringifyVariables(body.extensions));
 
   const finalUrl = url.toString();
-  if (finalUrl.length > 2047) {
+  if (finalUrl.length > 2047 && useGETMethod !== 'force') {
```
JoviDeCroock, line 43: Do we want to allow people to specify their own limit or are we just going to leave it to the escape-hatch?

JoviDeCroock (approved):

packages/core/src/internal/fetchOptions.ts
kitten: I don't really see personally why someone would have a different limit than the defacto IETF limit on the length of URLs, which also happens to be one that all browsers align on 🤔 
I think the main reason we're adding this is if someone knows their _entire_ stack can support variable length URLs, in which case, worst case, they're opting into HTTP stack errors, which are very observable. Most likely, I'd expect to see this in backend-only applications where likely no limit really applies.

Issue #2701, opened 2022-09-23
useQuery not returning stable references


sarink:
### Describe the bug

I hope that I'm simply misunderstanding or have misconfigured something, but I can't seem to figure out why objects inside `useQuery().data.MyQuery.someList[index]` are changing references. This means they cannot be memoized, and the application rerenders unnecessarily

* On initial load, the TodosList and every Todo renders three times
* Upon creating or updating a todo, each individual Todo in the list rerenders

Thanks :)

For skimming (if you don't want to pull the repro repo):

![Screen Shot 2022-09-23 at 4 05 03 PM](https://user-images.githubusercontent.com/1881761/192056778-58a3a68a-8e48-4fa5-9b14-6ff924b895c0.png)

![urql-memo](https://user-images.githubusercontent.com/1881761/192056828-fd2592e8-ada1-4f27-b4ff-8f4d7a20ad55.gif)

### Reproduction

https://github.com/sarink/urql-perf/pull/2

### Urql version

^2.2.3

### Validations

- [X] I can confirm that this is a bug report, and not a feature request, RFC, question, or discussion, for which [GitHub Discussions](https://github.com/FormidableLabs/urql/discussions) should be used
- [X] Read the [docs](https://formidable.com/open-source/urql/docs/).
- [X] Follow our [Code of Conduct](https://github.com/FormidableLabs/urql/blob/main/CODE_OF_CONDUCT.md)

JoviDeCroock:
We have gone over this quite a few times in issues/discussions, we do our best to provide stable references but at the end of the day it isn't the most important thing. i.e. a re-render isn't the expensive thing, memoizing everything and doing deep equality might be a lot more expensive if we'd do that on every query.

The best bet for us is to ensure you have the most up-to-date data and you can optimize the re-render process yourself by means of `memo` and `useMemo`, also a re-render isn't expensive as long as you aren't doing the most expensive operation, which is touching the DOM. 

Other issues https://github.com/FormidableLabs/urql/issues/425#issuecomment-530706043

sarink:
@JoviDeCroock the code _is_ fully memoized -- memoization works by equality comparison, so if the object references keep changing, memoizing has no effect. The above code has exactly the same number of rerenders with or without memoization.

sarink:
I don't really accept that this impossible or that it's it is unimportant. I just rebuilt the same demo with apollo client, and it behaves exactly as you'd expect, the memoization works and rerenders are prevented:

[...]

PR #2718, opened 2022-10-07
fix(graphcache): consistent __typename addition

JoviDeCroock:
## Summary

While investigating https://github.com/FormidableLabs/urql/issues/2701 I noticed that we invalidate the result for a cache-hit due to not having `__typename` there, on a closer look this is because when we [`foward`](https://github.com/FormidableLabs/urql/blob/main/exchanges/graphcache/src/cacheExchange.ts#L134) we add the `__typename` in `formatDocument`, this is also returned to the user because we have changed the operation itself, but when we get a result from cache we use the supplied query which might be missing these `__typename` additions. This operation is basically free because we have the operation cached either way.

## Set of changes

- add `__typename` consistently

### DIFF
--- a/exchanges/graphcache/src/cacheExchange.ts
+++ b/exchanges/graphcache/src/cacheExchange.ts
@@ -158,7 +158,19 @@ export const cacheExchange = <C extends Partial<CacheExchangeOpts>>(
   const operationResultFromCache = (
     operation: Operation
   ): OperationResultWithMeta => {
-    const result = query(store, operation, results.get(operation.key));
+    const result = query(
+      store,
+      makeOperation(
+        operation.kind,
+        {
+          key: operation.key,
+          query: formatDocument(operation.query),
+          variables: operation.variables,
+        },
+        operation.context
+      ),
+      results.get(operation.key)
+    );
     const cacheOutcome: CacheOutcome = result.data
       ? !result.partial
         ? 'hit'
--- a/.changeset/hot-turkeys-walk.md
+++ b/.changeset/hot-turkeys-walk.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-graphcache': patch
+---
+
+Ensure we are consistent in returning `__typename` in results


exchanges/graphcache/src/cacheExchange.ts
```diff
@@ -206,10 +210,16 @@ export const cacheExchange = <C extends Partial<CacheExchangeOpts>>(
...
       const queryResult = query(
         store,
-        operation,
-        operation.kind === 'query' ? results.get(operation.key) || data : data,
+        {
+          ...operation,
+          query: originalDocument || operation.query,
+        },
+        operation.kind === 'query'
+          ? results.get(operation.key) || undefined
```
JoviDeCroock, line 221: if we use `data` here we would not remove the `__typename` during `query`

exchanges/graphcache/src/cacheExchange.test.ts
```diff
@@ -206,7 +218,7 @@ describe('data dependencies', () => {
     const firstDataTwo = result.mock.calls[1][0].data;
     expect(firstDataOne).not.toBe(firstDataTwo);
     expect(firstDataOne.author).not.toBe(firstDataTwo.author);
-    expect(firstDataOne.unrelated).toBe(firstDataTwo.unrelated);
+    //expect(firstDataOne.unrelated).toBe(firstDataTwo.unrelated);
```
JoviDeCroock, line 221: It looks like currently this stopped working due to the change which seems peculiar to me as we have actually only changed the logic to use `queryResult.data` which should be equal on the subsequent invocation 😅

JoviDeCroock:
Superseded by https://github.com/FormidableLabs/urql/commit/da121d45512477844a651cf94da4854009adcb9d

kitten:
Testing a different approach this in #2736, which instead tries to preserve the original document and continue to prevent adding `__typename` when it's not in the original document (which was, I believe, an issue reported in the past)

JoviDeCroock:
@kitten did you mean to post this in #2731?

PR #2731, opened 2022-10-10
fix(graphcache): infinite loop in dependent queries

JoviDeCroock:
## Summary

When we have a data relationship between x queries we currently end up re-triggering them when we have a cache-miss, this because of how we collect `dependencies` and `reexecute` them. To circumvent this we keep track of the operation that creates these reexecutions and we keep track of the ones that were targeted by these reexecuting origins.

This means that A can trigger B but B can't re-trigger A if B is a cache-miss.

### DIFF
--- a/exchanges/graphcache/src/cacheExchange.ts
+++ b/exchanges/graphcache/src/cacheExchange.ts
@@ -55,8 +55,6 @@ export const cacheExchange = <C extends Partial<CacheExchangeOpts>>(
   const blockedDependencies: Dependencies = new Set();
   const requestedRefetch: Operations = new Set();
   const deps: DependentOperations = new Map();
-  let triggeringOperations = new Set();
-  let targetOperations = new Set();
 
   const isBlockedByOptimisticUpdate = (dependencies: Dependencies): boolean => {
     for (const dep of dependencies.values())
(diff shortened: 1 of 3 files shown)


exchanges/graphcache/src/cacheExchange.ts
```diff
@@ -79,17 +79,26 @@ export const cacheExchange = <C extends Partial<CacheExchangeOpts>>(
     operation: Operation,
     pendingOperations: Operations
   ) => {
+    if (operation.context.reexecuted) {
+      operation.context.reexecuted = false;
```
JoviDeCroock, line 83: We should remove this line and use it where we transform a cache and network invocation, possibly resulting from requested refetch, to network-only and remove the propety there

exchanges/graphcache/src/cacheExchange.ts
```diff
@@ -79,17 +79,26 @@ export const cacheExchange = <C extends Partial<CacheExchangeOpts>>(
...
+        // we should not reexecute it again
         if (op) {
           operations.delete(key);
           let policy: RequestPolicy = 'cache-first';
           if (requestedRefetch.has(key)) {
             requestedRefetch.delete(key);
             policy = 'cache-and-network';
           }
+          if (operation.kind === 'query')
+            op.context = { ...op.context, reexectued: true };
```
yankovalera, line 101: 🔧 typo in reexectued

JoviDeCroock:
Superseded by https://github.com/FormidableLabs/urql/commit/eec46e03a77aa67475def4f7ad2bd0ee075e86ad

kitten:
Testing a different approach to this in #2737

PR #2734, opened 2022-10-11
chore(docs): update comparison chart with Apollo Client defer support

alessbell:
## Summary

Hi there! This PR contains a small update to the comparison chart to reflect that Apollo Client now supports the `@defer` directive as of v3.7.0, though `@stream` is not yet supported ([release notes](https://github.com/apollographql/apollo-client/releases/tag/v3.7.0)).

Happy to format it differently - I went with ✅/🛑 to communicate both statuses since they're contained in a single cell.

## Set of changes

This is a docs-only change.


JoviDeCroock (approved):
Nice, congratulations on v3.7! Happy to see more clients landing this! Looks like we'll need to give this page an update as well in terms of bundle size etc

PR #2748, opened 2022-10-17
chore(workspace): stabilise reliability of e2e

JoviDeCroock:
## Summary

This pull request is aimed at increasing the reliability of our e2e, we were seeming some failures by our API answering too fast.

As a follow-up I would love to start adding graphCache, ... to this and test a real application with graphCache, another follow up would be to update to the new cypress major

## Set of changes

- increase speed of CI by only building the core package
- add missing cypress files
- add `delayExchange` to our exchanges in e2e

### DIFF
--- a/packages/react-urql/e2e-tests/useQuery.spec.tsx
+++ b/packages/react-urql/e2e-tests/useQuery.spec.tsx
@@ -12,23 +12,11 @@ import {
   fetchExchange,
   Exchange,
 } from '../src';
-import { mergeMap, pipe, fromPromise } from 'wonka';
+import { delay, pipe } from 'wonka';
 
 const delayExchange: Exchange = ({ forward }) => {
   return ops$ => {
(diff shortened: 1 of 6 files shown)


packages/react-urql/e2e-tests/useQuery.spec.tsx
```diff
@@ -2,7 +2,35 @@
...
+      mergeMap(op => {
+        return fromPromise(
+          new Promise(res => {
+            setTimeout(() => {
+              res(op);
+            }, 250);
+          })
+        );
+      }),
+      forward
```
kitten, line 30: We don't really want to delay results or teardown operations, but we're putting this exchange in front of `fetchExchange`, so we can exploit that and can likely just do:

```suggestion
    return pipe(
      ops$,
      forward,
      delay(250)
```

kitten (approved):

PR #2762, opened 2022-10-25
fix(core): build debug target correctly

### LINKED ISSUE #2690: Debug events all have the same event.source
### Describe the bug

No matter which exchange dispatch a debug event, all events seem to use the name of the last configured exchange as their `source`. For example the events from the dedup exchange look like this:

```
{timestamp: 1663681014865, source: "fetchExchange", type: "dedup", message: "An operation has been deduped.", operation: Object}
```

I tried to track it down and added a `console.log('dispatchDebug() exchange.name: ', exchange.name);` right into the `dispatchDebug(event) { ... }` implementation here:
https://github.com/FormidableLabs/urql/blob/0582f9b138ddb66c5164b00a978aac90e865c0dc/packages/core/src/exchanges/compose.ts#L14-L20

Weird enough the `exchange.name` is correct there! So not sure where or when the `source` prop could be overwritten.

### Reproduction

https://codesandbox.io/s/urql-duplicate-operation-yfp2ei

### Urql version

* @urql/core 3.0.3
* @urql/vue 1.0.2

### Validations

EDIT: Fixed broken sandbox link

JoviDeCroock:
Resolves #2690 

## Summary

When leveraging `transform-debug-target`  we fill in the `source` everywhere, this does not work for the `compose.ts` file as it's not an exchange. The bug is caused by us taking the latest transformed exchange and pushing `source: name` into the object invocation, resulting in [this](https://www.runpkg.com/?@urql/core@3.0.4/dist/urql-core.js#689). This PR effectively excludes `compose.ts` and hence fixes our future builds.

### DIFF
--- a/.changeset/quiet-turkeys-brake.md
+++ b/.changeset/quiet-turkeys-brake.md
@@ -1,0 +1,5 @@
+---
+'@urql/core': patch
+---
+
+Fix case where our transform-debug-target babel plugin would override the root dispatchDebug in `compose.ts` with the latest found exchange, in this case `fetchExchange`
--- a/scripts/babel/transform-debug-target.js
+++ b/scripts/babel/transform-debug-target.js
@@ -21,14 +21,14 @@ const plugin = ({ template, types: t }) => {
           if (/Exchange$/i.test(exportName)) name = exportName;
         }
       },
-      CallExpression(path) {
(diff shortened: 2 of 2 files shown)


kitten (approved):
I didn't even know that this is where the sources came from 🤣

PR #2782, opened 2022-11-07
fix(graphcache): set operation when updating the cache with a result

### LINKED ISSUE #2768: (graphcache) Query is not updated after cache update
### Describe the bug

Since graphcache v5.0.2 seems that queries with `requestPolicy: "network-only"` are no longer updated as before.

## Example
Query:
```
notes {
  id
  status # e.g. TODO
}
```

Mutation:
```
completeNote(id: 1, status: COMPLETED) {
  id
  status
}
```

After mutation, I see that cache is updated with `status: COMPLETED`, but query is not updated (to trigger React component rerender).

In graphcache v5.0.1 queries are updated even with `requestPolicy: "network-only"`.

### Reproduction

-

### Urql version

@urql/exchange-graphcache v5.0.2 (5.0.1 works fine)
urql v.3.0.3

### Validations

JoviDeCroock:
Resolves #2768 

## Summary

When we update the cache we aren't writing to the `operations` map anymore because we moved this setter from the `updateDependencies` loop to the relevant functions in https://github.com/FormidableLabs/urql/pull/2736.

### DIFF
--- a/.changeset/pink-melons-joke.md
+++ b/.changeset/pink-melons-joke.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-graphcache': patch
+---
+
+Set operations when updating the cache with a result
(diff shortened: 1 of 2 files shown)


Tigge:
Thank you :hugs:, this has been affecting us since we updated to the latest urql version a while back. Any chance this will make it into a release soon?

PR #2794, opened 2022-11-09
chore(graphcache): add way we could go forward in e2e testing exchanges/...

JoviDeCroock:
## Summary

Lately we have started encountering scenario's where our test-suite wan't doing an adequate job of reproducing the bugs we were seeing, this is a proposal to incorporate more e2e testing.
Just like in our React Suspense testing we will leverage Cypress Component Testing to guarantee that we add test scenario's for everything.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,45 @@ jobs:
         run: yarn run lint
       - name: Unit Tests
         run: yarn run test --maxWorkers=2
+
+  react-e2e:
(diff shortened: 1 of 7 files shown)


JoviDeCroock:
I have actually tried `playwright` but I wasn't able to write JSX inside of the test files so I resorted to what we already had in cypress which seems more stable. To their credit `playwright` component testing is marked as `experimental` so when they do go stable we can def consider replacing `cypress`.

kitten (approved):

PR #2803, opened 2022-11-14
fix(execute): end iterator on teardown

danielkaxis:
## Summary

When using useSubscription, I expected that when the teardown function runs in exchange execute, the iterator would also end. But it made one call to next() and then ended the iterator. By adding iterator.return() to the teardown function, the iterator will not make any extra call to next() when the subscription ends.

## Set of changes

- Call iterator.return() when teardown function runs in exhange execute

### DIFF
--- a/exchanges/execute/src/execute.ts
+++ b/exchanges/execute/src/execute.ts
@@ -49,6 +49,7 @@ const makeExecuteSource = (
   args: ExecuteParams
 ): Source<OperationResult> => {
   return make<OperationResult>(observer => {
+    let iterator: AsyncIterator<ExecutionResult>;
     let ended = false;
 
     Promise.resolve()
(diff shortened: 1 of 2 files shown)


exchanges/execute/src/execute.ts
```diff
@@ -106,6 +101,7 @@ const makeExecuteSource = (
       });
 
     return () => {
+      iterator && iterator.return && iterator.return();
```
kitten, line 104: Just to bring it in line with the codding style of the repo, I'd suggest the following:

```suggestion
      if (iterator && iterator.return) iterator.return();
```

kitten (commented):
Nice catch! Not sure how we missed that prior, ha 👍

exchanges/execute/src/execute.ts
danielkaxis: Thanks! Done :)

JoviDeCroock (approved):

kitten (approved):

PR #2806, opened 2022-11-14
Exchange execute/promise context

### LINKED ISSUE #2805: RFC: Allow exchange-execute to execute even if context returns a Promise Function
## Summary

In the current implementation, exchange-execute is expected to be an object or a function returning an object, but functions returning a Promise should also be supported

Specifically, I would like to get the following code working

```ts
export const serverSideUrqlLClient = createClient({
  exchanges: [
    // ...
    executeExchange({
      context: async () => {
        // Asynchronous processing of some kind.
        return {
          // ...
        }
      },
      schema: schema,
    }),
  ],
  url: `dummy`,
})
```

In the current implementation, even if contextValue is a Promise or a function that returns a Promise, the Promise is not properly handled.

## Proposed Solution

The contextValue must be resolved before the following is performed.

https://github.com/FormidableLabs/urql/blob/0b307a918242f5a8dadd977fd96b73b1b42d8c56/exchanges/execute/src/execute.ts#L55-L61

It looks like it would be better to add another `.then` just before and await the contextValue for that purpose.

## Requirements

Please also check the test cases and proposed implementation as a PR.

https://github.com/FormidableLabs/urql/pull/2806

## Thanks

Finally, thank you for creating a great library.

YutaUra:
Resolves #2805 

## Summary

As per the Issue, I have fixed the exchange-execute to handle correctly when the context is a function that returns a Promise in exchange-execute.

## Set of changes

- @urql/exchange-execute 

As far as I know, there should be no breaking changes.


kitten (approved):
Sorry for the delay!
I updated the changeset phrasing; let's get this merged 🙌

PR #2812, opened 2022-11-20
chore(workspace): upgrade to rollup v3

JoviDeCroock:
## Summary

This upgrades rollup and our associated plugins to v3.

Todo:

- [x] verify sizes

## Set of changes

- use `.mjs` files for our rollup/babel scripts as the rollup environment expects usage according to the `type` field in package.json


scripts/rollup/cjs-check-plugin.mjs
```diff
@@ -9,13 +9,14 @@ function cleanup(opts = {}) {
...
 
-    renderChunk(code, chunk) {
+    async renderChunk(code, chunk) {
       if (opts.extension !== '.js') {
         return null;
       } else if (!filter(chunk.fileName)) {
         return null;
       }
 
+      await init()
```
JoviDeCroock, line 19: Looks like when using `cjs-module-lexer` in ESM it needs us to do an `async` init

scripts/rollup/plugins.mjs
```diff
@@ -29,7 +30,7 @@ export const makePlugins = () => [
     ignoreGlobal: true,
     include: /\/node_modules\//,
     namedExports: settings.hasReact ? {
-      react: Object.keys(require('react'))
+      react: Object.keys(React)
```
JoviDeCroock, line 33: Moved this require to an upfront import as React is in our global deps anyway

kitten (approved):

PR #2814, opened 2022-11-21
(auth) separate initial and refresh auth

kadikraman:
## Summary
Currently, the auth exchange has a single function - `getAuth` - that gets called for both initial auth as well as refreshing. This is not great form a DX point of views, since "get initial auth" and "handle auth error" are very different things.

## Set of changes
- `auth-exhange`: remove `getAuth` and split it into `getInitialAuth` and `refreshAuth`.


kitten:
Superseded by #3012

PR #2823, opened 2022-11-24
chore(workspace): add devtools to the main repo

JoviDeCroock:
## Summary

Moving the devtools exchange to this repo, this needed a few changes to the build. We currently still have a few failing tests in relation to missing globals. Looking into that now

## Set of changes

- replace `__pkg_version__` with the package.json version
- add a few globals for vitest
- add the devtools-exchange to the `exchanges/` folder


exchanges/devtools/src/exchange.test.ts
```diff
@@ -0,0 +1,353 @@
...
+//         }`,
+//   };
+
+//   beforeEach(() => {
+//     pipe(source, devtoolsExchange({ client, forward, dispatchDebug }), publish);
+//     handler = addMessageListener.mock.calls[0][0];
+//   });
+
+//   it('executes request on client', () => {
+//    handler(requestMessage);
```
JoviDeCroock, line 312: I commented out this test as it doesn't feel like it makes sense, this relies on our mock of `client.executeRequestOperation` to return a source but it doesn't so this just crashes 😅

kitten:
@JoviDeCroock I assume the outstanding todo here are credentials for publishing and testing, right? So, we can ensure that this doesn't go out of sync with the actual extension repo

exchanges/devtools/src/declarations/urql.d.ts
```diff
@@ -0,0 +1,16 @@
...
+    /** An execute[query|mutation|subscription] call. */
+    execution: {
+      sourceComponent: string;
+    };
+    /** An operation result with data. */
+    update: { value: any };
+    /** An operation result with error. */
+    error: { value: any };
+  }
+}
```
kitten, line 16: I'd assume, this file may conflict with the "real" package, and I'm not sure if leaving it will cause issues with the repo wide `tsc` checks

packages/storybook-addon/src/global.d.ts
```diff
@@ -0,0 +1 @@
+declare const __pkg_version__: string;
```
kitten, line 1: unrelated change?

scripts/rollup/plugins.mjs
```diff
@@ -91,8 +91,11 @@ export const makeOutputPlugins = ({ isProduction, extension }) => {
     throw new Error('Missing option `extension` on makeOutputPlugins({ ... })');
 
   return [
-    isProduction && replace({
-      'process.env.NODE_ENV': JSON.stringify('production')
+    replace(isProduction ? {
+      'process.env.NODE_ENV': JSON.stringify('production'),
+      '__pkg_version__': JSON.stringify(settings.pkg.version),
```
kitten, line 96: I lost track of this, but not 100% sure where this would be used in the devtools :thinking:

tsconfig.json
```diff
@@ -9,7 +9,7 @@
         "exchanges/*/src"
       ],
       "@urql/core/*": ["node_modules/@urql/core/src/*", "packages/core/src/*"],
-      "@urql/devtools": ["node_modules/@urql/devtools"],
+      "@urql/devtools": ["node_modules/@urql/devtools", "exchanges/devtools/src"],
```
kitten, line 12: I'm not sure why this entry was already here, haha, just wondering if it's needed given the below entry, since we'd avoid conflicts by removing it

packages/storybook-addon/src/global.d.ts
JoviDeCroock: it broke without it 😅 something in the storybook addon is really flakey when it comes to the devtools exchange

PR #2831, opened 2022-11-25
fix(graphcache): fix loop bug when schema is present

### LINKED ISSUE #2830: When two components use same query and error exists, both components keep fetching indefinitely.
### Describe the bug

**Actual Result**
First component fetches data and receives error. Second component fetches data and and receives error.
First component fetches data again, and Second component fetches data again.
It's keep going like this.

**Expected Result**
First component fetches data and receives error. Second component fetches data and and receives error.
Both components should not fetch anymore.

I think it's have something to do with import { cacheExchange } from "@urql/exchange-graphcache";.
Because If I remove cacheExchange, this issue doesn't occur.

**Reproduce Steps**
1. Open https://codesandbox.io/s/reproduce-urql-issue-same-query-infinite-fetch-5vtduk
2. Click "Show"
3. Inspect Console or Chrome > Inspect > Network
4. Both components will keep fetching

Thank you for any help in advance

### Reproduction

Urql - https://codesandbox.io/s/reproduce-urql-issue-same-query-infinite-fetch-5vtduk Apollo Server - https://codesandbox.io/s/urql-sample-apollo-server-7hy69m

### Urql version

3.0.3

### Validations

JoviDeCroock:
Resolve #2830 

## Summary

When we have a partial result we should stop the query from re-fetching, this was a case that we hadn't caught yet. The issue here being that we will also eat the error as we aren't caching those. Basically the query will complete and we will get two results from cache without the errors rather than the plain result with an error added on top.

### DIFF
--- a/.changeset/cuddly-needles-work.md
+++ b/.changeset/cuddly-needles-work.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-graphcache': patch
+---
+
+Solve issue where partial data could cause loops between related queries
--- a/exchanges/graphcache/e2e-tests/query.spec.tsx
+++ b/exchanges/graphcache/e2e-tests/query.spec.tsx
@@ -1,0 +1,191 @@
+/// <reference types="cypress" />
+
+import * as React from 'react';
+import { mount } from '@cypress/react';
+import {
+  Provider,
+  createClient,
+  useQuery,
+  dedupExchange,
+  debugExchange,
+} from 'urql';
+import { executeExchange } from '@urql/exchange-execute';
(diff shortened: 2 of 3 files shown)


nandorojo:
Thank you for this! Big fix for us.

kitten (approved):
Just gave this another look over; re. simplifying some of the conditions we have across this file, I think it's fine currently, and I don't see anything we could batch with this change.

For context, originally the theory was that `cacheMissOps$`'s filter needed a change, however, `network-only` doesn't even arrive there due to the differentiation via `nonCacheOps$`.

PR #2833, opened 2022-11-26
fix(next-urql): Change import for init-urql-client to @urql/core to prevent it from using createContext

SleeplessOne1917:
## Summary

Currently, `initUrqlClient` cannot be used in Nextjs server components because it uses `createContext`. Per [a discussion](https://github.com/urql-graphql/urql/discussions/2829#discussioncomment-4234974) with @JoviDeCroock the other day, the solution to this seems to be to import from @urql/core instead of urql.

## Set of changes
The only change is changing the import for `init-urql-client.ts` from urql to @urql/core. This should not be a breaking change.

### DIFF
--- a/.changeset/beige-fishes-relate.md
+++ b/.changeset/beige-fishes-relate.md
@@ -1,0 +1,5 @@
+---
+'next-urql': patch
+---
+
+Change import for `createClient` to `@urql/core`, which helps Next not depend on `urql` and hence not cause `createContext` to be called when the import is treeshaken away.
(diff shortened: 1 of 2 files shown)


kitten:
Hiya :wave:
just leaving a note here that the changeset was missing, hence we didn't take a look at this earlier. I pushed a changeset here however for you, so let's get this across the line :v:
For reference, [there's some instructions here](https://github.com/urql-graphql/urql/blob/main/CONTRIBUTING.md#how-do-i-document-a-change-for-the-changelog) on how to add one, for the future.
Cheers! 🙌

kitten (approved):

PR #2860, opened 2022-12-06
chore(workspace): Fix detection of non-ESM builds

kitten:
This is a tiny issue that was introduced when adding missing dependencies in `@urql/exchange-graphcache`. Our non-ESM build mode is activated when React is installed, however, instead, we shouldn't check `devDependencies` for that, since currently the E2E tests require us to have it installed in `@urql/exchange-graphcache`.

### DIFF
--- a/scripts/rollup/settings.mjs
+++ b/scripts/rollup/settings.mjs
@@ -43,7 +43,7 @@ if (pkg.optionalDependencies)
 
 const prodDependencies = new Set([
   ...Object.keys(pkg.peerDependencies || {}),
-  ...Object.keys(pkg.dependencies),
+  ...Object.keys(pkg.dependencies || {}),
 ]);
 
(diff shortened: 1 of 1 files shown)


scripts/rollup/settings.mjs
```diff
@@ -41,6 +41,11 @@ if (pkg.dependencies)
 if (pkg.optionalDependencies)
   externalModules.push(...Object.keys(pkg.optionalDependencies));
 
+const prodDependencies = new Set([
+  ...Object.keys(pkg.peerDependencies),
+  ...Object.keys(pkg.dependencies),
```
JoviDeCroock, line 46:
```suggestion
  ...Object.keys(pkg.dependencies || {}),
```

JoviDeCroock (approved):

PR #2861, opened 2022-12-06
fix(graphcache): Fix deadlocked layers between deferred and optimistic layers

### LINKED ISSUE #2853: Subscription doesn't update records after mutation
### Describe the bug

Hello!

I believe I found a bug in urql subscriptions. I prepared a simple list, where users can add new records and rename existing ones. After a user creates a new record and after that they hit "Rename" button, updated results won't show up. The "Rename" button calls standalone gql client (not urql) just to simulate external source of change -- this way I'm sure urql doesn't know about mutation been called and so urql must update cache from subscription.

I also attached video: 

https://user-images.githubusercontent.com/1828134/205616233-c8c9d41e-e772-438f-a23f-c262d41e247e.mov

Or am I doing something wrong?

### Reproduction

https://github.com/cinan/urql-cache

### Urql version

@urql/exchange-graphcache: 5.0.5
urql: 3.0.3

### Validations

kitten:
Resolve #2853

## Summary

This fixes a deadlock that could occur in the layers of Graphcache, when a deferred layer (subscriptions or otherwise `hasNext` layers) would be moved past previously optimistic layers (like mutations). In such cases, it would become impossible for preceding layers (i.e. mutations) to ever squash and move "out of the way" causing a deadlock that'd make it impossible for the deferred layer's data to show up when the two layers had overlaps.

Instead, we now will try to clear out deferred layers more aggressively. This means that we will squash its data even when it's deferred, when data is written.

Furthermore, we then move the layer from its current position (or recreate it) when it's re-registered, i.e. when a new message/update comes in for the deferred layer.

This should keep the deferred layers data in the cache but still bump it as needed.

## Set of changes

- Allow deferred layers to be squashed
- Allow layers to only move past empty ones from its current position in the `optimisticOrder` list

### DIFF
--- a/exchanges/graphcache/src/cacheExchange.ts
+++ b/exchanges/graphcache/src/cacheExchange.ts
@@ -213,11 +213,8 @@ export const cacheExchange = <C extends Partial<CacheExchangeOpts>>(
       optimisticKeysToDependencies.delete(operation.key);
     }
 
-    reserveLayer(
-      store.data,
-      operation.key,
-      operation.kind === 'subscription' || result.hasNext
-    );
+    if (operation.kind === 'subscription' || result.hasNext)
+      reserveLayer(store.data, operation.key, true);
 
     let queryDependencies: void | Dependencies;
     let data: Data | null = result.data;
--- a/exchanges/graphcache/src/store/data.ts
+++ b/exchanges/graphcache/src/store/data.ts
@@ -161,11 +161,9 @@ export const clearDataState = () => {
     while (
       --i >= 0 &&
       data.refLock.has(data.optimisticOrder[i]) &&
-      data.commutativeKeys.has(data.optimisticOrder[i]) &&
-      !data.deferredKeys.has(data.optimisticOrder[i])
-    ) {
+      data.commutativeKeys.has(data.optimisticOrder[i])
+    )
       squashLayer(data.optimisticOrder[i]);
-    }
   }
 
   currentOwnership = null;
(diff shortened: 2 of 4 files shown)


exchanges/graphcache/src/cacheExchange.ts
```diff
@@ -213,11 +213,8 @@ export const cacheExchange = <C extends Partial<CacheExchangeOpts>>(
       optimisticKeysToDependencies.delete(operation.key);
     }
 
-    reserveLayer(
-      store.data,
-      operation.key,
-      operation.kind === 'subscription' || result.hasNext
-    );
+    if (operation.kind === 'subscription' || result.hasNext)
+      reserveLayer(store.data, operation.key, true);
```
kitten, line 217: This meant that we were:
- unnecessarily adding layers to `optimisticOrder` even when that wasn't needed
- where shifting non-deferred layers that already were registered before, which is unexpected. Our tests are relying on this though, so we can "cheat" here

exchanges/graphcache/src/store/data.ts
```diff
@@ -527,36 +525,34 @@ export const reserveLayer = (
...
+  // the order if it exists already
+  let index = data.optimisticOrder.indexOf(layerKey);
+  if (index > -1) data.optimisticOrder.splice(index, 1);
+
   if (hasNext) {
     data.deferredKeys.add(layerKey);
+    // If the layer has future results then we'll move it past any layer that's
+    // still empty, so currently pending operations will take precedence over it
+    for (
+      index = index > -1 ? index : 0;
```
kitten, line 538: This is the relevant changed line. We shift the layer based on where it was before if it's still around, rather than letting it "hop back in front"

exchanges/graphcache/src/store/data.ts
```diff
@@ -161,11 +161,9 @@ export const clearDataState = () => {
     while (
       --i >= 0 &&
       data.refLock.has(data.optimisticOrder[i]) &&
-      data.commutativeKeys.has(data.optimisticOrder[i]) &&
-      !data.deferredKeys.has(data.optimisticOrder[i])
-    ) {
+      data.commutativeKeys.has(data.optimisticOrder[i])
```
kitten, line 164: Here, as discussed, we realised that deferred layers should 100% be squasheable. We basially put deferred layers in the right place, write to them, and then ... we assume that they're never squasheable.

This means that for subscriptions (or other neverending result streams) this layer never goes away, and pending layers that then were on top can never go away either, and have the potential to infinitely be "in the way" of old deferred layers.

JoviDeCroock (approved):

JoviDeCroock:
As a follow up we could create an e2e test but wouldn't call it blocking for this PR as we can publish without

PR #2865, opened 2022-12-08
chore(actions): Remove frozen lockfile check from Release Workflow

kitten:


.github/workflows/release.yml
```diff
@@ -41,7 +41,7 @@ jobs:
             ${{ runner.os }}-pnpm-
 
       - name: Install Dependencies
-        run: pnpm install --frozen-lockfile --prefer-offline
+        run: pnpm install
```
ryan-roemer, line 44: Not sure if this is based in the fact that a changeset version impacts the lockfile, what we did in victory for this scenario (after we migrated to pnpm) was:

1. Add a lockfile update to our `version` script: https://github.com/FormidableLabs/victory/blob/main/package.json#L118 (`pnpm changeset version && pnpm install --no-frozen-lockfile`)
2. Call that for changeset gh action flow `version` field.

Hope that helps if that's related!

ryan-roemer: Oh, but remembering this more, these flags aren't needed in CI as it will check for frozen lockfile by default... 😂

Issue #2869, opened 2022-12-09
__typename is missing after upgrading from @urql/core 3.0.5 to 3.1.0


mech:
### Describe the bug

After upgrading, I noticed all my `__typename` are gone. Do I need to explicitly specify them in my query?

This work on 3.0.5 without specifying `__typename` in the selection:

```
query {
  notes {
    id
    name
  }
}
```

From the default test codesandbox you can see the `__typename` is missing.

### Reproduction

https://codesandbox.io/s/friendly-engelbart-sb00j4?file=/package.json

### Urql version

urql = 3.0.3
@urql/core = 3.1.0

### Validations

- [X] I can confirm that this is a bug report, and not a feature request, RFC, question, or discussion, for which [GitHub Discussions](https://github.com/urql-graphql/urql/discussions) should be used
- [X] Read the [docs](https://formidable.com/open-source/urql/docs/).
- [X] Follow our [Code of Conduct](https://github.com/urql-graphql/urql/blob/main/CODE_OF_CONDUCT.md)

mech:
<img width="1054" alt="fc6f3_16_18_50" src="https://user-images.githubusercontent.com/5850/206656802-637249f8-cff1-4ab2-82bc-a6f5ab69c376.png">

JoviDeCroock:
Fixed by #2871

PR #2871, opened 2022-12-09
fix(core): Fix regression causing stringifyDocument to not reprint after formatDocument

JoviDeCroock:
## Summary

Related to #2869 reproduces the issue.

[this](https://github.com/urql-graphql/urql/blob/main/packages/core/src/utils/request.ts#L43-L53) looks to be the problem, when we use `createRequest` we fill in `gql` as the source name and don't update the printed string when looping over it with `formatDocument`. When we use `gql` the name will be `graphql request` hence not hitting this clause.

The thing concerning me here is if people would be implementing local-only exchanges whether that would mess with i.e. them removing a field when seeing `@local`

### DIFF
--- a/packages/core/src/utils/typenames.ts
+++ b/packages/core/src/utils/typenames.ts
@@ -4,6 +4,7 @@ import {
   InlineFragmentNode,
   Kind,
   visit,
+  print,
 } from 'graphql';
 
 import { KeyedDocumentNode, keyDocument } from './request';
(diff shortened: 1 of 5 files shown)


packages/core/src/utils/typenames.ts
```diff
@@ -83,6 +84,11 @@ export const formatDocument = <T extends DocumentNode>(node: T): T => {
     });
 
     formattedDocs.set(query.__key, result);
+    if (typeof node !== 'string' && node.loc?.source.name === 'gql') {
+      const printed = print(result);
```
kitten, line 88: I'm not really sure about this approach 🤔 I see why this happens, i.e. if both sides are stringifier, but I think the deep update and print here seems a little odd to me.

kitten (approved):

PR #2909, opened 2022-12-27
fix(next): add empty pageprops object

### LINKED ISSUE #2892: On initial render `pageProps` are `undefined` rather than the expected empty object.
### Describe the bug

I was unable to use [next-i18next](https://github.com/i18next/next-i18next) with `withUrqlClient` due to `pageProps` being set to `undefined` on the initial render rather than the expected empty object. This caused error:

> Cannot read properties of undefined (reading '_nextI18Next')

Next.js [documentation](https://nextjs.org/docs/advanced-features/custom-app) states:

> pageProps is an object with the initial props that were preloaded for your page by one of our [data fetching methods](https://nextjs.org/docs/basic-features/data-fetching/overview), otherwise, it's an empty object.

And reviewing their code [`_app.tsx` code](https://github.com/vercel/next.js/blob/v13.0.7/packages/next/pages/_app.tsx#L30) indicates the same.

I believe the [line 163](https://github.com/urql-graphql/urql/blob/next-urql%404.0.2/packages/next-urql/src/with-urql-client.ts#L163) should change from:
```tsx
const appTreeProps = isApp ? props : { pageProps: props };
```
to
```tsx
const appTreeProps = isApp ? { ...props, pageProps: {} } : { pageProps: props };
```
or similar.

This appears to solve the issue and be more inline with the Next.js documentation.

A workaround I am employing for now until this gets addressed is to explicitly set pageProps:

```tsx
App.getInitialProps = async () => {
  return {
    pageProps: {},
  };
};
```

This is related to comment by @nruhe in https://github.com/urql-graphql/urql/issues/2573#issuecomment-1347597208

### Reproduction

https://codesandbox.io/p/sandbox/jovial-beaver-7sos8o

### Urql version

i18next@^22.4.5
next@13.0.0
next-i18next@^13.0.2
next-urql@^4.0.2
react@18.2.0
react-dom@18.2.0
react-i18next@^12.1.1
react-is@^18.2.0
urql@^3.0.

### Validations

JoviDeCroock:
Resolves #2892
Closes #2574

## Summary

This adds an empty `pageProps` object, the reason behind putting it before the spread is that if we have some HOC wrapping `next-urql` we don't want to override the existing pageProps.

### DIFF
--- a/.changeset/rare-panthers-sit.md
+++ b/.changeset/rare-panthers-sit.md
@@ -1,0 +1,5 @@
+---
+'next-urql': patch
+---
+
+Add empty pageProps object
--- a/.changeset/rare-panthers-sit.md
+++ b/.changeset/rare-panthers-sit.md
@@ -2,4 +2,4 @@
 'next-urql': patch
 ---
 
-Add empty pageProps object
+Add `pageProps: {}` entry to props on app components
--- a/packages/next-urql/src/with-urql-client.ts
+++ b/packages/next-urql/src/with-urql-client.ts
@@ -160,7 +160,9 @@ export function withUrqlClient(
         }
 
         const props = { ...pageProps, urqlClient };
-        const appTreeProps = isApp ? props : { pageProps: props };
+        const appTreeProps = isApp
+          ? { pageProps: {}, ...props }
+          : { pageProps: props };
 
         // Run the prepass step on AppTree. This will run all urql queries on the server.
         if (!options!.neverSuspend) {


kitten:
is this a duplicate of #2574?

JoviDeCroock:
@kitten a bit different as it only focusses on resolving the empty object issue 😅 but yes feel free to merge the other one if you prefer

kitten:
Ok, I see, I'm not inclined to add `getAppProps` since it's non-standard and would have to be documented, but can simply be replaced with an inner function being composed in. Since I haven't seen this before, lets' first merge this and discuss the alternative separately, since this is the less "committed" change, cc @Swiftwork

kitten (approved):

.changeset/rare-panthers-sit.md
```diff
@@ -0,0 +1,5 @@
+---
+'next-urql': patch
+---
+
+Add empty pageProps object
```
kitten, line 5:
```suggestion
Add `pageProps: {}` entry to props on app components
```

PR #2910, opened 2022-12-28
(request-policy) - configurable upgrade policy and operation persistency fix

kurkle:
## Summary

These are changes I needed to make for this to be usable with React and ssr. Makes sense?

### Allow configuring the upgraded requestPolicy

`ssrExchange` does not pass anything other than `network-only`, so needed to use that.

### Persist the lastOccurrence between calls

Seems like the exchange did not persist the lastOccurrence between calls, and thus always upgraded

### DIFF
--- a/exchanges/request-policy/src/requestPolicyExchange.test.ts
+++ b/exchanges/request-policy/src/requestPolicyExchange.test.ts
@@ -46,7 +46,7 @@ beforeEach(() => {
   ({ source: ops$, next } = makeSubject<Operation>());
 });
 
-it(`upgrades to cache-and-network`, async () => {
+it(`upgrades to cache-and-network by default`, async () => {
   const response = vi.fn(
     (forwardOp: Operation): OperationResult => {
       return {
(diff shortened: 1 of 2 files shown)


kitten:
I'm sorry there's not really an RFC and reasoning here to go on and I don't see why we'd add this. I'm following some of the reasoning here, potentially, but I believe, a) you may be better served making this a custom exchange, and b) we don't really provide options that may be confusing or don't further specify when it's obvious that they should be used.

In this case, we specifically chose sane defaults, and while we don't currently have an exchange that does everything "all in one" that we'd want that replaces this exchange (e.g. serialising mutations, tab-focus, cross-tab sync'ing and such), this is a very simple exchange that just conveys a pattern and alternative to something we don't recommend, which is to always use / change the default policy to `cache-and-network`

For now, with this reasoning, I'll close this PR. Sorry for the delay on writing this reasoning up.
Since we don't have an RFC issue, if you'd like, we can continue discussing this here though, if you think this should be reopened or you have a more concrete problem you want to talk about that you tried to solve. Cheers! :heart:

PR #2922, opened 2023-01-05
fix(core): Type response of CombinedError

inokawa:
## Summary

This PR changes type of `response` of `CombinedError` from `any` to `Response`.

I'm not sure this fix is OK but typed response is useful when check it in exchanges.

```ts
authExchange({
  ...
  didAuthError({ error }) {
    return error.response?.status === 401;
  }
});
```

## Set of changes

- Type response of CombinedError in `@urql/core`

### DIFF
--- a/.changeset/quick-chefs-happen.md
+++ b/.changeset/quick-chefs-happen.md
@@ -1,0 +1,5 @@
+---
+'@urql/core': patch
+---
+
+Type response of CombinedError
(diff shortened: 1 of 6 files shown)


kitten:
I'm unfortunately not too inclined to accept this. In urql, we basically don't assume that a fetch exchange will be the only way to communicate with a GraphQL API. While it's the default, we try to avoid any further assumptions on how a `Client` gets to an exchange that can get a response. An example would be the `subscriptionExchange`, which can also be used for queries and mutations with other transports.

In many instance, furthermore, the `Response` object isn't much more reusable beyond the status code once we've read out its body data.

We basically provide the optional `response` property because it's useful for debugging, and, in some cases, to read out the status code of an HTTP response (well, via the fetch API)

So, we've typed this as `response?: any` to discourage frequent use of it and make it clear that this could be absolutely _anything_. An implementor of a new transport or other exchange could decide to put another object into this property that is more relevant for their transport, but I'd rather not type it as `Response` and make this a "stable" property that's guaranteed ✌️

inokawa:
> So, we've typed this as response?: any to discourage frequent use of it and make it clear that this could be absolutely anything. 

I understood this design decision.
Thank you for your very kind explanation.

PR #2928, opened 2023-01-17
docs(showcase): add SSE urql exchange package

notrab:

### DIFF
--- a/docs/showcase.md
+++ b/docs/showcase.md
@@ -64,3 +64,4 @@ maintainers and supporters. This page is specifically dedicated to all of you!
   that helps you make typesafe hooks/components with urql.
 - [`urql-custom-scalars-exchange`](https://github.com/clentfort/urql-custom-scalars-exchange): An exchange
   to automatically convert scalars.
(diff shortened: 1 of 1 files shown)


kitten (approved):

notrab:
Thanks @kitten ❤️

kitten:
No worries! 🙌 Very excited for the first service, Grafbase, to my knowledge proliferating live queries, finally!

I'm currently not sure whether it complies with the `hasNext` flag and such in its current implementation, and was a little surprised to see  the JSON patch responses rather than the Incremental Delivery payloads as laid out in the GraphQL spec proposals 🤔

I'm hoping overlap here will increase over time, since this allows us to potentially support SSE for `@live` queries built into the core library.

But happy to learn more soon to dig into this. Currently, it seems the custom exchange is a good choice since multiple transports are obviously possible. I was just surprised to see there's less overlap than expected to `@defer`/`@stream`'s reference payloads & implementation 👀

notrab:
@kitten thanks! I believe we're the first to provide Live Queries with packages like this exchange that make it a breeze to adopt. I'm excited!

That said, we're still early days with Live Queries. We plan to add an arg to the `@live(...)` directive so we should be able to cover a range of different implementations.

We'll soon explore `@stream`/`@defer` with the above considerations too.

PR #2945, opened 2023-02-03
fix(graphcache): Fix onOnline data loss bug in offlineExchange

trcoffman:
## Summary

It's important that onOnline isn't established until after mutations are read from storage and processed by flushQueue. If the storage adapter's onOnline implementation calls flushQueue in the same task, then the storage will be wiped before it is read from, and all mutations that were in the metadata storage will be deleted before they can be read and processed.

Our team ran into this in our react-native project. The library we use for listening to network information state changes immediately invokes the event listener we install on it. Therefore, it's calling flushQueue before urql's offlineExchange even attempts to read from metadata storage.

### @urql/exchange-graphcache:
- Change how onOnline is set up in offlineExchange. Wait to install the onOnline callback on the StorageAdapter until after the mutations have been read from the StorageAdapter's metadata store.

### DIFF
--- a/exchanges/graphcache/src/offlineExchange.ts
+++ b/exchanges/graphcache/src/offlineExchange.ts
@@ -150,7 +150,7 @@ export const offlineExchange = <C extends Partial<CacheExchangeOpts>>(
           flushQueue();
         }
       })
-      .finally(() => storage?.onOnline?.(flushQueue));
+      .finally(() => storage.onOnline!(flushQueue));
 
     const cacheResults$ = cacheExchange({
       ...opts,
--- a/exchanges/graphcache/src/offlineExchange.ts
+++ b/exchanges/graphcache/src/offlineExchange.ts
@@ -134,21 +134,23 @@ export const offlineExchange = <C extends Partial<CacheExchangeOpts>>(
       );
     };
 
-    storage.onOnline(flushQueue);
-    storage.readMetadata().then(mutations => {
-      if (mutations) {
-        for (let i = 0; i < mutations.length; i++) {
-          failedQueue.push(
-            client.createRequestOperation(
-              'mutation',
-              createRequest(mutations[i].query, mutations[i].variables)
-            )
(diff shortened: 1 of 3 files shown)


trcoffman:
It would appear that I have to fix the unit tests first before this is ready to review. Hang tight.

exchanges/graphcache/src/offlineExchange.ts
```diff
@@ -134,21 +134,23 @@ export const offlineExchange = <C extends Partial<CacheExchangeOpts>>(
...
+            );
+          }
 
-        flushQueue();
-      }
-    });
+          flushQueue();
+        }
+      })
+      .finally(() => storage?.onOnline?.(flushQueue));
```
kitten, line 153:
```suggestion
      .finally(() => storage.onOnline!(flushQueue));
```

kitten: The check is redundant as we're already checking for `storage.onOnline` prior, and this is just TypeScript checking in a new scope, so we can guarantee that this is available

kitten (approved):
This is a really nice catch! Cheers! 🙌 
Not sure how this slipped through the net, but I'm glad y'all found this and decided to submit a fix ❤️ 

Just a quick note; Our PRs require a changeset. This basically is a file that just contains some metadata (i.e. `@urql/exchange-graphcache` being marked for a patch release) and a small entry for our changelog.
We have instructions on how to submit this here: https://github.com/urql-graphql/urql/blob/main/CONTRIBUTING.md#how-do-i-document-a-change-for-the-changelog

This would only require a small description, e.g.
> Fix `offlineExchange` calling `storage.onOnline` before the initial `storage.readMetadata` promise resolved, which could cause the saved queue of mutations to be erased prematurely.

I can get this pushed for you though, if you don't get to it, so no worries! I'll check back in

JoviDeCroock (approved):

PR #2951, opened 2023-02-10
feat(persisted-fetch): Add enableForMutation option

### LINKED ISSUE #1287: RFC: Support Persisted Queries for Mutations
## Summary
Currently the `@urql/exchange-persisted-fetch` only supports queries. Enabling persisted queries for mutations has a couple of advantages:

### Security

One of the challenges of GQL is that an API is exposed for arbitrary requests. The flexibility in the query language allows bad actors to try DDoS attacks by generating big expensive queries. By leveraging persisted queries at build time you can effectively limit the combinations and defend against these types of attacks. 

https://leapgraph.com/graphql-api-security#persist-queries

### Efficiency

While queries are usually larger than mutations there is still small improvements in not sending that information in the request. Here flexibility is the enabler, in use cases where mutations for business reasons have to be big in size persisted mutations could be a solution to reduce latency.

### Apollo-client compatibility

Apollo-client supports persisted queries for mutations and providing the same feature would help developers that are thinking to port projects into URQL.

## Proposed Solution

Enable mutations in the Persisted Fetch Exchange. It could be added as an opt in feature to provide backwards compatibility to projects that don’t need mutations to be hashed.

geekuillaume:
Resolve #1287

## Summary

This PR adds the `enableForMutation` option to `PersistedFetchExchangeOptions` to allow running the persisted operations logic on mutation in addition to queries.

## Set of changes

- Add `enableForMutation` option to `PersistedFetchExchangeOptions` (optional and default false to keep backward compatibility)
- Check `enableForMutation` option when filtering operations in the exchange


exchanges/persisted-fetch/src/persistedFetchExchange.ts
```diff
@@ -46,13 +47,14 @@ export const persistedFetchExchange = (
...
   const hashFn = options.generateHash || hash;
+  const enableForMutation = options.enableForMutation || false;
   let supportsPersistedQueries = true;
 
   return ops$ => {
     const sharedOps$ = share(ops$);
     const fetchResults$ = pipe(
       sharedOps$,
-      filter(operation => operation.kind === 'query'),
+      filter(operation => enableForMutation || operation.kind === 'query'),
```
JoviDeCroock, line 57:
```suggestion
      filter(operation => (enableForMutation && operation.kind === 'mutation') || operation.kind === 'query'),
```

exchanges/persisted-fetch/src/persistedFetchExchange.ts
```diff
@@ -128,7 +130,7 @@ export const persistedFetchExchange = (
 
     const forward$ = pipe(
       sharedOps$,
-      filter(operation => operation.kind !== 'query'),
+      filter(operation => !enableForMutation && operation.kind !== 'query'),
```
JoviDeCroock, line 133:
```suggestion
      filter(operation => (enableForMutation && operation.kind !== 'mutation') || operation.kind !== 'query'),
```

JoviDeCroock (commented):
A few things that stood out during review, will talk about this with a co-maintainer as we normally start with an RFC before introducing changes like this as detailed in the [contributing](https://github.com/urql-graphql/urql/blob/main/CONTRIBUTING.md#how-do-i-propose-changes)

exchanges/persisted-fetch/src/persistedFetchExchange.ts
geekuillaume: I refactored this part of the code to use a single `operationFilter` method. This will prevent mismatch between the filter in and filter out logic of the exchange

exchanges/persisted-fetch/src/persistedFetchExchange.ts
```diff
@@ -46,13 +47,18 @@ export const persistedFetchExchange = (
   const preferGetForPersistedQueries = !!options.preferGetForPersistedQueries;
   const enforcePersistedQueries = !!options.enforcePersistedQueries;
   const hashFn = options.generateHash || hash;
+  const enableForMutation = options.enableForMutation || false;
```
kitten, line 50:
```suggestion
  const enableForMutation = !!options.enableForMutation;
```

kitten (approved):
Looks good to me! 👍 Thanks for sending a detailed PR. I just have a small nit to add.
Checks are approved and running now.

JoviDeCroock (approved):

PR #2966, opened 2023-02-19
Update retryExchange.ts

DoisKoh:
## Summary

https://formidable.com/open-source/urql/docs/advanced/retry-operations/#:~:text=don%27t%20want%20to-,infinitely%20attempt,-an

The documentation implies that not passing in a `maxNumberAttempts` will result in retries occurring indefinitely. This is actually behaviour that I'm trying to get, but can't get because the code is not actually doing that - it defaults to `2`.

This simple modification allows for that behavior and is also consistent with the documentation.

## Set of changes

- Removed default of '2' to `maxNumberAttempts`
- If `maxNumberAttempts` is omitted, it will always proceed to retry

Note: People relying on the original behaviour might see have issues with their, so I guess this is a breaking change, though I believe it is more in line with what is expected.

### DIFF
--- a/exchanges/retry/src/retryExchange.ts
+++ b/exchanges/retry/src/retryExchange.ts
@@ -43,7 +43,7 @@ export const retryExchange = ({
 }: RetryExchangeOptions): Exchange => {
   const MIN_DELAY = initialDelayMs || 1000;
   const MAX_DELAY = maxDelayMs || 15000;
-  const MAX_ATTEMPTS = maxNumberAttempts || 2;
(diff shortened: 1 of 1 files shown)


kitten (commented):
I'm inclined not to accept this currently as it's an unnecessary breaking change, so I'm much more inclined to accept a PR that edits the docs 😅 Sorry!

It's much easier to pass `Infinity` or a high number than us introducing a breaking change that'll need to bump a major for what's ultimately an issue in the docs

DoisKoh:
Alright though I feel that the behaviour currently listed in the docs is more appropriate...

PR #2972, opened 2023-02-22
docs(core): Update type of OperationContext's preferGetMethod

inokawa:
## Summary

Updated type of OperationContext's preferGetMethod to be the same as Client's preferGetMethod.

## Set of changes

Docs only.


kitten (approved):
Nice catch! Cheers 🙌

PR #2978, opened 2023-02-24
fix(graphcache): Prevent cache writes during query operations

### LINKED ISSUE #2977: (graphcache): Writing to the cache fails during cache read operations
### Describe the bug

# Context
I am migrating an app from REST to GraphQL with URQL. I have run into a situation where calling `cache.invalidate` is throwing an undefined error in [data#setNode](https://github.com/urql-graphql/urql/blob/main/exchanges/graphcache/src/store/data.ts#L265). 

I am attempting to invalidate a paginated query response in a separate resolver. Here is the basic flow

## Reproduction

1. Query for a page's data, which includes a relay-compliant list of data. 
2. Create a ref to a specific field resolver in another query that is not linked to the paginated list result. When this value changes, we need to invalidate the list. We can't do this in a mutation because we are migrating to GraphQL, and the operations that would invalidate our list are not in the graph yet. When this ref changes, we know that the values in the list have changed.
3. In a simple `number` field resolver, I check if the resolved value from `parent.value` is different from the ref, and then call
    ```
    cache.invalidate(
        'Query',
        cache.keyOfField(...)
      );
    ```

When calling invalidate, `setNode` has `currentOptimisticKey` set, but `map.optimistic` is empty.  

### Reproduction

https://www.loom.com/share/d70f14f8dd76430190ed425e334210d4

### Urql version

3.0.3

### Validations

kitten:
Resolves #2977 

## Summary

This adds the missing invariants to `data.ts` that prevent cache writes during read/query operations, since users are not allowed to modify the cache during read operations.

## Set of changes

- Add dev-only invariant to prevent writes inside cache reads
- Exclude `squashLayer` operations from this restriction
- Update tests accordingly

### DIFF
--- a/docs/graphcache/errors.md
+++ b/docs/graphcache/errors.md
@@ -395,3 +395,17 @@ This error occurs when you are using an `interface` or `union` rather than an
 implemented type for these.
 
 Check the type mentioned and change it to one of the specific types.
+
+## (27) Invalid Cache write
+
+> Invalid Cache write: You may not write to the cache during cache reads.
+> Accesses to `cache.writeFragment`, `cache.updateQuery`, and `cache.link` may
+> not be made inside `resolvers` for instance.
+
+If you're using the `Cache` inside your `cacheExchange` config you receive it
+either inside callbacks that are called when the cache is queried (e.g.
+`resolvers`) or when data is written to the cache (e.g. `updates`). You may not
+write to the cache when it's being queried.
+
+Please make sure that you're not calling `cache.updateQuery`,
+`cache.writeFragment`, or `cache.link` inside `resolvers`.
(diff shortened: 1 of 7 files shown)


docs/graphcache/errors.md
```diff
@@ -395,3 +395,17 @@ This error occurs when you are using an `interface` or `union` rather than an
 implemented type for these.
 
 Check the type mentioned and change it to one of the specific types.
+
+## (27) Invalid Cache write
```
PrescottJRynewicz, line 399: Should we also add some notes in the Local Resolvers section of the graphcache docs? That is where I want to initially find if this was an invariant

PrescottJRynewicz (commented):
All makes sense to me - just one nit pick about comments

JoviDeCroock (approved):

docs/graphcache/errors.md
kitten: That's a good point 👍

PR #2994, opened 2023-03-05
fix(multipart-fetch): support passing in custom file detector function to `multipartFetchExchange`

giseburt:
## Summary

Making a client in a VSCode extension, which runs in the node environment where `File` and `FormData` don't exist, the `multipartFetchExchange` uses the default `extract-files` call `extractFiles` without passing in a `isExtractable` function, which leaves the default one in place, where neither `File` nor `Blob` are available globally (again, in the Node environment).

## Set of changes

In the patch I adjusted `multipartFetchExchange` to be `multipartFetchExchangeWithOptions()`, with a new optional parameters list. Old code will work as-is since `multipartFetchExchange` still yields the same results, but now you can pass in a function that tests for extractable file objects (from the updated docs):

```typescript
import { createClient, dedupExchange, cacheExchange } from 'urql';
import { multipartFetchExchange } from '@urql/exchange-multipart-fetch';

export class CustomFile extends Blob {
  constructor(sources: Array<Blob>, public readonly name: string) {
    super(sources);
  }
}
export const isCustomFile = (value: unknown | undefined) =>
  value instanceof CustomFile;

const client = createClient({
  url: 'http://localhost:3000/graphql',
  exchanges: [
    dedupExchange, 
    cacheExchange, 
    multipartFetchExchangeWithOptions({ customFileCheck: isCustomFile })
  ],
});
```

I also adjusted the tests to include the new functionality (leaving the old tests in place, no regressions here 😄), and updated the docs as well. The tests were only using snapshots and didn't catch that the resulting `body` was not a proper `FromData` object so I added that test as well.

### DIFF
--- a/exchanges/multipart-fetch/src/multipartFetchExchange.ts
+++ b/exchanges/multipart-fetch/src/multipartFetchExchange.ts
@@ -9,7 +9,13 @@ import {
   makeFetchSource,
 } from '@urql/core/internal';
 
-export const multipartFetchExchange: Exchange = ({
+export interface MultipartFetchExchangeOpts {
+  customFileCheck?: (maybeFile: unknown | undefined) => boolean;
+}
+
+export const multipartFetchExchangeWithOptions = ({
+  customFileCheck,
+}: MultipartFetchExchangeOpts = {}): Exchange => ({
   forward,
   dispatchDebug,
 }) => ops$ => {
@@ -26,9 +32,13 @@ export const multipartFetchExchange: Exchange = ({
       );
 
       // Spreading operation.variables here in case someone made a variables with Object.create(null).
-      const { files, clone: variables } = extractFiles({
-        ...operation.variables,
-      });
+      const { files, clone: variables } = extractFiles(
+        {
+          ...operation.variables,
+        },
+        '',
+        customFileCheck
+      );
       const body = makeFetchBody({ query: operation.query, variables });
 
       let url: string;
@@ -104,3 +114,5 @@ export const multipartFetchExchange: Exchange = ({
 
   return merge([fetchResults$, forward$]);
 };
+
+export const multipartFetchExchange: Exchange = multipartFetchExchangeWithOptions();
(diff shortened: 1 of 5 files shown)


exchanges/multipart-fetch/src/multipartFetchExchange.ts
```diff
@@ -9,7 +9,13 @@ import {
...
 } from '@urql/core/internal';
 
-export const multipartFetchExchange: Exchange = ({
+export interface MultipartFetchExchangeOpts {
+  customFileCheck?: (maybeFile: unknown | undefined) => boolean;
+}
+
+export const multipartFetchExchangeWithOptions = ({
+  customFileCheck,
+}: MultipartFetchExchangeOpts = {}): Exchange => ({
```
kitten, line 18: other exchanges have also gone through this breaking change, but maybe it's worth considering making `multipartFetchExchangeWithOptions` the only available API and naming it `multipartFetchExchange` (i.e. making that the factory function)

That'd also give us an opportunity to maybe look at refactoring some of it, since this librar hasn't had a "true" breaking change almost since its creation https://github.com/urql-graphql/urql/blob/main/exchanges/multipart-fetch/CHANGELOG.md

giseburt: Well, I'm good with it either way, honestly. 😄

If you decide you'd prefer a breaking change to get the API cleaner, I'll happily make that adjustment.

kitten:
> where neither File nor Blob are available globally (again, in the Node environment).

Just following up on this; I'm currently investigating adding multipart support to `@urql/core`, i.e. it'd not be an opt-in feature anymore, but built-in, and am looking at actually not implementing an option for this and stumbled upon the statement above.

This isn't strictly speaking true, I believe. `Blob` is available as a global in Node.js, while `File` isn't. However, `Blob` can still be used in place, is available globally, and is equivalent to `require('buffer').Blob` in Node as well.
The addition seems to have taken place in Node 14.18 and 15.7, so the LTS does seem to have it, and has been unmarked as experimental in Node 16.7 and 18.

kitten:
Update: This will be closed by #3051 for now, since it changes where this is implemented.

PR #3012, opened 2023-03-07
refactor(auth): Implement new initializer function API

### LINKED ISSUE #2815: RFC (auth): separate initial and refresh auth
## Summary

In the `auth-exchange`, we currently have one function - `getAuth` that handles fetching both initial auth as well as refreshing the auth in case of errors.

This proposal is to separate this function into two separate functions: one to handle the initial auth only, and the other to handle all subsequent calls.

These should be separated because currently users will have to keep track of their own auth state to know when to fetch the token from storage and when to trigger the refresh flow, and it's easy to make mistakes.

Furthermore, "get initial auth" and "refresh auth" are very different actions from a library user's point of view andit would reduce the mental load to keep them separated.

## Proposed Solution

Separate the `getAuth` into two functions:
1. `getInitialAuth`: called only once when the exchange first loads. It should be used to get any initial auth state (e.g. form a cookie or async storage)
2. `refreshAuth`: called when `willAuthError` returns true, or when a request fails with an auth error (as determined by `didAuthError`)

## Requirements

The proposed changes can be seen in this PR: https://github.com/urql-graphql/urql/pull/2814

kitten:
Resolves #2815

## Summary

**This is a breaking change, which updates the `authExchange` API.**

This PR eliminates the need for an `authState` (i.e. the authentication state) to be kept with the `authExchange`, and moves it into the responsibility of the user.
It also eliminates `getAuth`'s "double duty" of being called, both, to initialize the authentication state, and to refresh the authentication state after an authentication error.

The `authExchange` now accepts an initializer function, which is an async function (or a function returning a promise). This function is user-defined and will receive a utilities object and _must_ return an authentication configuration object.

This configuration object is unchanged but now contains `refreshAuth` rather than `getAuth`. Several arguments that its functions accept have been simplified.

```js
authExchange(async utils => {
  let token = localStorage.getItem('token');
  let refreshToken = localStorage.getItem('refreshToken');

  return {
    addAuthToOperation(operation) {
      return utils.appendHeaders(operation, {
        Authorization: `Bearer ${token}`,
      });
    },
    willAuthError(_operation) {
      return false;
    },
    didAuthError(error, _operation) {
      return error.graphQLErrors.some(e => e.extensions?.code === 'FORBIDDEN');
    },
    async refreshAuth() {
      const result = await utils.mutate(REFRESH, { token });
      if (result.data?.refreshLogin) {
        token = result.data.refreshLogin.token;
        refreshToken = result.data.refreshLogin.refreshToken;
        localStorage.setItem('token', token);
        localStorage.setItem('refreshToken', refreshToken);
      }
    },
  };
});
```

As we can see, the initializer function itself now initializes the authentication state, and the state is just kept in local variables (which also now makes it easier to keep the state in some "third object" outside of the `authExchange` without any hacky code)

The function returns the configuration and can use `utils.mutate` to send GraphQL mutations for authentication purposes, and now also has `utils.appendHeaders` to add headers to operations!

## Set of changes

- Simplify `addAuthToOperation`, `willAuthError`, and `didAuthError` arguments
- Remove `authState` and `getAuth` and replace with initializer
- Replace `getAuth` flow with `refreshAuth` call
- Update tests
- Add TSDocs to new exchange and APIs (generates a single `d.ts` output)
- Update documentation

**NOTE:** This isn't super final yet, but the goal here is to remove the API docs on the website and move those over to TSDocs. Hence, the website docs for the `authExchange` API are very lean now.

### DIFF
--- a/.changeset/perfect-lobsters-kiss.md
+++ b/.changeset/perfect-lobsters-kiss.md
@@ -1,0 +1,38 @@
+---
+'@urql/exchange-auth': major
+---
+
+Implement new `authExchange` API, which removes the need for an `authState` (i.e. an internal authentication state) and removes `getAuth`, replacing it with a separate `refreshAuth` flow.
+
+The new API requires you to now pass an initializer function. This function receives a `utils`
+object with `utils.mutate` and `utils.appendHeaders` utility methods.
+It must return the configuration object, wrapped in a promise, and this configuration is similar to
+what we had before, if you're migrating to this. Its `refreshAuth` method is now only called after
+authentication errors occur and not on initialization. Instead, it's now recommended that you write
+your initialization logic in-line.
+
+```js
+authExchange(async utils => {
+  let token = localStorage.getItem('token');
+  let refreshToken = localStorage.getItem('refreshToken');
+  return {
+    addAuthToOperation(operation) {
+      return utils.appendHeaders(operation, {
+        Authorization: `Bearer ${token}`,
+      });
+    },
+    didAuthError(error) {
+      return error.graphQLErrors.some(e => e.extensions?.code === 'FORBIDDEN');
+    },
+    async refreshAuth() {
+      const result = await utils.mutate(REFRESH, { token });
+      if (result.data?.refreshLogin) {
+        token = result.data.refreshLogin.token;
+        refreshToken = result.data.refreshLogin.refreshToken;
+        localStorage.setItem('token', token);
+        localStorage.setItem('refreshToken', refreshToken);
+      }
+    },
+  };
+});
+```
--- a/docs/advanced/authentication.md
+++ b/docs/advanced/authentication.md
@@ -59,100 +59,104 @@ const client = createClient({
   exchanges: [
     dedupExchange,
     cacheExchange,
-    authExchange({
-      /* config */
+    authExchange(async utils => {
+      return {
+        /* config... */
+      };
     }),
     fetchExchange,
   ],
 });
 ```
 
+You pass an initialization function to the `authExchange`. This function is called by the exchange
+when it first initializes. It'll let you receive an object of utilities and you must return
+a (promisified) object of configuration options.
+
 Let's discuss each of the [configuration options](../api/auth-exchange.md#options) and how to use them in turn.
 
-### Configuring `getAuth` (initial load, fetch from storage)
+### Configuring the initializer function (initial load)
 
-The `getAuth` option is used to fetch the auth state. This is how to configure it for fetching the tokens at initial launch in React:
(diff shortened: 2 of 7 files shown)


kitten:
Here's a small guide on how to map things from the "old" API to this "new" API, as proposed.
Comments in `ALL_CAPS` are meant to mark common sections.

### Legacy API

```js
authExchange({
  addAuthToOperation({ authState, operation }) {
    if (!authState || !authState.token) {
      return operation; // DONT_ADD_AUTH
    }
    return operation; // ADD_AUTH
  },
  willAuthError({ authState, operation }) {
    return false; // WILL_AUTH_ERROR
  },
  didAuthError({ error }) {
    return false; // DID_AUTH_ERROR
  },
  async getAuth ({ authState, mutate }) {
    if (!authState) {
      // INITIALIZE_AUTH
      const token = localStorage.getItem('token');
      const refreshToken = localStorage.getItem('refreshToken');
      return token ? { token, refreshToken } : null;
    }

    // REFRESH_AUTH
    const result = await mutate(refreshMutation, { token: authState?.refreshToken });
    if (result.data?.refreshLogin) {
      localStorage.setItem('token', result.data.refreshLogin.token);
      localStorage.setItem('refreshToken', result.data.refreshLogin.refreshToken);
      return {
        token: result.data.refreshLogin.token,
        refreshToken: result.data.refreshLogin.refreshToken,
      };
    }

    // BAIL_REFRESH_AUTH
    localStorage.clear();
    logout();
    return null;
  },
});
```

### Proposed API

```js
authExchange(async utils => {
  // INITIALIZE_AUTH
  let token = localStorage.getItem('token');
  let refreshToken = localStorage.getItem('refreshToken');

  return {
    addAuthToOperation(operation) {
      if (!token) {
        return operation; // DONT_ADD_AUTH
      }
      // ADD_AUTH
      return utils.appendHeaders(operation, {
        Authorization: `Bearer ${token}`,
      });
    },
    willAuthError(_operation) {
      return false; // WILL_AUTH_ERROR
    },
    didAuthError(error, _operation) {
      return false; // DID_AUTH_ERROR
    },
    async refreshAuth() {
      // REFRESH_AUTH
      const result = await mutate(refreshMutation, { token: authState?.refreshToken });
      if (result.data?.refreshLogin) {
        token = result.data.refreshLogin.token;
        refreshToken = result.data.refreshLogin.refreshToken;
        localStorage.setItem('token', token);
        localStorage.setItem('refreshToken', refreshToken);
      } else {
        // BAIL_REFRESH_AUTH
        localStorage.clear();
        logout();
      }
    },
  };
});
```

JoviDeCroock (approved):
This is a huge improvement

kadikraman (approved):

PR #3020, opened 2023-03-08
feat(graphcache): Add isOfflineError option to offlineExchange

robertherber:
## Summary

When using the offline plugin I think it'd make sense to simplify the isOffline check to apply to all network errors.

## Set of changes

Simplify the isOffline check to apply for all network errors.

### DIFF
--- a/.changeset/popular-carpets-compare.md
+++ b/.changeset/popular-carpets-compare.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-graphcache': minor
+---
+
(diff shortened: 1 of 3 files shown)


kitten:
That's not really the goal of this function. We don't really have data on either approaches, or feedback that is, to determine that one is objectively better than the other. However, I'm open to make this customizable.

We can add an extension of `CacheExchangeOpts` and enable `offlineExchange` to accept an `isOfflineError` attribute, with the default set to the current function.

kitten (approved):

PR #3033, opened 2023-03-13
major(core): remove default exchanges

JoviDeCroock:
## Summary

This removes our notion of `defaultExchanges` and replaces it with the notion of having to add them all separately. This improves the tree-shakeability of urql and introduces people earlier to our concept of exchanges, highlighting that we aim to be an extensible client.

## Set of changes

- remove `defaultExchanges`
- make `exchanges` mandatory in TS


packages/preact-urql/src/context.ts
```diff
@@ -4,7 +4,7 @@ import { Client, createClient } from '@urql/core';
 
 // We assume some default options here; mainly not to actually be used
 // but not to error catastrophically if someone is just playing around
-const defaultClient = createClient({ url: '/graphql' });
+const defaultClient = createClient({ url: '/graphql', exchanges: [] });
```
JoviDeCroock, line 7: Currently this gives a functional set of exchanges, do we want to keep this?

exchanges/auth/src/authExchange.test.ts
```diff
@@ -40,7 +43,10 @@ const makeExchangeArgs = () => {
           tap(op => operations.push(op)),
           map(result)
         ),
-      client: new Client({ url: '/api' }),
+      client: new Client({
+        url: '/api',
+        exchanges: [dedupExchange, cacheExchange, fetchExchange],
```
JoviDeCroock, line 48: TODO: remove exchanges, as we noticed in #3043 we are dispatching requests that error

.changeset/calm-buckets-scream.md
```diff
@@ -0,0 +1,18 @@
...
+meaning they can be removed if not used.
+
+A migration would look as follows if you are currently creating a client without exchanges
+
+```js
+import { createClient, dedupExchange, cacheExchange, fetchExchange } from '@urql/core'
+
+const client = createClient({
+  url: '',
+  exchanges: [dedupExchange, cacheExchange, fetchExchange]
```
kitten, line 16: See #3058, if we want to remove the `dedupExchange` entirely.

packages/core/src/client.ts
```diff
@@ -111,13 +111,13 @@ export interface ClientOptions {
...
-   * When this option is left out, the `Client` defaults to a list of {@link defaultExchanges}. By default, these
-   * will implement deduping, caching (via a document cache), and fetching (GraphQL over HTTP).
+   * To easily get started you should consider using the {@link dedupExchange}, {@link cacheExchange} and {@link fetchExchange}
+   * these are all exported from the core package.
    *
    * @see {@link https://formidable.com/open-source/urql/docs/architecture/#the-client-and-exchanges} for more information
    * on what `Exchange`s are and how they work.
    */
-  exchanges?: Exchange[];
+  exchanges: Exchange[];
```
kitten, line 120: We could define this as `[Exchange, ...Exchange[]]` to enforce that at least one exchange should be passed. Further, we can then add an `@example` TSDoc here, which explains the default for new-comers ✨

kitten (approved):

packages/core/src/client.ts
kitten: Ha, I just realised that that'd mean updating a gazillion tests, so maybe not 🤷

packages/react-urql/src/context.ts
```diff
@@ -3,7 +3,7 @@ import { Client, createClient } from '@urql/core';
 
 // We assume some default options here; mainly not to actually be used
 // but not to error catastrophically if someone is just playing around
-const defaultClient = createClient({ url: '/graphql' });
+const defaultClient = createClient({ url: '/graphql', exchanges: [] });
```
kitten, line 6: We may want to remove the defaults while we're at it, finally 🤔 
I just realised that it's technically possible to upgrade `@urql/core` without upgrading `urql`, which pretty much just fails on `composeExchanges(undefined)` then.

I'd propose that we allow `composeExchanges` to handle that, but also remove the defaults finally.

JoviDeCroock: this would already be handled by the `fallbackExchange` I removed the defaults from react/preact

PR #3041, opened 2023-03-14
feat(core): Replace formatDocument's graphql.visit usage

kitten:
## Summary

This replaces `graphql.visit` with a small alternative that is functionally equivalent. Many GraphQL apps actually never touch the GraphQL document or use `visit`, so this can be treeshaken away if `urql` doesn't rely on it.

This does increase our bundlesize by `40B` gzipped, but may be worth it overall.

## Set of changes

- Replace `graphql.visit` with custom alternative

### DIFF
--- a/.changeset/tame-pumas-promise.md
+++ b/.changeset/tame-pumas-promise.md
@@ -1,0 +1,5 @@
+---
+'@urql/core': patch
+---
+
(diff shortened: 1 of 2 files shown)


kitten:
Closing since this doesn't make much sense.
I forgot that `print` was relying on `visit` unless it's replaced with an alternative while bundling.

PR #3046, opened 2023-03-14
fix(graphcache): fix restoring variables on mutation operations

JoviDeCroock:
## Summary

Before we send out the mutation we filter the variables so extraneous variables don't cause an issue on the server. However this makes it so that we don't restore them afterwards as they are not part of the `operations` Map. This gives us a way to restore variables on mutations.

Works now: https://codesandbox.io/s/empty-water-6tm6yy?file=/src/index.js

### DIFF
--- a/.changeset/fifty-jeans-invite.md
+++ b/.changeset/fifty-jeans-invite.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-graphcache': patch
+---
+
+Restore variables correctly on mutations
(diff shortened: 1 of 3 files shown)


kitten (approved):
Seems good :+1: I think we can think of alternatives in the future. Generally, I'm not opposed to maybe creating an `original` property poibnting at what `makeOperation`'s second argument was.
This could potentially then also be used for other cases.

PR #3058, opened 2023-03-15
feat(core): Deprecate the dedupExchange and absorb hasNext checks into Client

kitten:
## Summary

This deprecates the `dedupExchange` and turns it into a noop exchange.

The logic that `dedupExchange` applied was already effectively taken care of by the `Client` and we only needed small modifications to prevent dispatching when `hasNext: true` is present on any given subscription or query result, i.e. when we're still waiting for updates.

## Set of changes

- Add `hasNext` dispatch stop to `Client`
- Replace `dedupExchange` with noop logic


kitten:
This may vibe well with #3033. I have no idea if we'll get major resistance if we were to drop a major with that, but if we are to get rid of the `dedupExchange`, then there's really not much left for most apps.

We'd start avoiding `[...defaultExchanges, x]` mistakes automatically and by default, there's only two exchanges to add really. So, I think we may want to drop both at the same time.

JoviDeCroock (approved):

PR #3059, opened 2023-03-16
fix(core/graphcache): Add missing hasNext/stale flags to cached results

kitten:
## Summary

Ensures that the `ssrExchange` and Graphcache's `cacheExchange` don't accidentally remove `hasNext` (or `stale`) flags.

## Set of changes

- Add `hasNext` passthrough to Graphcache
- Add `hasNext` passthrough to `ssrExchange`

### DIFF
--- a/.changeset/polite-penguins-notice.md
+++ b/.changeset/polite-penguins-notice.md
@@ -1,0 +1,6 @@
+---
+'@urql/exchange-graphcache': patch
+'@urql/core': patch
+---
+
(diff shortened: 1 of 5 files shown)


kitten:
cc @JoviDeCroock: Pretty trivial, so I went ahead and merged it.
On a side note, I was trying to actually make `hasNext` and `stale` required fields. We can still do that, but it'd mean updating like 50–100 lines of failing tests. Not a big deal, but we may want to check whether we really want to do this.
I think it's potentially worth it though.

PR #3067, opened 2023-03-16
chore(storybook): Deprecate & Remove @urql/storybook-addon

kitten:
## Summary

Remove & Deprecate `@urql/storybook-addon`.
I think we can eventually make a better `mockExchange` as a replacement, but the plugin itself doesn't serve much of a purpose right now and has gone unmaintained.

## Set of changes

- Remove `packages/storybook-addon`


kitten:
@dkolman-vendavo: I haven't written up an RFC yet, but the likely successor is a bundled `@urql/testing` package. It's looking likely that storybook helpers will move there, if we can find more convenient ways to ensure that they'll work for all UI bindings equally.

For now, I'm treating this as tracked by #3133 until a new RFC replaces it.

It's a bit of an unfortunate situation that we didn't foresee in time. The package wasn't previously documented or had a place for docs where we could track this.
But in short, for now you can still rely on it until an alternative becomes available.

The longer story here is that it didn't receive maintenance and the new version of Storybook came around. However, it didn't live up to our standards either because the package was only targeting React properly, so a replacement is and was needed.

PR #3071, opened 2023-03-17
chore(examples): Update package versions in examples and keep up-to-date

kitten:
## Summary

This attempts to update the dependencies of all of our examples, but also adds a version that'll cause them to be updated automatically with the "Version Packages" changeset PR (i.e. the release workflow), which will then become part of each release's PR.

## Set of changes

- Add version script that also updates versions in examples
- Update all versions across examples


examples/pnpm-workspace.yaml
```diff
@@ -0,0 +1,2 @@
+packages:
+  - '*'
```
kitten, line 2: This seems to be the only way for us to get `pnpm` to work again. When inside a workspace, this will only install as another sub-workspace with separate `node_modules`. Otherwise, ignoring examples in the root will just cause the examples to never install at all.

JoviDeCroock (approved):

PR #3077, opened 2023-03-18
docs(graphcache): Add example to "Variables for Optimistic Updates"

romain-trotard:
Hello :)

## Summary

Following comments here https://github.com/urql-graphql/urql/discussions/2279#discussioncomment-5266279 and the fix that will be available soon https://github.com/urql-graphql/urql/pull/3046 <3
I still think it would be nice to have some code example in the documentation.

### DIFF
--- a/docs/graphcache/cache-updates.md
+++ b/docs/graphcache/cache-updates.md
@@ -543,31 +543,8 @@ variables using our query document definitions, which means that our API will ne
 variables other than the ones we've defined.
 
 However, we're able to pass additional variables to our mutation, e.g. `{ extra }`, and since
-`$extra` isn't defined it will be filtered once the mutation is sent to the API.
-
-```jsx
-const MyComponent = () => {
-  const [_, executeMutation] = useMutation(UpdateTodo);
-
-  return (
-    <button
-      type="button"
-      onClick={() =>
-        executeMutation({
-          id: 'anId',
-          text: 'New value',
-          // Let's pass an extra variable
-          extra: 'An extra value',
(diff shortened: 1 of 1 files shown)


docs/graphcache/cache-updates.md
```diff
@@ -543,8 +543,43 @@ variables using our query document definitions, which means that our API will ne
...
+          // Let's pass an extra variable
+          extra: 'An extra value',
+        })
+      }
+    >
+      Update todo
+    </button>
+  );
+};
+```
```
romain-trotard, line 568: I'm not really sure, we want to have React specific code in the **Graphcache** documentation. 
I don't know how to show the mutation in a framework/library agnostic way. Do you have an idea?

kitten: I would opt to simply not have a code sample in here, sorry! 😅
We generally try to prefer to show no specific bindings in any part of the documentation that is kept generic. While this is sometimes ignored in some "Basics" documentation or when it's absolutely necessary. From experience, I don't think a code sample here helps much.

I've seen that code samples don't help discoverability. They simply take up more space. While that seems like it's more likely for someone skimming the docs to discover it, we can't have all parts of the documentation compete for space and attention.

So, I'd prefer to leave this as is. Sorry! ❤️

romain-trotard: I understand, I don't think this code is that helpful :)
I remove this one. 
But I think the other one is useful

kitten (approved):

PR #3091, opened 2023-03-23
Experiment: feat(core): Switch from graphql to @0no-co/graphql.web import

kitten:
**Note:** This currently only tests feasibility and outcome of what it would look like if `@urql/core` (et al) shipped with a leaner default implementation of `graphql` internals via `@0no-co/graphql.web` while still staying TS-type compatible with `graphql`.

### DIFF
--- a/.changeset/tame-pumas-promise.md
+++ b/.changeset/tame-pumas-promise.md
@@ -1,0 +1,5 @@
+---
+'@urql/core': patch
+---
+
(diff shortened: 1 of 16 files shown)


kitten:
We likely will never really want to force people to use a different implementation of `graphql` (even if it only affects `parse`, `print`, and `visit), but we can make `@0no-co/graphql.web` a little more intercompatible for now; see: https://github.com/0no-co/graphql.web/pull/10

PR #3097, opened 2023-03-24
major(core/graphcache): Remove graphql imports by default

### LINKED ISSUE #3096: RFC: Reduce size of shipped GraphQL.js code by default
## Summary

As it stands today, most apps ship at least `10kB` gzip of JS code to include common utilities from `graphql` in their apps for `@urql/core` and all of `urql`’s dependencies include a peer dependency on `graphql`.

This is because `@urql/core`, and most of the client-side GraphQL ecosystem, depends on the following utilities in `graphql` to drive support of the query language and AST:
```js
import { parse, print, visit, Kind, GraphQLError } from 'graphql';
```

Recently, we’ve experimented with bringing this size down with the [`graphql-web-lite` project](https://github.com/0no-co/graphql-web-lite) which is a “rebuilt” package of `graphql` that replaces the common utilities we use and is a full shim of `graphql` that can be aliased in for it. This brings the above size down from `10kB` to `~3kB`, a change of at least `-7kB` gzip.

Most of this change is driven by the [`@0no-co/graphql.web` project](https://github.com/0no-co/graphql.web), which reimplements the above utilities while maintaining the same performance and a 100% test coverage, while reimplementing the parts of `graphql` that `@urql/core` depends on.

If we shipped this by default, all currently deployed `urql` apps that were to upgrade immediately shrink in size. In fact, the base bundlesize impact of importing `@urql/core` would reduce to `10kB` gzip _in total_, meaning, that while these seem like small reductions to some, we’d introduce an impact on bundlesize of over `-50%`.

We would have to maintain type compatibility and escape hatches nonetheless.

## Proposed Solution

This change would impact all three of the aforementioned packages, plus all packages in the `urql` monorepo currently.

### Changees in `@urql/core`

- All imports in `@urql/core` must switch from `graphql` to `@0no-co/graphql.web`.
- Its `peerDependency` on `graphql` is removed.
- When it accepts a `DocumentNode`, `DefinitionNode`, or `GraphQLError` it must accept types that are either from `graphql` or from `@0no-co/graphql.web` in case of type incompatibilities
- It must export an input type that accepts `DocumentNode | TypedDocumentNode` as per the above type aliases

### Changes in `@0no-co/graphql.web`

- It must replace its custom AST types with `graphql`’s AST types when `graphql` is installed
  - See the PR for this change: https://github.com/0no-co/graphql.web/pull/10

### Changes in `graphql-web-lite`

- It must mark `@0no-co/graphql.web` as an external dependency, excluded from its bundle
  - This is so that `graphql` can still be aliased to `graphql-web-lite` to prevent code duplication

### Outcome

All three changes together mean the following:
- `@urql/core` will expose a type and accept a type of `DocumentNode` (& others) that can either come from the `graphql` package or the “internal” parser (i.e. `@0no-co/graphql.web`).
- `graphql` will not be included anymore and all peer dependencies on it are removed.
- `graphql-web-lite` can be aliased into an app, so the entire app makes use of the shared `@0no-co/graphql.web`, if so desired.
- `@0no-co/graphql.web` can be aliased to `graphql` to switch back to the official package. (Considered a rare use-case)

This effectively provides type compatibility and reduces the risk of `graphql` still being used and installed in an app to a total of the usual 10kB penalty of including it, while maintaining type compatibility. This also provides intercompatibility with type generation tools that use `graphql` to type their outputs.

kitten:
Resolves #3096

## Summary

This removes `graphql` from `peerDependencies` on all packages. `@urql/core` and `@urql/exchange-graphcache` will now depend on `@0no-co/graphql.web` instead and switch to `graphql` types dynamically where needed, and otherwise either accept `@0no-co/graphql.web`’s types (which will also switch to `graphql`; See: https://github.com/0no-co/graphql.web/pull/10) or the native `graphql` types.

This effectively replaces all imports we use from `graphql` with smaller alternatives from `@0no-co/graphql.web` by default as per the RFC at #3096.

## Set of changes

- Replace `graphql` with `@0no-co/graphql.web`
- Add types shim for common imports to `@urql/core` which is marked with the `/*!@ts-ignore*/` pragma in the type output
- Replace `formatDocument` in `@urql/core` with an implementation that avoids using `visit`
- Remove `graphql` from `peerDependencies` in all packages
- Replace `graphql` usage in `@urql/exchange-graphcache` and add shim types for schema


JoviDeCroock (approved):

scripts/rollup/cleanup-plugin.mjs
```diff
@@ -37,7 +38,9 @@ function cleanup() {
           babelrc: false
         });
       } else if (dtsFilter(chunk.fileName)) {
-        return code.replace(emptyImportRe, '');
+        return code
+          .replace(emptyImportRe, '')
+          .replace(gqlImportRe, x => '/*!@ts-ignore*/\n' + x);
```
kitten, line 43: **NOTE:** This build change marks `graphql` imports in out `d.ts` typings files with a `/*!@ts-ignore*/` pragma. While imports from `graphql` will change to `any` when the package isn’t installed, we still need to mark this in case a user has `skipLibCheck: false` set in their tsconfig

PR #3112, opened 2023-03-31
fix(auth): avoid infinite loop when didAuthError keeps returning true

JoviDeCroock:
Discovered in https://github.com/urql-graphql/urql/discussions/3111

## Summary

In the fall-through case (edited now to not fall through) we would set the `authAttempt` back to false. Instead now we check whether it's true and just add auth to the operation rather than manipulating the context.

The test provided in the discussion won't work as we pass back `result.operation` without the manipulation happening in the `authExchange` working on that.

### DIFF
--- a/exchanges/auth/src/authExchange.ts
+++ b/exchanges/auth/src/authExchange.ts
@@ -270,6 +270,7 @@ export function authExchange(
           operation.key,
           addAuthAttemptToOperation(operation, true)
         );
+
         // check that another operation isn't already doing refresh
         if (config && !authPromise) {
           authPromise = config.refreshAuth().finally(flushQueue);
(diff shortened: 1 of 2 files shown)


exchanges/auth/src/authExchange.ts
```diff
@@ -310,7 +311,9 @@ export function authExchange(
       const opsWithAuth$ = pipe(
         merge([retries.source, pendingOps$]),
         map(operation => {
-          if (bypassQueue.has(operation)) {
+          if (operation.context.authAttempt) {
```
kadikraman, line 314: Worth adding [this test](https://github.com/urql-graphql/urql/discussions/3111#discussioncomment-5488476) case as well?

JoviDeCroock: Got it in https://github.com/urql-graphql/urql/pull/3112/commits/dcd44f9dd9eb5f49bd4a1daadbe3b72e055af5c0 was harder than anticipated cause I missed the result mocking :joy:

kitten (approved):

PR #3118, opened 2023-03-31
fix(core): Fix multipart FormData map field format

kitten:
## Summary

The multipart request spec actually specifies that `map` is of type `Record<string, string[]>` rather than `Record<string, string>`. While this is easy to miss, it's likely for upload deduplication, which we don't do but have to respect.

## Set of changes

- Convert `FormData` `map` field to `[prop: string]: [string]`


negezor:
My bad, thanks for the quick fix 😅

PR #3124, opened 2023-04-01
fix(core): Fix Incremental Delivery payloads deep merging and root patches

### LINKED ISSUE #3122: deferred fragments on Query produces  `undefined` key in data response
### Describe the bug

Hi, i've been exploring adding the `@defer` directive to our graphql server and believe I've come across an issue with `urql`.

When deferring on the root `Query` fragment, `urql` will respond with an `undefined` key in the data response.

Versions:
* `"@graphql-yoga/plugin-defer-stream": "^1.7.1"`
* `"@urql/core": "^4.0.1"`
* `"graphql": "17.0.0-alpha.2"`
* `"urql": "^4.0.0"`

**Expected behavior**

Entire query is resolved and merged together
```
  data: {
      song: { firstVerse: "Now I know my ABC's.", secondVerse: "Next time won't you sing with me?" },
  },
```

**Actual behavior**

Query is split via an `undefined` key
```
  data: {
      song: { firstVerse: "Now I know my ABC's." },
      undefined: { song: [Object] }
  },
```
---
This was verified by using the `with-defer-stream-directives` example with the following script below

```
import { Client, gql, debugExchange, fetchExchange } from 'urql';
import fetch from 'node-fetch';

const SONGS_QUERY = gql`
  query App_Query {
    ... on Query @defer {
      song {
        secondVerse
      }
    }
    song {
      firstVerse
    }
  }
`;

const client = new Client({
  url: 'http://localhost:3004/graphql',
  fetch: fetch,
  exchanges: [debugExchange, fetchExchange],
});

client.query(SONGS_QUERY).then(response => {
  console.log('LOG response: ', response);
});

```

This is resolved if we do not `@defer` the root `Query` fragment. E.g.
```
  query App_Query {
    song {
      firstVerse
      ... on Song @defer {
        secondVerse
      }
    }
  }
```

However I'm unable to do so ATM due to lack of schema stitching support (https://github.com/ardatan/graphql-tools/pull/1941)

I believe this to be a bug from the side of `urql`

### Reproduction

Apologies, found this difficult to setup

### Urql version

urql v4.0.0

### Validations

kitten:
Resolves #3122

## Summary

This deeply merges (while cloning) result sets from Incremental Delivery results, and handles deep merges on the root payload as well, which we previously didn't realise was a legal usage of `@defer`.

## Set of changes

- Add deep-clone merging of incremental delivery payloads
- Add handling of merging patches into root (i.e. `path: []`)

### DIFF
--- a/packages/core/src/internal/fetchSource.test.ts
+++ b/packages/core/src/internal/fetchSource.test.ts
@@ -493,7 +493,7 @@ describe('on text/event-stream', () => {
                   incremental: [
                     {
                       path: [],
-                      data: { author: { name: 'Steve' }  },
+                      data: { author: { name: 'Steve' } },
                     },
                   ],
                   hasNext: true,
@@ -523,7 +523,9 @@ describe('on text/event-stream', () => {
 
     const AuthorFragment = gql`
       fragment authorFields on Query {
-        author { name }
+        author {
+          name
+        }
       }
     `;
 
--- a/packages/core/src/internal/fetchSource.test.ts
+++ b/packages/core/src/internal/fetchSource.test.ts
@@ -459,4 +459,122 @@ describe('on text/event-stream', () => {
       },
     });
   });
+
+  it('merges deferred results on the root-type', async () => {
+    fetch.mockResolvedValue({
+      status: 200,
+      headers: {
+        get() {
+          return 'text/event-stream';
+        },
+      },
+      body: {
+        getReader: function () {
+          let cancelled = false;
+          const results = [
+            {
+              done: false,
+              value: Buffer.from(
+                wrap({
+                  hasNext: true,
+                  data: {
+                    author: {
+                      id: '1',
+                      __typename: 'Author',
+                    },
+                  },
+                })
+              ),
+            },
+            {
+              done: false,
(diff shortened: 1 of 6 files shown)


packages/core/src/internal/fetchSource.test.ts
```diff
@@ -459,4 +459,124 @@ describe('on text/event-stream', () => {
...
+    `;
+
+    const streamedQueryOperation: Operation = makeOperation(
+      'query',
+      {
+        query: gql`
+          query {
+            author {
+              id
+              ...authorFields @defer
```
tompuric, line 539: I don't think this is valid gql? You can't spread a Query fragment inside Author. Shouldn't it be

```
query: gql`
  query {
    author {
      id
    }
    ...authorFields @defer
  }
  ${AuthorFragment}
`,
```

Similar to the previous PR https://github.com/urql-graphql/urql/pull/3123

Though I could be missing context from this repo

kitten: It says `fragment authorFields on Query {`, so it's indeed spread into the right place.
That said, this isn't executed, so the query doesn't have to be valid

PR #3130, opened 2023-04-05
Handle network throw error - Resolves #3128

### LINKED ISSUE #3128: fetchOperation throws "Cannot read properties of undefined" when client is offline
### Describe the bug

TypeError: Cannot read properties of undefined (reading 'status')

The [response](https://github.com/urql-graphql/urql/blob/8b1d16418a36b54ad5a027268869a36410e11303/packages/core/src/internal/fetchSource.ts#L121) gets lost 
when the fetch throws an error causing the offline logic to fail since it cant check status etc. 

Repro steps: 
1) Open the repro codesandbox. https://codesandbox.io/s/beautiful-leakey-fg4px1
2) Open F12 and select Offline in network settings
3) Execute an mutation for example clicking the todo li item. 

Expected: 
The makeErrorResult to return an error down the pipe and not to throw a "Cannot read properties of undefined" error. 

Actual: 
The makeErrorResult fails to process the fetch response and instead throw "Cannot read properties of undefined" when trying to get the response.status. 

Observations: 
I have not created a full offline example here but just trying to replicate the "undefined" response issue. 
The fix is to handle the throw more gracefully in the catch. 

### Reproduction

https://codesandbox.io/s/beautiful-leakey-fg4px1

### Urql version

"@urql/exchange-graphcache": "^6.0.1",
"@urql/core": "^4.0.2",
"urql": "^4.0.0"

### Validations

Psvensso:
Resolves #3128

## Summary
Doesent throw on network fail

## Set of changes
If network throws (when client is offline etc) it still return an internal error. 

## Missing
Should probaly have some tests.

### DIFF
--- a/packages/core/src/internal/fetchSource.ts
+++ b/packages/core/src/internal/fetchSource.ts
@@ -149,6 +149,13 @@ async function* fetchOperation(
       throw error;
     }
 
+    if(!response){
+      yield makeErrorResult((
+        operation,
+        error
+      ))
+    }
+
     yield makeErrorResult(
       operation,
       (response!.status < 200 || response!.status >= 300) &&
--- a/packages/core/src/internal/fetchSource.ts
+++ b/packages/core/src/internal/fetchSource.ts
@@ -154,16 +154,18 @@ async function* fetchOperation(
         operation,
         error
       ))
+    } else {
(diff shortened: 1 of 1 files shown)


kitten:
Superseded by #3131

PR #3137, opened 2023-04-06
fix(core): Fix early closing of subscriptions

kitten:
## Summary

Fixes an issue where `hasNext` would be updated to `false` and a subscription would be ended early when the `subscriptionExchange`’s sink doesn't actually set `hasNext: true` on a following result.

This is a similar issue to the missing `defaultHasNext` in `makeResult` we had to add.

## Set of changes

- Add default `hasNext` value to merged results
- Add unit tests

### DIFF
--- a/.changeset/nice-flowers-think.md
+++ b/.changeset/nice-flowers-think.md
@@ -1,0 +1,5 @@
+---
+'@urql/core': patch
+---
+
+Fix `hasNext` being defaulted to `false` when a new subscription event is received on the `subscriptionExchange` that doesn't have `hasNext` set.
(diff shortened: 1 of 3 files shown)


TheMightyPenguin:
I was getting the same issue that "subscriptions" where closed after 2-3 events, and went down a rabbit hole to read the source for urql and grapqhl-ws, which led me to this PR 😅 awesome to see it is fixed 🚀

Yuripetusko:
This fix didn't make it to the general `urql` npm package yet right? As if I install @urql/core separately I get a lot of type mismatches between `@urql/exchange-auth@2.1.2` `urql@4.0.2` and `@urql-core@4.0.7` but the issue is fixed. Hard to understand how to get the fix and have all the different packages work between each other

PR #3147, opened 2023-04-13
Pass-through the context to fetchBody

i404788:
This enables runtime-customizable exchange handlers, this available in previous version but in the latest it doesn't seem to be available anymore. Let me know if this pattern was replaced, but I couldn't find a reference (or alternate way through debugging) .

## Summary
Allows for runtime-customizable exchanges in the framework (vue) versions, example of my use-case:
```ts
     ...
      subscriptionExchange({
        forwardSubscription: (request) => {
          if (request.context.endpoint === 'Hasura') {
            // graphql-ws usage
            const input = { ...request, query: request.query || '' };
            return {
              subscribe(sink) {
                const unsubscribe = Hasura.subscribe(input, sink);
                return { unsubscribe };
              },
            };
          } else if (request.context.endpoint === 'dGraph') {
            // subscriptions-transport-ws usage
            return dGraph.request(request)
          }
        },
      }),
```

```ts
const {fetching: fetchingUpdate, data: lastUpdateWorkplaceEvents, isPaused, pause, resume} = useOnWorkplaceEventUpdateSubscription({context: {endpoint: 'dGraph'}});
```

## Set of changes
* Passed through the context argument to the fetchBody, this was the behavior in certain prior version iirc

EDIT: it seems I misread the debugger, I'll edit it if it's still possible

### DIFF
--- a/packages/core/src/internal/fetchOptions.ts
+++ b/packages/core/src/internal/fetchOptions.ts
@@ -5,7 +5,7 @@ import {
   extractFiles,
 } from '../utils';
 
-import { AnyVariables, GraphQLRequest, Operation } from '../types';
+import { AnyVariables, Operation } from '../types';
 
 /** Abstract definition of the JSON data sent during GraphQL HTTP POST requests. */
 export interface FetchBody {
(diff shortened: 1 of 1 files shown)


i404788:
Hmm I see from the unit tests that makeFetchBody is also used for HTTP body (makes sense), it could be done on https://github.com/urql-graphql/urql/blob/35657e771633be43b850f0a1f7fdbac94b075dd7/packages/core/src/exchanges/subscription.ts#L143
but then it would only  work for subscription exchanges :thinking:

i404788:
Oh I totally missed that change of signature :sweat_smile:, thanks for the clarification.

PR #3151, opened 2023-04-14
fix(docs): urql vue plugin exchanges

JoviDeCroock:
## Summary

Missed it here

### DIFF
--- a/docs/basics/vue.md
+++ b/docs/basics/vue.md
@@ -103,7 +103,7 @@ const app = createApp(Root);
 
 app.use(urql, {
   url: 'http://localhost:3000/graphql',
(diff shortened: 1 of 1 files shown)


docs/basics/vue.md
```diff
@@ -98,11 +98,13 @@ importing its default export and using it [as a plugin](https://v3.vuejs.org/gui
 import { createApp } from 'vue';
 import Root from './App.vue';
 import urql from '@urql/vue';
+import { cacheExchange, fetchExchange } from '@urql/core';
```
boboldehampsink, line 101: These can also both be imported from `@urql/vue`, so it's not necessary to add `@urql/core` to your dependencies.

PR #3164, opened 2023-04-19
fix(auth): prevent retrying the operation produced by `mutate`

JoviDeCroock:
## Summary

There were scenario's where the `operation` would get altered and hence we wouldn't see it in the bypassQueue, now to make sure the first thing we do is we add the operation in the `bypassQueue` by the mutation identity.

Secondly we prevent ever calling `didAuthError` on operations produced by `utils.mutate` and last but not least we have started cleaning up the Set when results do come in.

## Set of changes

- assert on a `operationc.context._identity` match
- key the bypassQueue by `operationc.context._identity`
- clear the queue after a result comes in

### DIFF
--- a/.changeset/strange-elephants-serve.md
+++ b/.changeset/strange-elephants-serve.md
@@ -2,4 +2,4 @@
 '@urql/exchange-auth': patch
 ---
 
-Fix case where we would retry the operation created by `mutate` when it failed auth
+Fix operations created by `utilities.mutate()` erroneously being retried and sent again like a regular operation
--- a/.changeset/strange-elephants-serve.md
+++ b/.changeset/strange-elephants-serve.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-auth': patch
+---
+
+Fix case where we would retry the operation created by `mutate` when it failed auth
--- a/exchanges/auth/src/authExchange.ts
+++ b/exchanges/auth/src/authExchange.ts
@@ -200,7 +200,7 @@ export function authExchange(
   init: (utilities: AuthUtilities) => Promise<AuthConfig>
 ): Exchange {
   return ({ client, forward }) => {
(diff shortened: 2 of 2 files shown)


exchanges/auth/src/authExchange.ts
```diff
@@ -233,10 +234,17 @@ export function authExchange(
                 result$,
                 onStart(() => {
                   const operation = addAuthToOperation(baseOperation);
```
JoviDeCroock, line 236: it's possible that we need to remove `addAuthToOperation` here not entirely sure, it does make less sense to call this (we also say this in the docs)

exchanges/auth/src/authExchange.ts
```diff
@@ -233,10 +234,17 @@ export function authExchange(
...
+                    operation.context._instance as OperationInstance
+                  );
                   retries.next(operation);
                 }),
-                filter(result => result.operation.key === baseOperation.key),
+                filter(
+                  result =>
+                    result.operation.key === baseOperation.key &&
+                    baseOperation.context._instance ===
+                      result.operation.context._instance
```
kitten, line 246: I'm pretty sure that this can become `0` but we could probably just do the `.has` check either way and allow `undefined` on the `Set` type since we can just make sure not to add `undefined` to the set lol

.changeset/strange-elephants-serve.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/exchange-auth': patch
+---
+
+Fix case where we would retry the operation created by `mutate` when it failed auth
```
kitten, line 5:
```suggestion
Fix operations created by `utilities.mutate()` erroneously being retried and sent again like a regular operation
```

Just clarifying here and removing the "we" from the change entry

kitten (approved):

PR #3167, opened 2023-04-19
fix(graphcache): Fix defer field state becoming sticky and affecting future fields

### LINKED ISSUE #3161: cache issues when using `@defer` directive and `@urql/exchange-graphcache`
### Describe the bug

I’ve run into a pretty weird issue while using urql and I could really use some help.

The main problem is that when my app sends a query with defer or when it receives chunked data, it seems to be the last request that’s ever fired.

For example, when I open my app and navigate through random pages, everything works just fine. But when I go to a page that uses defer and then move to another random page, I get a bunch of runtime errors due to missing data. It looks like the query was resolved (cache hit), but an API request was never fired. When I refresh the page, everything works again.

After some debugging, I narrowed it down to have something to do with `@urql/exchange-graphcache` (if I remove the exchange it works fine). Based on the logs and urql devtools, there’s a cache hit for those queries when I there shouldn’t be. Also, when I remove defer usage, logs from the same flow show cache misses.

Do you have any suggestions what could be causing this?

Urql dev tools:
* There's no API call for the last four queries
* The fifth one (from the bottom) is the one with `defer` directive
CleanShot 2023-04-18 at 17 01 19@2x

CleanShot 2023-04-18 at 17 02 33@2x

Here's a PR to GraphQL Hive with adding defer to one of the queries (and necessary setup): https://github.com/kamilkisiela/graphql-hive/pull/1964

### Reproduction

https://github.com/kamilkisiela/graphql-hive/pull/1964

### Urql version

@urql/core: 4.0.4

@urql/exchange-graphcache: 6.0.1

### Validations

kitten:
Resolves #3161

**Note:** This will need a rebase after #3165 is merged and should be merged on top.

## Summary

A rather old regression would cause `@defer` directives to manipulate the `deferRef` state and then become “sticky”, meaning that all future fields would also be considered deferred.

Instead, we need to make sure that this field resets properly and that the state returns to normal.

On a side-note, I tried to convert `makeSelectionIterator` to a generator function, since this would make it more elegant, but an initial rough implementation decreased overall cache read/write performance by 15–20%, so we'll have to investigate whether there's more we can do here. In general, traversal is the most expensive part of the cache operations, so if we can instead make it faster we'd profit from that quite a lot.

## Set of changes

- Refactor `makeSelectionIterator` to reset `deferRef`
- Add missing test cases for `makeSelectionIterator`

### DIFF
--- a/.changeset/real-queens-crash.md
+++ b/.changeset/real-queens-crash.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-graphcache': patch
+---
+
+Fix regression which caused `@defer` directives from becoming “sticky” and causing every subsequent cache read to be treated as if the field was deferred.
--- a/exchanges/graphcache/src/operations/query.ts
+++ b/exchanges/graphcache/src/operations/query.ts
@@ -137,7 +137,13 @@ const readRoot = (
     return input;
   }
 
-  const iterate = makeSelectionIterator(entityKey, entityKey, select, ctx);
+  const iterate = makeSelectionIterator(
+    entityKey,
+    entityKey,
+    deferRef,
+    select,
+    ctx
+  );
 
   let node: FieldNode | void;
   let hasChanged = InMemoryData.currentForeignData;
@@ -334,7 +340,13 @@ const readSelection = (
   }
 
   const resolvers = store.resolvers[typename];
-  const iterate = makeSelectionIterator(typename, entityKey, select, ctx);
+  const iterate = makeSelectionIterator(
+    typename,
+    entityKey,
+    deferRef,
+    select,
+    ctx
+  );
 
   let hasFields = false;
   let hasPartials = false;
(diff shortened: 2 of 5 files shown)


kitten:
Unanswered question that may go along with traversal performance; cc @JoviDeCroock,
when fixing the regression I also realised that we potentially don't cover overlapping deferred and non-deferred selection sets. When a fragment is deferred but not fulfilled, however, an overlapping selection does provide part of the nested, deferred fragment, we'd currently fail with an unexpected cache miss, e.g.:

```graphql
{
  child { id }
  ... @defer {
    # `child` would be traversed here...
    child {
      id # ...because this object is present because of `id`
      extra # however, here we'd run into a cache miss
    }
  }
}
```

In the above case, we can see that we have a need for defer to become recursive. I can submit a small fix for this after this comment, however, this highlights the need for us to optimise traversal further

JoviDeCroock (approved):

PR #3172, opened 2023-04-20
fix(core): Handle both variations of multipart/mixed preamble

kitten:
## Summary

This implements a fix for a regression in #3155.

Basically, we now split the incoming `multipart/mixed` response by `CRLF` and the boundary. However, our tests send the preamble split by `CRLF` + boundary, while some newer variations would only send the boundary as the preamble and only after use `CRLF`

This caused the first payload to be malformed and disregarded.

## Set of changes

- Add preamble recovery depending on `CRLF` splitting

### DIFF
--- a/.changeset/itchy-queens-scream.md
+++ b/.changeset/itchy-queens-scream.md
@@ -1,0 +1,5 @@
+---
+'@urql/core': patch
+---
+
+Handle `multipart/mixed` variations starting with boundary rather than CRLF and a boundary.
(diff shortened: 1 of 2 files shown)


JoviDeCroock (approved):

maraisr:
You could just pull in meros and get this kind of stuff for free

kitten:
@maraisr We can't really any longer 😄 At least looking at it from what we did in the last major, the implementation now also throws out some smaller things we don't need and is combined with event streams, so I'm fine with fixing a regression from the refactor at the benefit of the combined implementation 👀

PR #3186, opened 2023-04-27
Fix typo in docs/advanced/debugging.md

adamsoderstrom:
## Summary

The changes in this PR updates a typo in documentation, regarding a function call.
This could cause confusion for the readers of the documentation.

The statement is the following: 
> While it's possible to call
  `client.subscsubscribeToDebugTarget` in an exchange it's only valuable when creating a debugging

Besides the typo, the statement is true. Though, the typo assumes that the `subscsubscribeToDebugTarget` method is available on the `client` object, [which it isn't](https://github.com/urql-graphql/urql/blob/ff5d7397a14497193e28bfc69b719abe8dcdcaf8/packages/core/src/client.ts#L720).

## Set of changes

- Altered the content of `docs/advanced/debugging.md`

### DIFF
--- a/docs/advanced/debugging.md
+++ b/docs/advanced/debugging.md
@@ -133,7 +133,7 @@ custom exchanges:
 - ✅ **Create unique event types** : Key events should be easily identifiable and have a unique
   names.
 - ❌ **Don't listen to debug events inside your exchange**: While it's possible to call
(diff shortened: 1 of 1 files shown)


JoviDeCroock (approved):
Thanks!

PR #3192, opened 2023-04-27
fix(persisted): persisted queries can use GETs within URL limits

NWRichmond:
## Summary

This PR aims to resolve https://github.com/urql-graphql/urql/issues/3185.

Currently, `@urql/exchange-persisted`'s `options.preferGetForPersistedQueries` doesn't allow for `persistedOperation.context.preferGetMethod` to be set to `true` or `'within-url-limit'`, and instead sets `persistedOperation.context.preferGetMethod` to `'force'` when `options.preferGetForPersistedQueries` is `true`. This leads to broken GET requests when queries become massive.

Another issue was that although the TSDocs suggest that [the default value](https://github.com/urql-graphql/urql/blob/cb628ca4e9207b1928a5e137fddbd14c33afb6e4/exchanges/persisted/src/persistedExchange.ts#L42) of `options.preferGetForPersistedQueries` is `true`, this wasn't reflected in the logic.

## Set of changes

In `@urql/exchange-persisted`:
- The type of `options.preferGetForPersistedQueries` has changed from `boolean | undefined` to match the value of the [Client](https://github.com/urql-graphql/urql/blob/main/docs/api/core.md#client)'s `preferGetMethod` option, which includes `'force'` and `'within-url-limit'`.
- Sets a default value of `options.preferGetForPersistedQueries` (to `true`) to match the TSDocs.
- When `options.preferGetForPersistedQueries` is `true`, we no longer set `operation.context.preferGetMethod` to `'force'`. Instead, `operation.context.preferGetMethod` is set to `true`.
  - This extends to all values of `options.preferGetForPersistedQueries`. The value set there will be the value set in `operation.context.preferGetMethod`.
- Removes a [check on the value of `preferGetForPersistedQueries `](https://github.com/urql-graphql/urql/blob/cb628ca4e9207b1928a5e137fddbd14c33afb6e4/exchanges/persisted/src/persistedExchange.ts#L164) before setting `persistedOperation.context.preferGetMethod`. This allows `persistedOperation.context.preferGetMethod` to be set to `false` if `options.preferGetForPersistedQueries` is set to `false`.
- Adds unit tests to validate the changes described here.

### DIFF
--- a/.changeset/fair-eels-march.md
+++ b/.changeset/fair-eels-march.md
@@ -1,5 +1,5 @@
 ---
-'@urql/exchange-persisted': patch
+'@urql/exchange-persisted': major
 ---
 
 Update the `preferGetForPersistedQueries` option to include the `'force'` and `'within-url-limit'` values from the Client's `preferGetMethod` option. The default value of `true` no longer sets `OperationContext`'s `preferGetMethod` setting to `'force'`. Instead, the value of `preferGetForPersistedQueries` carries through to the `OperationContext`'s `preferGetMethod` setting for persisted queries.
--- a/.changeset/fair-eels-march.md
+++ b/.changeset/fair-eels-march.md
@@ -2,4 +2,4 @@
 '@urql/exchange-persisted': patch
 ---
 
-Update the `preferGetForPersistedQueries` option to include the 'force' and 'within-url-limit' values from the Client's `preferGetMethod` option. The default value of `true` no longer sets `OperationContext`'s `preferGetMethod` setting to `'force'`. Instead, the value of `preferGetForPersistedQueries` carries through to the `OperationContext`'s `preferGetMethod` setting for persisted queries.
+Update the `preferGetForPersistedQueries` option to include the `'force'` and `'within-url-limit'` values from the Client's `preferGetMethod` option. The default value of `true` no longer sets `OperationContext`'s `preferGetMethod` setting to `'force'`. Instead, the value of `preferGetForPersistedQueries` carries through to the `OperationContext`'s `preferGetMethod` setting for persisted queries.
--- a/.changeset/fair-eels-march.md
+++ b/.changeset/fair-eels-march.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-persisted': patch
+---
+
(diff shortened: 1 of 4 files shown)


exchanges/persisted/src/persistedExchange.ts
```diff
@@ -159,11 +165,9 @@ export const persistedExchange =
                 sha256Hash,
               },
             };
-            if (
-              persistedOperation.kind === 'query' &&
-              preferGetForPersistedQueries
```
NWRichmond, line 164: Removing the check on `preferGetForPersistedQueries` allows `persistedOperation.context.preferGetMethod` to be set to `false`.

exchanges/persisted/src/persistedExchange.ts
```diff
@@ -118,7 +123,8 @@ export const persistedExchange =
   ({ forward }) => {
     if (!options) options = {};
 
-    const preferGetForPersistedQueries = !!options.preferGetForPersistedQueries;
+    const preferGetForPersistedQueries =
+      options.preferGetForPersistedQueries ?? true;
```
kitten, line 127: Two things 🤔 I think we shouldn't change the default here. I believe I simply typo'd the default value in the TSDocs since I wrote many of them all at once 😅

The other thing is that we currently don't use the bullish coalescing operator or chaining operators. Currently, due to some JS engine edge cases transpilation for them is still turned on, and that leads to some unnecessary code in the bundle. Either way, I suppose, since the default was wrong this could be removed ✌️


```suggestion
      options.preferGetForPersistedQueries;
```

.changeset/fair-eels-march.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/exchange-persisted': patch
```
kitten, line 2:
```suggestion
'@urql/exchange-persisted': major
```

As discussed, we'll have to mark this as a major, since it's a breaking change

NWRichmond: Indeed. Fixed in https://github.com/urql-graphql/urql/pull/3192/commits/7ea17f77112b7e595fb4a107abe650c18c711554.

exchanges/persisted/src/persistedExchange.ts
NWRichmond: Thanks! That makes sense. Fixed in https://github.com/urql-graphql/urql/pull/3192/commits/a9d74916122a4056d7faf27ea3e3fd1b6160ed41.

I also added the `&& preferGetForPersistedQueries` check back to the `persistedOperation.context.preferGetMethod` assignment:
1. So that we don't set `persistedOperation.context.preferGetMethod` to `undefined`
2. To continue treating a falsey value as the default

kitten (approved):

PR #3195, opened 2023-04-28
Fix re-rendering when another component reuses the same query

nathan-knight:
## Summary
When component A uses `useQuery` and has already loaded a result and then component B mounts and uses the same query, it currently causes component A to re-render because `operation` is not shallowly equal even though nothing about the result is changing.

This is a performance issue because it causes unnecessary re-renders.

It also can cause an infinite render loop with lazily loaded components because it can trigger a state update further up the tree during the initial render of the lazily loaded component which makes React back out of the render (which happened in my application, although I've had trouble creating a minimal reproduction). This issue is resolved with this change in my application.

## Set of changes
- Added a special case to `isShallowDifferent` for `operation` where it checks the operation key if the `operation`s are not equal
- Fixed a typo in `CONTRIBUTING.md`

### DIFF
--- a/.changeset/fifty-sheep-battle.md
+++ b/.changeset/fifty-sheep-battle.md
@@ -2,4 +2,4 @@
 'urql': patch
 ---
 
-Fix re-rendering when another component reuses the same query without a change in the result
+Avoid unnecessary re-render when two components use the same query but receive unchanging results, due to differing operations
--- a/.changeset/fifty-sheep-battle.md
+++ b/.changeset/fifty-sheep-battle.md
@@ -1,0 +1,5 @@
+---
+'urql': patch
+---
+
+Fix re-rendering when another component reuses the same query without a change in the result
--- a/packages/react-urql/src/hooks/state.ts
+++ b/packages/react-urql/src/hooks/state.ts
@@ -9,6 +9,14 @@ export const initialState = {
   operation: undefined,
 };
 
+// Two operations are considered equal if they have the same key
+const areOperationsEqual = (
+  a: { key: number } | undefined,
+  b: { key: number } | undefined
+) => {
+  return a === b || !!(a && b && a.key === b.key);
+};
+
 /**
  * Checks if two objects are shallowly different with a special case for
  * 'operation' where it compares the key if they are not the otherwise equal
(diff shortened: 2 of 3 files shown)


packages/react-urql/src/hooks/state.ts
```diff
@@ -9,10 +9,21 @@ export const initialState = {
...
+  for (const key in a) if (!(key in b)) return true;
+  for (const key in b) {
+    if (a[key] !== b[key]) {
+      // Two operations are considered equal if they have the same key
+      if (key === 'operation' && a.operation?.key === b.operation?.key) {
+        continue;
+      }
+      return true;
+    }
+  }
```
kitten, line 26: Ha, nice catch 😅 I didn't expect this old code to have such a severe lingering issue.
Re. the implementation, we currently avoid nullish chaining and coalescing operators. We don't have a lint set up for that right now, so obviously you couldn't have known, but it's basically to avoid some issues with quirky JS engines that are still in range of our support. Currently, this will lead to some excessively large polyfill code

We could just avoid them by type casting but we can also do

```suggestion
const isShallowDifferent = <T extends Record<string, any>>(a: T, b: T) => {
  for (const key in a) if (!(key in b)) return true;
  for (const key in b) {
    if (
      (key === 'operation' && a[key].key === b[key].key) ||
      a[key] !== b[key]
    ) return true;
  }
```

...since we don't have exact types here anyway

.changeset/fifty-sheep-battle.md
```diff
@@ -0,0 +1,5 @@
+---
+'urql': patch
+---
+
+Fix re-rendering when another component reuses the same query without a change in the result
```
kitten, line 5:
```suggestion
Avoid unnecessary re-render when two components use the same query but receive unchanging results, due to differing operations
```

packages/react-urql/src/hooks/state.ts
nathan-knight: If `a[key].key === b[key].key` then we don't want to return true, I can adjust it to not use the operator

nathan-knight: I introduced a helper function to keep it readable but if you'd prefer not to have an extra function call for performance reasons then I can inline it

kitten (approved):

PR #2612, opened 2022-08-16
major(graphcache): stop buffering operations while the cache hydrates

JoviDeCroock:
Resolves #1512 

## Summary

This enables graph-cache to forward operations while hydration is ongoing, this benefits application in various ways

- they can benefit from synchronous rehydration due to the `ssrExchange`
- does not cause an extra delay for empty caches or missing operations

## Set of changes

- remove buffering of operations

PR #3196, opened 2023-04-28
fix(graphcache): Add missing retries after offline rehydration

### LINKED ISSUE #3093: Using `graphcache@5.0.0` or 6 the cache in the default storage is not loaded anymore on the first page load
After upgrading from 4 to 5 (or 6) `@urql/exchange-graphcache` doing some tests with a fake delay on the backend I discovered that when I refresh the page (F5 or Ctrl+R) the cache present in the IndexedDB is not used.

**The REPL** (with SvelteKit too) is here: https://codesandbox.io/p/sandbox/romantic-feather-xmc826

**Steps**:

1. Go to https://xmc826-5173.csb.app/todos

1. the todos are downloaded, showed and stored in the IndexedDB after 3 seconds of [fake delay on the server resolver](https://codesandbox.io/p/sandbox/apollo-server-urql-svelte-crud-h1pcl?file=%2Fschema.js&selection=%5B%7B%22endColumn%22%3A25%2C%22endLineNumber%22%3A132%2C%22startColumn%22%3A7%2C%22startLineNumber%22%3A132%7D%5D)

1. Click on the reload button in the navbar (or the browser's one)

1. The "Loading listStore..." message appears during the 3 fake delay seconds

1. I expect instead to see the todos immediately (during the `cache-and-network` call which is still happening)

If I rollback only `"@urql/exchange-graphcache"` to `"4.4.3"` this "cold cache" behavior works again.

**What could be causing the problem?**

Thank you in advance.

### Urql version

[package.json](https://codesandbox.io/p/sandbox/romantic-feather-xmc826?selection=%5B%7B%22endColumn%22%3A13%2C%22endLineNumber%22%3A19%2C%22startColumn%22%3A13%2C%22startLineNumber%22%3A19%7D%5D&file=%2Fpackage.json)

### Validations

kitten:
Resolves #3093

## Summary

Adds missing retries of query operations when offline rehydration has completed and fix the timing of flushing the failed queue of operations after rehydration.

The former can lead to operations only awaiting network requests even after finishing storage rehydration, since no additional operations would be issued.

The latter can lead to stalled operations when the device is offline on load.

## Set of changes

- Add query operations to `failedQueue` while rehydrating
- Delay timing of `flushQueue` to after storage rehydration
- Move `onOnline` register to `flushQueue`

### DIFF
--- a/.changeset/wet-crabs-dream.md
+++ b/.changeset/wet-crabs-dream.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-graphcache': patch
+---
+
+Retry operations against offline cache and stabilize timing of flushing failed operations queue after rehydrating the storage data.
--- a/exchanges/graphcache/src/cacheExchange.ts
+++ b/exchanges/graphcache/src/cacheExchange.ts
@@ -71,6 +71,7 @@ export const cacheExchange =
     const store = new Store<C>(opts);
 
     if (opts && opts.storage) {
+      store.data.hydrating = true;
       opts.storage.readData().then(entries => {
         hydrateData(store.data, opts!.storage!, entries);
       });
--- a/exchanges/graphcache/src/offlineExchange.test.ts
+++ b/exchanges/graphcache/src/offlineExchange.test.ts
@@ -5,7 +5,6 @@ import {
   Operation,
   OperationResult,
 } from '@urql/core';
-import { print } from 'graphql';
 import { vi, expect, it, describe, beforeAll } from 'vitest';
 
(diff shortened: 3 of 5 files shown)


trcoffman:
This looks to me like it would solve #3093 but would it possibly reintroduce some kind of issues for the ssrExchange that #2612 fixed? I don't really have any experience with React SSR or the ssrExchange. Could re-executing these queries against the cache after storage hydration cause stale data from the cache to show up in query results, overriding the data from the ssrExchange?

Issue #3199, opened 2023-04-29
Multiple duplicated requests using `cache-and-network` or `netowrk-only`


frederikhors:
### Describe the bug

I have an issue with duplicated requests (I remove the now deprecated `dedupExchange`: I read all CHANGELOGs and announcement issues like https://github.com/urql-graphql/urql/issues/3114).

If I use this code:

```ts
import { getContextClient, queryStore } from '@urql/svelte';

const currentPlayerStore = queryStore({
  client: getContextClient(),
  query: CurrentPlayerDocument,
  // the default requestPolicy is 'cache-first' // one request only, ok
  // requestPolicy: 'cache-and-network', // two requests, duplicated
  // requestPolicy: 'network-only' // THREE requests! Duplicated!
});

$: currentPlayer = $currentPlayerStore.data?.currentPlayer;
```

As you can read in the comments I isolated the issue: if I use the `'cache-first'` policy it only send one request.

If I use the `requestPolicy: 'cache-and-network'` the requests are 2, duplicated, the same ones.

If I use the `requestPolicy: 'network-only'` the requests are 3 now! Crazy!

What am I doing wrong?

### Reproduction

This repl is using your modified `graphcache` version that fixes https://github.com/urql-graphql/urql/issues/3093.

**REPL**: https://x2y2ub-5173.csb.app/todos

**Steps**:

1. click on "Todos list"

1. open the browser dev tools

1. reload the page

1. the requests are 2!

1. If you use `network-only` in [the list file](https://codesandbox.io/p/sandbox/issue-urql-6-duplicated-requests-x2y2ub?file=%2Fsrc%2Froutes%2Ftodos%2F%2Bpage.svelte&selection=%5B%7B%22endColumn%22%3A35%2C%22endLineNumber%22%3A14%2C%22startColumn%22%3A5%2C%22startLineNumber%22%3A14%7D%5D) the requests are 3 now

### Urql version

I upgraded:

```diff            
"devDependencies": {        
  "@graphql-codegen/cli": "3.3.1",
  "@graphql-codegen/typed-document-node": "4.0.1",
  "@graphql-codegen/typescript": "3.0.4", 
- "@urql/core": "2.6.1",  
+ "@urql/core": "4.0.7",  
  "@urql/devtools": "2.0.3",
- "@urql/exchange-graphcache": "4.4.3", 
+ "@urql/exchange-graphcache": "6.0.4", 
- "@urql/exchange-persisted-fetch": "1.3.4",  
[...]

PR #3200, opened 2023-04-29
fix(graphcache): Prevent offlineExchange from issuing duplicate operations

### LINKED ISSUE #3093: Using `graphcache@5.0.0` or 6 the cache in the default storage is not loaded anymore on the first page load

kitten:
Follow-up to #3199
Resolves #3093

## Set of changes

- Add a crude teardown+seen routine to `failedQueue` in `offlineExchange`'s `flushQueue`

### DIFF
--- a/.changeset/hungry-zebras-count.md
+++ b/.changeset/hungry-zebras-count.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-graphcache': patch
+---
+
+Prevent `offlineExchange` from issuing duplicate operations.
--- a/exchanges/graphcache/src/offlineExchange.ts
+++ b/exchanges/graphcache/src/offlineExchange.ts
@@ -151,11 +151,16 @@ export const offlineExchange =
       const flushQueue = () => {
         if (!isFlushingQueue) {
           isFlushingQueue = true;
+
+          const sent = new Set<number>();
           for (let i = 0; i < failedQueue.length; i++) {
             const operation = failedQueue[i];
-            if (operation.kind === 'mutation')
-              next(makeOperation('teardown', operation));
-            next(operation);
+            if (operation.kind === 'mutation' || !sent.has(operation.key)) {
+              if (operation.kind !== 'subscription')
+                next(makeOperation('teardown', operation));
+              sent.add(operation.key);
+              next(operation);
+            }
           }
 
           failedQueue.length = 0;
--- a/exchanges/graphcache/src/offlineExchange.ts
+++ b/exchanges/graphcache/src/offlineExchange.ts
@@ -159,7 +159,7 @@ export const offlineExchange =
               if (operation.kind !== 'subscription')
                 next(makeOperation('teardown', operation));
               sent.add(operation.key);
-              next(operation);
+              next(toRequestPolicy(operation, 'cache-first'));
             }
           }
 


frederikhors:
I forgot to write here too that this still doesn't fix https://github.com/urql-graphql/urql/issues/3093 as is unfortunately.

kitten:
@frederikhors: Should be fixed now. In short, coming back to this, I think it's best to just force the flushed operations to always use `cache-first`.

frederikhors:
@kitten thank you very much. I'll try in an hour-ish.

frederikhors:
I'm trying right now...

frederikhors:
Please do not release the others, give me 5 minutes.

frederikhors:
Ok. I tried. It works, but.

The cached data is used on the cold load now.

The are no multiple requests for the same call.

But I realized that now on cold start on page reload there is no longer the loading indicator I had before, using the following custom exchange:

```ts
const pendingRequestsExchange: Exchange = ({ forward }) => {
  function before(op: Operation) {
    if (op.kind === 'subscription') return;
    if (op.kind && op.kind === 'teardown') {
      console.log('call deletePendingRequest');
      deletePendingRequest(op.key);
    } else {
      console.log('call addPendingRequest');
      addPendingRequest(op.key);
    }
  }
  function after(op: OperationResult) {
    deletePendingRequest(op.operation.key);
  }
  return (ops$) => {
    const forward$ = pipe(ops$, tap(before));
    return pipe(forward(forward$), tap(after));
  };
};
```

This is happening because it now calls immediately `addPendingRequest` and `deletePendingRequest` without waiting the pending request.

This is not what I expect during the cold start with the cache data already there.

frederikhors:
The amazing @kitten is working on this [here](https://github.com/urql-graphql/urql/pull/3200).

With the latest commit the things are working now, except [this](https://github.com/urql-graphql/urql/pull/3200#issuecomment-1569845321):

REPRODUCTION: https://codesandbox.io/p/sandbox/issue-urql-6-duplicated-requests-forked-0ybcue.

Steps:

1. click on "Todos list"

1. open the browser dev tools

1. reload the page

1. see in the console `call addPendingRequest` and immediately (not after the fake delay of the server) `call deletePendingRequest`

frederikhors:
Added reproduction for the latest issue here: https://github.com/urql-graphql/urql/issues/3093#issuecomment-1569869068.

kitten:
@frederikhors: I can't debug your exchange for you, but you will have to adjust your `after` call to be more precise. If this is just to keep track of global loading states, I'd suggest to be more accurate with `cache-only` operations in `after` and/or maybe adding a delay or debounce here.

frederikhors:
This new behavior is reproduced here: https://codesandbox.io/p/sandbox/issue-urql-6-duplicated-requests-forked-0ybcue. TOu can see the `after` call is never issued.

PR #3214, opened 2023-05-12
major(next): support Next 13 and React Server Components

### LINKED ISSUE #3208: RFC (next-urql): Support for Next.js app router
## Summary

Next.js has just released [version 13.4](https://nextjs.org/blog/next-13-4), and with it the new app directory is now stable. I'd like to see `next-urql` updated to support the new app router, with updated documentation on how to set it up.

## Proposed Solution

The existing `withUrqlClient` HOC function currently accepts a component of type `NextPage` or `NextApp`, both of which are no longer relevant with the new app directory. The `next-urql` package should support the new app directory (with a [`layout.js` file](https://nextjs.org/docs/getting-started/react-essentials#rendering-third-party-context-providers-in-server-components)) and ensure that it uses the [`use client`](https://nextjs.org/docs/getting-started/react-essentials#the-use-client-directive) directive where appropriate.

### LINKED ISSUE #3219: TS Error when trying to use the next-urql default exchanges
### Describe the bug

Hello! 👋 
I noticed the [Next.js](https://formidable.com/open-source/urql/docs/advanced/server-side-rendering/#nextjs) docs mentions:
> One added caveat is that these options may not include the exchanges option because next-urql injects the ssrExchange automatically at the right location. If you're setting up custom exchanges you'll need to instead provide them in the exchanges property of the returned client object.

I also see where this is [handled in the code](https://github.com/urql-graphql/urql/blob/1470a9c7474c75b65303b503c0a3a51e895f1a38/packages/next-urql/src/with-urql-client.ts#L111-L114). 

However, when I try to forgo passing `exchanges`:

```ts
export default withUrqlClient(() => ({
  url: 'http://localhost:3000/graphql',
}))(App);
```

I receive the following TS error: 

```
Property 'exchanges' is missing in type '{ url: string; }' but required in type 'ClientOptions'.ts(2741)
```

To reproduce this, please see the reproduction link. You'll need to open a terminal and run `yarn tsc` to see the error.

### Reproduction

https://codesandbox.io/s/vibrant-wozniak-7lhu9m

### Urql version

urql v4.0.2
next-urql 5.0.1

### Validations

JoviDeCroock:
Resolves #3208
Resolves #3219

## Summary

This creates a new package `@urql/next` (and replaces `next-urql`), the goal is to achieve good Next 13 support.

In essence this creates two new entry points

- `@urql/next` the entry points for components rendered on the server and hydrated on the client, this uses a few next-internals to hydrate data as DOM streams to the client
- `@urql/next/rsc` this solely intends to create a client that can be used and shared across
server components. The sharing method is `react.cache`

You can play with an example [here](https://github.com/JoviDeCroock/next-app-dir-data-rehydration-exploration/) the main idea is that in Next we need a way to retrieve context values on the client so they can be shared. During the stream the values will be flushed, we use this to ensure we have all the data we need. In the example you can see that we have two suspense boundaries, neither will initiate a fetch on the client but instead you'll see a `script` tag get populated twice, one for each boundary. We read these `<script` tags and add them to the ssr-exchange.

TODO:

- [x] resolve the build warning
- [x] add an example
- [x] check if we can find a good way of testing this (spoiler: nope)
- [x] update documentation
- [x] find a way around the key warning for `<script />`
- [x] find a way to add `"use client"` to the `@urql/next` entry-point
- [x] think about handling `text/event-stream` and `multipart/mixed`, do we just remove `@defer` and `@stream` during ssr? Or can we add a mechanism that ensures these are handled correctly i.e. by means of a `useFragment`-like mechanism throwing promises as well

Having an awkward problem where it's not letting me create a changeset on my fork 😅 

People can still use `next-urql` for the old next supported ways


packages/next-urql/src/index.ts
```diff
@@ -1,3 +1,3 @@
-export { withUrqlClient } from './with-urql-client';
-export { initUrqlClient } from './init-urql-client';
-export * from './types';
+export * from 'urql';
+export { useQuery } from './useQuery';
+export { UrqlProvider } from './Provider';
```
JoviDeCroock, line 3: This is currently erroring during build

packages/next-urql/src/useQuery.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  const [result, execute] = orig_useQuery(args);
+
+  const transported: any = {};
+  for (const key of transportKeys) {
+    // @ts-ignore
+    transported[key] = result[key];
+  }
+
+  useUrqlValue(request.key, transported);
```
JoviDeCroock, line 22: this reads a bit wonky, I agree but basically we use `useUrqlValue` the first time to read whether or not there's data to append to the ssrExchange and the second time is mainly used on the server to write data. We could split this out into two hooks, I didn't want to conditionally call mainly as that is agains the rules of hooks

packages/next-urql/src/useUrqlValue.ts
```diff
@@ -0,0 +1,46 @@
...
+  if (!ssrExchange) {
+    throw new Error('forgot to pass an "ssrExchange" to the UrqlProvider.');
+  }
+
+  if (typeof window == 'undefined') {
+    if (rehydrationContext && value) {
+      rehydrationContext.operationValuesByKey[operationKey] = value;
+    }
+  } else {
+    const stores = (window[urqlTransportSymbol as any] ||
```
JoviDeCroock, line 26: On the client we can have multiple stores, one for each Suspense boundary we have traversed. Eventually we'll need to clean up stores, we could however do something along the lines of just consuming all stores into the ssrExchange and removing them from the window object.

exchanges/graphcache/default-storage/package.json
```diff
@@ -8,9 +8,9 @@
   "source": "../src/default-storage/index.ts",
   "exports": {
     ".": {
-      "types": "../dist/urql-exchange-graphcache-default-storage.d.ts",
       "import": "../dist/urql-exchange-graphcache-default-storage.mjs",
       "require": "../dist/urql-exchange-graphcache-default-storage.js",
+      "types": "../dist/urql-exchange-graphcache-default-storage.d.ts",
```
JoviDeCroock, line 13: I think this moved itself when running build or prepare 😅

scripts/rollup/banner-plugin.mjs
```diff
@@ -0,0 +1,8 @@
+export default function banner () {
+  return {
+    name: 'banner',
+    renderChunk (code, chunk) {
+      return chunk.name === 'urql-next' ? `"use client"\n\n${code}` : code
```
kitten, line 5: The only thing I'm a little concerned about here is that we're not using Rollup's banner feature. Mainly, there's nothing wrong with not using it, but I suspect this will mess with sourcemaps

After #3201 I'm now even avoiding any transformations (outside of Rollup plugins) to not having to deal with returning modified sourcemaps or having to use `magic-string` for simple changes. But I suspect, just using the Rollup banner feature could maybe help?

https://github.com/urql-graphql/urql/issues/3182 has more information on how I tested sourcemaps to be working in the longer bundles, but I can test this here as well

packages/next-urql/src/Provider.ts
```diff
@@ -0,0 +1,25 @@
+'use client';
+
+import React from 'react';
+import { Provider, SSRExchange, Client } from 'urql';
```
kitten, line 4: I haven't changed `createContext` yet to use `import * as React from 'react'; React.createContext`. Should this be done in this same PR? https://github.com/urql-graphql/urql/blob/2107283a7fbe5328641efb67e9784d4d60541dbc/packages/react-urql/src/context.ts#L1

packages/next-urql/src/rsc.ts
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
```
kitten, line 1:
```suggestion
import * as React from 'react';
```

Does this still cause trouble in some cases?

kitten (commented):
Just some minor questions. Sourcemaps question is the only relevant for publishing and correctness.

We're missing TSDocs for the public APIs, but I think I can add those later on before publishing as well

packages/next-urql/src/Provider.ts
JoviDeCroock: Importing contexts is only relevant in server-components, this means that as long as people use the client when they leverage `/rsc` we are good. This is demonstrated in the examples as well, here we create a `React.cached` client and we get it when we want to use it. The good thing here is that in RSC, folks can just use `await` and leverage the data, this can afterwards be passed down to the client over props.

scripts/rollup/banner-plugin.mjs
JoviDeCroock: Gotcha, I'll see how the banner option works, for reference https://rollupjs.org/configuration-options/#output-banner-output-footer

JoviDeCroock:
Should I copy the `useQuery` TSDocs from the react implementation or is there a way to inherit?

packages/next-urql/src/useUrqlValue.ts
```diff
@@ -0,0 +1,47 @@
...
+export const urqlTransportSymbol = Symbol.for(symbolString);
+
+export type UrqlResult = { data: any; error: any };
+
+export function useUrqlValue(operationKey: number, value?: UrqlResult): void {
+  const ssrExchange = React.useContext(SSRContext);
+  const rehydrationContext = useDataHydrationContext();
+
+  if (!ssrExchange) {
+    throw new Error('forgot to pass an "ssrExchange" to the UrqlProvider.');
```
sylv, line 17: This can also be caused by UrqlProvider not being used/not being a parent component, mentioning that in the error would have saved me an embarassing amount of time

kitten (approved):
LGTM :+1: I haven't tested this in code, but I think we can release this as a prerelease and see what happens 😄

DeclanBoller:
Would be cool to see this shipped, would help me with migrating to App Router for some of my company's NextJS projects 😃

JoviDeCroock:
@kitten I agree, we can test it in the example with a beta or alpha tag

kitten:
@JoviDeCroock: Published as `@urql/next@1.0.0-beta.1`

docs/advanced/server-side-rendering.md
```diff
@@ -177,254 +177,156 @@ we'll have to import from. `preact-ssr-prepass`.
...
+We now have two ways to leverage `@urql/next`, one being part of a Server component
+or being part of the general `app/` folder.
 
-We're now able to wrap any page or `_app.js` using the `withUrqlClient` higher-order component. If
-we wrap `_app.js` we won't have to wrap any individual page.
+In a server component we will import from `@urql/next/rsc`
 
-```js
-// pages/index.js
+```ts
```
seanparmelee, line 196: the syntax highlighting is a bit off, maybe `tsx` would work better
image

docs/advanced/server-side-rendering.md
```diff
@@ -177,254 +177,156 @@ we'll have to import from. `preact-ssr-prepass`.
...
-    url: 'your-url',
-  })
-  // Cannot specify { ssr: true } here so we don't wrap our component in getInitialProps
-)(Todos);
 ```
 
-The above example will make sure the page is rendered as a static-page, It's important that
-you fully pre-populate your cache so in our case we were only interested in getting our todos,
-if there are child components relying on data you'll have to make sure these are fetched as well.
+It is important that we pas both a client as well as the `ssrExchange` to the `Provider`
```
seanparmelee, line 252:
```suggestion
It is important that we pass both a client as well as the `ssrExchange` to the `Provider`
```

docs/advanced/server-side-rendering.md
```diff
@@ -177,254 +177,156 @@ we'll have to import from. `preact-ssr-prepass`.
...
+these will also get flushed as they complete and re-hydrated.
 
-### Stale While Revalidate
+> When data is used throughout the application we advise against
+> rendering this as part of a server-component so you can benefit
+> from the client-side cache.
 
-If we choose to use Next's static site generation (SSG or ISG) we may be embedding data in our initial payload that's stale on the client. In this case, we may want to update this data immediately after rehydration.
-We can pass `staleWhileRevalidate: true` to `withUrqlClient`'s second option argument to Switch it to a mode where it'll refresh its rehydrated data immediately by issuing another network request.
+### Invalidating data from a server-component
```
seanparmelee, line 307: fwiw, this doc goes back and forth between "server-component" and "server component" in case you wanted to consolidate to one or the other

docs/advanced/server-side-rendering.md
```diff
@@ -177,254 +177,156 @@ we'll have to import from. `preact-ssr-prepass`.
...
+    </main>
+  );
+}
 ```
 
-Be aware that wrapping the `_app` component using `withUrqlClient` with the `{ ssr: true }`
-option disables Next's ["Automatic Static
-Optimization"](https://nextjs.org/docs/advanced-features/automatic-static-optimization) for
-**all our pages**. It is thus preferred to enable server-side rendering on a per-page basis.
+When we aren't leveraging server components we will import the things we will
```
seanparmelee, line 227: Should this be
> When we aren't leveraging server components ~we will import the things~ we will need to do a bit more setup

?

docs/advanced/server-side-rendering.md
```diff
@@ -177,254 +177,156 @@ we'll have to import from. `preact-ssr-prepass`.
...
-npm install --save next-urql react-is urql graphql
+npm install --save @urql/next urql graphql
 ```
 
-The peer dependency on `react-is` is inherited from `react-ssr-prepass` requiring it.
-
-Note that if you are using Next before v9.4 you'll need to polyfill fetch, this can be
-done through [`isomorphic-unfetch`](https://www.npmjs.com/package/isomorphic-unfetch).
+We now have two ways to leverage `@urql/next`, one being part of a Server component
+or being part of the general `app/` folder.
```
seanparmelee, line 192: For folks that are using Next 13, but haven't migrated to use the `app` folder yet, is the recommendation to continue using the `next-urql` package? or is `@urql/next` meant for those use cases as well?

seanparmelee: Apologies, just saw the PR description mentions:

> People can still use `next-urql` for the old next supported ways

I think it'd be good if these docs included mention of this. I also think the old docs (i.e. the ones for when using the [Pages Router](https://nextjs.org/docs/pages)) should probably stick around for the people who aren't able to upgrade to Next 13 and/or the app router yet. Perhaps this page ends up having two Next.js sections:
* Next.js (App Router)
* Next.js (Pages Router)

?

kitten: This is a WIP, but eventually, I'm in favour of actually preserving the legacy docs in place, so people can still find the legacy package. So TBD

packages/next-urql/src/useQuery.ts
```diff
@@ -0,0 +1,221 @@
...
+    args.query,
+    (args.variables || {}) as AnyVariables
+  );
+  useUrqlValue(request.key);
+
+  const [result, execute] = orig_useQuery(args);
+
+  const transported: any = {};
+  for (const key of transportKeys) {
+    transported[key] = result[key];
```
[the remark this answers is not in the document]
JoviDeCroock: Yes, tbh this is a demo ideally we add `extractData` in the `useUrqlValue` and disregard the second argument

JoviDeCroock:
Ported over to `extractData()` feel free to issue another beta

kitten:
@JoviDeCroock: Published as `@urql/next@1.0.0-beta.2` under the `beta` tag

adikari:
Would this allow you to load the query in RSC then mutate it using the client? Basically populate the ssr cache data in the client component.

JoviDeCroock:
> when using @urql/exchange-graphcache, it show warning `Couldn't find __typename when writing`

That's fixed in the upcoming release of `@urql/core` https://github.com/urql-graphql/urql/pull/3294

focux:
Thanks for your work adding support for the Next's app folder. The only issue that I've encountered so far is that the `getClient` function is not properly typed, it returns `any` but besides that, this is awesome!

This is the workaround that I'm using to make the types work:
```ts
const { getClient: getClientUntyped } = registerUrql(makeClient);
const getClient = getClientUntyped as () => Client;
```

focux:
I'm using `@urql/next:1.0.0-beta.2` and for some reason, I can't use the `UrqlProvider` component on my layout component which is a server component. This is the error that I'm getting:


> TypeError: createContext only works in Client Components. Add the "use client" directive at the top of the file to use it. Read more: https://nextjs.org/docs/messages/context-in-server-component

I thought that it was related to the `UrqlProvider` file not having the `use client` directive but I saw that it's already there. The workaround, for now, is to create a new component wrapping the `UrqlProvider` component and adding the `use client` directive into that file.

```ts
"use client";

import { client, ssr } from "@/lib/urql-client";
import { UrqlProvider } from "@urql/next";
import { ReactNode } from "react";

export function Provider({ children }: { children: ReactNode }) {
  return (
    <UrqlProvider client={client} ssr={ssr}>
      {children}
    </UrqlProvider>
  );
}

```

docs/advanced/server-side-rendering.md
```diff
@@ -177,254 +177,156 @@ we'll have to import from. `preact-ssr-prepass`.
...
+It is important that we pas both a client as well as the `ssrExchange` to the `Provider`
+this way we will be able to restore the data that Next streams to the client later on
+when we are hydrating.
 
-The `getServerSideProps` and `getStaticProps` functions only run on the **server-side** — any
-code used in them is automatically stripped away from the client-side bundle using the
-[next-code-elimination tool](https://next-code-elimination.vercel.app/). This allows **executing
-our schema directly** using `@urql/exchange-execute` if we have access to our GraphQL server:
+The next step is to query data in your client components by means of the `useQuery`
+method defined in `@urql/next`.
```
ppold, line 257: does it _have_ to import them from `@urql/next` rather than the base urql package? I'd prefer to have it import the base one to make the component re-usable in contexts outside of Next.js (like React Native).

docs/advanced/server-side-rendering.md
```diff
@@ -177,254 +177,156 @@ we'll have to import from. `preact-ssr-prepass`.
...
-import { withUrqlClient, initUrqlClient } from 'next-urql';
-import { ssrExchange, dedupExchange, cacheExchange, fetchExchange, useQuery } from 'urql';
-
-const TODOS_QUERY = `
-  query { todos { id text } }
-`;
+const ssr = ssrExchange();
+const client = createClient({
+  url: 'https://trygql.formidable.dev/graphql/web-collections',
+  exchanges: [cacheExchange, ssr, fetchExchange],
```
ppold, line 240: Since this is using 'use client', what's the best way to get a reference to the context like `withUrqlClient` provides? This is for cases where a need for a different method of passing the Authorization header is required during SSR.

docs/advanced/server-side-rendering.md
JoviDeCroock: This is a valid point, it's pretty hard to circumvent this however as we need a trigger to check the data being streamed to the client. i.e. when a suspense boundary resolves on the server and hydration continues it will naturally call `useQuery` else we would need an additional hook. Maybe there are ways to hook into next/React telling us more data is available but I haven't found it yet 😅

docs/advanced/server-side-rendering.md
JoviDeCroock: That's a good question, `withUrqlClient` obtains this only when it's called through `getInitialProps` next does not expose it elsewhere (apart from `next/headers`) so this might be a question for their side 😅

jermatic:
Excited to see this coming. I havent used the new app router specifically because I refuse to give up my use of urql.

docs/advanced/server-side-rendering.md
ppold: true, I hope they have a proper answer because this is really puzzling, even just passing credentials for an API on the same domain isn't sending the auth cookies when the query is executed!

Enalmada:
I am attempting to migrate a next 13 app dir todo list app from apollo to urql but having trouble getting graphqlcache to work.  It feels like something is getting cached in the ssr of urql when using graphqlcache.  

The app has a list page and another page to create/edit/delete which does a router.push back to list page on success.  Components with queries all "use client".  There is a normalized graphqlcache setup to [push/filter](https://formidable.com/open-source/urql/docs/graphcache/cache-updates/#updating-lists-or-links) create and deletes.   

When adding or deleting a task and navigating back to the list page, the results don't show.  They continue not to show on hard refresh of the list page.  I find it very interesting I can't see the new/deleted tasks in the page source inconsistent with db state).   To get it to refresh, one must navigate to another page, hard refresh, and then they show on return to the list page.  I am hoping this is a clue to what is going on.  Reading through the directions here, I have tried adding  router.refresh() to page, cache: 'no-store' to fetchOptions, `export const dynamic = 'force-dynamic'; export const revalidate = 0;` to the page with no effect.

Any ideas on what I or the code might be doing wrong or that I could try?

kitten:
@Enalmada just fyi, I marked this as off topic, since it doesn't concern this PR. However, feel free to ask a question in a GH discussion or in our Discord ✌️
This doesn't sound like an app dir issue to me but maybe a config issue? I'd love to give it more space to investigate rather than comments here

Enalmada:
@kitten Sorry, I thought it might just be relevant enough because I am trying to use the specific code and documentation from this PR to see if there were any issues I could identify and get fixed in my Next 13 and React Server Components test project.  I am sure you are right though there must be some config issue and will be better about reaching out on discord vs github in the future.

kitten:
@Enalmada no stress 👍 it just doesn't sound like this would be easily resolved or investigated within two replies or so

kitten (approved):

kitten (approved):

PR #3217, opened 2023-05-17
Vercel for Docs Deployment

gksander:
This PR tweaks the react-static docs config _a tiny bit_ so that we can deploy the urql docs on Vercel. The end result:

- Vercel for staging previews on PRs. [See here](https://urql-docs-m6qsbtehz-formidable-labs.vercel.app/) for an example from this PR. For staging previews, the docs will be deployed at root of deployment site.
- We'll be using Vercel for production deploys here soon as well (instead of S3), so production Vercel deployment will still be nested at `/open-source/urql` so that it can be "stitched" into the upcoming formidable.com site.
- Removing all traces of Surge for deployments.
- Once formidable.com is migrated to new Next.js site, we'll follow up and remove formideploy logic as well.

### DIFF
--- a/.github/workflows/site.yml
+++ b/.github/workflows/site.yml
@@ -70,13 +70,3 @@ jobs:
         working-directory: ./packages/site
         if: github.ref != 'refs/heads/main'
         run: pnpm run build --staging
-
-      #- name: Deploy Staging Site
-      #  working-directory: ./packages/site
-      #  if: github.ref != 'refs/heads/main'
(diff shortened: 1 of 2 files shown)


carloskelly13 (approved):

kitten:
My only problem here is that we may have to disable their GitHub integration and replace it with an action based on changed files 😅
I'm happy to look into that next week, but as it stands, the output is likely going to be a bit spammy on PRs

PR #3224, opened 2023-05-28
feat(graphcache): Add globalIDs option

### LINKED ISSUE #3176: RFC: Graphcache support changing `__typename` component of the cache
## Summary

The format for keys in the graphcache index is `<__typename>:<id>`. The `id` component is a bit ambiguous, as different apis will use different identifiers. Consequently, graphcache allows specifying a `keys` field on the options object that allows users to override how that id portion is computed for the key.

However, there is no way to do the same for the `__typename` component. This makes sense -- it is a standard part of the GQL api and there's little reason to override.

Unfortunately, the current caching model has no internal notion of inheritance (or unions). What this means is that if we have a resolver that returns `Car` whose two inhabitants are `FourWheeler` and `TwoWheeler`, a `FourWheeler` with id `123` will be cached as `FourWheeler:123`. 

Now let's say we have a mutation that unceremoniously rips the back two wheels off our car and returns it to the frontend. The id remains `123` and the entity is still a `Car`, but the underlying `__typename` would change to `TwoWheeler`. Graphcache will see this and insert a **brand new** cache entry called `TwoWheeler:123`, and not update the status of `FourWheeler:123`.

In a perfect world, this feature request would be to include better handling of that (maybe by generating multiple cache keys for each type and then having a separate type of link that joins them? not sure -- out of scope here).  However, there is a much simpler solution that handles the 95 percentile of cases (at least, that I personally encounter).

## Proposed Solution

Even if graphcache isn't really able to understand what the relationship between elements of a common inheritance structure is, the programmer does. In the same way that we are allowed to use the `keys` option on graphcache to re-key (really re-id) a record, we could also expose a `typenames` option that let's us "re-type" a record.

I don't know if this is a perfect solution. It really only works if cache keys are write-only (i.e. never parsed). As far as I can tell, this is the case, but I only gave a cursory glance.

For what it's worth, I've implemented this naive solution for our codebase, and so far it seems to be working exactly as expected. However I would love to hear the thoughts of the maintainers on whether this is a feasible (and safe!) solution. 

## Requirements

For me personally, the primary requirement is that I can use mutations that safely re-type an entity of a given inheritance chain (or union) and have the cache keep the link between it and previous versions.

## POC

https://gist.github.com/zwade/64b96dea8b483720be32eba51713a8bc

kitten:
Resolves #3176

## Summary

Add `globalIDs` option to omit typenames from entity cache keys.

This allows schemas with global ID guarantees and no ID overlap to combine entities in the cache. While this isn't recommended per se, since it's often a schema design problem, rather than a cache problem, it can be used to create strategic overlaps in the cache, when a type of an object implementing an interface “changes”

## Set of changes

- Add `globalIDs` option and update `keyOfEntity` function

### DIFF
--- a/exchanges/graphcache/src/types.ts
+++ b/exchanges/graphcache/src/types.ts
@@ -581,6 +581,17 @@ export type CacheExchangeOpts = {
    * the full keys docs.
    */
   keys?: KeyingConfig;
+  /** Enables global IDs for keying GraphQL types.
+   *
+   * @rematks
+   * When `globalIDs` are enabled, GraphQL object type names will not contribute
+   * to the keys of entities and instead only their ID fields (or `keys` return
+   * values will be used.
+   *
+   * This is useful to overlap types of differing typenames. While this isn’t recommended
+   * it can be necessary to represent more complex interface relationships.
+   */
+  globalIDs?: boolean;
   /** Configures Graphcache with Schema Introspection data.
    *
    * @remarks
--- a/exchanges/graphcache/src/types.ts
+++ b/exchanges/graphcache/src/types.ts
@@ -590,8 +590,11 @@ export type CacheExchangeOpts = {
    *
    * This is useful to overlap types of differing typenames. While this isn’t recommended
    * it can be necessary to represent more complex interface relationships.
+   *
+   * If this should only be applied to a limited set of type names, a list of
+   * type names may be passed instead.
    */
-  globalIDs?: boolean;
+  globalIDs?: string[] | boolean;
   /** Configures Graphcache with Schema Introspection data.
    *
    * @remarks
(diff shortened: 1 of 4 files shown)


exchanges/graphcache/src/types.ts
```diff
@@ -581,6 +581,17 @@ export type CacheExchangeOpts = {
    * the full keys docs.
    */
   keys?: KeyingConfig;
+  /** Enables global IDs for keying GraphQL types.
+   *
+   * @rematks
```
JoviDeCroock, line 586:
```suggestion
   * @remarks
```

kitten: Cheers! Nice catch 😅 time to put the linter back in

PR #3302, opened 2023-06-30
chore(svelte): Increase peer dependency range for Svelte 4

### LINKED ISSUE #3300: RFC: Support Svelte@4.0.0
## Summary

Update the package dependencies this can be installed with svelte 4

## Proposed Solution

I don't believe there were any breaking changes in svelte 4.0.0.  Therefore you might just be able to update the peerDependency from "^3.0.0" "^3.0.0 || ^4.0.0"

ategen3rt:
Resolves #3300

## Summary

This allows the package to be installed in projects using svelte 4

## Set of changes

Changed @urql/svelte
Added Svelte ^4.0.0 as a peer dependency (in addition to the ^3.0.0 that was already there.

### DIFF
--- a/.changeset/rich-crews-wink.md
+++ b/.changeset/rich-crews-wink.md
@@ -1,0 +1,5 @@
+---
+'@urql/svelte': patch
+---
+
+Allowed Svelte 4.0 as a peer dependency
--- a/.changeset/rich-crews-wink.md
+++ b/.changeset/rich-crews-wink.md
@@ -2,4 +2,4 @@
 '@urql/svelte': patch
 ---
 
-Allowed Svelte 4.0 as a peer dependency
+Updated peer dependency range to include support for Svelte `^4.0.0`
--- a/examples/with-svelte/package.json
+++ b/examples/with-svelte/package.json
@@ -11,10 +11,10 @@
     "@urql/core": "^4.0.10",
     "@urql/svelte": "^4.0.2",
     "graphql": "^16.6.0",
-    "svelte": "^4.0.0"
+    "svelte": "^4.0.5"
   },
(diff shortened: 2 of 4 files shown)


ptrxyz:
I am using Svelte 4 since it's release and things seem to be fine. However, there are some breaking changes in Svelte 4, but I can't tell if they are relevant for urql:  https://svelte.dev/docs/v4-migration-guide#other-breaking-changes

Mostly they are about typing and components and some minor internal. Should not affect urql.

kitten:
Hiya,
I don't think the breaking changes affect us.

However, can you please upgrade the dev dependency for Svelte here, so we're actually using the newest against the range?

We also have an example here that can be bumped and tested, so if that's updated I'll have a look before merging: https://github.com/urql-graphql/urql/tree/main/examples/with-svelte

Lastly, every PR that changes code will need a changeset, as per the contributing guide: https://github.com/urql-graphql/urql/blob/main/CONTRIBUTING.md#how-do-i-document-a-change-for-the-changelog
This is to add changelog entries and to specify what version bump is required.
In this case, a patch entry with a description along the lines of "Updated peer dependency range to include support for Svelte 4" will be enough.

Happy to take either of these over, but please let me know.

ategen3rt:
I believe I did the tasks you mentioned correctly.

frederikhors:
Hi @ategen3rt and @ptrxyz, I saw you're using Svelte too.

I can be an accurate and fast tester of `svelte@4` with `urql` but I'm stucked with https://github.com/urql-graphql/urql/discussions/3304.

Can I ask for your help? **Even just to figure out if I'm doing something wrong.**

kitten (approved):

PR #3306, opened 2023-07-05
feat(graphcache): local directives

### LINKED ISSUE #3191: RFC(graphcache): Configurable Cache Directives
## Summary

We have three categories of features and problems that have come up repeatedly and have a potentially common solution:
- When using `relayPagination`/`simplePagination`/custom resolvers that merge fields, it's impossible to tell when these resolvers should actually be _not_ used
  - if a paginated field is used for infinite pagination and regular pagination, and both happen in the same query, they conflict.
  - if a paginated field shouldn't be infinitely paginated in another query, a custom wrapper must be written around the helpers to prevent this manually
- We have considered the addition of `@required` and `@optional` directives, which “preempt” the support for nullability operators on the client-side, and are an alternative to schema awareness
  - They further become more important when we consider future scenarios where fragments may be partially loading, while the full query has only partially resolved against the cache
- We have past requests for being able to selectively apply resolvers or not: #2650

We could consider solving these problems in one go by adding custom directive support to Graphcache. With custom directives, instead of adding additional “switching/toggling” logic to Graphcache, we make potentially arbitrary resolvers configurable.

e.g. `relayPagination` could be included as an `@relayPagination` directive.

## Proposed Solution

```ts
cacheExchange({
  directives: {
    relayPagination,
    required,
    optional,
    asDate: (parent, args, cache, info) {
      return new Date(cache.resolve(parent, info.parentFieldKey));
    },
  },
});
```

Some questions that require further consdieration (**TBD**):
- Will directives simply ever be `resolvers`?
- Will they have the same shape and should there be a shortcut for `cache.resolve(parent, info.parentFieldKey)`?
- How does this affect typing and is `@required` and `@optional` feasibly implemented as a resolver?
- How do we deal with directives on fragments?

## Requirements

**TBD**

JoviDeCroock:
Resolves #3191

## Summary

This introduces a new property to the `graphcache` config named `directives`, this will function as a more selective way to apply resolvers. i.e. you can do `@relayPagination` only when you want it.

Additionally this introduces two default directives `optional` and `required` as a means to facilitate optional query selections without the spec being pushed through 😅 this bridges the gap between schemaless and schema-aware graphcache.

TODO:

- [ ] tests for `relayPagination` and `required`/`optional`
- [ ] documentation on website and TSDoc
- [ ] stripping directives from the query-string


docs/graphcache/local-directives.md
```diff
@@ -0,0 +1,4 @@
+---
+title: Local Directives
+order: 2
+---
```
JoviDeCroock, line 4: Should this be merged with local resolvers as the signature is almost identical?

exchanges/graphcache/src/cacheExchange.ts
```diff
@@ -182,7 +183,7 @@ export const cacheExchange =
         operation.kind,
         {
           key: operation.key,
-          query: formatDocument(operation.query),
+          query: formatDocument(operation.query, directivesToRemove),
```
JoviDeCroock, line 186: This won't remove the document as it's cached in `formatDocument`, would it be okay to remove the cache or bust it here by removing `operation.query.__key`? An alternative would have been to use i.e. `_` as a prefix to graphcache-directives but then we'd have to store them on the AST after formatting the document as we would lose it too early on in our chain.

exchanges/graphcache/src/types.ts
```diff
@@ -692,6 +707,11 @@ export type ResolverConfig = {
   } | void;
 };
 
+// TODO: docs
```
JoviDeCroock, line 710: The TSDocs for this are going to be highly similar to resolvers but I think that's all right as this might replace resolvers over time

exchanges/graphcache/src/operations/query.ts
```diff
@@ -356,6 +356,9 @@ const readSelection = (
   let node: FormattedNode<FieldNode> | void;
   const output = InMemoryData.makeData(input);
   while ((node = iterate()) !== undefined) {
+    const fieldDirectives = Object.keys(node._directives || {}).map(x => x);
+    const storeDirective = fieldDirectives.find(x => store.directives[x]);
```
kitten, line 360: Hm 🤔 This makes me think a bit. I'm not a huge fan of the extra `find` and `keys` work here.

The former should just be replaceable by `getDirectives(node)` which now exists to always return the object.

I'd argue, we should then `for (const name in)` loop over the `fieldDirectives` (as an object) and potentially allow multiple directives to be executed.

This would also require us to cover:
- the field directive not having an entry in `store.directives`
- multiple directives being used

I'd argue multiple directives being used simultaneously should either be a warning or it should be allowed. I'm not sure if there are cases where composition makes sense, but it's possible.

We could just not allow that for now, reuse the same warning code even and add a different message for it.

The minimum here is though I think to just discard the `fieldDirectives` / `storeDirective` split and just find the directive function immediately:

```
function getFieldDirective(node: FormattedNode<FieldNode>, store: Store) {
  for (const name in getDirectives(node)) {
    if (name === 'include' || name === 'skip') continue;
    if (store.directives[name])
      return store.directives[name];
  }
}
```

wdyt?

exchanges/graphcache/src/store/store.ts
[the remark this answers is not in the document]
JoviDeCroock: Would do it in isolation for the changelog

exchanges/graphcache/src/operations/query.ts
JoviDeCroock: Handling multiple directives is technically allowed according to the spec so I don't mind implementing it

exchanges/graphcache/src/operations/query.ts
```diff
@@ -373,7 +392,7 @@ const readSelection = (
     ctx.__internal.path.push(fieldAlias);
     // We temporarily store the data field in here, but undefined
     // means that the value is missing from the cache
-    let dataFieldValue: void | DataField;
+    let dataFieldValue: void | DataField = undefined;
```
JoviDeCroock, line 395: had to do this because TS kept thinking it wasn't being assigned

kitten:
```suggestion
    let dataFieldValue: undefined | DataField;
```

This usually does the trick. Not sure when TS changed this, could've been a long time ago or always been there, but I know that `undefined` allows unassigned undefined-defaulted variables, `void` does not

exchanges/graphcache/src/operations/query.ts
kitten: took a look at the code, I think I can't find a way to reconcile how to actually chain values, i.e. how to pipe the output of one directive into the next as an input, since that's not how resolvers work, so I reckon, we just drop it for now 🤔 

re. the former, I think there's a slight refactor that can be made to help with that, but also something that can be done after this is merged, since it's probably a good time for me to scan the file for other changes that can be made again 👀

kitten (approved):

exchanges/graphcache/src/operations/query.ts
JoviDeCroock: @kitten if you want me to refactor other things feel free to point htem out

PR #3327, opened 2023-07-23
Add solid

### LINKED ISSUE #3303: RFC: Add solid
## Summary
Add new package `solid-urql` providing [solid-js](https://www.solidjs.com/) bindings

## Proposed Solution
Implement following hooks:
- `createQuery`
- `createMutation`
- `createSubscription`

### Suspense
Solution used in [tanstack solid](https://github.com/TanStack/query/blob/main/packages/solid-query/src/createBaseQuery.ts) should be applicable for `urql`

## Note
I'm willing to provide implementation, got already `createQuery` working along with some tests

Nvos:
Resolves #3303 

## Summary

Add new `urql-solid` package implementing `solid-js` bindings.

Api is nearly same as for react/preact with difference being that some props are passed as union of value and accessor. 

## Set of changes

Adds following `solid-js` hooks
- `createQuery`
- `createMutation`
- `createSubscription`

### DIFF
--- a/packages/solid-urql/src/createQuery.ts
+++ b/packages/solid-urql/src/createQuery.ts
@@ -1,0 +1,230 @@
+import {
+  type AnyVariables,
+  type OperationContext,
+  createRequest,
+  type DocumentInput,
+  type OperationResult,
+  type RequestPolicy,
+} from '@urql/core';
+import {
+  batch,
+  createComputed,
+  createMemo,
+  createResource,
+  onCleanup,
+} from 'solid-js';
+import { createStore, unwrap } from 'solid-js/store';
+import { useClient } from './context';
+import { asAccessor } from '@solid-primitives/utils';
+import {
+  concat,
+  fromValue,
+  makeSubject,
+  map,
+  never,
+  pipe,
+  scan,
+  subscribe,
+  switchMap,
+} from 'wonka';
+import { type MaybeAccessor } from '@solid-primitives/utils';
+
+export type QueryExecuteArgs<
+  Data = any,
+  Variables extends AnyVariables = AnyVariables
+> = {
+  query: DocumentInput<Data, Variables>;
+  requestPolicy?: RequestPolicy;
+  context?: Partial<OperationContext>;
+  pause?: boolean;
+  variables?: Variables;
+};
+
+/** State of the current query, your {@link createQuery} hook is executing.
+ *
+ * @remarks
+ * `UseQueryState` is returned (in a tuple) by {@link createQuery} and
+ * gives you the updating {@link OperationResult} of GraphQL queries.
+ *
+ * Even when the query and variables passed to {@link createQuery} change,
+ * this state preserves the prior state and sets the `fetching` flag to
+ * `true`.
+ * This allows you to display the previous state, while implementing
+ * a separate loading indicator separately.
+ */
+export type CreateQueryState<
+  Data = any,
+  Variables extends AnyVariables = AnyVariables
+> = OperationResult<Data, Variables> & {
+  /** Indicates whether `createQuery` is waiting for a new result.
+   *
+   * @remarks
+   * When `createQuery` is passed a new query and/or variables, it will
+   * start executing the new query operation and `fetching` is set to
+   * `true` until a result arrives.
+   *
+   * Hint: This is subtly different than whether the query is actually
+   * fetching, and doesn’t indicate whether a query is being re-executed
+   * in the background. For this, see {@link CreateQueryState.stale}.
+   */
+  fetching: boolean;
+};
+
+export type CreateQueryArgs<
+  Data = any,
+  Variables extends AnyVariables = AnyVariables
+> = {
+  query: DocumentInput<Data, Variables>;
+  requestPolicy?: MaybeAccessor<RequestPolicy>;
+  context?: MaybeAccessor<Partial<OperationContext>>;
+  pause?: MaybeAccessor<boolean>;
+  variables?: MaybeAccessor<Variables>;
+};
+
+export const createQuery = <
+  Data = any,
+  Variables extends AnyVariables = AnyVariables
+>(
+  args: CreateQueryArgs<Data, Variables>
+) => {
+  const client = useClient();
+  const optionsSubject = makeSubject<QueryExecuteArgs<Data, Variables>>();
+  const getContext = asAccessor(args.context),
+    getPause = asAccessor(args.pause),
+    getRequestPolicy = asAccessor(args.requestPolicy),
+    getVariables = asAccessor(args.variables);
+
+  const isSuspense = createMemo(() => {
+    const ctx = getContext();
+    if (ctx !== undefined && ctx.suspense !== undefined) {
+      return ctx.suspense;
+    }
+
+    return client.suspense;
+  });
+
+  const request = createRequest(args.query, getVariables() as any);
+  const context: Partial<OperationContext> = {
+    requestPolicy: getRequestPolicy(),
+    ...getContext(),
+  };
+  const operation = client.createRequestOperation('query', request, context);
+  const initialResult: CreateQueryState<Data, Variables> = {
+    operation: operation,
+    fetching: false,
+    data: undefined,
+    error: undefined,
+    extensions: undefined,
+    hasNext: false,
+    stale: false,
+  };
+
+  const [result, setResult] =
+    createStore<CreateQueryState<Data, Variables>>(initialResult);
+
+  const [dataResource, { refetch, mutate }] = createResource<
+    CreateQueryState<Data, Variables>
+  >(() => {
+    return new Promise(resolve => {
+      const got = unwrap(result);
+      if (got.fetching) {
+        return;
+      }
+
+      resolve(got);
+    });
+  });
+
+  const sub = pipe(
+    optionsSubject.source,
+    switchMap(options$ => {
+      if (options$.pause) return never;
+
+      const request = createRequest<Data, AnyVariables>(
+        options$.query,
+        options$.variables
+      );
+
+      return concat<Partial<CreateQueryState<Data, Variables>>>([
+        fromValue({ fetching: true, stale: false }),
+        pipe(
+          client.executeQuery(request, {
+            requestPolicy: options$.requestPolicy,
+            ...options$.context,
+          }),
+          map(({ stale, data, error, extensions, operation }) => ({
+            fetching: false,
+            stale: !!stale,
+            data,
+            error,
+            operation,
+            extensions,
+          }))
+        ),
+        fromValue({ fetching: false }),
+      ]);
+    }),
+    scan(
+      (result: CreateQueryState<Data, Variables>, partial) => ({
+        ...result,
+        ...partial,
+      }),
+      initialResult
+    ),
+    subscribe(result => {
+      batch(() => {
+        setResult(result);
+        mutate(result);
+        refetch();
+      });
+    })
+  );
+
+  createComputed(() => {
+    const pause = getPause();
+    if (pause === true) {
+      return;
+    }
+
+    optionsSubject.next({
+      query: args.query,
+      context: getContext(),
+      pause: pause,
+      requestPolicy: getRequestPolicy(),
+      variables: getVariables(),
+    });
+  });
+
+  onCleanup(() => sub.unsubscribe());
+
+  batch(() => {
+    mutate(() => unwrap(result));
+    refetch();
+  });
+
+  const refetchFn = (refetchArgs: Partial<OperationContext>) => {
+    optionsSubject.next({
+      query: args.query,
+      requestPolicy: refetchArgs.requestPolicy ?? getRequestPolicy(),
+      context: refetchArgs ?? getContext(),
+      pause: false,
+      variables: getVariables(),
+    });
+  };
+
+  const handler = {
+    get(
+      target: CreateQueryState<Data, Variables>,
+      prop: keyof CreateQueryState<Data, Variables>
+    ): any {
+      if (isSuspense() && prop === 'data') {
+        return dataResource()?.data;
+      }
+
+      return Reflect.get(target, prop);
+    },
+  };
+
+  const proxy = new Proxy(result, handler);
+  return [proxy, refetchFn] as const;
+};
--- a/packages/solid-urql/src/createQuery.ts
+++ b/packages/solid-urql/src/createQuery.ts
@@ -4,29 +4,20 @@ import {
   type DocumentInput,
   type OperationResult,
   type RequestPolicy,
-  type OperationResultSource,
   createRequest,
 } from '@urql/core';
 import {
-  batch,
   createComputed,
   createMemo,
   createResource,
+  createSignal,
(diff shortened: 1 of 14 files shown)


Nvos:
Seems that CI is failing on typecheck in project root, given that solid requires different `jsx` and `jsxImportSource` than react typecheck with current configuration will fail as solid package extends base tsconfig and base tsconfig includes all packages thus runs typecheck with `jsx` set to `react` and without `jsxImportSource`.

packages/solid-urql/src/createQuery.ts
```diff
@@ -0,0 +1,334 @@
...
+  const [dataResource, { refetch, mutate }] = createResource<
+    CreateQueryState<Data, Variables>
+  >(() => {
+    return new Promise(resolve => {
+      const got = unwrap(result);
+      if (got.fetching) {
+        return;
+      }
+
+      resolve(got);
```
kitten, line 238: I don't quite feel this will be the best approach to be honest. Especially, since the inputs can change while we're waiting to resolve a promise.

I also haven't checked this in detail, but I'm very hesitant to use this approach of resolving + re-triggering this as per the `refetch` call. This seems a little brittle to me.

You can see how the `vue` bindings do this: https://github.com/urql-graphql/urql/blob/344b54410aaaf3927c70174ad8e810ccceff336d/packages/vue-urql/src/useQuery.ts#L389-L403

The only difference is that we wouldn't necessarily want to check for `stale`. That depends on how background fetches should be handled.

Maybe it does make sense though to include this, since not using suspense would be the escape hatch here. But since in these bindings there's no concept of conditionally using the resource, it's likely better not to check for `stale`

The important part here is that the promise _actually_ only reacts to the result. While the `source` isn't shared in the Vue implementation and instead relies on the semaphore logic in the `Client`, it's also fine to try to share it

packages/solid-urql/src/createQuery.ts
```diff
@@ -0,0 +1,334 @@
...
+    });
+  });
+
+  const sub = pipe(
+    resultSourceSubject.source,
```
kitten, line 243: I'm not sure whether using a subject here makes sense. It may make more sense to actually put the source in a store, similar to Vue again. This depends, but generally, I've found that using a subject here makes it actually a little harder to handle the control flow.

packages/solid-urql/src/createQuery.ts
```diff
@@ -0,0 +1,334 @@
...
+
+  batch(() => {
+    mutate(() => unwrap(result));
+    refetch();
+  });
+
+  const executeQuery: CreateQueryExecute = opts => {
+    const request = createRequest(args.query, getVariables() as any);
+    const context: Partial<OperationContext> = {
+      requestPolicy: opts?.requestPolicy ?? getRequestPolicy(),
```
kitten, line 311: There's a couple of partially stylistic, partially transpilation/browser-support related issues in this PR. Instead of listing them out, I merged a PR into `main`. So, once you rebase it'll tell you about them automatically: https://github.com/urql-graphql/urql/pull/3350

packages/solid-urql/src/createQuery.ts
```diff
@@ -0,0 +1,334 @@
...
+import {
+  batch,
+  createComputed,
+  createMemo,
+  createResource,
+  onCleanup,
+} from 'solid-js';
+import { createStore, unwrap } from 'solid-js/store';
+import { useClient } from './context';
+import { type MaybeAccessor, asAccessor } from '@solid-primitives/utils';
```
kitten, line 19: I don't really see why we'd have to depend on `@solid-primitives/utils` for a single, primitive helper, so I'd much prefer this to be re-implemented/vendored in `@urql/solid` itself, and keep this independent.

kitten:
This is only a preliminary review, but I'll get back to the details once we get closer to having a final implementation. I'm fine with fixing up the Vitest and TypeScript configs to be inline with what we'd expect in this repo, but I'll get back to that closer to having it done as a last step ✌️

packages/solid-urql/src/createQuery.ts
Nvos: Tried to implement it in similar way as in vue, technically it seems to be working (you can take a look at https://github.com/urql-graphql/urql/pull/3327/commits/b0725fea415f60a49ccf22720dbfa7004404820e). Thought after working a bit on it I'm not sure if this is good idea as it results in dependency on subscription order as subscription to fetch data has to be made before subscription in `createResource` fetcher.

Furthermore I'm not 100% sure if there wont be cases where order is different

packages/solid-urql/src/createQuery.ts
Nvos: I don't have preferences both solutions have some tradeoffs - source state  + effect + cleanup / subject + switchMap. You can take a look at source state implementation in https://github.com/urql-graphql/urql/pull/3327/commits/604c6a52bc2b4b93d5cfc784d3236ebcf6a5a13b

packages/solid-urql/src/createQuery.ts
Nvos: Fixed, didn't notice that I had invalid line endings git config on my current station...

packages/solid-urql/src/createQuery.ts
Nvos: Sure, extracted relevant code from `@solid-primitives/utils`

packages/solid-urql/src/createQuery.ts
```diff
@@ -0,0 +1,337 @@
...
+    error: undefined,
+    extensions: undefined,
+    hasNext: false,
+    stale: false,
+  };
+
+  const [result, setResult] =
+    createStore<CreateQueryState<Data, Variables>>(initialResult);
+
+  createComputed(() => {
```
kitten, line 214: The implementation looks obviously quite similar between [`createComputed`](https://github.com/solidjs/solid/blob/f79ba4d9087d74ef7e55fd0dc336c98ce2ec8526/packages/solid/src/reactive/signal.ts#L282-L290) and [`createRenderEffect`](https://github.com/solidjs/solid/blob/f79ba4d9087d74ef7e55fd0dc336c98ce2ec8526/packages/solid/src/reactive/signal.ts#L313-L321).

The difference being a small `pure` flag that's being set.

However, I'm curious what the difference is because I find it a little odd to perform effects in what's essentially a utility that's supposed to be used to derive new signals

packages/solid-urql/src/createQuery.ts
```diff
@@ -0,0 +1,337 @@
...
+  });
+
+  const [dataResource, { refetch }] = createResource<
+    CreateQueryState<Data, Variables>,
+    Source<OperationResult<Data, Variables>> | undefined
+  >(source, source => {
+    let sub: Subscription | void;
+    if (source === undefined) {
+      return Promise.resolve(result);
+    }
```
kitten, line 283: I'm not sure whether this comment you left about ordering is accurate as long as the computed effect runs before the resource, which should be the case: https://github.com/urql-graphql/urql/pull/3327#discussion_r1292479692

However, `@urql/vue` does get around this with effect ordering

(See other comment re. effect start)

packages/solid-urql/src/createSubscription.ts
```diff
@@ -0,0 +1,318 @@
...
+ * const Notifications = () => {
+ *   const [result, executeSubscription] = createSubscription(
+ *     { query: NotificationsSubscription },
+ *     combineNotifications,
+ *   );
+ *   // ...
+ * };
+ * ```
+ */
+export const createSubscription = <
```
kitten, line 212: Obvious TODO (mostly for my own benefit); This will have to match the `createQuery` implementation closely before merging

packages/solid-urql/src/createQuery.ts
```diff
@@ -0,0 +1,337 @@
...
+      pipe(
+        s,
+        onEnd(() => {
+          setResult(
+            produce(draft => {
+              draft.fetching = false;
+              draft.stale = false;
+            })
+          );
+        }),
```
kitten, line 259: Small question: Are we sure this triggers _before_ the new effect is set up, so we don't get `fetching: false` after the new effect started?

Just double checking that cleanups are called before the next source is set up.

packages/solid-urql/src/createQuery.ts
```diff
@@ -0,0 +1,337 @@
...
+    const request = createRequest(args.query, getVariables() as any);
+    const context: Partial<OperationContext> = {
+      requestPolicy: getRequestPolicy(),
+      ...getContext(),
+    };
+
+    setSource(() => client.executeQuery(request, context));
+  });
+
+  createComputed(() => {
```
kitten, line 229: As per the other comment, I can see why you're concerned that this may effect the order of effects. Essentially, `createComputed` here obviously also causes an immediate effect, which happens to be called after the one that calls `setSource`.

This isn't really a concern in `@urql/vue` due to the same reason it isn't here: call order. However, if you're concerned about this, the tests should keep this intact anyway, and in theory, we could even try to make the code bind this guarantee more tightly by making a small utility that combines the memo-like computation of `source` with `createSignal`, but I think that's unnecessary

packages/solid-urql/src/createQuery.ts
Nvos: Didn't delve deep into solid-js code base, but from my understanding
- `createComputed` - called immediataly on upadate, similar to `createMemo`
- `createRenderEffect` - behavior similar to `createEffect` difference is first execution primarly to be able to access refs during initial render

packages/solid-urql/src/createQuery.ts
Nvos: From what I see it is always called after dependencies (only dependency is `source` signal) change

packages/solid-urql/src/createQuery.ts
Nvos: After some testing I think it should be fine, `createResource` internally uses `createComputed` to listen for source changes thus order should be correct

packages/solid-urql/src/createSubscription.ts
Nvos: Changed `createSubscription` to store source in state

packages/solid-urql/src/createQuery.ts
Nvos: Order should be fine after some testing I don't think there should be any edge cases

Nvos:
Is there anything else from my side to do to push this PR forward?

kitten:
@Nvos: Hiya 👋 Sorry for the delay. Mainly, what I got stuck on here is the setup. While I haven't normalised, rebased and checked everything for discrepancies yet, I started with aligning some of the setup changes.

Firstly, the `tsconfig` changes are pretty trivial and we can even avoid JSX to not have to deal with it that often. React is a little special in that regards and Typescript doesn't have many great options to isolate builds without becoming the “sole/primary build tool”

Anyway, from a Vitest standpoint, the config for it for Solid should obviously not be a snowflake and unique. Having it all run as one thing (besides some of the E2E tests which are mostly to prevent common regressions instead; added _as an experiment_) would be neat.

That said, the tests are failing now and I don't have too much time right now to debug this myself (Sorry!)
One thing I discovered is that there's several `waitFor` checks that aren't await-ed properly. I didn't get rid of these in any test besides `createQuery`'s.

When I'm looking at the tests triggering effects, the primary thing I'm noticing is that Solid's test framework lacks an `act` function or another mechanism of flushing effects. Even waiting for a random timeout or micro-tick doesn't seem to really convince it to consistently execute effects in a predictable manner that I'd expect from tests. It's a little puzzling.

On a side-note, we're missing a small change that tests that the `createSubscription` scan function actually executes. But that's a small detail.
Overall, I haven't tried this in a Vite sandbox, and set up an example yet, but from the tests, my confidence right now in them isn't particularly high until I can understand how Solid expects effects to flush in tests 😅

Nvos:
I aggree that having 1 vitest config is good, thought thought I still would like to have solid plugin in vitest instead of using `solid-js/h` maybe it would be good idea to merge main config if some changes are required:
```ts
// other imports...
import rootConfig from '../../vitest.config';

export default mergeConfig(
  rootConfig,
  defineProject({
    plugins: [solidPlugin()],
    test: {
      globals: true,
    },
  })
);
```
Though plugin is required only to handle JSX, we could remove those tests (those are only required to test suspense) and instead introduce some E2E tests via playwright/cypress. I could write playwright tests, as for cypres would require to do some research as didn't use it for long time now.

I'm not that proficient in writing unit tests in solid. Tests in this PR were based on my initial research. Though I noticed that there's test helper `testEffect` which might be what we need. I'll take a look at what we can do in regard to handling effects in tests this week.

In regard to:

> That said, the tests are failing now and I don't have too much time right now to debug this myself (Sorry!)

Do you mean CI or tests were failing when running locally?

Nvos:
Updated all tests to use `testEffect`, it should be more reliable now.

Tried few things and as for now I'm not sure how to handle running tests from repository root without larger changes as those do not respect configurations in nested packages. 

Tried to change `solid/h` as you did on your branch but for some reason reactivity was not working inside components. Will take a look at this today as it seems to be only way to use `vitest.config` from repository root

Nvos:
After some research I come to conclusion that with current setup it seems unlikely to run solid tests from project root via `test` script due to `vite-plugin-solid` seemingly being necessary to run hook tests (seems that `createResource` reads `sharedConfig.context.id` and context is never set without plugin). I have tried to handle this in 2 ways:
- vitest workspaces - I'm not exactly sure why but it seems that dependencies are not resolved correctly when running solid tests inside vitest workspace. Will try to research it bit more
- using `vite-plugin-solid` in main `vitest.config.ts` using plugin `includes` which doesn't work as well due to conflicting configurations which break `preact` and `react` tests: `resolve.conditions` and `esbuild`

From my perspective best way to handle such cases globally would be to switch from running tests via single config with includes to respective configs in sub-packages which could be ran by `pnpm` or preferably `turbo`. There could be single base config which could be used as default if there's no changes needed.

Another way could be handling `solid-urql` separately by excluding it from current config and running it along side current tests but this makes `solid-urql` special case which might not be preferable

stefanmaric:
Fixing the TypeScript errors was fairly easy: https://github.com/urql-graphql/urql/pull/3607/commits/398267e70e42dced7867627795b916eb6c4bbe98

On the other hand, I spent quite some time trying to get tests running with Vitest's Workspaces, but to no avail. I tried different combinations of inline vs. standalone project configs, and not even with the [v2 beta which forks by default](https://github.com/vitest-dev/vitest/discussions/5210), was I able to get it to work.

This thread might be relevant for this issue: https://github.com/vitest-dev/vitest/issues/5301.

I rebased this branch and got it to kinda work: https://github.com/urql-graphql/urql/actions/runs/9435968855/job/25990068426.

Not without some downsides: in CI, it will run tests for each package individually (but in parallel), while locally `pnpm run test` will run the tests from the top level in watch mode as it has been doing so far, but it will skip solid-urql's tests.

So, if your changes break the Solid integration, you will only notice after pushing. Not ideal, but it might be worth it.

Solid + urql is a magic combo I would love to have. ❤️

JoviDeCroock:
Superseded by https://github.com/urql-graphql/urql/pull/3607

PR #3364, opened 2023-08-30
docs(next): Enable suspense in non-rsc example

spiceee:
## Summary

The non-rsc example is missing the `suspense: true` argument, so the query never runs on the server like the copy suggests.

## Set of changes

Just enabling suspense in the docs, like it's done in the Next non-rsc example in https://github.com/urql-graphql/urql/blob/main/examples/with-next/app/non-rsc/layout.tsx


kitten (approved):
nice catch!

PR #3378, opened 2023-09-12
fix(core): allow controll by signal from fetchOptions

### LINKED ISSUE #3377: Signal from fetchOptions not passed to fetch
### Describe the bug

```js
import { createClient, cacheExchange, fetchExchange } from "@urql/core"
export const client = createClient({
	url: "https://api/graphql",
	exchanges: [cacheExchange, fetchExchange],
	requestPolicy: "cache-and-network",
	fetch,
})
const { signal } = new AbortController
function fetch(input, init) {
    console.assert(init.signal === signal)
}
await client.query(gql`...`, {}, { fetchOptions: {signal} })
```

Expect the signal from `fetchOptions` should be passed to fetch and user can control/abort the query.
Actually the `signal` has been override from `fetchSource.ts`
```ts
export function makeFetchSource(
  operation: Operation,
  url: string,
  fetchOptions: RequestInit
): Source<OperationResult> {
  let abortController: AbortController | void;
  if (typeof AbortController !== 'undefined') {
    fetchOptions.signal = (abortController = new AbortController()).signal;
  }
  return pipe(
    fromAsyncIterable(fetchOperation(operation, url, fetchOptions)),
    filter((result): result is OperationResult => !!result),
    onEnd(() => {
      if (abortController) abortController.abort();
    })
  );
}
```
Suggest to also abort if the signal passed from user aborted.
```diff
export function makeFetchSource(
  operation: Operation,
  url: string,
  fetchOptions: RequestInit
): Source<OperationResult> {
  let abortController: AbortController | void;
  if (typeof AbortController !== 'undefined') {
+    if (fetchOptions.signal) fetchOptions.signal.addEventListener('abort', () => abortController?.abort());
    fetchOptions.signal = (abortController = new AbortController()).signal;
  }
  return pipe(
    fromAsyncIterable(fetchOperation(operation, url, fetchOptions)),
    filter((result): result is OperationResult => !!result),
    onEnd(() => {
      if (abortController) abortController.abort();
    })
  );
}
```

### Reproduction

https://stackblitz.com/edit/github-b9nibt?file=src%2Fclient.js

### Urql version

urql v4.1.2

### Validations

rexpan:
Resolves #3377

## Summary

The `signal` passed from `fetchOptions` by user be ignored and overrided. 

## Set of changes

Listen to `abort` event from user passed `signal` and abort the current request.

### DIFF
--- a/packages/core/src/internal/fetchSource.ts
+++ b/packages/core/src/internal/fetchSource.ts
@@ -237,14 +237,17 @@ export function makeFetchSource(
   fetchOptions: RequestInit
 ): Source<OperationResult> {
   let abortController: AbortController | void;
+  function abort() {
+    if (abortController) abortController.abort();
+  }
   if (typeof AbortController !== 'undefined') {
+    if (fetchOptions.signal)
+      fetchOptions.signal.addEventListener('abort', abort, { once: true });
     fetchOptions.signal = (abortController = new AbortController()).signal;
   }
   return pipe(
     fromAsyncIterable(fetchOperation(operation, url, fetchOptions)),
     filter((result): result is OperationResult => !!result),
-    onEnd(() => {
-      if (abortController) abortController.abort();
-    })
+    onEnd(abort)
   );
 }


tatianajiselle:
I can confirm that I am experiencing the bug myself, would appreciate this being resolved.

tatianajiselle:
I can confirm that I am experiencing the bug myself, would appreciate this being resolved.

kitten:
@tatianajiselle I left a comment on #3377 trying to answer all three streams of this discussion, but basically, I don't consider this a bug or missing feature right now, and instead, would recommend to write a `fetch` wrapper and pass that to the `Client`. There's a further explanation in the issue's comment, but I wouldn't consider using the `AbortSignal` that's passed in as more useful or more correct, given that it isn't more practical than either a custom exchange or a `fetch` wrapper, and also isn't a practical public API

tatianajiselle:
all good! moving convo to #3377 for simplicity

kitten:
As per the conversation in #3377, since `signal` is not part of the public API (1) because HTTP is not assumed and hence `fetchOptions` is limited, and (2) since it's not expected to set up a timeout or otherwise pass an `AbortSignal` in the UI layer to the `Client`, what this change adds is not expected to work.

PR #3389, opened 2023-09-26
feat(core): Update defer/stream protocol as per spec

JoviDeCroock:
## Summary

In the [new June 2023 spec](https://github.com/graphql/defer-stream-wg/discussions/69) we get the introduction of `pending` on both incremental and execution results, pending contains an array of id and path.

The id will be used in incremental payloads, we can then do a lookup in `pending` and find a path, the path indicates where the fragment got spread starting at the root.

Another addition here is the `subPath` which can be present on an incremental payload and builds on the path we can retrieve through `id`, when deferred selection-sets are mixed with the non-deferred one this subPath will point at where we have to start merging.

### DIFF
--- a/.changeset/fast-jeans-pay.md
+++ b/.changeset/fast-jeans-pay.md
@@ -1,5 +1,5 @@
 ---
-"@urql/core": minor
+"@urql/core": patch
 ---
 
-implement new defer/stream spec
+Implement new `@defer` / `@stream` transport protocol spec changes.
--- a/.changeset/fast-jeans-pay.md
+++ b/.changeset/fast-jeans-pay.md
@@ -1,0 +1,5 @@
+---
+"@urql/core": minor
+---
+
+implement new defer/stream spec
--- a/packages/core/src/internal/fetchSource.ts
+++ b/packages/core/src/internal/fetchSource.ts
@@ -175,9 +175,13 @@ async function* fetchOperation(
       throw new Error(await response.text());
     }
 
+    let pending: ExecutionResult['pending'];
     for await (const payload of results) {
+      if (payload.pending) {
+        pending = payload.pending;
+      }
       result = result
-        ? mergeResultPatch(result, payload, response)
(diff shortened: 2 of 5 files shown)


packages/core/src/utils/result.test.ts
```diff
@@ -38,7 +38,127 @@ describe('makeResult', () => {
   });
 });
 
-describe('mergeResultPatch', () => {
+describe.only('mergeResultPatch (defer/stream latest', () => {
```
JoviDeCroock, line 41:
```suggestion
describe('mergeResultPatch (defer/stream latest', () => {
```

packages/core/src/internal/fetchSource.ts
```diff
@@ -175,9 +175,15 @@ async function* fetchOperation(
       throw new Error(await response.text());
     }
 
+    let pending: ExecutionResult['pending'];
     for await (const payload of results) {
+      if (payload.pending && !result) {
+        pending = payload.pending;
+      } else if (payload.pending) {
+        pending = [...pending!, ...payload.pending];
+      }
```
JoviDeCroock, line 184: Initially pending will be present on the ExecutionResult, however when we have nested incremental results then the IncrementalResult will contain new pending information which we have to add to our array here so we can keep building on our result

.changeset/fast-jeans-pay.md
```diff
@@ -0,0 +1,5 @@
+---
+"@urql/core": minor
+---
+
+implement new defer/stream spec
```
kitten, line 5:
```suggestion
---
"@urql/core": patch
---

Implement new `@defer` / `@stream` transport protocol spec changes.
```

kitten: I'd love to designate this as a patch, if you don't mind, since it's not stable yet and the minor version update doesn't do much as long as there's no corresponding minor change in the GraphQL package 😅

kitten (approved):

PR #3403, opened 2023-10-16
Fix `preferGetMethod` client option

Burbenog:
Urql client was ignore 'force' and 'within-url-limit' options for `preferGetMethod` by simply casting it to boolean.

Now `preferGetMethod` works as expected


kitten (approved):
Hiya 👋 Nice catch! Not sure how we missed this.
On a side-note, please don't forget to read [Contributing guides](https://github.com/urql-graphql/urql/blob/main/CONTRIBUTING.md) before opening PRs. Just leaving a note since most other repos will have similar ones. In our case it contains some notes on changesets for changelog entries.
Either way, I simply pushed one quickly here :v:

PR #3428, opened 2023-11-10
feat(graphcache): add onCacheHydrated event

JoviDeCroock:
Resolves https://github.com/urql-graphql/urql/issues/3422

## Summary

We used to buffer operations while the cache hydrates from storage, this got disabled in our latest release however we can enable folks to still have this. We now support the `cacheExchange` calling `storage.onCacheHydrated` this way folks can load their shell but leave the content in a loading state while the storage hydrates the data.


elilabes:
This is awesome! Thank you 🙏

PR #3430, opened 2023-11-12
chore(core): try to parse a text/plain content-type response as json

JoviDeCroock:
## Summary

As seen in https://github.com/urql-graphql/urql/discussions/3429, some servers send back the response as `text/plain` which can still be valid JSON.

### DIFF
--- a/.changeset/silver-planets-sniff.md
+++ b/.changeset/silver-planets-sniff.md
@@ -2,4 +2,4 @@
 '@urql/core': minor
 ---
 
-Try to parse no content-type as JSON because we are dealing with GraphQL
+Try to parse `text/plain` content-type as JSON before bailing out with an error.
--- a/.changeset/silver-planets-sniff.md
+++ b/.changeset/silver-planets-sniff.md
@@ -1,0 +1,5 @@
+---
+'@urql/core': minor
+---
+
+Try to parse no content-type as JSON because we are dealing with GraphQL
--- a/packages/core/src/internal/fetchSource.ts
+++ b/packages/core/src/internal/fetchSource.ts
@@ -169,7 +169,7 @@ async function* fetchOperation(
       results = parseMultipartMixed(contentType, response);
     } else if (/text\/event-stream/i.test(contentType)) {
       results = parseEventStream(response);
-    } else if (!/text\//i.test(contentType)) {
+    } else if (!/text\//i.test(contentType) || !contentType) {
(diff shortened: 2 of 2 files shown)


.changeset/silver-planets-sniff.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/core': minor
+---
+
+Try to parse no content-type as JSON because we are dealing with GraphQL
```
JoviDeCroock, line 5:
```suggestion
Try to parse `text/plain` content-type as JSON before bailing out with an error.
```

packages/core/src/internal/fetchSource.ts
```diff
@@ -169,10 +169,28 @@ async function* fetchOperation(
       results = parseMultipartMixed(contentType, response);
     } else if (/text\/event-stream/i.test(contentType)) {
       results = parseEventStream(response);
-    } else if (!/text\//i.test(contentType)) {
+    } else if (!/text\//i.test(contentType) || !contentType) {
```
JoviDeCroock, line 172:
```suggestion
    } else if (!/text\//i.test(contentType)) {
```

packages/core/src/internal/fetchSource.ts
```diff
@@ -172,7 +172,25 @@ async function* fetchOperation(
     } else if (!/text\//i.test(contentType)) {
       results = parseJSON(response);
     } else {
-      throw new Error(await response.text());
+      if (contentType === 'text/plain') {
+        const text = await response.text();
+        if (text.startsWith('{')) {
+          try {
+            results = JSON.parse(text);
```
JoviDeCroock, line 179: My generator foo is lacking here, does this count, because `yield JSON.parse()` seems invalid

packages/core/src/internal/fetchSource.ts
```diff
@@ -172,7 +172,17 @@ async function* fetchOperation(
     } else if (!/text\//i.test(contentType)) {
       results = parseJSON(response);
     } else {
-      throw new Error(await response.text());
+      const text = await response.text();
+      try {
+        results = JSON.parse(text);
```
kitten, line 177: Small issue here, `results` is expected to be an async iterator (for the for-await loop below). It's likely not warning here since the result of the parse is `any` and it may still appear to work, but that likely will run into a runtime error 🤔

kitten (approved):

PR #3434, opened 2023-11-16
Make useQuery refetch async

davidfant:
## Summary

In my React app I want to be able to `await` the refetch function, for example to call another function after refetch is done

## Set of changes

Mainly the React package's `useQuery.ts`. Let me know if there's a better way to promisify this using wonka

### DIFF
--- a/packages/react-urql/src/hooks/useQuery.spec.ts
+++ b/packages/react-urql/src/hooks/useQuery.spec.ts
@@ -267,7 +267,9 @@ describe('useQuery', () => {
     expect(client.executeQuery).toBeCalledTimes(1);
 
     const [, executeQuery] = result.current;
-    act(() => executeQuery());
+    act(() => {
+      executeQuery();
+    });
     await waitForNextUpdate();
     expect(client.executeQuery).toBeCalledTimes(2);
   });
--- a/packages/react-urql/src/hooks/useQuery.ts
+++ b/packages/react-urql/src/hooks/useQuery.ts
@@ -158,7 +158,9 @@ export interface UseQueryState<
  * };
  * ```
  */
-export type UseQueryExecute = (opts?: Partial<OperationContext>) => Promise<unknown>;
+export type UseQueryExecute = (
+  opts?: Partial<OperationContext>
+) => Promise<unknown>;
 
 /** Result tuple returned by the {@link useQuery} hook.
  *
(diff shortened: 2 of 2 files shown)


packages/react-urql/src/hooks/useQuery.spec.ts
```diff
@@ -267,7 +267,9 @@ describe('useQuery', () => {
     expect(client.executeQuery).toBeCalledTimes(1);
 
     const [, executeQuery] = result.current;
-    act(() => executeQuery());
+    act(() => {
```
davidfant, line 270: As this returns a promise now, tsc complained

packages/react-urql/src/hooks/useQuery.ts
```diff
@@ -355,25 +357,30 @@ export function useQuery<
...
-              onPush(result => {
-                cache.set(request.key, result);
-              })
-            )
-          : client.executeQuery(request, context);
-        return [source, state[1], deps];
-      });
-    },
+        deferDispatch(setState, state => {
+          const source = suspense
```
davidfant, line 369: I also did `source.then(resolve)` below everything (which works fine when running locally), but Typescript complains

packages/react-urql/src/hooks/useQuery.ts
```diff
@@ -355,25 +357,30 @@ export function useQuery<
...
+            ? pipe(
+                client.executeQuery(request, context),
+                onPush(result => {
+                  cache.set(request.key, result);
+                  resolve();
+                })
+              )
+            : pipe(client.executeQuery(request, context), result => {
+                resolve();
```
JoviDeCroock, line 378: in a `cache-and-network` call this would most likely return the stale equivalent. We do already have a streaming util to convert this to a Promise conditionally in [https://github.com/urql-graphql/urql/blob/main/packages/core/src/utils/streamUtils.ts#L15].

So what we could do is create the `source`, do a `deferDispatch` to set the state to React and return the `source`

packages/react-urql/src/hooks/useQuery.ts
```diff
@@ -158,7 +158,9 @@ export interface UseQueryState<
  * };
  * ```
  */
-export type UseQueryExecute = (opts?: Partial<OperationContext>) => void;
+export type UseQueryExecute = (
+  opts?: Partial<OperationContext>
+) => Promise<unknown>;
```
JoviDeCroock, line 163: I am wondering whether this could turn into a breaking change, i.e. something complaining about not awaiting/... when doing `onClick={() => executeQuery()}`

kitten:
Closing due to inactiviy.

There's also past threads here where we rejected similar proposals and similar changes, since it can lead both to state issues that are hard to test and maintain, and also isn't a desirable pattern in our opinion. The canonical way is to maintain state on inputs to `useQuery`, or to use the client's methods directly. Making an in-between solution has lead to undesirable usage bugs and usage patterns in other libraries, from what we could tell when this decision was made

PR #3444, opened 2023-12-02
feat(graphcache): add logger interface

### LINKED ISSUE #3404: RFC: Option to hide warnings when in development
## Summary

Add a way to hide warnings, even in development. This is helpful when you are using urql in a library and the end library users are subjected to urql warnings that they can't do anything about while in development. For example, `Heuristic Fragment Matching` or `Invalid undefined` warnings.

## Proposed Solution

Have an option hide warnings or a log-level option in the client initializer

## Requirements

- warnings can be optionally hidden while in development

JoviDeCroock:
Resolves #3404

## Summary

This adds an option to the `graphCache` constructor named `Logger` we will use the logger to convey information that is encountered during development. In doing so a user can choose to remove the warnings by stubbing the logger with `() => {}` or adding other custom logic to filter things out.

## Set of changes

- add `Logger` interface for graph-cache

### DIFF
--- a/.changeset/tender-files-tie.md
+++ b/.changeset/tender-files-tie.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-graphcache': minor
+---
+
+Add optional `logger` to the options, this allows you to filter out warnings or disable them all together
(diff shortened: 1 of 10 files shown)


kitten (approved):

johnpmitsch:
Thank you for this! 🙏🏻

PR #3446, opened 2023-12-02
feat(graphcache): debug logs for cache misses

### LINKED ISSUE #3382: RFC: Operation context cache miss information
## Summary

Having an additional response `operation.context.meta.cacheOutcome` field to know what field(s) caused the cache miss.

I've spent so many hours trying to debug a local cache resolver to find why a field would return `null`(with partial miss), or a request not return anything at all(without partial miss).

There's no error or warning messages if a cache field doesn't return a value or a correct value beside the "should have returned a scalar type" warning.

On simple applications this doesn't really happen, if if your application complexity increase a bit and your resolvers query other resolvers, this start to happen pretty easily.

Having a better workflow to work with the cache in general would be great.

## Proposed Solution

Add a new field in the context meta `cacheOutcome` object with the fields paths that caused the cache miss.
Additionally it could also be useful to get the returned value of those fields.

JoviDeCroock:
Resolves #3382

## Summary

This will log when we are missing a field on a selection, this can be useful for debugging cache misses in development. This is opt-in by defining the new `logger` interface and listening for `debug` logs.

### DIFF
--- a/.changeset/mean-onions-share.md
+++ b/.changeset/mean-onions-share.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-graphcache': patch
+---
+
+Debugging cache misses
--- a/exchanges/graphcache/src/cacheExchange.test.ts
+++ b/exchanges/graphcache/src/cacheExchange.test.ts
@@ -110,6 +110,82 @@ describe('data dependencies', () => {
     expect(result.mock.calls[1][0].data).toBe(result.mock.calls[0][0].data);
   });
 
+  it('logs cache misses', () => {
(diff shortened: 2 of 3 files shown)


.changeset/mean-onions-share.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/exchange-graphcache': patch
```
JoviDeCroock, line 2:
```suggestion
'@urql/exchange-graphcache': minor
```

kitten (approved):

Issue #3448, opened 2023-12-06
Saving a component in Next.js development mode casues infinite requests


lswinblad:
### Describe the bug

I'm not sure if I'm implementing this correctly, but saving a component when using the Next.js development mode causes queries to be made in an infinite loop.

It seems to be caused by having different URLs for server / client side. The only reference I could find to having different URLs depending on the context is in the [old repository](https://github.com/FormidableLabs/next-urql#different-client-configurations-on-the-client-and-the-server).

To reproduce

1. `npm run dev`
2. Open devtools in e.g. chrome
3. Save a file after e.g. inserting a newline somewhere
4. Observe infinite requests being made

### Reproduction

https://stackblitz.com/edit/github-unakrx?file=app%2Furql-provider.tsx

### Urql version

See package.json in reproduction

### Validations

- [X] I can confirm that this is a bug report, and not a feature request, RFC, question, or discussion, for which [GitHub Discussions](https://github.com/urql-graphql/urql/discussions) should be used
- [X] Read the [docs](https://formidable.com/open-source/urql/docs/).
- [X] Follow our [Code of Conduct](https://github.com/urql-graphql/urql/blob/main/CODE_OF_CONDUCT.md)

Issue #3408, opened 2023-10-20
Infinite initialization? (with Graphql-codegen and Next.js 13)


shkim:
### Describe the bug

When I use the Provider for urql client, it seems that it is infinitely called when any query is activated.

I prepared the reproducible example code: https://github.com/shkim/urql-next13-error
Remarkable files are:

src/app/GqlProvider.tsx 
src/app/layout.tsx 
src/app/query/page.tsx 
src/graphql/index.ts
src/graphql/documents.tsx

src/graphql/generated.tsx file is generated with graphql-codegen
it can be regenerated by "graphql-codegen --config gql-config.yml"

Thank you.

### Reproduction

https://github.com/shkim/urql-next13-error

### Urql version

@urql/next 1.1.0 (latest at the moment)

### Validations

- [X] I can confirm that this is a bug report, and not a feature request, RFC, question, or discussion, for which [GitHub Discussions](https://github.com/urql-graphql/urql/discussions) should be used
- [X] Read the [docs](https://formidable.com/open-source/urql/docs/).
- [X] Follow our [Code of Conduct](https://github.com/urql-graphql/urql/blob/main/CODE_OF_CONDUCT.md)

JoviDeCroock:
That's because [this line](https://github.com/shkim/urql-next13-error/blob/main/src/graphql/generated.ts#L2) should be importing from `@urql/next`

michaelzadra1:
I'm experiencing the same issue. I tried the above solution but it didn't resolve the problem. 

Strangely enough, what resolves the infinite fetch problem for me is removing `export const runtime = "edge";` from my Page components. There seems to be some issue with serving pages from edge fns w/ urql

PR #3449, opened 2023-12-06
chore(next): warn for missing suspense boundary

JoviDeCroock:
Related to #3448 and #3408

## Summary

We have seen a few issues where folks forget their Suspense boundary which leads to infinite requests. We will shield from infinite requests by providing our own underneath the `urql` provider so the client doesn't get re-created and give the user a warning when we are not in production.

### DIFF
--- a/packages/next-urql/src/Provider.ts
+++ b/packages/next-urql/src/Provider.ts
@@ -9,6 +9,15 @@ export const SSRContext = React.createContext<SSRExchange | undefined>(
   undefined
 );
 
+const SuspenseWarning = () => {
+  if (process.env.NODE_ENV !== 'production') {
+    console.warn(
+      'urql suspended but there was no boundary to catch it, add a <Suspense> component under your urql Provider.'
+    );
+  }
+  return null;
+};
+
 /** Provider for `@urql/next` during non-rsc interactions.
  *
  * @remarks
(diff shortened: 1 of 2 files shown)


packages/next-urql/src/Provider.ts
```diff
@@ -59,7 +68,15 @@ export function UrqlProvider({
     React.createElement(
       SSRContext.Provider,
       { value: ssr },
-      React.createElement(DataHydrationContextProvider, { nonce }, children)
+      React.createElement(
+        DataHydrationContextProvider,
+        { nonce },
+        React.createElement(
+          React.Suspense,
```
kitten, line 75: Hm, my only concern here is that there could be legitimate uses for this. I'm imagining that some router systems may use suspense in unexpected ways, so not sure if we can somehow make detection work without a custom boundary?

kitten (approved):

packages/next-urql/src/Provider.ts
SoraKumo001: If this Suspense is inserted on the server side, the Promise resolution in the React component will change to streaming instead of waiting.

I'm sorry, but SSR in urql without Suspense will be impossible.

https://next-blog.croud.jp/contents/7m4dShGLIpFkc0GKHIrW

Can you please adjust the server side so that Suspense is not interrupted?

PR #3454, opened 2023-12-09
fix(next): only inject `<Suspense` and warning during client-side rendering

JoviDeCroock:
## Summary

In response to a comment on https://github.com/urql-graphql/urql/pull/3449#discussion_r1418516774 this opts you into streaming automatically 😅 which is probably not what we want by default. This moves the defaulting to only in the browser

CC @SoraKumo001

### DIFF
--- a/.changeset/funny-hornets-heal.md
+++ b/.changeset/funny-hornets-heal.md
@@ -1,0 +1,5 @@
+---
+'@urql/next': patch
+---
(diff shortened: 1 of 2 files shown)


SoraKumo001:
@JoviDeCroock 
Thank you for your response.

kitten (approved):

PR #3459, opened 2023-12-12
fix(core): return a result in cache-only scenarios

### LINKED ISSUE #3457: Request Policy set to cache-only throws error during build of next.js app with document cache
### Describe the bug

First of all thanks a lot for your work on this project!

The [docs](https://formidable.com/open-source/urql/docs/basics/document-caching/#request-policies) for request policies on document cache mention that cache-only should return the cached data or null.

When using useQuery in a next.js app and setting the request policy to cache-only everything looks as expected in dev mode. However when trying to create a production build with `npm run build` I get this error: `TypeError: Cannot read properties of undefined (reading 'operation')`.

The expected behaviour would be no build errors and the same result as in dev mode.

Maybe for context why this is an issue for my use case: I tried to leverage the ssr prefetching of queries and want to only display data that has been prefetched server side. This worked fine with the graphCache, but I recently had to move back to document caching. So I'm not sure with which version this bug got introduced.

### Reproduction

https://github.com/Thoud/urql-cache-only-issue-demo

### Urql version

4.0.6

### Validations

JoviDeCroock:
Resolves #3457

## Summary

When we are dealing with a `cache-only` request-policy we should return a result on cache-miss, if we don't do this defaulting we end up returning `undefined` which crashes the client entirely


packages/core/src/exchanges/cache.ts
```diff
@@ -79,7 +80,14 @@ export const cacheExchange: Exchange = ({ forward, client, dispatchDebug }) => {
...
-        let result: OperationResult = cachedResult!;
+        let result: OperationResult;
+        if (!cachedResult) {
+          result = makeResult(operation, {
+            data: null,
+          });
+        } else {
+          result = cachedResult;
+        }
         if (process.env.NODE_ENV !== 'production') {
```
JoviDeCroock, line 91: This line here makes it work in `dev` where it will create a result either way

.changeset/three-pandas-invent.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/core': patch
+---
+
+When we are dealing with a `cache-only` request-policy we should return a result on cache-miss, if we don't do this defaulting we end up returning `undefined` which crashes the client entirely
```
kitten, line 5: just shortening this up to start with `Fix`

```suggestion
Fix the default `cacheExchange` crashing on `cache-only` request policies with cache misses due to `undefined` results.
```

kitten (approved):

PR #3464, opened 2023-12-13
fix(core): add back cacheOutcome

### LINKED ISSUE #3462: request-policy-exchange behavior differs between dev & production
### Describe the bug

The exchange has this basic logic:
```
const processIncomingResults = (result: OperationResult): void => {
  const meta = result.operation.context.meta;
  const isMiss = !meta || meta.cacheOutcome === 'miss';
  if (isMiss) {
    operations.set(result.operation.key, new Date().getTime());
  }
};
```
`meta` does not exist in production (as mentioned https://github.com/urql-graphql/urql/discussions/3439) resulting in this always being viewed as a miss. This will keep a request alive much longer than anticipated and won't be the same behavior between dev and production.

### Reproduction

Is this needed here?

### Urql version

4.0.6, core 4.2.2

### Validations

JoviDeCroock:
Resolves #3462 

## Summary

In the document cache we put the `cache-outcome` behind a dev-flag, which is inconsistent as we rely on it in other exchanges and it's not behind a flag in our normalized and ssr cache.

### DIFF
--- a/.changeset/silly-pigs-return.md
+++ b/.changeset/silly-pigs-return.md
@@ -1,0 +1,5 @@
+---
+'@urql/core': patch
+---
+
+Add back our cache-outcome on the document-cache, this was behind a development flag however in our normalized cache we always add it already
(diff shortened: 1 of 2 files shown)


kitten:
Just noting down a TODO here for myself since I forgot (@kitten). This should be reimplemented without relying on internal flags being set by other exchanges

PR #3493, opened 2024-02-01
fix(graphcache): Restore `stale` being set to `true` on blocked (but in-flight) operations

kitten:
> **Note:** I wasn't reasonably able to write tests for the looping protection yet that hit this particular case. Manual testing is needed!

## Summary

For partial cache results, `stale: false` was being set when the looping protection kicks in. In simple words, if an operation is triggered twice, once as part of an unrelated query updating, causing a partial result, then again, the second trigger will create a cache result with `stale: false`, even if the operation would be in flight.

This PR restores `stale: true` being set, if we reasonably believe the operation is still in flight.

## Set of changes

- Add separate `stale: true` override if operation is in flight

### DIFF
--- a/.changeset/giant-beans-act.md
+++ b/.changeset/giant-beans-act.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-graphcache': patch
+---
+
+Set `stale: true` on cache results, even if a reexecution has been blocked by the loop protection, if the operation is already pending and in-flight.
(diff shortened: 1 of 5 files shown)


JoviDeCroock (approved):

AndrewIngram:
Seems to address the issue I was having (mentioned here: https://twitter.com/andrewingram/status/1752787499053723667), but I also haven't been able to make a simplified reproduction :/

kitten:
The test I added should sufficiently reproduce this scenario. It doesn't quite match the manner in which this case is triggered, but it should represent both the overall conditions that triggered this bug/issue and the fix itself

PR #3499, opened 2024-02-02
feat(core): Support Apollo Federation subscriptions multipart payloads

JoviDeCroock:
## Summary

A missing feature for `multipart/mixed` was pointed out where we were missing support for the apollo subscriptions spec which adds a `payload` property

### DIFF
--- a/.changeset/small-cheetahs-jam.md
+++ b/.changeset/small-cheetahs-jam.md
@@ -1,0 +1,5 @@
+---
+'@urql/core': minor
+---
+
+Support the [Apollo federation spec](https://www.apollographql.com/docs/router/executing-operations/subscription-multipart-protocol/) for dispatching subscriptions
--- a/.changeset/small-cheetahs-jam.md
+++ b/.changeset/small-cheetahs-jam.md
@@ -2,4 +2,4 @@
 '@urql/core': minor
 ---
 
-Support the [Apollo federation spec](https://www.apollographql.com/docs/router/executing-operations/subscription-multipart-protocol/) for dispatching subscriptions
+Support [Apollo Federation's format](https://www.apollographql.com/docs/router/executing-operations/subscription-multipart-protocol/) for subscription results in `multipart/mixed` responses (result properties essentially are namespaced on a `payload` key)
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -223,7 +223,7 @@ export interface ExecutionResult {
    * @see {@link https://github.com/graphql/graphql-wg/blob/main/rfcs/DeferStream.md#payload-format} for the DeferStream spec
    */
   hasNext?: boolean;
-  payload?: ExecutionResult;
+  payload?: Omit<ExecutionResult, 'payload'>;
 }
 
 /** A source of {@link OperationResult | OperationResults}, convertable to a promise, subscribable, or Wonka Source.
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -223,6 +223,7 @@ export interface ExecutionResult {
    * @see {@link https://github.com/graphql/graphql-wg/blob/main/rfcs/DeferStream.md#payload-format} for the DeferStream spec
    */
   hasNext?: boolean;
+  payload?: ExecutionResult;
 }
 
 /** A source of {@link OperationResult | OperationResults}, convertable to a promise, subscribable, or Wonka Source.
--- a/packages/core/src/utils/result.ts
+++ b/packages/core/src/utils/result.ts
@@ -153,7 +153,7 @@ export const mergeResultPatch = (
     withData.data = (nextResult.payload || nextResult).data || prevResult.data;
     errors =
       (nextResult.errors as any[]) ||
-      (nextResult.payload && nextResult.payload.errors) || 
+      (nextResult.payload && nextResult.payload.errors) ||
       errors;
   }
 
--- a/packages/core/src/utils/result.ts
+++ b/packages/core/src/utils/result.ts
@@ -92,8 +92,12 @@ export const mergeResultPatch = (
   pending?: ExecutionResult['pending']
 ): OperationResult => {
   let errors = prevResult.error ? prevResult.error.graphQLErrors : [];
-  let hasExtensions = !!prevResult.extensions || !!nextResult.extensions;
-  const extensions = { ...prevResult.extensions, ...nextResult.extensions };
+  let hasExtensions =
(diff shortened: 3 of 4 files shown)


.changeset/small-cheetahs-jam.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/core': minor
+---
+
+Support the [Apollo federation spec](https://www.apollographql.com/docs/router/executing-operations/subscription-multipart-protocol/) for dispatching subscriptions
```
kitten, line 5:
```suggestion
Support [Apollo Federation's format](https://www.apollographql.com/docs/router/executing-operations/subscription-multipart-protocol/) for subscription results in `multipart/mixed` responses (result properties essentially are namespaced on a `payload` key)
```

packages/core/src/types.ts
```diff
@@ -223,6 +223,7 @@ export interface ExecutionResult {
    * @see {@link https://github.com/graphql/graphql-wg/blob/main/rfcs/DeferStream.md#payload-format} for the DeferStream spec
    */
   hasNext?: boolean;
+  payload?: ExecutionResult;
```
kitten, line 226:
```suggestion
  payload?: Omit<ExecutionResult, 'payload'>;
```

packages/core/src/utils/result.ts
```diff
@@ -146,7 +150,7 @@ export const mergeResultPatch = (
       }
     }
   } else {
-    withData.data = nextResult.data || prevResult.data;
+    withData.data = (nextResult.payload || nextResult).data || prevResult.data;
     errors = (nextResult.errors as any[]) || errors;
```
kitten, line 154: needs to combine `result.errors` with `result.payload.errors` potentially

JoviDeCroock: `result.payload` doesn't have the `errors` property right?

JoviDeCroock: https://github.com/urql-graphql/urql/pull/3499/commits/fd6b15b10f1991bface15814387784443f93dc68

packages/core/src/utils/result.ts
```diff
@@ -146,8 +150,12 @@ export const mergeResultPatch = (
...
     }
   } else {
-    withData.data = nextResult.data || prevResult.data;
-    errors = (nextResult.errors as any[]) || errors;
+    withData.data = (nextResult.payload || nextResult).data || prevResult.data;
+    errors =
+      [
+        ...((nextResult.errors || []) as any[]),
+        ...((nextResult.payload && nextResult.payload.errors) || []),
+      ] || errors;
```
kitten, line 158: I think the protocol assumes that `payload.errors` and `errors` are mutually exclusive 🤔 I think a protocol-level error will mean that `payload` isn't defined, looking at the spec again, so maybe we're fine with:

```suggestion
    errors =
      (nextResult.errors as any[]) ||
      (nextResult.payload && nextResult.payload.errors) || 
      errors;
```

This also solves the problem of us potentially creating an empty `errors` array here, even without any errors 🤔

packages/core/src/utils/result.ts
```diff
@@ -146,8 +150,11 @@ export const mergeResultPatch = (
       }
     }
   } else {
-    withData.data = nextResult.data || prevResult.data;
-    errors = (nextResult.errors as any[]) || errors;
+    withData.data = (nextResult.payload || nextResult).data || prevResult.data;
+    errors =
+      (nextResult.errors as any[]) ||
+      (nextResult.payload && nextResult.payload.errors) ||
```
JoviDeCroock, line 156:
```suggestion
      (nextResult.payload && nextResult.payload.errors) ||
```

kitten (approved):

packages/core/src/utils/result.ts
```diff
@@ -146,8 +150,11 @@ export const mergeResultPatch = (
       }
     }
   } else {
-    withData.data = nextResult.data || prevResult.data;
-    errors = (nextResult.errors as any[]) || errors;
+    withData.data = (nextResult.payload || nextResult).data || prevResult.data;
+    errors =
```
alessbell, line 154: In addition to GraphQL errors there can also be fatal transport-level errors:

> If a fatal transport-level error occurs, the router sends a message with a top-level errors field and null payload field, then closes the connection

We opted to [stash protocol errors in `extensions`](https://github.com/apollographql/apollo-client/pull/10592/files#diff-d180d739ac5085e3556db4fb80641f30389c9958f0eb702e88475d6253506e18R88-R96) (using a symbol property to avoid possible collisions) to pass the error through to client internals and then handle them where we process subscription errors.

JoviDeCroock: Hmm, is there a need for that? As essentially it will deliver the errors on that property and there is no possibility for an accompanying payload. I guess a recipe we should look into is re-establishing the connection if that occurs due to i.e. an interrupted connection

kitten: If this is desirable, we can throw an error in here as well, which will, in the `fetchExchange` at least, be caught here: https://github.com/urql-graphql/urql/blob/5b6449e62d8c3eb7f0f863ae98b3994871c2da38/packages/core/src/internal/fetchSource.ts#L213

i.e. be wrapped in a `CombinedError`.

If this is a protocol-level error, that's the right place to put it, I'd say, if it necessitates destroying the response entirely.

Just to clarify this _will_ make prior data, errors, and extensions irrecoverable in most cases for the UI bindings and following exchanges handling results.

However, the reason why this is potentially problematic is:
- `errors` mirrors the GraphQL result spec and can hence be an array and isn't limited, i.e. `path` and such may still be provided
- throwing is only acceptable if `'payload' in result` hence it's hard to make a decision here without knowing more about protocol implementers

@alessbell: Thanks for pointing this out! 🙌 Is there any case where `payload: null` isn't set in a multipart subscription response and/or the root (not `.payload`) errors array will ever contain more than one protocol-level error that prohibits keeping the rest of the data around?

alessbell: @kitten: thanks for your patience here - we're in the process of clarifying one edge case that came up whereby a subgraph emits a heartbeat-like chunk of `payload: null` (without `errors`), which we may amend the spec text to account for. I should have a clearer answer in the next day or so, thanks!

kitten:
@JoviDeCroock: I think we're good on getting this released. We an always adjust this later on, if needed, but I don't think any future protocol changes here would critically break this, especially if they're clarifications of the current spec.

PR #3502, opened 2024-02-04
feat(graphcache): allow for defining inline-fragment/fragment-definition client controlled nullability directives

JoviDeCroock:
## Summary

This allows for defining `@_optional` and `@_required` on both a fragment-definition as well as an inline-fragment. This to facilitate patterns like a `refetchable-fragment` where you can refetch parts of a document.

Example

```js
query {
  todo {
    id
    text
    author { id name }
    ... on Todo @_optional { seenBy(first: 10, after: $after) }
    
  }
}
```

Now we can paginate over `seenBy` without losing the data surrounding it, this is good for composing fragments from i.e. a list with the parent.


exchanges/graphcache/src/operations/write.ts
```diff
@@ -239,6 +239,7 @@ const writeSelection = (
     typename,
     entityKey || typename,
     deferRef,
+    false,
```
kitten, line 242:
```suggestion
    optionalRef,
```

I might be wrong, but this could accidentally cascade to this being reset to `false`, right?

This is a mistake on my part because, ultimately, `makeSelectionIterator` shouldn't accept either defer or optional as arguments, but that was carried over from past buggy behaviour.
It's just inconvenient for testing

.changeset/happy-peas-sin.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/exchange-graphcache': minor
+---
+
+Allow for defining `@_optional` and `@_required` on fragment-definitions and inline-fragments
```
kitten, line 5:
```suggestion
Allow `@_optional` and `@_required` to be placed on fragment definitions and inline fragments
```

exchanges/graphcache/src/operations/query.ts
```diff
@@ -529,6 +532,7 @@ const readSelection = (
       !deferRef &&
       dataFieldValue === undefined &&
       (directives.optional ||
+        (optionalRef && !directives.required) ||
```
kitten, line 535: This doesn't seem right.
We're of course not tracking `optionalRef` per field, but per “selection iteration”, which is fine but `directives.required` here isn't the only usage.

We also have `directives.required` below on `dataFieldValue === null`, so that error isn't cascading:
https://github.com/urql-graphql/urql/blob/153a250bd5a0ac7eb0b73397a3f5452eb2c2cb93/exchanges/graphcache/src/operations/query.ts#L543-L556

It should probably be `(directives.required || optionalRef === false)` rather than `directives.required` to also cover that case.

Leaving the comment here, since I can't comment on the unmodified line

exchanges/graphcache/src/operations/shared.ts
```diff
@@ -203,10 +213,15 @@ export const makeSelectionIterator = (
...
                 pushDebugNode(typename, fragment);
+
+              const isFragmentOptional = isOptional(select);
               child = makeSelectionIterator(
                 typename,
                 entityKey,
                 defer || isDeferred(select, ctx.variables),
+                isFragmentOptional === undefined
+                  ? optional
+                  : isFragmentOptional,
```
kitten, line 224: I think this _might_ potentially be incorrect.

We pass `optionalRef` back into `optional` as an argument here, i.e.
`optionalRef = optional`, as per the above.

This means that _any sub-selection_ of an optional selection remains optional. In other words, these two cases should be equivalent:

```graphql
query {
  optionalField @_optional {
    nonOptionalField
  }
}

query {
  ... @_optional {
    optionalField {
      nonOptionalField
    }
  }
}
```

However, they're not, as far as I can tell.
We inherit `optionalRef` through sub-iterations, which is fine, since those resolve fragments, however, passing in `optional` as an argument to `optionalRef` seems to break assumptions here.

I previously didn't notice this. Sorry!

exchanges/graphcache/src/operations/query.ts
JoviDeCroock: `optionalRef` is false by default so changing that line will have a much broader effect than what we are aiming for I reckon 😅 we explicitly check whether this has a required directive and else allow ourselves to fall through to the default `optional = false` case as we did before.

kitten (approved):

PR #3503, opened 2024-02-06
docs(graphcache): remove mutative operations

JoviDeCroock:
It's confusing that we use mutative operations in our updaters when that can alter the base-layer when we i.e. are working with optimistic layers.

### DIFF
--- a/docs/graphcache/cache-updates.md
+++ b/docs/graphcache/cache-updates.md
@@ -260,8 +260,10 @@ cacheExchange({
         `;
 
         cache.updateQuery({ query: TodoList }, data => {
-          data.todos.push(result.createTodo);
-          return data;
+          return {
+            ...data,
+            todos: [...data.todos, result.createTodo],
+          };
         });
       },
     },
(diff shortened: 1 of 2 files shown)


kitten:
I see the issue here, so basically, a user could accidentally modify a link on a layer by mutating it directly, since `cache.resolve` returning links is an exception, i.e.:
https://github.com/urql-graphql/urql/blob/f06ef3499b69a20896057c0027c5116f4e73c54a/exchanges/graphcache/src/store/store.ts#L161-L162

Does it make sense for us to instead clone links in `cache.resolve`? It should be the only place that can cause this problem and we already have an appropriate function that can clone links deeply:
https://github.com/urql-graphql/urql/blob/f06ef3499b69a20896057c0027c5116f4e73c54a/exchanges/graphcache/src/operations/shared.ts#L226-L249

JoviDeCroock:
I would personally prefer hammering on the immutable route in the docs either way

kitten:
I'm fine with changing the docs, yea, let's get this merged as a docs change then, and I'll file the fix above separately, so we don't let this bug become a longer term problem 👍

PR #3505, opened 2024-02-10
fix(core): support Headers being passed in fetchOptions

### LINKED ISSUE #3504: Passing a `Headers` object results in an error
### Describe the bug

urql accepts `HeadersInit` (via `RequestInit`), it appears that a `Headers` object can be passed.
But when I actually do, it causes errors.

```ts
const headers = new Headers();

const client = new Client({
  url: "https://localhost:3000/graphql",
  exchanges: [fetchExchange],
  fetchOptions: { headers },
});

const response = await client.query("{ ping }", {});
```

```
[Network] Headers.append: "append(name, value) { /* omit */ }" is an invalid header value.
```

I think: 🤔

- the type might be incorrect
- the implementation might be incorrect

### Reproduction

https://github.com/konomae/urql-headers-repro

### Urql version

urql 4.2.3
node 20.11

### Validations

JoviDeCroock:
Resolves #3504 

## Summary

This aims to solve the problem of supporting `Headers` I opted not to do do a `instanceof` check for this to support node versions that don't have this by default. The downside of doing the `forEach` check is that we also get the `[[string, string]]` `HeadersInit` in there which is a tad harder to support

### DIFF
--- a/packages/core/src/internal/fetchOptions.ts
+++ b/packages/core/src/internal/fetchOptions.ts
@@ -134,7 +134,11 @@ export const makeFetchOptions = (
       (extraOptions.headers as Headers | [[string, string]]).forEach(
         (value, key) => {
           if (Array.isArray(value)) {
-            headers[value[0]] = value[1];
+            if (headers[value[0]]) {
+              headers[value[0]] = `${headers[value[0]]},${value[1]}`;
+            } else {
+              headers[value[0]] = value[1];
+            }
           } else {
             headers[key] = value;
           }
--- a/packages/core/src/internal/fetchOptions.ts
+++ b/packages/core/src/internal/fetchOptions.ts
@@ -102,6 +102,9 @@ const serializeBody = (
   }
 };
 
+const isHeaders = (headers: HeadersInit): headers is Headers =>
+  'has' in headers && !Object.keys(headers).length;
+
 /** Creates a `RequestInit` object for a given `Operation`.
  *
  * @param operation - An {@link Operation} for which to make the request.
@@ -130,8 +133,12 @@ export const makeFetchOptions = (
       ? operation.context.fetchOptions()
       : operation.context.fetchOptions) || {};
   if (extraOptions.headers) {
-    if (extraOptions.headers.forEach) {
-      (extraOptions.headers as Headers | [[string, string]]).forEach(
+    if (isHeaders(extraOptions.headers)) {
+      extraOptions.headers.forEach((value, key) => {
+        headers[key] = value;
+      });
+    } else if (Array.isArray(extraOptions.headers)) {
+      (extraOptions.headers as Array<[string, string]>).forEach(
         (value, key) => {
           if (Array.isArray(value)) {
             if (headers[value[0]]) {
--- a/packages/core/src/internal/fetchOptions.ts
+++ b/packages/core/src/internal/fetchOptions.ts
@@ -129,9 +129,18 @@ export const makeFetchOptions = (
     (typeof operation.context.fetchOptions === 'function'
       ? operation.context.fetchOptions()
       : operation.context.fetchOptions) || {};
-  if (extraOptions.headers)
-    for (const key in extraOptions.headers)
-      headers[key.toLowerCase()] = extraOptions.headers[key];
+  if (extraOptions.headers) {
+    if (extraOptions.headers.forEach) {
+      (extraOptions.headers as Headers).forEach((value, key) => {
+        headers[key] = value;
+      });
+    } else {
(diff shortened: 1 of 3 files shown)


packages/core/src/internal/fetchOptions.ts
```diff
@@ -129,9 +129,24 @@ export const makeFetchOptions = (
...
       : operation.context.fetchOptions) || {};
-  if (extraOptions.headers)
-    for (const key in extraOptions.headers)
-      headers[key.toLowerCase()] = extraOptions.headers[key];
+  if (extraOptions.headers) {
+    if (extraOptions.headers.forEach) {
+      (extraOptions.headers as Headers | [[string, string]]).forEach(
+        (value, key) => {
+          if (Array.isArray(value)) {
+            headers[value[0]] = value[1];
```
kitten, line 137: This doesn't seem right and should be comma-separated 🤔

kitten (commented):
Should this be "inverted"? Seems to me that this is an opportunity to simply ditch the object representation and instead pass a `Headers` object directly since we're already assuming web-globale support

JoviDeCroock:
That would be a breaking change as that is not compatible with node-fetch environments

packages/core/src/internal/fetchOptions.ts
kitten: Oh, sorry, I meant, `[string, string[]]`, but I'm not sure whether that'll actually be issued as repeated `[string, string]` entries. Sorry!

JoviDeCroock: According to the `HeadersInit` type, which we support here it's `[[string, string]]` which I guess makes it so that we have to merge into existing headers

packages/core/src/internal/fetchOptions.ts
```diff
@@ -129,9 +129,28 @@ export const makeFetchOptions = (
     (typeof operation.context.fetchOptions === 'function'
       ? operation.context.fetchOptions()
       : operation.context.fetchOptions) || {};
-  if (extraOptions.headers)
-    for (const key in extraOptions.headers)
-      headers[key.toLowerCase()] = extraOptions.headers[key];
+  if (extraOptions.headers) {
+    if (extraOptions.headers.forEach) {
+      (extraOptions.headers as Headers | [[string, string]]).forEach(
+        (value, key) => {
```
kitten, line 135: I'm a bit concerned that the `forEach`’s function doesn't infer from usage (i.e. `value` and `key` are `any`)

Also `[[string, string]]` is incorrect, as it's `[string, string][]`, but this doesn't fix the inference here anyway, so it's a bit besides the point.

The main problem here is that both cases aren't equivalent however. `forEach` on the array of tuples will have `value` be `[string, string]`, while on `Headers#forEach` it'll be just `string` and the second argument becomes `string` too.

We can supplement this with `headers.entries()` but that invalidates the type cast.
It also is a little awkward since we'd have to check `Array.isArray` and otherwise check for something that definitively tells us we have a `Headers` instance.
We can still rely on `forEach`, but we could also do `'has' in headers && !Object.keys(headers).length` which is pretty definitive.

But this comes down to `for (const entry of Array.isArray(headers) ? headers : (headers as Headers).entries())`, which will also require `dom.iterable` to be added to our `tsconfig.json`.

---

Another alternative is to reuse whatever we receive:

```ts
  const defaultHeaders = {
    accept:
      operation.kind === 'subscription'
        ? 'text/event-stream, multipart/mixed'
        : 'application/graphql-response+json, application/graphql+json, application/json, text/event-stream, multipart/mixed',
  };

  const extraOptions =
    (typeof operation.context.fetchOptions === 'function'
      ? operation.context.fetchOptions()
      : operation.context.fetchOptions) || {};

  const isHeaders = (headers: HeadersInit): headers is Headers =>
    'has' in headers && !Object.keys(headers).length;

  let headers: HeadersInit | undefined = extraOptions.headers;
  if (!headers) {
    headers = defaultHeaders;
  } else if (Array.isArray(headers)) {
    if (!headers.some((entry) => entry[0].toLowerCase() === 'accept'))
      headers.push(['accept', defaultHeaders.accept]);
  } else if (isHeaders(headers)) {
    if (!headers.has('accept'))
      headers.set('accept', defaultHeaders.accept);
  } else {
    if (!Object.keys(headers).some((key) => key.toLowerCase() === 'accept'))
      headers.accept = defaultHeaders.accept;
  }
```

which also avoids the discrepancy below of `${headerPrevValue},${headerValue}` rather than `${headerPrevValue}, ${headerValue}` and any similar issues we may run into.

Ultimately either is fine though 🤔

kitten (approved):

PR #3514, opened 2024-03-02
fix(core): Allow path-only URLs to be used with GET requests

akrantz01:
## Summary

Changes query parameter handling to allow path-only URLs (i.e. `/api/graphql`) passed to `fetch` to be used with GET requests. This was previously broken as [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL) would throw a `TypeError: Invalid URL` as it only works with fully formed URLs.

Extracting the unparsed query parameters is now done manually, so as to not depend on `URL`. Important to note is that the current method will not handle URLs with fragments (i.e. `/api/graphql#fragment`) in them, but it is easy to add if desired.

### DIFF
--- a/.changeset/curvy-sheep-drive.md
+++ b/.changeset/curvy-sheep-drive.md
@@ -2,4 +2,4 @@
 '@urql/core': patch
 ---
 
-Allow URLs consisting of only a path (i.e. `/api/graphql`) to be passed to `fetch` when using GET requests
+Allow `url` to be a plain, non-URL pathname (i.e. `/api/graphql`) to be used with `preferGetMethod`
--- a/.changeset/curvy-sheep-drive.md
+++ b/.changeset/curvy-sheep-drive.md
@@ -1,0 +1,5 @@
+---
+'@urql/core': patch
+---
+
+Allow URLs consisting of only a path (i.e. `/api/graphql`) to be passed to `fetch` when using GET requests
--- a/packages/core/src/internal/fetchOptions.ts
+++ b/packages/core/src/internal/fetchOptions.ts
@@ -55,19 +55,17 @@ export const makeFetchURL = (
     operation.kind === 'query' && operation.context.preferGetMethod;
   if (!useGETMethod || !body) return operation.context.url;
 
-  const [url, params] = extractParams(operation.context.url);
+  const urlParts = splitOutSearchParams(operation.context.url);
   for (const key in body) {
     const value = body[key];
     if (value) {
-      params.set(
+      urlParts[1].set(
         key,
         typeof value === 'object' ? stringifyVariables(value) : value
       );
     }
   }
-
-  const serialized = params.toString();
-  const finalUrl = serialized.length > 0 ? `${url}?${serialized}` : url;
+  const finalUrl = urlParts.join('?');
   if (finalUrl.length > 2047 && useGETMethod !== 'force') {
     operation.context.preferGetMethod = false;
     return operation.context.url;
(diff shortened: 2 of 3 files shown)


packages/core/src/internal/fetchOptions.ts
```diff
@@ -75,6 +76,15 @@ export const makeFetchURL = (
   return finalUrl;
 };
 
+/** Extract the query parameters from a URL, without using {@link URL} to support partial URLs */
+const extractParams = (url: string): [string, URLSearchParams] => {
```
kitten, line 80:
```suggestion
const extractParams = (url: string): readonly [string, URLSearchParams] => {
```

nit: Let's omit the TSDoc comment here, since it doesn't add much given the short length of the code, and isn't public

I'd also rename the function from `extractParams` to `splitOutSearchParams`, or something similar for clarity

The `readonly` addition is for a different comment, just to make sure that always works

packages/core/src/internal/fetchOptions.ts
```diff
@@ -75,6 +76,15 @@ export const makeFetchURL = (
...
+  const start = url.indexOf('?');
+  if (start === -1) return [url, new URLSearchParams()];
+
+  const raw = url.slice(start + 1);
+  return [url.slice(0, start), new URLSearchParams(raw)];
```
kitten, line 85:
```suggestion
  return start > -1
    ? [url.slice(0, start), new URLSearchParams(url.slice(start + 1))]
    : [url, new URLSearchParams()]
```

nit: this is just a little different than our typical code style. Nothing major, but since I've got more suggestions we can align this with the rest

packages/core/src/internal/fetchOptions.ts
```diff
@@ -55,18 +55,19 @@ export const makeFetchURL = (
...
+      params.set(
         key,
         typeof value === 'object' ? stringifyVariables(value) : value
       );
     }
   }
 
-  const finalUrl = url.toString();
+  const serialized = params.toString();
+  const finalUrl = serialized.length > 0 ? `${url}?${serialized}` : url;
```
kitten, line 70:
```suggestion
  const urlParts = extractParams(operation.context.url);
  for (const key in body) {
    const value = body[key];
    if (value) {
      urlParts[1].set(
        key,
        typeof value === 'object' ? stringifyVariables(value) : value
      );
    }
  }

  const finalUrl = urlParts.join('?');
```

nit: since we can assume that we'll always have search params we can do something a little fancier here ❤️ mostly we try to sometimes find these little changes to save on some bytes in the output bundle, just to add context on why I'm suggesting this

kitten (commented):
Great work! I only have some nits here.

We indeed previously assumed that a full URL will be passed, but of course, this doesn't give a user the most options, so I appreciate the suggestion! The test is great too ❤️ 

I've left the nits mostly for code style / byte savings, so that's why they're nits

.changeset/curvy-sheep-drive.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/core': patch
+---
+
+Allow URLs consisting of only a path (i.e. `/api/graphql`) to be passed to `fetch` when using GET requests
```
kitten, line 5:
```suggestion
Allow `url` to be a plain, non-URL pathname (i.e. `/api/graphql`) to be used with `preferGetMethod`
```

Just suggesting this change too, while I'm here for clarity.

kitten (approved):

PR #3515, opened 2024-03-02
feat: Add support for persisted documents via the `documentId` property

kitten:
## Summary

This PR adds support for persisted documents, as long as:
- An input GraphQL AST contains a `documentId` property
- A GraphQL API accepts the `documentId` parameter as a replacement for the `query` parameter

When we detect the `documentId` property on a document which has no definitions (either an empty array or not defined), we send a `documentId` instead of a `query`.

This is similar to Automatic Persisted Queries (APQs), but those require the `@urql/exchange-persisted` exchange, to facilitate generating persisted queries on the fly (as per the spec) and retrying them.

This is to comply with: https://github.com/graphql/graphql-over-http/pull/264

## Set of changes

- Add `documentId` property to detected GraphQL document types
- Send `documentId` on `FetchBody` when a persisted document is passed

### DIFF
--- a/.changeset/two-bananas-grow.md
+++ b/.changeset/two-bananas-grow.md
@@ -1,0 +1,5 @@
+---
+'@urql/core': minor
+---
+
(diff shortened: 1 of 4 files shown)


JoviDeCroock (approved):

kitten:
Merging since it technically doesn't matter when we land this exactly

PR #3564, opened 2024-04-17
feat(graphcache): only add dependency if we are changing the value

JoviDeCroock:
## Summary

This is kind of an optimisation, say we have three operations that are all watching different fields of `Todo`, only one of these operations watches `Todo.completed`. When a mutation changes `Todo.completed` we should only invoke that one operation, this is slightly more performant. Not sure if worth it but it could also enable us to create a granular re-rendering mechanism with a first-class `fragment` citizen.

### DIFF
--- a/exchanges/graphcache/src/cacheExchange.test.ts
+++ b/exchanges/graphcache/src/cacheExchange.test.ts
@@ -480,7 +480,7 @@ describe('data dependencies', () => {
     );
   });
 
-  it.only('does not notify related queries when a mutation update does not change the data', () => {
+  it('does not notify related queries when a mutation update does not change the data', () => {
     vi.useFakeTimers();
 
     const balanceFragment = gql`
@@ -639,12 +639,6 @@ describe('data dependencies', () => {
     expect(updates.Mutation.updateBalance).toHaveBeenCalledTimes(1);
 
     expect(reexec).toHaveBeenCalledTimes(0);
-    expect(reexec.mock.calls[0][0].key).toBe(1);
-
-    expect(result.mock.calls[2][0]).toHaveProperty(
-      'data.author.balance.amount',
-      100
-    );
   });
 
   it('does nothing when no related queries have changed', () => {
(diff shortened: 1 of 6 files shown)


exchanges/graphcache/src/cacheExchange.test.ts
```diff
@@ -1448,15 +1609,17 @@ describe('optimistic updates', () => {
 
     expect(response).toHaveBeenCalledTimes(1);
     expect(optimistic.concealAuthor).toHaveBeenCalledTimes(2);
-    expect(reexec).toHaveBeenCalledTimes(2);
+    expect(reexec).toHaveBeenCalledTimes(1);
```
JoviDeCroock, line 1612: Have to verify these changes first though 😅

kitten: The change in test behaviour seems as intended.

Above we have two mutations that both trigger the same optimistic change, hence, one doesn't actually change the value, so the swallowed change is correct.

However, I'm a little curious whether we're at risk of violating some layering if we don't analyse that first since there's a slight possibility of shifted layers (layers hosited up) not seeing an actual change, even if they have the side-effect of overriding a field back to an updated value.
i.e. from the `write`’s perspective nothing changes, but from the `read` perspective a value _does_ change.

Shouldn't be too much of a big deal though if that does happen (?)
But does highlight that, potentially, the layering can be simplified as we talked about before

exchanges/graphcache/src/operations/write.ts
```diff
@@ -343,19 +346,28 @@ const writeSelection = (
...
-        fieldKey,
-        (fieldValue !== null || !getFieldError(ctx)
-          ? fieldValue
-          : undefined) as EntityField
+        fieldKey
       );
+      const value = (
+        fieldValue !== null || !getFieldError(ctx) ? fieldValue : undefined
+      ) as EntityField;
+      if (!isEqualLinkOrScalar(existingRecord, value)) {
```
kitten, line 367: Are these checks too high in the code path?
We could in theory shift them down to `writeRecord` and `writeLink`, no? Or am I missing something?

JoviDeCroock: If we shift them down it also affects `squashLayer` which I wanted to avoid

kitten (approved):

PR #3567, opened 2024-04-18
fix(core): Fix `@ts-ignore` directive on TS import in type declaration output

kitten:
See: https://github.com/0no-co/graphql.web/pull/27

### DIFF
--- a/.changeset/lucky-timers-hear.md
+++ b/.changeset/lucky-timers-hear.md
@@ -1,0 +1,5 @@
+---
+'@urql/core': patch
+---
+
+Fix `@ts-ignore` on TypeScript peer dependency import in typings not being applied due to a leading `!` character.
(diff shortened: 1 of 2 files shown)


JoviDeCroock (approved):

IvanUkhov:
I think you forgot to update the dependency on `0no-co/graphql.web`. It still depends on the version with that problem.

kitten:
@IvanUkhov It's within range: https://github.com/urql-graphql/urql/blob/de4081b2637d2a882425c9395ee5f95a4a6220d3/packages/core/package.json#L57
Typically, I wouldn't bump that unless there's an explicit reliance on changes in the sub-package (like API changes). Since it's in range and the fix isn't preventing anyone from using it (i.e. if it's discovered, I'd hope the first thing someone does is to upgrade the broken dependency transitively). especially since `skipLibCheck` in TS is common I'd just leave it be

IvanUkhov:
Good point. People will figure out to update if case they bump into this issue.

PR #3570, opened 2024-04-22
feat(react): useFragment hook (BETA)

### LINKED ISSUE #1408: RFC: Fragment Suspense boundaries in React bindings
## Summary

Overall, our new Suspense implementation is working a lot better, as expected, after getting some feedback on it from the React team. This gives us the room to experiment with more uses of Suspense in our React bindings.

Today when we load a query with Graphcache we may run into cases where [Schema Awareness](https://formidable.com/open-source/urql/docs/graphcache/schema-awareness/) kicks in, when it's enabled.

d13d7d912b7dd35e82cb6c98b36d9c65-2000-1

Today this already allows people to have a page-wide query that loads some data and displays partial data when the cache doesn't have enough information to fulfil the query in its entirety but so far that it can leave out optional fields.

This poses a challenge for implementations, where some sub-components of a page may have to switch between a partial and a non-partial state. Assuming that these have types in a TypeScript environment, dealing with partial data on its own isn't too difficult. However, implementing many cascading loading states can be a difficult task and Suspense can help here.

## Proposed Solution

It should be possible for a component to _check data for completion_ using a `useFragment` suspense hook. This hook accepts a fragment and a piece of data and suspends when the fragment isn't fully fulfiled and partial.

**TBD: Exact API details pending #1221**

## Requirements

- A `useFragment` hook should accept data and a fragment and return masked data
- The `useFragment` hook should suspend when its data is incomplete

JoviDeCroock:
Resolves #1408

## Summary

This introduces a new hook to the React package that has two main purposes. On one hand it will mask over the data it gets from its parent component given a fragment, so in the following scenario:

```graphql
query GetTodo { todo { ...TodoFields completed } }
fragment TodoFields on Todo { id text }
```

When we'd pass the resulting data from `GetTodo.todo` into a component and leverage `useFragment` we wouldn't be able to get to the `completed` parameter.

The second goal is to establish a `loading` paradigm for `deferred` queries, this means that given the following document

```graphql
query {
  todo {
    id
    text
    author { ...AuthorFields @defer }
  }
}

fragment AuthorFields on Author {
  id
  name
}
```

When the initial payload comes in we can display the `id` and `text`, during that we display this query as `stale` however we can make this hook in deeper with React paradigms by leveraging `suspense` when the deferred boundary isn't complete and suspense is enabled we can suspend. Even without `suspense` this brings the benefit of not having to check `stale` and being able to render independently of the `result.stale` flag.

## Notes

The heuristic used to assess whether we are using is the presence of `undefined` basically `data` can only be `undefined` when it has not been queried or while it's being streamed in.

I opted for a slightly different approach to `useQuery`, in `useQuery` we are subscribed to the `request` and when the resolved query comes in we call `resolve` on the promise we threw to React. In this one I opted to purely base it on `props`, I tested this in this [demo](https://stackblitz.com/edit/vitejs-vite-bg9pmk?file=src%2FApp.tsx&terminal=dev).

I opted to create the request with our code we use for operations but replaced `variables` with `data` to ensure every entity gets its own cache entry.

A lacking piece at the moment would be directives, if a user leverages `@include()` with a variable then we'd need access to the parent variables to uphold our heuristic of when to set this to `loading`. To counter-act this we check whether the field is `undefined` and when it had one of these directives we accept the absence of the data.

In the future we'll need to find a way to `resolve()` the promise, this is currently impossible as we have no way of establishing a stable identifier across concurrent invocations of a fragment given unique sets of data. 

## Todo

- [x] support `gql.tada` `FragmentOf` dance for `data`
   - Decision: unwrap with `readFragment` before using `useFragment`
- [x] verify whether `TypedDocumentNode` types work correctly
- [x] handle nested fragments in the `FragmentDefinitionNode`

### DIFF
--- a/examples/with-defer-stream-directives/src/Songs.jsx
+++ b/examples/with-defer-stream-directives/src/Songs.jsx
@@ -31,7 +31,6 @@ const Song = React.memo(function Song({ song }) {
 });
 
 const DeferredSong = ({ data }) => {
-  console.log(data, SecondVerseFragment)
   const result = useFragment({
     query: SecondVerseFragment,
     data,
--- a/examples/with-defer-stream-directives/src/Songs.jsx
+++ b/examples/with-defer-stream-directives/src/Songs.jsx
@@ -1,5 +1,5 @@
-import React from 'react';
-import { gql, useQuery } from 'urql';
+import React, { Suspense } from 'react';
+import { gql, useQuery, useFragment } from 'urql';
 
 const SecondVerseFragment = gql`
   fragment secondVerseFields on Song {
@@ -13,9 +13,6 @@ const SONGS_QUERY = gql`
       firstVerse
       ...secondVerseFields @defer
     }
-    alphabet @stream(initialCount: 3) {
-      char
-    }
   }
 
   ${SecondVerseFragment}
@@ -25,10 +22,22 @@ const Song = React.memo(function Song({ song }) {
   return (
     <section>
       <p>{song.firstVerse}</p>
+      <Suspense fallback={'Loading song 2...'}>
+        <DeferredSong data={song} />
+      </Suspense>
       <p>{song.secondVerse}</p>
     </section>
   );
 });
+
+const DeferredSong = ({ data }) => {
+  console.log(data, SecondVerseFragment)
+  const result = useFragment({
+    query: SecondVerseFragment,
+    data,
+  });
+  return <p>{result.secondVerse}</p>;
+};
 
 const LocationsList = () => {
   const [result] = useQuery({
@@ -42,9 +51,6 @@ const LocationsList = () => {
       {data && (
         <>
           <Song song={data.song} />
-          {data.alphabet.map(i => (
-            <div key={i.char}>{i.char}</div>
-          ))}
         </>
       )}
     </div>
--- a/examples/with-defer-stream-directives/src/Songs.jsx
+++ b/examples/with-defer-stream-directives/src/Songs.jsx
@@ -13,6 +13,9 @@ const SONGS_QUERY = gql`
       firstVerse
       ...secondVerseFields @defer
     }
+    alphabet @stream(initialCount: 3) {
+      char
+    }
   }
 
   ${SecondVerseFragment}
@@ -50,6 +53,9 @@ const LocationsList = () => {
       {data && (
         <>
           <Song song={data.song} />
+          {data.alphabet.map(i => (
+            <div key={i.char}>{i.char}</div>
+          ))}
         </>
       )}
     </div>
--- a/packages/react-urql/src/hooks/useFragment.ts
+++ b/packages/react-urql/src/hooks/useFragment.ts
@@ -240,6 +240,21 @@ const maskFragment = <Data extends Record<string, any>>(
         }
       }
       maskedData[selection.name.value] = data[selection.name.value];
+    } else if (selection.kind === Kind.INLINE_FRAGMENT) {
+      if (
+        selection.typeCondition &&
+        selection.typeCondition.name.value !== data.__typename
+      ) {
+        return;
+      }
+
+      const result = maskFragment(data, selection.selectionSet);
+      if (!result.fulfilled) {
+        isDataComplete = false;
+      }
+      Object.assign(maskedData, result.data);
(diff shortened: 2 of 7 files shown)


packages/react-urql/src/hooks/useFragment.ts
```diff
@@ -0,0 +1,262 @@
...
+  return currentResult;
+}
+
+const maskFragment = <Data extends Record<string, any>>(
+  data: Data,
+  selectionSet: SelectionSetNode
+): { data: Data; fulfilled: boolean } => {
+  const maskedData = {};
+  let isDataComplete = true;
+  selectionSet.selections.forEach(selection => {
```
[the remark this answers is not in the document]
JoviDeCroock: There are a few things bothering me about doing that, we should def do `FragmentSpread` as alluded to in the todo, however what bothers me is that these spreads could be suspended themselves. I wanted to create the assumption that a `FragmentSpread` with `@defer` wouldn't be registered in `useFragment` but there would be an expectation that this is passed on to a new component.

For `InlineFragment` I have implemented it similar to `Field` and `defer` would trigger internal to the component.

In writing this I realised that we actually can't do `include`/`skip` as it stands because we don't have access to the variables the parent-operation was executed with, this could be a limitation for now but weakens the heuristic we use to see whether something is deferred significantly.

packages/react-urql/src/hooks/useFragment.ts
```diff
@@ -0,0 +1,339 @@
...
+ * };
+ * ```
+ */
+export function useFragment<
+  Data extends Record<string, any> = Record<string, any>,
+>(args: UseFragmentArgs<Data>): UseFragmentState<Data> {
+  const client = useClient();
+  const cache = getCacheForClient(client);
+  const suspense = isSuspense(client, args.context);
+  const request = useRequest(args.query, args.data);
```
JoviDeCroock, line 127: Not sure whether this is too expensive, we could just rely on `args.data` and `args.query`

examples/with-defer-stream-directives/src/Songs.jsx
```diff
@@ -25,11 +22,22 @@ const Song = React.memo(function Song({ song }) {
...
       <p>{song.firstVerse}</p>
+      <Suspense fallback={'Loading song 2...'}>
+        <DeferredSong data={song} />
+      </Suspense>
       <p>{song.secondVerse}</p>
     </section>
   );
 });
 
+const DeferredSong = ({ data }) => {
```
JoviDeCroock, line 33: I will revert the example changes, mainly using this for testing

packages/react-urql/src/hooks/useFragment.ts
```diff
@@ -0,0 +1,345 @@
...
+        const newResult = maskFragment<Data>(
+          data,
+          fragment.selectionSet,
+          fragments
+        );
+
+        if (newResult.fulfilled) {
+          return { data: newResult.data, fetching: false };
+        } else if (suspense) {
+          const promise = new Promise(() => {});
```
JoviDeCroock, line 175: Being able to resolve this Promise might be important for streamed-SSR, I am not sure 😅 the issue here being that we can't keep a stable reference to this cache-entry as we face two challenges

- This could be a series of useFragment, think of a list of todos. This means we'd be able to create a key for the document, however we'd then share this key across all siblings which could be streamed in out of order.
- When we add `data` to the `key` this means that when the previous entry is completed due to the missing part streaming in that we update the key to a new state. This becomes problematic as we never actually resolve the promise, storing the previous reference in a ref isn't possible as React tears down the previous state.
- Leveraging `useId` is also not possible as that isn't stable across Suspense invocations

One solution I can see is deriving a `data` that follows along the document and purposefully omits any `include/skip` and `defer` from the data and uses that to derive the key. This would essentially be similar to creating a `__typename:id` as they key, which could also be relevant here.

JoviDeCroock:
This proposal has run into kind of a dead-end after trying it out with streamed rendering, the issue resides in the fact that during SSR React does not do any re-rendering which means `useFragment` will never be re-evaluated.

For this to work we'd have to establish a `subscription` in our urql-client where this fragment is used, this subscription would need to bring the data to the `useFragment`, this adds additional complexity to this API proposal because we'd need to be aware on which entity we are listening for changes _and_ hook into the correct query. to evaluate this fragment against.

Issue #3254, opened 2023-06-07
Invalidate doesn't work as expected for some of the queries, or components not getting rerendered with new data


RIP21:
### Describe the bug

Basically, I invalidate a bunch of root queries on mutation. It goes through successfully, but then queries that got invalidated are not getting reexeucted from the network. Or it does, and one of the components (freshly rendered) is updated, but the rest of the subscribed to the same query components hold on to old data.

This is how it looks in the dev tools.

Some did refetch as you can see with the green thing, but some are instant weird reexecutions that lead to nothing.
<img width="694" alt="CleanShot 2023-06-07 at 22 03 59@2x" src="https://github.com/urql-graphql/urql/assets/3940079/7be44be8-a144-4ab6-9ab3-a2617462c230">

It looks like a series (15-20 of them) of teardown, execute, update, execute, update, teardown, execute, update.

It's hard to make a reproduction, but this is what my update function looks like.

```ts
  updates: {
    Mutation: {
      organizationCreateWithExistingSafe: (_parent, _args, cache, _info) => {
        cache.invalidate("Query", "organizations" as keyof Query)
        cache.invalidate("Query", "safes" as keyof Query)
      },

```
Maybe I don't understand something?

It's important to note that these two queries are shared extensively in UI through a hook like.
```
export const useSafesQuery = () => {
  return useQueryData({ query: UseSafes_SafesDocument })
}
```
And there are at least 2 consumer components on the page rendered at the point of invalidation.
Since there is no warning against such a thing in docs, it's quite a common pattern in the codebase for very reusable stuff like user info etc. that makes no sense to re-query all the time.

Also, it's important to note that network policy everywhere is the default one, so `cache-first` and not `cache-and-network` or `network-only`.

REPRO ATTACHED HERE IS NOT A REPRO. I don't have time to build a repro, especially since it DOES invalidate others correctly rather 90% of the time so to build one is virtually impossible.

I can hop on a call anytime to show it off to anyone interested if that would be easiest.

### Reproduction

https://codesandbox.io/s/urql-issue-template-client-iui0o

### Urql version

    "@urql/core": "~4.0.10",
    "@urql/devtools": "~2.0.3",
    "@urql/exchange-graphcache": "~6.1.1",
    "@urql/exchange-refocus": "~1.0.2",
    "urql": "~4.0.3",
    "wonka": "~6.3.2",

### Validations

[...]

PR #3573, opened 2024-04-24
fix(client): eager unblocking of in-flight operations

### LINKED ISSUE #3565: Multiple requests when using react suspense
### Describe the bug

The bug is that urql is somehow spamming multiple requests.
Our application works like this basically. It has a set of modules and we iterate them returning lazy loaded components wrapped in React Suspense.

```
<Suspense>
  modules.map((m) => {
    component = lazy(m.component)
    return <Route 
      ...
    />
  })
</Suspense>
```

We don't know why this happens. But we suspect that it has something to do with how we structured one of our tree. We also suspect it has something to do with the lazy loading and react suspense In the example we have been able to reproduce the issue.

Consider this structure. It only happens when it is like this.
```
Query
  vehicles
Vehicles
  car
  cars
  bicycle
  bicycles
```

### Reproduction
https://github.com/danielkaxis/urqlsuspensetest/tree/multiplerequests

__mocked_generated__ folder has the schema and urql resolvers.

yarn
yarn serve
localhost:8000
Open developer tools
reload page

All logs show the query that has been called with console.count().
`Car:name` should be called 4 times but is called 14 times.

Why is it called so many times?

image

Comment out the mutation in App.tsx line 13.
```js
await client.mutation(userMutation, {}).toPromise()
```
reload the page
`Car:name` is now called only 8 times.

Why did that decrease the number of requests?

### Urql version

```
"@graphql-tools/schema": "10.0.3",
"@urql/devtools": "2.0.3",
"@urql/exchange-execute": "2.2.2",
"@urql/exchange-graphcache": "7.0.1",
"graphql": "16.8.1",
"urql": "4.0.7",
```

### Validations

JoviDeCroock:
Resolves #3565

## Summary

In the initial issue we were faced with the following issue, imagine a query that returns a list with values `[Todo:1, Todo:2]` we use this list to return components that in turn has its own query to fetch the details of this `Todo`. Both this list and the details have a shared parent field let's name it `Query.author`, this is important for graph-cache to re-dispatch operations.

This diagram of this looks like the following

```
Operation List
Completed Operation List
Operation Todo:1
Operation Todo:2
Completed Operation Todo:1
Operation Todo:2
Completed Operation Todo:2
Completed Operation Todo:2
```

Notice that we re-dispatch `Todo:2` even though it is already in-flight, this is problematic as in the cache it has not resolved to a response yet and hence we send off multiple network requests. This issue becomes more and more apparent given more queries being in-flight. The lines causing this issue can be found [here](https://github.com/urql-graphql/urql/blob/main/packages/core/src/client.ts#L701-L706) since there is no operation queued we will always remove the operation from our in-flight operations.

The solution here is to change this logic to only remove `dispatched` on a queued operation that either is already in `dispatched` or is `network-only`.

> I did notice that https://github.com/urql-graphql/urql/pull/3564 already makes this issue less severe due to dispatching less operations from graphcache.

The above fix however re-introduces the issue we faced in https://github.com/urql-graphql/urql/issues/3254 this due to our optimistic mutation re-triggering the operation and resulting in `stale` and then our real data re-triggering the query but it already having been dispatched. This because we have similar logic in the [`onPush`](https://github.com/urql-graphql/urql/blob/main/packages/core/src/client.ts#L645-L653) part of our `operation`. When it's stale we'll check the queue whether there is an operation in there, however there won't be as we are dealing with an optimistic mutation re-trigger, to fix this I added a check whether the stale result has a next payload coming, if not it gets deleted form dispatched either way.

I think the above `stale` flag fix is the root cause that needed to be addressed in #3254 but I am not a 100% sure as I got a bit overloaded while leveraging `console.log` in all of these reproductions 😅

### DIFF
--- a/.changeset/tiny-pots-allow.md
+++ b/.changeset/tiny-pots-allow.md
@@ -1,0 +1,8 @@
+---
+'@urql/core': patch
+---
+
+Fix issue where a reexecute on an in-flight operation would lead to multiple network-requests.
+This issue was observable when using graphcache on pages where we'd have multiple queries that are inter-dependent, i.e.
+they shared entities that had queries running in parallel, one of them completing would result in
+graphcache calling `reexecute` while the other ones were still in-flight leading to duplicate network requests.
--- a/.changeset/tiny-pots-allow.md
+++ b/.changeset/tiny-pots-allow.md
@@ -3,6 +3,4 @@
 ---
 
 Fix issue where a reexecute on an in-flight operation would lead to multiple network-requests.
-This issue was observable when using graphcache on pages where we'd have multiple queries that are inter-dependent, i.e.
-they shared entities that had queries running in parallel, one of them completing would result in
-graphcache calling `reexecute` while the other ones were still in-flight leading to duplicate network requests.
+For example, this issue presents itself when Graphcache is concurrently updating multiple, inter-dependent queries with shared entities. One query completing while others are still in-flight may lead to duplicate operations being issued.
--- a/packages/core/src/client.test.ts
+++ b/packages/core/src/client.test.ts
@@ -19,6 +19,9 @@ import {
   onPush,
   tap,
   take,
+  fromPromise,
+  fromValue,
+  switchMap,
 } from 'wonka';
 
 import { gql } from './gql';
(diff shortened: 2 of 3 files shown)


packages/core/src/client.ts
```diff
@@ -697,13 +702,25 @@ export const Client: new (opts: ClientOptions) => Client = function Client(
...
         for (let i = 0; i < queue.length; i++)
           queued = queued || queue[i].key === operation.key;
-        if (!queued) dispatched.delete(operation.key);
-        queue.push(operation);
-        Promise.resolve().then(dispatchOperation);
+
+        if (
+          !queued &&
+          (!dispatched.has(operation.key) ||
+            operation.context.requestPolicy !== 'network-only')
```
JoviDeCroock, line 716:
```suggestion
            operation.context.requestPolicy === 'network-only')
```

packages/core/src/client.test.ts
```diff
@@ -732,13 +778,82 @@ describe('deduplication behavior', () => {
...
+    pipe(client.executeRequestOperation(mutation), subscribe(onResult));
+    await Promise.resolve();
+
     expect(onOperation).toHaveBeenCalledTimes(2);
     expect(onResult).toHaveBeenCalledTimes(1);
+
+    resolve();
+    await Promise.resolve();
+    expect(onOperation).toHaveBeenCalledTimes(2);
+    // TODO: figure out why this isn't called a second time
```
JoviDeCroock, line 855: This should be called due to us resolving the promise, however it isn't

.changeset/tiny-pots-allow.md
```diff
@@ -0,0 +1,8 @@
+---
+'@urql/core': patch
+---
+
+Fix issue where a reexecute on an in-flight operation would lead to multiple network-requests.
+This issue was observable when using graphcache on pages where we'd have multiple queries that are inter-dependent, i.e.
+they shared entities that had queries running in parallel, one of them completing would result in
+graphcache calling `reexecute` while the other ones were still in-flight leading to duplicate network requests.
```
kitten, line 8:
```suggestion
For example, this issue presents itself when Graphcache is concurrently updating multiple, inter-dependent queries with shared entities. One query completing while others are still in-flight may lead to duplicate operations being issued.
```

kitten (approved):

PR #3574, opened 2024-04-26
chore(workspace): automatically publish urql-core to JSR

JoviDeCroock:
## Summary

This adds a publishing pipeline to [jsr](https://jsr.io/), for now it will only publish `@urql/core` because we have a lot of "slow types" where it doesn't infer from the generic or is implicit. I'd solve these one by one rather than creating one big PR. The PR already adds the infrastructure in each workspace to publish to JSR by means of the `jsr.json` file. This also adds a new script during versioning where it will bump all the `jsr` packages.

A point for improvement that I'm considering in scope of this PR is a transformation script for our TS codebase where we transform all the `process.env.NODE_ENV` to `import.meta.env.NODE_ENV` or we strip it out. Not doing so will cut out deno support for a while, supporting both Deno as well as Node in JSR would require `globalThis.Deno?.env.get("NODE_ENV") || process.env.NODE_ENV`. I think with regards to https://github.com/graphql/graphql-js/issues/4075 it might be a good idea to transform our ESM bundles to `import.meta.env` sometime either way.

Currently we do not support canaries on jsr.

TODO

- [x] add dry run step to the PR so we know when slow types are added
- [x] release the current version by adding it to the PR workflow for a bit

### DIFF
--- a/packages/core/src/utils/request.ts
+++ b/packages/core/src/utils/request.ts
@@ -23,15 +23,21 @@ const SOURCE_NAME = 'gql';
 const GRAPHQL_STRING_RE = /("{3}[\s\S]*"{3}|"(?:\\.|[^"])*")/g;
 const REPLACE_CHAR_RE = /(?:#[^\n\r]+)?(?:[\r\n]+|$)/g;
 
-const replaceOutsideStrings = (str: string, idx: number) =>
+const replaceOutsideStrings = (str: string, idx: number): string =>
   idx % 2 === 0 ? str.replace(REPLACE_CHAR_RE, '\n') : str;
 
 /** Sanitizes a GraphQL document string by replacing comments and redundant newlines in it. */
 const sanitizeDocument = (node: string): string =>
   node.split(GRAPHQL_STRING_RE).map(replaceOutsideStrings).join('').trim();
 
-const prints = new Map<DocumentNode | DefinitionNode, string>();
-const docs = new Map<HashValue, KeyedDocumentNode>();
+const prints: Map<DocumentNode | DefinitionNode, string> = new Map<
+  DocumentNode | DefinitionNode,
+  string
+>();
+const docs: Map<HashValue, KeyedDocumentNode> = new Map<
+  HashValue,
+  KeyedDocumentNode
+>();
 
 /** A cached printing function for GraphQL documents.
  *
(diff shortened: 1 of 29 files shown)


packages/core/src/utils/request.ts
```diff
@@ -23,15 +23,21 @@ const SOURCE_NAME = 'gql';
 const GRAPHQL_STRING_RE = /("{3}[\s\S]*"{3}|"(?:\\.|[^"])*")/g;
 const REPLACE_CHAR_RE = /(?:#[^\n\r]+)?(?:[\r\n]+|$)/g;
 
-const replaceOutsideStrings = (str: string, idx: number) =>
+const replaceOutsideStrings = (str: string, idx: number): string =>
```
kitten, line 26: This, and annotations like `Map` below, seem a little concerning to me.
These are all internal functions and this seems to imply to me that JSR is enforcing a rule based on publishing raw TypeScript.

This may mean that we'd have to maintain a lot of changes across the code base for JSR and make a huge amount of changes. I think this would be reasonable if JSR/Deno would be the more popular platform over Node/npm, but it isn't.

We can implement an eslint rule for this (and there may even be an auto fixing one), but assuming declaration support is present in Deno, and we can publish prebuilts to JSR, wouldn't that be an easier way forward here?

If JSR doesn't allow this though, I think this seems a tad aggressive. Isolated declaration-like changes make sense to me, but changing all annotations to be statically analysable (except for functions where no assignment type is fine, but no return type isn't) seems a little aggressive 😅

JoviDeCroock: I've shared that feedback with them, not much more I can do

kitten: Gotcha! If we can't publish declarations yet, let's maybe publish this one package then figure out whether we want to do these changes for all packages.

Don't want to block this, but also want to make sure we don't have to make wide ranging changes across all packages of course 💙

kitten (approved):

packages/core/src/utils/request.ts
JoviDeCroock: So to handle publishing JS we have to add a reference to the JS file

```
- Point the exports path to the js file
- Add a /// <reference path="my-script.d.ts" /> line at the top of the JS file to include the types
```

PR #3582, opened 2024-05-03
fix(vue): Support getter-functions on `MaybeRef`-typed inputs

### LINKED ISSUE #3556: RFC: @urql/vue - Allow getter syntax in `useQuery`
## Summary

In Vue, if we use component props in the `variable` object of the `useQuery` composable, their reactivity is lost in the process and this results the same as if it was passed as a constant. To avoid this, the solution is to wrap the `variable` object in a `toRef`, using a getter function.

Allowing to pass directly the getter function to the `variable` property, as VueUse is doing with their [`MaybeRefOrGetter`](https://github.com/vueuse/vueuse/blob/main/packages/shared/utils/types.ts#L37) type, would be more fluent.

## Proposed Solution

Check if the passed `variable` arg is a function, and if yes, unwrap it when needed (see [VueUse `useFetch` composable](https://github.com/vueuse/vueuse/blob/1558cd2b5b019abc1feda6d702caa1053a182903/packages/core/useFetch/index.ts#L436) and [how they does it](https://github.com/vueuse/vueuse/blob/1558cd2b5b019abc1feda6d702caa1053a182903/packages/shared/toValue/index.ts#L9)?)

kitten:
Resolves #3556

## Summary

Supports getter functions on `MaybeRef`-typed inputs.

## Set of changes

- Add `(() => T)` to `MaybeRef` and handle in `unref`

### DIFF
--- a/packages/vue-urql/src/useQuery.ts
+++ b/packages/vue-urql/src/useQuery.ts
@@ -1,7 +1,7 @@
 /* eslint-disable react-hooks/rules-of-hooks */
 
 import type { WatchStopHandle, Ref } from 'vue';
-import { shallowRef, ref, watchEffect, reactive, isRef } from 'vue';
+import { shallowRef, ref, watchEffect, reactive } from 'vue';
 
 import type { Subscription, Source } from 'wonka';
 import { pipe, subscribe, onEnd } from 'wonka';
@@ -19,10 +19,8 @@ import type {
 import { createRequest } from '@urql/core';
 
 import { useClient } from './useClient';
-import { unwrapPossibleProxy, updateShallowRef } from './utils';
-
-type MaybeRef<T> = T | Ref<T>;
-type MaybeRefObj<T extends {}> = { [K in keyof T]: MaybeRef<T[K]> };
+import type { MaybeRef, MaybeRefObj } from './utils';
+import { unref, updateShallowRef } from './utils';
 
 /** Input arguments for the {@link useQuery} function.
  *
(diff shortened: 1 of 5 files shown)


JoviDeCroock (approved):

packages/vue-urql/src/useSubscription.ts
```diff
@@ -254,17 +252,11 @@ export function callUseSubscription<
   const operation: Ref<Operation<T, V> | undefined> = ref();
   const extensions: Ref<Record<string, any> | undefined> = ref();
 
-  const scanHandler: Ref<SubscriptionHandler<T, R> | undefined> = ref(handler);
-
-  const isPaused: Ref<boolean> = isRef(_args.pause)
-    ? _args.pause
-    : ref(!!_args.pause);
+  const scanHandler = ref(handler);
+  const isPaused = ref(!!unref(args.pause));
```
negezor, line 256: The reactivity for pause stopped working because of this change

negezor: Might be better to use toRef

negezor:
Maybe it's better to use [toRef](https://vuejs.org/api/reactivity-utilities#toref) instead of using unref?

packages/vue-urql/src/useQuery.ts
```diff
@@ -255,16 +253,11 @@ export function callUseQuery<T = any, V extends AnyVariables = AnyVariables>(
   const operation: Ref<Operation<T, V> | undefined> = ref();
   const extensions: Ref<Record<string, any> | undefined> = ref();
 
-  const isPaused: Ref<boolean> = isRef(_args.pause)
-    ? _args.pause
-    : ref(!!_args.pause);
+  const isPaused = ref(!!unref(args.pause));
```
negezor, line 256: Reactivity for pause is lost due to this change.

negezor: @kitten can you look at this? Since this is a breaking change. Since we now cannot work with pause from the hook itself, without calling pause() and resume() or modifying isPaused.

PR #3589, opened 2024-05-16
update docs on composeExchanges

aamir-mns:
The code and code-docs says its right to left (and tested it locally, it is right left) but the official docs says its left to right

Helpful links:

- https://github.com/urql-graphql/urql/blob/88bed21c015dc673d900bf1a806e43ccb2900879/packages/core/src/exchanges/compose.ts#L12
- https://github.com/urql-graphql/urql/commit/e6710d29ed426df9d21d9d9daa18ca182647db53
- https://commerce.nearform.com/open-source/urql/docs/api/core/#composeexchanges
- 

## Summary

I was using composeExchanges. Read official docs but it works in the opposite direction.

## Set of changes

Update docs

### DIFF
--- a/docs/api/core.md
+++ b/docs/api/core.md
@@ -511,7 +511,7 @@ This utility is used by the [`cacheExchange`](#cacheexchange) and by
 ### composeExchanges
 
 This utility accepts an array of `Exchange`s and composes them into a single one.
-It chains them in the order that they're given, left to right.
+It chains them in the order that they're given, right to left.
 
 ```js
 function composeExchanges(Exchange[]): Exchange;


kitten:
Both are correct, actually. Sorry about the confusion!
Let me explain.

The TSDocs state:

> [...] calling them from
> right to left, with the prior `Exchange`’s `ExchangeIO` function
> as the `ExchangeInput.forward` input.

The docs site states:

> It chains them in the order that they're given, left to right.

Both are correct.
The exchanges themselves are chained from left to right, meaning, the operations run through from left to right, just like the `exchanges` array on the `Client` are given (i.e. natural order, basically)

However, the TSDocs explain the **code** and what's going on there. What's going on in the code is that the exchanges are **called** from right to left, which composes them to be chained in the natural order, as given.
This is further complicated by the operation results running in reverse order basically, so I know it's confusing, but the docs **must** state left-to-right to clarify that that's the expected order

aamir-mns:
@kitten thanks for the quick reply. I got your point. I will close this PR but I think the docs not be a bit more clear as you said :)
thanks

aamir-mns:
@kitten Actually, I'm still confused. The docs say, 'It chains them in the order that they're given, left to right.' What does that mean? One would assume that it executes in the order they are provided, i.e., the first item will execute first and so on. By reading this, one can never assume that they will execute from right to left, i.e., the last will execute first and then it travels towards the left

JoviDeCroock:
It means that the requests that come in will be executed in the order of the exchanges left-to-right, and then return as a result from right-to-left

princekumar-d:
If I understand this correctly, is it something similar to this diagram?

image

If this is the case then maybe the docs could be updated to make it a bit more obvious?

kitten:
That's a great suggestion :+1:
Originally, this diagram served to explain this, but it isn't quite up-to-date and good enough anymore: https://urql.dev/goto/docs/architecture/#the-client-and-exchanges

Generally, we'll need to work on some documentation changes to get things back into shape

PR #3590, opened 2024-05-18
chore: change `global` -> `globalThis` in tests

marvinhagemeister:
## Summary

This is more of a maintenance change. The Node docs officially recommend to switch to `globalThis` instead of `global`, see https://nodejs.org/api/globals.html#global . `globalThis` is available in all other runtimes as well which makes the test suite easier to run outside of node. I'm working on getting vitest to run in Deno and I'm using urql's test suite for that.

## Set of changes

Replaces usages of `global` in the test suite to `globalThis`.


JoviDeCroock (approved):
Looks great, thank you!

PR #3595, opened 2024-05-25
fix(vue): Fix regression causing `pause` argument to not be reactive

arkandias:
## Summary

Solves https://github.com/urql-graphql/urql/issues/3594#issue-2316403034

## Set of changes

This is just a quick fix for getting isPaused reactive, but there might be some issues with the reactivity of other parameters.

### DIFF
--- a/.changeset/eight-mirrors-sing.md
+++ b/.changeset/eight-mirrors-sing.md
@@ -1,0 +1,5 @@
+---
+'@urql/vue': patch
+---
+
+Fix regression causing `pause` argument on `useQuery` and `useSubscription` to not be reactive
--- a/packages/vue-urql/src/useQuery.ts
+++ b/packages/vue-urql/src/useQuery.ts
@@ -1,6 +1,7 @@
 /* eslint-disable react-hooks/rules-of-hooks */
 
 import type { WatchStopHandle, Ref } from 'vue';
+import { isRef } from 'vue';
 import { shallowRef, ref, watchEffect, reactive } from 'vue';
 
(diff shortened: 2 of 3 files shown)


kitten:
Hiya :wave: We'll need a patch changeset here to merge this, i.e. a changelog entry, e.g.
> Fix regression causing `pause` argument on `useQuery` and `useSubscription` to not be reactive

I tried to quickly push one but push access for maintainers may not be enabled on your PR.
You can read more about this here though ❤️ https://github.com/urql-graphql/urql/blob/main/CONTRIBUTING.md#how-do-i-document-a-change-for-the-changelog

JoviDeCroock (approved):

negezor:
The type `MaybeRef<T> = T | (() => T) | Ref<T>;` is specified to support getters. 

This is not working as it should at the moment:
```ts
const skip = ref(0);

const pokemons = await useQuery({
  query: gql`
    query ($skip: Int!) {
      pokemons(limit: 10, skip: $skip) {
        id
        name
      }
    }
  `,
  variables: { skip },
  pause: () => {
    return skip.value > 1000 || skip.value === -1
  },
});
```

Of course this can be solved using manual `toRef()` or `computed()`. But maybe we can replace it with [toRef()](https://vuejs.org/api/reactivity-utilities#toref) by default? The only thing is that it will require vue 3.3+

arkandias:
Hi @negezor, I'm working on it. I'll make a PR very soon (with some other minor updates).

PR #3598, opened 2024-06-01
fix(vue): pause not working with getters

arkandias:
## Summary

This PR fixes a bug noted by @negezor: the `pause` property was not working with a getter, see https://github.com/urql-graphql/urql/pull/3595#issuecomment-2142285519.

## Set of changes

1) Minor fixes in `useQuery.ts` and `useSubscription.ts` to manage the case where pause is a getter.

2) Corrected some typos in the docs of `useQuery.ts` and `useSubscription.ts`.

3) Modified a test in `useQuery.test.ts` to use a reactive `pause` property with a getter.

### DIFF
--- a/packages/vue-urql/src/useSubscription.ts
+++ b/packages/vue-urql/src/useSubscription.ts
@@ -36,7 +36,7 @@ export type UseSubscriptionArgs<
    * @remarks
    * `pause` may be set to `true` to stop {@link useSubscription} from starting
    * its subscription automatically. This will pause the subscription until
-   * {@link UseSubscriptonState.resume} is called, or, if `pause` is a reactive
+   * {@link UseSubscriptionResponse.resume} is called, or, if `pause` is a reactive
    * ref of a boolean, until this ref changes to `true`.
    */
   pause?: MaybeRef<boolean>;
@@ -143,7 +143,7 @@ export interface UseSubscriptionResponse<
    *
    * @remarks
    * This is the subscription {@link Operation} that is currently active.
-   * When {@link UseQueryState.fetching} is `true`, this is the
+   * When {@link UseSubscriptionResponse.fetching} is `true`, this is the
    * last `Operation` that the current state was for.
    */
   operation: Ref<Operation<T, V> | undefined>;
--- a/packages/vue-urql/src/useSubscription.ts
+++ b/packages/vue-urql/src/useSubscription.ts
@@ -3,8 +3,8 @@
 import type { Source } from 'wonka';
 import { pipe, subscribe, onEnd } from 'wonka';
 
-import type { WatchStopHandle, Ref } from 'vue';
-import { reactive, ref, shallowRef, watch, watchEffect } from 'vue';
+import type { Ref, WatchStopHandle } from 'vue';
+import { isRef, reactive, ref, shallowRef, watch, watchEffect } from 'vue';
 
 import type {
   Client,
@@ -255,10 +255,9 @@ export function callUseSubscription<
 
   const scanHandler = ref(handler);
   const isPaused: Ref<boolean> = ref(!!unref(args.pause));
-  watch(
-    () => !!unref(args.pause),
-    value => (isPaused.value = value)
-  );
+  if (isRef(args.pause) || typeof args.pause === 'function') {
+    watch(args.pause, value => (isPaused.value = value));
+  }
 
   const input = shallowRef({
     request: createRequest<T, V>(unref(args.query), unref(args.variables) as V),
--- a/packages/vue-urql/src/useSubscription.ts
+++ b/packages/vue-urql/src/useSubscription.ts
@@ -4,8 +4,7 @@ import type { Source } from 'wonka';
 import { pipe, subscribe, onEnd } from 'wonka';
 
 import type { WatchStopHandle, Ref } from 'vue';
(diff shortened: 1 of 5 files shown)


packages/vue-urql/package.json
```diff
@@ -52,7 +52,7 @@
     "@urql/core": "workspace:*",
     "@vue/test-utils": "^2.3.0",
     "graphql": "^16.0.0",
-    "vue": "^3.2.47"
+    "vue": "^3.4.27"
```
JoviDeCroock, line 55: This is kind of a breaking change if we require folks to be above 3.4 as our peerDeps hint at > 3

arkandias: Oh right, sorry I was not very careful about that. I will make another PR to resolve the problem with pause only (along with the test and the small typos in the doc).

JoviDeCroock: Can you do it in this PR so we don't merge a breaking change

pnpm-lock.yaml
JoviDeCroock, line 1: This diff looks suspicious 😅

packages/vue-urql/package.json
arkandias: Sure (I meant I would make a PR -- or correct this one -- just to correct the issue with pause and leave the update to Vue 3.4).

packages/vue-urql/src/useSubscription.ts
```diff
@@ -254,9 +254,12 @@ export function callUseSubscription<
   const extensions: Ref<Record<string, any> | undefined> = ref();
 
   const scanHandler = ref(handler);
-  const isPaused: Ref<boolean> = isRef(_args.pause)
-    ? _args.pause
-    : ref(!!_args.pause);
+  const isPaused: Ref<boolean> =
+    typeof _args.pause === 'function'
+      ? computed(_args.pause)
```
kitten, line 259: Does `computed` give us a mutable ref that can still be modified after this conversion by `pause()` and `resume()`?

negezor: @kitten no. Also computed will give additional overhead unnecessarily. Why don't we just do ref + watch?

arkandias: @kitten You're right, sorry for not being careful about that!

@negezor are you suggesting the following?
```
const isPaused: Ref<boolean> = ref(!!unref(args.pause));
watch(
  () => !!unref(args.pause),
  value => (isPaused.value = value)
);
```
Should we add the watcher to `stops` then?

Another possibility is to compute the value of `isPaused` from `args.pause`
```
const isPaused: Ref<boolean> = computed(() => !!unref(args.pause));
```
and have the latter updated by `pause` and `resume`:
```
pause() {
  args.pause = true;
},
resume() {
  args.pause = false;
},
```
Not sure which one is better...

arkandias: I pushed the ref + watch option. I guess the watcher should be added to `stops`, but I'm not sure yet.
Actually I do not understand why stops are being called (before unmounting) with `useClientHandle()` but not with `useClient()`...

negezor: We can do a little optimization for watch, what do you think? It wouldn't require an unref in watch.
```ts
if (isRef(args.pause) || typeof args.pause === 'function') {
    watch(args.pause, value => (isPaused.value = value));
}
```

arkandias: You're right! What about adding it to the stops?

negezor: It makes sense to add it to stops since it is used for `useClientHandle()`
https://github.com/urql-graphql/urql/blob/d95baf18d34e6337ecc9f7c9ce372028308c1b7b/packages/vue-urql/src/useClientHandle.ts#L130-L146

negezor: @kitten maybe we can remove stops from the parameters and use [effectScope](https://vuejs.org/api/reactivity-advanced.html#effectscope)? This will eliminate the unnecessary need to store the stops array for most cases, which will have a positive impact on performance, as we generally do not use `useClientHandle()`.

arkandias: Thank you @negezor, I added the watchers to `stops`. Still, I do not understand: Why is it needed since all watchers are called synchronously? And why with useClientHandle specifically?

arkandias:
@kitten I think the PR is ready now. It reverts the changes from my previous PR (that restored `pause` reactivity but broke the usage of a getter) and added a watcher instead (as suggested by @negezor). Let me know if I can do anything.

kitten (approved):

kitten:
Can you add a changeset to this please? Guidance for this can be found here: https://github.com/urql-graphql/urql/blob/main/CONTRIBUTING.md#how-do-i-document-a-change-for-the-changelog

arkandias:
> Can you add a changeset to this please? Guidance for this can be found here: https://github.com/urql-graphql/urql/blob/main/CONTRIBUTING.md#how-do-i-document-a-change-for-the-changelog

I pushed a change log following the instructions. Let me know if you need something else.

JoviDeCroock (approved):

arkandias:
Thank you!

PR #3605, opened 2024-06-06
fix(vue): Fix regression causing variables to not be reactive

yurks:
## Summary

Restore reactivity typings for `variables` input.

This regression caused a [number](https://stackoverflow.com/questions/76163956/how-to-allow-urql-typescript-to-accept-vue-reactive-variables-for-queries-crea) of [issues](https://github.com/dotansimha/graphql-code-generator-community/issues/194) using `graphql-code-generator`

## Set of changes

Reuse `MaybeRef` type for variables typing.

### DIFF
--- a/packages/vue-urql/src/useQuery.ts
+++ b/packages/vue-urql/src/useQuery.ts
@@ -75,7 +75,7 @@ export type UseQueryArgs<
    * documentation on the `pause` option.
    */
   pause?: MaybeRef<boolean>;
-} & MaybeRefObj<GraphQLRequestParams<Data, Variables>>;
+} & MaybeRefObj<GraphQLRequestParams<Data, MaybeRefObj<Variables>>>;
 
 /** State of the current query, your {@link useQuery} function is executing.
  *
--- a/packages/vue-urql/src/useSubscription.ts
+++ b/packages/vue-urql/src/useSubscription.ts
@@ -58,7 +58,7 @@ export type UseSubscriptionArgs<
    * ```
    */
   context?: MaybeRef<Partial<OperationContext>>;
-} & MaybeRefObj<GraphQLRequestParams<Data, Variables>>;
+} & MaybeRefObj<GraphQLRequestParams<Data, MaybeRefObj<Variables>>>;
 
 /** Combines previous data with an incoming subscription result’s data.
  *
--- a/packages/vue-urql/src/utils.ts
+++ b/packages/vue-urql/src/utils.ts
@@ -3,7 +3,9 @@ import type { Ref, ShallowRef } from 'vue';
 import { isRef } from 'vue';
 
 export type MaybeRef<T> = T | (() => T) | Ref<T>;
-export type MaybeRefObj<T extends {}> = { [K in keyof T]: MaybeRef<T[K]> };
+export type MaybeRefObj<T> = T extends {}
+  ? { [K in keyof T]: MaybeRef<T[K]> }
+  : T;
 
 export const unref = <T>(maybeRef: MaybeRef<T>): T =>
   typeof maybeRef === 'function'


JoviDeCroock (approved):

JoviDeCroock:
Darn I missed that this doesn't have a changeset, do you mind adding one or should I?

yurks:
> Darn I missed that this doesn't have a changeset, do you mind adding one or should I? More info [here](https://github.com/urql-graphql/urql/blob/main/CONTRIBUTING.md#what-are-the-pr-conventions) or the bot comment above

thanks, will do within another PR

JoviDeCroock:
Pushed a changeset so I can add it to a release @yurks

yurks:
@JoviDeCroock just prepared the same, but too late) Thanks!

negezor:
Now it breaks down queries that don't have variables:
```ts
error TS2345: Argument of type '{}' is not assignable to parameter of type 'Omit<UseQueryArgs<never, Exact<{ [key: string]: never; }>>, "query">'.
  Property 'variables' is missing in type '{}' but required in type 'Omit<UseQueryArgs<never, Exact<{ [key: string]: never; }>>, "query">'.

29 const { data } = useFetchUserProvidersQuery({});
                                               ~~

  node_modules/@urql/core/dist/urql-core-chunk.d.ts:848:5
    848     variables: Variables;
            ~~~~~~~~~
    'variables' is declared here.
```

yurks:
@negezor you are using codegen, and that's a typing issue in `@graphql-codegen/typescript-vue-urql`. There was an attempt to fix typings according `urql-vue` types refactor someday, but it was done incorrectly, as variables lost reactivity.

If you can try to `urql-vue` methods directly, you'll see empty variables working as expected.

A bit more mess: based on your example you're using `useFetchUserProvidersQuery({})`, where `{}` is kind of workaround. With `urql-vue@1.0.4` it worked just like `useFetchUserProvidersQuery()`, which is correct.

Looks like we are using the same stack, so I can recommend you temporary resolve this issue with `useFetchUserProvidersQuery({ variables: {}})`, until
`@graphql-codegen/typescript-vue-urql` will be fixed.

yurks:
> @negezor you are using codegen, and that's a typing issue in `@graphql-codegen/typescript-vue-urql`. There was [an attempt](https://github.com/dotansimha/graphql-code-generator-community/pull/494) to fix typings according `urql-vue` types refactor someday, but it was done incorrectly, as variables lost reactivity.
> 
> If you can try to use `urql-vue` methods directly, you'll see empty variables working as expected.
> 
> A bit more mess: based on your example you're using `useFetchUserProvidersQuery({})`, where `{}` is kind of workaround. With `urql-vue@1.0.4` it worked just like `useFetchUserProvidersQuery()`, which is correct.
> 
> Looks like we are using the same stack, so I can recommend you temporary resolve this issue with `useFetchUserProvidersQuery({ variables: {}})`, until `@graphql-codegen/typescript-vue-urql` will be fixed.



> Now it breaks down queries that don't have variables:
> 
> ```ts
> error TS2345: Argument of type '{}' is not assignable to parameter of type 'Omit<UseQueryArgs<never, Exact<{ [key: string]: never; }>>, "query">'.
>   Property 'variables' is missing in type '{}' but required in type 'Omit<UseQueryArgs<never, Exact<{ [key: string]: never; }>>, "query">'.
> 
> 29 const { data } = useFetchUserProvidersQuery({});
>                                                ~~
> 
>   node_modules/@urql/core/dist/urql-core-chunk.d.ts:848:5
>     848     variables: Variables;
>             ~~~~~~~~~
>     'variables' is declared here.
> ```

@negezor https://github.com/dotansimha/graphql-code-generator-community/pull/801

arkandias:
@yurks It seems that the modification of `MaybeRefObj` in `utils.ts` made `useQuery` completely loose variables typing, see [this issue](https://github.com/urql-graphql/urql/issues/3733). Do you remember why you did this modification, and if it can be reverted or otherwise?

arkandias:
> @negezor you are using codegen, and that's a typing issue in `@graphql-codegen/typescript-vue-urql`. There was [an attempt](https://github.com/dotansimha/graphql-code-generator-community/pull/494) to fix typings according `urql-vue` types refactor someday, but it was done incorrectly, as variables lost reactivity.
> 
> If you can try to use `urql-vue` methods directly, you'll see empty variables working as expected.
> 
> A bit more mess: based on your example you're using `useFetchUserProvidersQuery({})`, where `{}` is kind of workaround. With `urql-vue@1.0.4` it worked just like `useFetchUserProvidersQuery()`, which is correct.
> 
> Looks like we are using the same stack, so I can recommend you temporary resolve this issue with `useFetchUserProvidersQuery({ variables: {}})`, until `@graphql-codegen/typescript-vue-urql` will be fixed.

@negezor @yurks Actually I think the problem comes from `MaybeRefObj` because `Exact<{ [key: string]: never; }>` extends `{}`. The definition of `GraphQLRequestParams` takes care to make `variables` optional in certain cases (though I have some questions [here](https://github.com/urql-graphql/urql/issues/3733#issuecomment-2578786229)), but wrapping it into `MaybeRefObj` breaks this behavior.

This could be fixed by something like:
```typescript
type MaybeRefObj<T> = T extends Record<string, never> ? T
  : T extends {} ? { [K in keyof T]: MaybeRef<T[K]> }
    : T;
```

I'll investigate and maybe add this to PR #3734

PR #3607, opened 2024-06-09
Add SolidJS integration

stefanmaric:
This is an attempt to fix the setup issues encountered by @Nvos on https://github.com/urql-graphql/urql/pull/3327

### DIFF
--- a/.changeset/forty-cameras-check.md
+++ b/.changeset/forty-cameras-check.md
@@ -1,0 +1,5 @@
+---
+'@urql/solid': minor
+---
+
+Initial release
--- a/exchanges/auth/package.json
+++ b/exchanges/auth/package.json
@@ -40,7 +40,7 @@
     "dist/"
   ],
   "scripts": {
-    "test": "vitest --config ../../vitest.config.ts",
+    "test": "vitest",
     "clean": "rimraf dist extras",
     "check": "tsc --noEmit",
     "lint": "eslint --ext=js,jsx,ts,tsx .",
@@ -57,7 +57,8 @@
   },
   "devDependencies": {
     "@urql/core": "workspace:*",
-    "graphql": "^16.0.0"
+    "graphql": "^16.0.0",
+    "vitest": "^0.30.1"
   },
   "publishConfig": {
     "access": "public",
--- a/exchanges/auth/vitest.config.ts
+++ b/exchanges/auth/vitest.config.ts
@@ -1,0 +1,4 @@
+import { mergeConfig } from 'vitest/config';
+import baseConfig from '../../vitest.config';
+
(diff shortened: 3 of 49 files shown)


XiNiHa:
Are there any blockers for reviewing and merging the PR? I was also interested in creating the Solid integration and found this PR waiting reviews for months 😢 I'm open to both code contributing and experimenting in a real project (I guess it'll be more convenient to try this if we have [pkg.pr.new](https://github.com/stackblitz-labs/pkg.pr.new) set up?)

JoviDeCroock:
The main issue with this PR is what is described, it requires so many changes to our workspace setup. I'll try to free up time to look at alternatives/... but it comes down to the same thing as when I initially tried to get solid setup, the testing story in solid seems lacking.

XiNiHa:
Since it looks like it's the JSX transform that causes the issue, how about using [tagged template literals](https://github.com/solidjs/solid/tree/e4b2c668096c47a99e815ad202cb27dfa35f2ec5/packages/solid/html) or [HyperScript API](https://github.com/solidjs/solid/tree/e4b2c668096c47a99e815ad202cb27dfa35f2ec5/packages/solid/h) instead of JSX? They exist for supporting no-build scenarios and I think it matches pretty well with what we want 🤔

JoviDeCroock:
@XiNiHa do you want to pick that up? Reviving https://github.com/urql-graphql/urql/pull/3327 with either of those two solutions

XiNiHa:
Sure! I'll make a PR shortly 😋

XiNiHa:
Well I got blocked by some issues:
- Both tagged template literals and HyperScript wrappers are broken 🤦‍♂️ so it'll be the only way to put precompiled code directly in the test code if we really want to have no-build
- Solid Testing Library requires JSDOM environments to run, and this breaks one of the current persistent exchange tests in the repo
  - Using JSDOM everywhere will also harm performance and make SSR tests unreliable
- Solid requires `resolve.conditions` in the Vite config to be set as `["development", "browser"]` at least and this breaks Preact Testing Library 😵‍💫
After going through these, I just felt that it's natural to have different test environments for each package, as they could run on entirely different environments. Imagine supporting React Server components and bringing all the bundling stuffs; maybe breaking up the config would be the only option there. So I'd like to suggest to go with what we have in this PR, and make it as an opportunity to microtune each test environment per package to reflect the actual environment the package would run. Thoughts?

JoviDeCroock:
@XiNiHa I'll merge this branch as is and we can look at [vitest workspaces](https://vitest.dev/guide/workspace.html) as an optimisation to reduce the configs

JoviDeCroock (approved):

PR #3608, opened 2024-06-10
docs: Add `urql-exhaustive-additional-typenames-exchange` to showcase

MH4GF:
## Summary

I have been using the exchange referred to here https://github.com/urql-graphql/urql/discussions/3440#discussioncomment-8768504 for several months and it's useful, I would like to add it to the documentation.


MH4GF:
@kitten thanks!!

PR #3610, opened 2024-06-18
Use effectScope instead of manually stopping effects

negezor:
## Summary

This PR eliminates the need to manually stop effects, as Vue has an efficient `effectScope` API. Manual stopping is less readable and consumes more memory unnecessarily, since Vue will still write the effects to scope.

## Set of changes

- Removes `stops` in `useQuery.ts` and `useSubscription.ts`
- Replaces `stops` with `effectScope` in `useClientHandle.ts`

This should be a minor change. However, for most users this won't matter because everything basically uses methods directly instead of `useClientHandle`, which is much more efficient.

### DIFF
--- a/.changeset/kind-bugs-do.md
+++ b/.changeset/kind-bugs-do.md
@@ -1,0 +1,5 @@
+---
+'@urql/vue': minor
+---
+
+Use `effectScope` instead of manually stopping effects
--- a/packages/vue-urql/src/useClientHandle.ts
+++ b/packages/vue-urql/src/useClientHandle.ts
@@ -1,7 +1,11 @@
 import type { DocumentNode } from 'graphql';
 import type { AnyVariables, Client, TypedDocumentNode } from '@urql/core';
-import type { WatchStopHandle } from 'vue';
-import { getCurrentInstance, onMounted, onBeforeUnmount } from 'vue';
+import {
+  getCurrentInstance,
+  onMounted,
+  onBeforeUnmount,
+  effectScope,
+} from 'vue';
 
 import { useClient } from './useClient';
 
@@ -129,11 +133,11 @@ export interface ClientHandle {
  */
 export function useClientHandle(): ClientHandle {
   const client = useClient();
-  const stops: WatchStopHandle[] = [];
+
+  const scope = effectScope();
 
   onBeforeUnmount(() => {
-    let stop: WatchStopHandle | void;
-    while ((stop = stops.shift())) stop();
+    scope.stop();
   });
 
   const handle: ClientHandle = {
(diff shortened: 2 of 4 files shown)


JoviDeCroock:
Adding this would break our compatibility with older Vue versions https://github.com/vuejs/core/blob/main/changelogs/CHANGELOG-3.2.md#reactivity - we have to stick to the minimum peerDep version described in our package.json or it will be a breaking change.

negezor:
Indeed, I didn't notice. Perhaps a major change might be worth considering then? For example, a router for Vue 3 asks for a minimum of vue@3.2.0
https://github.com/vuejs/router/blob/249bf2da5382758fc55a6690c4179a63313f0f24/packages/router/package.json#L111

At the same time, Vue 2 has already reached its EOL https://v2.vuejs.org/eol/

A good option would be to install the minimum version 3.3. We could use native `toRef` and `unref`

kitten:
@yurks I don't want to sound too harsh, but please don't comment assuming that something that's been required is there for no reason 😅❤️
This is relevant for `useClientHandle`, which, due to its API absolutely has the ability to set up subscriptions that don't happen during mounting and hence aren't cleaned up automatically https://github.com/urql-graphql/urql/blob/1bb05f58d91fee9c20368bd808c4832ead318179/packages/vue-urql/src/useClientHandle.ts#L134

yurks:
@kitten 
I have no doubt that there was a strong reason for using this particular code. But I can see also `useClient()` in deeper usage, which is using `getCurrentInstance()` and `inject()` inside, which in turn means, this couldn't be used outside component. Also, there is a comment:

> In development, if `useClient` is called outside of a Vue `setup` function or no {@link Client} was provided, an error will be thrown.

Based on the all above, I can conclude that it's impossible to use this method outside vue context and leve a comment here. Am I wrong?

kitten:
Feel free to browse past issues and discussions regarding this topic. The function instantiates in component setup functions, but is then meant to assist with contexts that are outside of components and Vue's tracking contexts after. This in fact becomes pretty apparent when you try to call `useQuery` without this in more contexts. You'll see that it straight up errors. You'll also find that without taking care of stop conditions, the subscriptions remain active indefinitely

yurks:
I may be wrong and I would appreciate it if you could help me realize this, but based on current code watchers are **always** connected to component instance. How it possible to stay active, as unwathing should be handled internally by vue? I can see only one possible reason, based on date this code parts was authored - it was developed with beta version of vue 3, where was a number of similar issues. 

PS: I just did a [little refactoring](https://github.com/urql-graphql/urql/pull/3619) with care of reactivity handling, and propose to remove `flush: 'pre'` for watchers there, as it's default setting. But having in mind, nothing is accidental here, I can assume the [beta vue3](https://github.com/vuejs/docs/issues/2350#issuecomment-1549943418) was used in development stage.

yurks:
@kitten I found your [use case](https://github.com/urql-graphql/urql/issues/1596) - handling several awaits when calling several `await useQuery()`:
```ts
 const handle = useClientHandle();
 const resultOne = await handle.useQuery(/* ... */);
 const resultTwo = await handle.useQuery(/* ... */);
```
But it looks like incorrect code pattern, as it's known vue issue and even vue lint [trying](https://eslint.vuejs.org/rules/no-lifecycle-after-await.html) to be aware of it. 

It's ok to handle such cases manually for sure, but I think it perpetuates the misuse of the framework. I would suggest documenting this feature and propose correct pattern for such cases like:

```ts
const resultOne = useQuery(/* ... */);
const resultTwo = useQuery(/* ... */);

// sequential execution
await resultOne;
await resultTwo;

// or parallel execution
await Promise.all([resultOne, resultTwo])
```

negezor:
Closed in favor of #3619

PR #3612, opened 2024-06-18
Remove wrapping request args in reactive to fix memory leak

negezor:
## Summary

This eliminates the memory leak since reactive always assigned new `watchEffects` to the static request object.

Resolves: https://github.com/urql-graphql/urql/issues/3507#issuecomment-2174017806

Memory leak appeared in this PR https://github.com/urql-graphql/urql/pull/1151

## Set of changes

- Removes wrapping `args` in `reactive` in `useQuery.ts` and `useSubscription.ts`

This change is not breaking, since we still do `unref` as before #1151 PR.

### DIFF
--- a/.changeset/quick-pens-battle.md
+++ b/.changeset/quick-pens-battle.md
@@ -1,0 +1,5 @@
+---
+'@urql/vue': minor
+---
+
+Remove wrapping request `args` in `reactive` to fix memory leak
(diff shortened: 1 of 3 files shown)


JoviDeCroock (approved):

JoviDeCroock:
@yurks mate chill... first and foremost you are looking at the wrong PR and a tone like that when not adding tests/... is really not appreciated

yurks:
(how you can identify the tone by text question? PR was approved and immediately released)

that's not about types but about reactivity.

PR #3619, opened 2024-06-18
feat(vue): refactor composable functions

yurks:
Resolves https://github.com/urql-graphql/urql/issues/3507
Resolves https://github.com/urql-graphql/urql/issues/3633

## Summary

This includes https://github.com/urql-graphql/urql/pull/3612, but fixed reactivity loosing here.

In addition, a little refactoring for `use*` composable functions for reducing code duplication and adding more test cases.

## Set of changes

* Remove wrapping args in reactive in useQuery.ts and useSubscription.ts
* Add `createRequestWithArgs()`, `useClientState()` and `useRequestState()` utils to avoid code duplication.
* Refactor `useQuery()`, `useSubscription()` and `useMutation()` to use shared  utils mentioned above.
* Add `isReadonly()` check before changing `isPaused` variable, as it could be readonly.
* ~~Remove `onBeforeUnmount` and unwatch handlers, as [watcher automatically stopped when the owner component is unmounted](https://vuejs.org/guide/essentials/watchers.html#stopping-a-watcher)~~
* Remove `flush: 'pre'` in watcher options, as this is default setting
* Rename `unref()` to `unwrap()` to avoid confusions with vue function
* Reuse `then()` handler for `useQuery()` for cleaner code and avoid duplications
* Cover of reactive arguments usage with test cases

### DIFF
--- a/.changeset/chatty-mice-join.md
+++ b/.changeset/chatty-mice-join.md
@@ -1,0 +1,5 @@
+---
+'@urql/vue': minor
+---
+
+Refactor composable functions with care of reactivity to fix possible memory leaks
--- a/.changeset/chatty-mice-join.md
+++ b/.changeset/chatty-mice-join.md
@@ -2,4 +2,4 @@
 '@urql/vue': minor
 ---
 
-Refactor composable functions with care of reactivity to fix possible memory leaks
+Refactor composable functions with a focus on avoiding memory leaks and Vue best practices
--- a/packages/vue-urql/src/useQuery.test.ts
+++ b/packages/vue-urql/src/useQuery.test.ts
@@ -1,5 +1,9 @@
-import { OperationResult, OperationResultSource } from '@urql/core';
-import { nextTick, reactive, ref } from 'vue';
+import {
+  OperationResult,
+  OperationResultSource,
+  RequestPolicy,
+} from '@urql/core';
+import { computed, nextTick, reactive, ref } from 'vue';
 import { vi, expect, it, describe } from 'vitest';
 
 vi.mock('./useClient.ts', async () => ({
@@ -10,9 +14,27 @@ vi.mock('./useClient.ts', async () => ({
 
 import { pipe, makeSubject, fromValue, delay } from 'wonka';
 import { createClient } from '@urql/core';
-import { useQuery } from './useQuery';
+import { useQuery, UseQueryArgs } from './useQuery';
 
 const client = createClient({ url: '/graphql', exchanges: [] });
+
+const createQuery = (args: UseQueryArgs) => {
+  const executeQuery = vi
+    .spyOn(client, 'executeQuery')
+    .mockImplementation(request => {
+      return pipe(
+        fromValue({ operation: request, data: { test: true } }),
+        delay(1)
+      ) as any;
+    });
+
+  const query$ = useQuery(args);
+
+  return {
+    query$,
+    executeQuery,
+  };
+};
 
 describe('useQuery', () => {
   it('runs a query and updates data', async () => {
@@ -79,18 +101,9 @@ describe('useQuery', () => {
   });
 
   it('runs queries as a promise-like that resolves even when the query changes', async () => {
-    const executeQuery = vi
-      .spyOn(client, 'executeQuery')
-      .mockImplementation(request => {
-        return pipe(
-          fromValue({ operation: request, data: { test: true } }),
-          delay(1)
-        ) as any;
-      });
-
     const doc = ref('{ test }');
 
-    const query$ = useQuery({
+    const { executeQuery, query$ } = createQuery({
       query: doc,
     });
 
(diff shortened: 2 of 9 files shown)


negezor:
This is a really good refactoring. The only thing that bothers me is why the effects in `useClientHandle()` were initially manually stopped; I couldn’t find the reason (or I didn’t look well).

yurks:
Thanks @negezor ☀️

> why the effects in `useClientHandle()` were initially manually stopped

The reason of having `useClientHandle()` and manually stopping watchers I found is here https://github.com/urql-graphql/urql/pull/3610#issuecomment-2178843530. And the reason it was removed in current PR - I consider this as excessive concern for rare use cases and attempt to bypass vue limitations, which should be completely on users responsibility imho.

But I can agree this kind of removal is overkill for mine "little refactoring", so I'll revert this functionality here.
I guess we can found an optimal solution in scope of [that PR](https://github.com/urql-graphql/urql/pull/3610).

negezor:
@kitten could you please review the PR?

packages/vue-urql/src/utils.ts
```diff
@@ -1,45 +1,142 @@
...
-    if (ref.value[key] !== next[key]) {
-      ref.value = next;
-      return;
+  let vars = unwrap(args.variables);
+  // unwrap possible nested reactive variables with `readonly()`
+  for (const prop in vars) {
+    if (Object.hasOwn(vars, prop)) {
+      const value = vars[prop];
+      if (value && (typeof value === 'object' || isRef(value))) {
+        vars = readonly(vars);
```
negezor, line 52: Is it really necessary to wrap variables in readonly every time? I think we could avoid the additional overhead.

yurks:
Not every time, but only in case it contains some ref deeply. For this case
it will be "unwrapped" with `readonly()` and it all will be "listen" inside
computed var.

```
{
  variables: {
     somevar: ref()
  }
}
```

packages/vue-urql/src/utils.ts
yurks: @negezor
Not every time, but only in case it contains some ref deeply. For this case it will be "unwrapped" with `readonly()` and it all will be "listen" inside computed var.

```
{
  variables: {
     somevar: ref()
  }
}
```

yurks: In comparing to original implementation (where all arguments were passed to `reactive()`), unwrapping deeply nested variables with `readonly()` shouldn't create a lot of additional overhead, as complex variables object is already overhead ;)

yurks: @negezor that piece of code has been rewritten to not use `readonly()` but simply iterate over the object and unwrap possible refs

negezor:
@kitten gentle ping 👀

packages/vue-urql/src/utils.ts
```diff
@@ -1,45 +1,150 @@
...
-  next: T
-) => {
-  for (const key in next) {
-    if (ref.value[key] !== next[key]) {
-      ref.value = next;
-      return;
+  if (input && typeof input === 'object') {
+    const isArray = Array.isArray(input);
+    const out = (isArray ? [] : {}) as T;
+    for (const prop in input) {
```
negezor, line 39:
```ts
if (input[prop] instanceof Blob) {
  out[prop] = input[prop];
  continue;
}
```

packages/vue-urql/src/utils.ts
```diff
@@ -1,45 +1,150 @@
...
+      if (
+        isArray ||
+        (Object.hasOwn || Object.prototype.hasOwnProperty.call)(input, prop)
```
negezor, line 42: This polyfill will fail because `.call()` is a virtual method.
```
> (false || Object.prototype.hasOwnProperty.call)({}, 'test')

TypeError: (false || Object.prototype.hasOwnProperty.call) is not a function
    at <anonymous>:1:48
```

negezor: I think a good option would be to basically remove `Object.hasOwn` for now. Plus it is slower according to this benchmark https://jsperf.app/labega/2.
image

packages/vue-urql/src/utils.ts
```diff
@@ -26,6 +26,12 @@ const unwrap = <T>(maybeRef: MaybeRef<T>): T =>
     ? maybeRef.value
     : maybeRef;
 
+const _toString = Object.prototype.toString;
+const isPlainObject = (obj: any): boolean => {
```
JoviDeCroock, line 30: Ï think we might do better in re-using/copying our existing checks https://github.com/urql-graphql/urql/blob/main/packages/core/src/utils/variables.ts#L23-L24

.changeset/chatty-mice-join.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/vue': minor
+---
+
+Refactor composable functions with care of reactivity to fix possible memory leaks
```
JoviDeCroock, line 5:
```suggestion
Refactor composable functions with a focus on avoiding memory leaks and Vue best practices
```

yurks:
> @yurks I deployed these changes to production, found one problem. Uploading files does not work in mutations, it sends `{}` instead of `File`

@negezor thanks for the review! I optimised that piece of code, and added test case for checking different types of variables.

packages/vue-urql/src/utils.ts
yurks: thanks, removed

packages/vue-urql/src/utils.ts
yurks: fixed by rewrite

packages/vue-urql/src/utils.ts
```diff
@@ -1,45 +1,168 @@
...
+    if (isArray(input)) {
+      const length = input.length;
+      const out = new Array(length) as T;
+      let i = 0;
+      for (; i < length; i++) {
+        out[i] = unwrapDeeply(input[i]);
+      }
+
+      return out;
+    } else if (isPlainObject(input)) {
```
JoviDeCroock, line 52: We'll need a similar check to `Object.getPrototypeOf(x).constructor !== Object.prototype.constructor` so we don't clobber built-in classes like `Date` might be good to add a test for that

.changeset/chatty-mice-join.md
yurks: thanks, updated

packages/vue-urql/src/utils.ts
yurks: `Date.toString() !== Object.toString()`, so this condition would be false.
I checked a couple of code patterns for `isPlainObject()` implementation, and found [vue way](https://github.com/vuejs/vue/blob/main/src/shared/util.ts) reasonable. Wdyt?

packages/vue-urql/src/utils.ts
yurks: yes, that checks is for Files/Blobs only. I just think this might be overkill here, as finally urql/core is used with its internal validations.
What I am trying to do here - is just unwrap possible refs and pass everything as it is. That's why I found `isPlainObject()` way more generic: urql/core could be kind of "source of truth" for input validation

packages/vue-urql/src/useQuery.test.ts
```diff
@@ -148,23 +363,86 @@ describe('useQuery', () => {
...
 
     subject.next({ data: { test: true } });
 
-    expect(query.fetching).toBe(false);
-    expect(query.data).toEqual({ test: true });
+    expect(query.fetching.value).toBe(false);
+    expect(query.data.value).toHaveProperty('test', true);
+  });
+
+  it('pauses query with ref variable', async () => {
```
JoviDeCroock, line 383: looks like this resolves https://github.com/urql-graphql/urql/issues/3633

yurks: it is

packages/vue-urql/src/utils.ts
yurks: added, thanks!

packages/vue-urql/src/utils.ts
```diff
@@ -1,45 +1,175 @@
...
-export function createRequestState<
-  Data = any,
-  Variables extends AnyVariables = AnyVariables,
->(
-  request: GraphQLRequest<Data, Variables>,
-  isPaused: boolean
-): RequestState<Data, Variables> {
-  return { request, isPaused };
-}
+  if (_toString.call(value) !== '[object Object]') return false;
```
JoviDeCroock, line 33:
```suggestion
```

yurks: done

JoviDeCroock (approved):

PR #3632, opened 2024-07-15
perf(core): remove duplicate `JSON.stringify` call on `data` in `ssrExchange`

negezor:
## Summary

Removes the unnecessary `JSON.stringify` when serializing, since it will be called below anyway

image

## Set of changes

Removing the `data` pre-property in exchange/ssr during serialization

### DIFF
--- a/packages/core/src/exchanges/ssr.ts
+++ b/packages/core/src/exchanges/ssr.ts
@@ -109,7 +109,6 @@ const serializeResult = (
   includeExtensions: boolean
 ): SerializedResult => {
   const serialized: SerializedResult = {
-    data: JSON.stringify(result.data),
     hasNext: result.hasNext,
   };
 
--- a/.changeset/sweet-goats-judge.md
+++ b/.changeset/sweet-goats-judge.md
@@ -1,0 +1,5 @@
+---
+'@urql/core': patch
+---
+
+Removes double serialization of `data` in `ssrExchange`


packages/core/src/exchanges/ssr.ts
```diff
@@ -109,7 +109,6 @@ const serializeResult = (
   includeExtensions: boolean
 ): SerializedResult => {
   const serialized: SerializedResult = {
-    data: JSON.stringify(result.data),
```
JoviDeCroock, line 112: I'd prefer removing the `if` branch then as `JSON.stringify(undefined)` still has a result.

negezor: On the other hand, calling `JSON.stringify` with undefined is slower
https://jsperf.app/vuwofo

JoviDeCroock: Yes, but the property missing isn't really according to spec so we might even want to default that to null https://graphql.github.io/graphql-over-http/draft/#sec-application-json

negezor: The `SerializedResult` type allows the data to be a `string` only or `undefined`. `deserializeResult` will return `undefined` if there is no data.
https://github.com/urql-graphql/urql/blob/6406fd5fa30180648439e9f9ba9f18caf16e2497/packages/core/src/exchanges/ssr.ts#L154

Looking at the implementation closer, I'd like to use something more efficient than `JSON.stringify`, like [devalue](https://github.com/Rich-Harris/devalue). Because I can reduce 2mb JSON to 1.1MB, I just checked this out

kitten:
Haven't looked at this closely yet, but just to point this out ahead of time, this may necessitate a major, or at least a minor bump. We can't be sure of how this is used and hence we can't guarantee that all apps that have been built ensure that the versions of serialisation are the same in the SSR process and on the client-side.

Going for backwards compatibility here would also work.

negezor:
It doesn't require any breaking changes, as the result is identical, because we are just doing the same action twice. I have already tested in two applications at my place, one on Nuxt and one on custom SSR.

On the other hand, the changes worth introducing are custom serializers and deserializers. For example something like this:
```js
import { uneval } from 'devalue';
// ...

const ssr = ssrExchange({
    isClient: !import.meta.env.SSR,

    // By default
    // serialize: JSON.stringify,
    // deserialize: JSON.parse,

    // Custom for devalue
    serialize: data => data,
    deserialize: data => data,
});

// ...

if (import.meta.env.CLIENT) {
  ssr.restoreData(window.__GRAPHQL__ || {});
}

// ...

if (import.meta.env.SSR) {
  const graphqlState = uneval(ssr.extractData());
  return render('index.html', { graphqlState })
}
```

This will be backwards compatible for the reason that we will default to JSON methods

kitten (approved):

kitten:
Had another look, I agree with the removal of the property entry over the `if` branch, to avoid empty data being stringified. I don't think custom serializers/deserializers are worth it, since it would only come into play when non-JSON data is preserved, which implies that the `ssrExchange` would simply have to be re-ordered, since custom deserialization can only occur in other exchanges. Merging

PR #3641, opened 2024-07-26
perf(vue): use shallowRef for data

yurks:
## Summary

As graphql response usually contains a quite huge nested objects structure, wrapping it with `ref()` is not memory efficient.
Instead of doing this, it's more efficient to use `shallowRef()` for `data` reactive variable, where the payload is stored.

However, in https://github.com/urql-graphql/urql/pull/3611 while optimizing other variables, @negezor missed the `data` variable. I can assume there was a reason for that, so I consider this PR as a suggestion/discussion. Personally, I don't see any negative side effects from using `shallowRef()`, but maybe you guys have some thought on it.

I see only one possible scenario in which this would be a "breaking change": updating some nested `data` parts manually outside of this plugin, but I'd say - this is a quite specific pattern.

## Set of changes
- Replaces `ref()` with `shallowRef()` for the `data` variable

### DIFF
--- a/.changeset/old-humans-knock.md
+++ b/.changeset/old-humans-knock.md
@@ -1,0 +1,5 @@
+---
+'@urql/vue': patch
+---
+
+Use `shallowRef` for data variable to avoid extra overhead for heavy objects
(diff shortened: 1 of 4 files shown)


negezor:
Makes sense, I wasn't originally looking at data as to how field level changes might be happening there. But I forgot that we always replace the contents of ref 😅

kitten (approved):
Don't really see what could go wrong here, since types aren't relevant and `computed` was already mentioned and recommended across several issues/discussions 👍

PR #3643, opened 2024-07-28
chore(SSRExchange): Remove 2nd data serialization

ccouzens:
It is unnecessary to set it a 2nd time.

The preceding code is

```
  const serialized: SerializedResult = {
    data: JSON.stringify(result.data),
    hasNext: result.hasNext,
  };
```

which has initialized `serialized.data` with the very same value that the removed code is setting.


kitten:
Closing since this is a duplicate in intention of #3632, with prior discussion there. Also please follow the contributing guidelines next time, if you do choose to contribute again ❤️ They contain notes on changesets

PR #3660, opened 2024-09-03
fix(debug): use debug verbosity in debugExchange

onx2:
## Summary

The debug exchange currently outputs at a verbosity of `log`. However I think it makes more sense to use `debug` because that is the name of this package and the console gets littered with log messages without some other approach to minimize that output. 

This change should hopefully promote more usage of this exchange as it can easily be toggled from the console.

## Set of changes

- Changes the console output verbosity from `log` to `debug` in the `debugExchange`


onx2:
I tried to follow the instructions from the bot for changeset but it isn't working. Neither `npx changeset` or `yarn changeset` are working for me. Could someone please add the changeset for me or show me how to get that automated process working?

onx2:
Friendly ping on this. It would really help my workflow to separate debug messages from log messages.

kitten:
Seems the commit here has caused a split in what's reconcilable and the added push has broken mergeability here: https://github.com/onx2/urql/commit/d578b5b297a7f59ee98e7a53c0bcccb401cbb896

I'm a bit late on this, I realise, the problem here however is that the PR has been opened from a `main` branch, and the merge commit makes it hard to add commits for maintainers as per what GitHub enables me to send to the branch.

PR #3665, opened 2024-09-09
fix(graphcache): mutation would cause dependent operations and reexecuting operations become the same set

xuanduc987:
## Summary

The PR #2761 moved `reexecutingOperations = dependentOperations` assignment into `if (operation.kind === 'query')` guard. Thus after a completion of mutation, `dependentOperations = _reexecutingOperations` make reexecuting and dependent operations the same set object.

I noticed this issue after debugging the bug reported in discord https://discord.com/channels/1082378892523864074/1082386845553406002/1281542643670712334

### DIFF
--- a/.changeset/tidy-spies-turn.md
+++ b/.changeset/tidy-spies-turn.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-graphcache': patch
+---
+
+fix bug that mutation would cause dependent operations and reexecuting operations to become the same set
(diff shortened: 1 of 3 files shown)


JoviDeCroock:
Mind adding a changeset and a test that shows this behavior?

xuanduc987:
@JoviDeCroock I had added a new changeset.
About test, I took a look at `cacheExchange.test.ts` but still not sure how to test this.

xuanduc987:
@JoviDeCroock I added test to assert that the number of reexec after a query would not change after executing a mutation

JoviDeCroock (approved):

PR #3669, opened 2024-09-13
Enable vitest workspaces

kitten:


JoviDeCroock (approved):

JoviDeCroock:
Currently it works already when running inside of the `solid-urql` folder in isolation but not from the root

PR #3672, opened 2024-09-13
fix(solid): Correctly track query data reads with suspense

XiNiHa:
## Summary

This PR fixes a bug in `@urql/solid` that the change tracking for `query.data` is not working when the client has `suspense: true`.

## Set of changes

Read data from the store instead of the resource, so further changes on store can be tracked and reflected to the returned value

### DIFF
--- a/.changeset/healthy-years-ring.md
+++ b/.changeset/healthy-years-ring.md
@@ -1,0 +1,5 @@
+---
+'@urql/solid': patch
+---
+
+fix(solid): Correctly track query data reads with suspense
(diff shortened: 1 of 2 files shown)


.changeset/healthy-years-ring.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/solid': patch
+---
+
+fix(solid): Correctly track query data reads with suspense
```
JoviDeCroock, line 5:
```suggestion
Correctly track query data reads with suspense
```

PR #3685, opened 2024-10-02
Fix @_required directive when the schema is available

AndrewIngram:
If graphcache was configured with the full schema, the `@_required` directive would be ignored on nullable fields. This diff tweaks the condition to fix that behaviour.

### DIFF
--- a/exchanges/graphcache/src/operations/query.ts
+++ b/exchanges/graphcache/src/operations/query.ts
@@ -534,7 +534,7 @@ const readSelection = (
       (directives.optional ||
         (optionalRef && !directives.required) ||
         !!getFieldError(ctx) ||
-        (store.schema &&
+        (!directives.required && store.schema &&
           isFieldNullable(store.schema, typename, fieldName, ctx.store.logger)))
     ) {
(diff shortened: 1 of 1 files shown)


JoviDeCroock (commented):
Mind adding a test and changeset?

AndrewIngram:
Should be able to, will actually need to check out the code first 😅 

Can I just get confirmation that this is actually fixing unintended behaviour? The docs are a little confusing on this point, even though required _seems_ most useful when there's a schema saying a field is optional:

> Graphcache comes with two directives built-in by default. The optional and required directives. These directives can be used as an alternative to [the Schema Awareness feature’s](https://commerce.nearform.com/open-source/urql/docs/graphcache/schema-awareness/) ability to generate partial result

AndrewIngram:
(test isn't good yet, plz don't review yet!)

kitten (approved):

kitten:
> Can I just get confirmation that this is actually fixing unintended behaviour?

I'd say it's unintended behaviour, yes. It's definitely not something we had in mind, since the directives were mostly tested as an alternative to schema awareness.

JoviDeCroock (approved):

PR #3690, opened 2024-10-19
fix(core): Remove `for-of` syntax from core helpers for JSC memory reduction 

kitten:


packages/core/src/gql.ts
```diff
@@ -78,8 +78,9 @@ function gql(parts: string | TemplateStringsArray /* arguments */) {
   }
 
   source.unshift(keyDocument(body));
-  for (const document of source) {
-    for (const definition of document.definitions) {
+  for (let i = 0; i < source.length; i++) {
+    for (let j = 0; j < source[i].definitions; j++) {
```
JoviDeCroock, line 82: Could probably save some work by hoisting this length but that's a nit

```suggestion
    for (let j = 0; j < source[i].definitions.length; j++) {
```

JoviDeCroock (approved):

PR #3712, opened 2024-11-18
docs: add retry info for idempotency

bahrmichael:


kitten:
Note here: the docs haven't been in a super deployable and maintainable state here, so currently their deployments aren't necessarily working and we'd have to fix this first to be able to merge this afaik. We haven't really checked on them and they are in state where it's either easier to replace them or to double-check that they still deploy and work as expected.

PR #3727, opened 2025-01-02
(chore) dedupe pnpm lockfile

schiller-manuel:
## Summary

this resolved the following obscure error when running the tests:

> The requested module 'vite' does not provide an export named 'parseAstAsync'


JoviDeCroock (commented):
I think this also updated a number of deps, namely the one in our introspection minifier. This leads to the typescript issue you are seeing here

PR #3728, opened 2025-01-02
allow to provide underlying storage for ssr exchange

schiller-manuel:
## Summary

This change allows to optionally provide the storage that is used in the ssr exchange for storing the serialized results.

## Detailed description

The goal is to enable streaming SSR with @tanstack/start.
When it is possible to pass in a custom storage into the ssr exchange, a proxy can be passed in.
This can be used on the server to stream the serialized results to the client as soon as they are available, as well as retrieve them on the client.

A demo implementation is available here: https://github.com/schiller-manuel/start-urql/blob/main/app/router.tsx#L17

### DIFF
--- a/packages/core/src/exchanges/ssr.ts
+++ b/packages/core/src/exchanges/ssr.ts
@@ -36,6 +36,10 @@ export interface SSRData {
   [key: string]: SerializedResult;
 }
 
+export interface SSRDataStorage {
+  [key: string]: SerializedResult | null;
+}
+
 /** Options for the `ssrExchange` allowing it to either operate on the server- or client-side. */
 export interface SSRExchangeParams {
   /** Indicates to the {@link SSRExchange} whether it's currently in server-side or client-side mode.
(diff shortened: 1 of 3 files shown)


packages/core/src/exchanges/ssr.ts
```diff
@@ -188,7 +197,8 @@ const revalidated = new Set<number>();
 export const ssrExchange = (params: SSRExchangeParams = {}): SSRExchange => {
   const staleWhileRevalidate = !!params.staleWhileRevalidate;
   const includeExtensions = !!params.includeExtensions;
-  const data: Record<string, SerializedResult | null> = {};
+  const data: SSRDataStorage =
```
JoviDeCroock, line 200: Why can't we do this with initialState? For instance when we run extractData we extract into this custom storage and on the client we can leverage the restoreData method and initialState to restore additional and the initial entries.

This is similar to our nextjs solution

schiller-manuel:
closed in favor of https://github.com/urql-graphql/urql/pull/3729

PR #3730, opened 2025-01-05
fix(core,graphcache): Update optional graphql types with TS 5.5 compatible switch

### LINKED ISSUE #3724: graphql is not declared as peer dependency
### Describe the bug

`graphql` is not declared as a peer dependency neither for the `urql` package nor `@urql/core`.  While it happens to work with npm and pnpm because of the fs lookup, it breaks for yarn pnp.

While so far, for me, the only thing that broke was the type inference because all graphql types will fallback to any, there is runtime code also which won't work at all.

Solution:
`@urql/core` and all packages that rely on it must have graphql as required peer dependency.

### Reproduction

.

### Urql version

v4.2.1

### Validations

andreisergiu98:
Resolves #3724

## Summary

1. TypeScript changed the behaviour of "unresolved" any from 5.5 so doing `type Doc = OrNever<UnresolvedType> | LocalType` will always be any.
2. Without `graphql` as optional peer dependency in `@urql/core`, `graphql.web` will be unable to resolve it when using a strict package manager like Yarn Berry.

Both problems are covered by #3724 

## Set of changes

1. Changed OrNever to 
```typescript
type OrNever<T> = void extends T ? never : T;
```
Which is compatible with both newer and older TypeScript versions.

2. Added `graphql` as optional peer dependency for packages that are using `graphql.web`

As a sidenote, there is a `graphql` import without fallback in `urql/packages/preact-urql/src/hooks/useRequest.ts` that will also break inference, but I am not sure how would you like to handle that.

### DIFF
--- a/exchanges/graphcache/package.json
+++ b/exchanges/graphcache/package.json
@@ -66,6 +66,7 @@
     "@urql/core": "^5.0.0"
   },
   "dependencies": {
+    "@0no-co/graphql.web": "^1.0.13",
     "@0no-co/graphql.web": "^1.0.5",
     "@urql/core": "workspace:^5.1.1",
     "wonka": "^6.3.2"
--- a/exchanges/graphcache/package.json
+++ b/exchanges/graphcache/package.json
@@ -67,7 +67,6 @@
   },
   "dependencies": {
     "@0no-co/graphql.web": "^1.0.13",
-    "@0no-co/graphql.web": "^1.0.5",
     "@urql/core": "workspace:^5.1.1",
     "wonka": "^6.3.2"
   },
--- a/.changeset/old-ears-hunt.md
+++ b/.changeset/old-ears-hunt.md
@@ -1,0 +1,6 @@
+---
+'@urql/exchange-graphcache': patch
+'@urql/core': patch
+---
+
+Fix compatibility with Typescript >5.5 (See: https://github.com/0no-co/graphql.web/pull/49)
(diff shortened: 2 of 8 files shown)


andreisergiu98:
Might be an idea to wait for this release https://github.com/0no-co/graphql.web/pull/50 and also include a version bump with this PR.

exchanges/graphcache/package.json
```diff
@@ -63,10 +63,16 @@
     "prepublishOnly": "run-s clean build"
   },
   "peerDependencies": {
-    "@urql/core": "^5.0.0"
+    "@urql/core": "^5.0.0",
+    "graphql": "^14.0.0 || ^15.0.0 || ^16.0.0"
```
kitten, line 67: Hm, I understand that Yarn PnP may have problems here, but since this is a typings change, I'm not sure it'll actually be of huge concern, especially since there's overlap here with what `graphql.web` does.

I'm a little concerned that this change would cause other package managers (or also Yarn Berry) to install `graphql` automatically. The problem is that peers have changed in how they're treated by package managers a lot recently, and auto-installations are now a thing, and this might now have the unintended side-effect of always installing `graphql`, even if it's marked as optional.

I think it'd be nice to split this out and only port the typings change for now. Happy to go ahead and make that change, since I dropped the ball on getting the main typings fix in first

andreisergiu98: > I think it'd be nice to split this out and only port the typings change for now. Happy to go ahead and make that change, since I dropped the ball on getting the main typings fix in first

That makes sense. You can go ahead and close this PR then.

I'll test the inference again after the typing fix is released. If it's still not working, though as you said there should be sufficient overlap, then I'll open another PR with the peer dependency change. My understanding is that if it's also an optional one package managers won't auto install it, but I'll need to do some digging to confirm that.

kitten (approved):

Issue #3633, opened 2024-07-17
(vue) useQuery pause option is not reactive


mrvik:
### Describe the bug

When a reactive property (a ref, computed or reactive) variable is used in the `pause` property, it keeps using the initial value.

### Reproduction

https://stackblitz.com/edit/github-uroxw1-qpmhta?file=package.json

### Urql version

Bug introduced again in >= v1.3.1

### Validations

- [X] I can confirm that this is a bug report, and not a feature request, RFC, question, or discussion, for which [GitHub Discussions](https://github.com/urql-graphql/urql/discussions) should be used
- [X] Read the [docs](https://formidable.com/open-source/urql/docs/).
- [X] Follow our [Code of Conduct](https://github.com/urql-graphql/urql/blob/main/CODE_OF_CONDUCT.md)

angelov-a:
Experiencing the same issue.

PR #3734, opened 2025-01-07
fix(vue): fix variables typing

### LINKED ISSUE #3733: (vue) useQuery loses typing of variables
### Describe the bug

I'm using GraphQL codegen for typing. Here is an MWE (I also made a repo to show the issue -- see the reproduction link):

```typescript
const AllFilmsQuery = graphql(/* GraphQL */ `
  query AllFilmsWithVariablesQuery($first: Int!) {
    allFilms(first: $first) {
      edges {
        node {
          id
        }
      }
    }
  }
`)

// These variables should raise a type error:
useQuery({
  query: AllFilmsQuery,
  variables: { first: null } // wrong type
})
useQuery({
  query: AllFilmsQuery,
  variables: {  } // missing field
})
useQuery({
  query: AllFilmsQuery,
  variables: { foo: 'bar' } // extra field
})
```

The problem comes from [this commit](https://github.com/urql-graphql/urql/commit/d07602d23db0f8ef528181bef467de6bae5acee8). If you revert
```typescript
type MaybeRefObj<T> = T extends {} ? {
    [K in keyof T]: MaybeRef<T[K]>;
} : T;
```
to 
```typescript
type MaybeRefObj<T extends {}> = { [K in keyof T]: MaybeRef<T[K]> };
```
then the type of the variables is correctly inferred, and the above examples produce a type error.

Not sure if one can revert the aforementioned commit without breaking something else... (I can investigate if need be.)

### Reproduction

https://github.com/arkandias/urql-vue-variables-typing-issue

### Urql version

  "dependencies": {
    "@urql/vue": "^1.4.2",
    "graphql": "^16.10.0",
    "vue": "^3.5.13"
  },
  "devDependencies": {
    "@graphql-codegen/cli": "^5.0.3",
    "@graphql-codegen/client-preset": "^4.5.1",
    "@graphql-typed-document-node/core": "^3.2.0",
    "@vitejs/plugin-vue": "^5.2.1",
    "@vue/tsconfig": "^0.7.0",
    "typescript": "~5.6.3",
    "vite": "^6.0.5",
    "vue-tsc": "^2.2.0"
  }

### Validations

arkandias:
## Summary

Resolves #3733 

## Set of changes

Revert type `MaybeRefObj` to its definition before commit [d07602d](https://github.com/urql-graphql/urql/commit/d07602d23db0f8ef528181bef467de6bae5acee8).


arkandias:
Need more discussion to decide what to do (see #3733).

arkandias:
I think this last commit is a good candidate: it fixes the typing problem and does a simple unwrapping of `variables` and `query` as discussed in the related issue.

The CI does not pass because of some incompatibility between Vite versions (not related to the changes I made, I'm baffled...).

Something that bothers me is that I could not help TS to correctly infer the type of `_args.variables` in `createRequestWithArgs`, so I used a type assertion there (it's the only place). I'm pretty sure the typing is correct, but if someone can double check this (as well as the definition of `GraphQLRequest<T, V>`, cf. my comment in the related issue).

Also, I left the possibility to use a reactive `query` field, but I really don't understand the use case for that.

Finally, the doc should be updated to explicitly say what reactive fields can be passed to `useQuery` et al.
If the changes in this PR are accepted I can update the doc accordingly.

arkandias:
(N.B. I also fixed the doc of `createRequest`: the variables names were incorrect and there was no description for one of them.)

arkandias:
> The CI does not pass because of some incompatibility between Vite versions (not related to the changes I made, I'm baffled...).

Regenerating `pnpm-lock.yaml` fixed this error.

I also updated the tests as follows:
- Revert test changes introduced in 068df71f
- Fix 'reacts to variables changing' test by wrapping variables in reactive()  since deep unwrapping is no longer performed

Now the CI checks pass!

JoviDeCroock:
Overall these changes look good to me, however the changes to the tests concern me a bit. Also I'd love for us to have some TypeScript based tests to prevent regressions

arkandias:
Hi @JoviDeCroock, I received on email from GitHub on January 23th with a response from you, but I can't see any in this thread, did you delete it?

Just some more info (since you were concerned with the modifications to the tests): basically I reverted the test to their state before 068df71. Many of the test introduced in this commit are testing the deep unwrapping behavior, which is IMO not desirable, that's why I reverted most of it.

I'm not exactly sure how to test that type safety is working, but I can see two solutions:
- using a dependency like `tsd` (which I'm not familiar with), or
- writing `.ts` files which would raise errors during TS transpilation if type safety is not working; these tests would be run during `tsc` step, not `vitest` though.

arkandias:
Sorry to be a bit pushy about this Issue/PR, but it's been two months and I haven't heard from anyone... Having a broken typing system seems like a major issue for a production-grade GraphQL engine... Even more since it was working before, so people might be relying on it (I did find this bug after pushing some erroneous code which would have been caught if the typing had been working as expected). I noticed other people have run into this issue too.

I did my best to document the issue and propose a solution (I also raised some questions in this PR and the related issue). If that's not good enough or if you are too busy to dig into it right now, could we at least revert the changes from the faulty commit while thinking about a better solution?

@JoviDeCroock @kitten

arkandias:
So it's been 3 months since I posted this PR and the related issue, and I haven't heard from anyone. The regression is 8 months old now and I'm losing hope to see it fixed.

For those interested in having a lightweight vue-compatible graphql client with non-broken typing, I suggest moving to Villus as I did, because it does not seem to be a priority here.

kitten:
The problems here are several layers deep.

We unfortunately don't use Vue or `@urql/vue` ourselves in any projects. This means that testing these changes manually, whenever large changes are made, is arduous and takes a long time, so we delay it for as long as possible.
(In this case, I dropped the ball on this entirely, and I'm sorry about that, but usually I'm the one testing these changes)

The other problem here is that this PR needs a changeset, as per the contributing guidelines and the bot comment above: https://github.com/urql-graphql/urql/blob/main/CONTRIBUTING.md#how-do-i-document-a-change-for-the-changelog
Without it, it won't be documented for the release (or flagged for release)

The final problem here is that huge PRs like #3619 with a lot of back and forth and no context on what direction some Vue patterns have moved to cause more problems for us since it compounds the first problem. When we then have PRs that basically introduce a "apply and revert" loop where changes go beyond what's needed in the moment, it increases the surface area of what's checked and lowers my confidence in the change.
For context, the past external PRs to `@urql/vue` have fixed some issues, but regressed on others, simply because they touched too much code. This PR again makes a huge amount of changes that probably go beyond what's needed.
It also regenerates the lockfile in the same PR, which is a big unrelated change. Gradual updates are important and we're affecting tens of packages here that are affected in a PR that's not for those packages.

Lastly, re. the typings change. We aren't yet making use of extensive typings tests in this repo, since Vitest didn't have that feature when we would've added it. We also haven't added alternatives. This meant more regressions in some patterns than we would've liked and also increases testing burden.

I didn't have the time to write all of this out before, but I hope this is useful nonetheless.
If you'd like to see this quicker, a fork is always safest for yourself and for us in situations like these. I'm still happy to merge this, but I'd have to break down what changes belong to this PR, which changes are correct, what other changes should be split out, etc.
But, yes, `villus` is the Vue-community maintained library here, and if it suits your need, there's obviously nothing wrong with switching.

arkandias:
Thank you for your reply @kitten. I totally understand that you don't have the time to review my PR, especially if you don't use Vue. Sorry if my last message sounded bitter somehow. I really like urql, so I'd really like to help fix this issue eventually.

Just to be clear, I was not expecting you nor anybody else to merge my PR as is (since I explicitly left some points to be discussed). That's why I did not provide a changeset yet. Also, I regenerated the lockfile to pass one of the checks (this one might be quite hazardous).

Now regarding the main issue, I tried to describe extensively the problems and possible workarounds in this PR and in the related issue, but here is a sum up:
- https://github.com/urql-graphql/urql/commit/d07602d23db0f8ef528181bef467de6bae5acee8 modified the definition of `MaybeRefObj`, which caused the typing of variables to be broken
- https://github.com/urql-graphql/urql/commit/068df71f65506445e26e5b57366bae691a6c3554 introduced "deep unwrapping" of `variables` with an incorrect typing (which passes typecheck with a type assertion); I argued in the issue that this behavior is (1) not necessary (it can be achieved with a simple `ref` or `reactive`), (2) very hard (if not impossible) to type correctly, (3) not documented.

This is mainly what the PR is about. The rest is just some renaming of utils (to be consistent with Vue current implementations) and reverting the tests that use this deep unwrapping.

I'd really like to have the opinion of someone more knowledgeable about this... Maybe @yurks or @negezor if you have some time to take a look?

arkandias:
I must confess I share some responsibility in that long chain as I did temporarily broke a functionality in #3633 😬
So yes, having at least another Vue dev's opinion on these matters would be great!

Also, I think the Vue bindings documentation should be updated afterwards (for example the functionality I broke was not documented -- and still isn't I think). I'd be happy to help with this too.

yurks:
I'm unable to check this right now, but want to mention - deep variables unwrapping was _not_ introduced in https://github.com/urql-graphql/urql/pull/3619. That PR is about fixing memory leaks with unwrapping, but this feature was in place from the point we decided to use this library, and that was about 3 years ago, so this is kind of original functionality.

I'd say removing deep unwrapping is a real breaking change, which costs at least major version bump and deprecation notice.

Personally, I didn't get the point of removing this, but it definitely concerns me not understanding the context. I'll try again a bit later, or kindly ask you for details why you are removing.

arkandias:
Thanks for taking the time to comment!
Seems to me that `unwrapDeeply` was introduced in this PR (see [here](https://github.com/urql-graphql/urql/pull/3619/files#diff-e14cd5e784e3c6f2e8279a48f04666e7d41cb10abe57a1b2494bd8870f4fb510))

As to whether this a breaking change or not, I'm not sure, because this behavior has never been documented (to the best of my knowledge). The doc merely states:
```
All inputs that are passed to useQuery may also be [reactive state](https://v3.vuejs.org/guide/reactivity-fundamentals.html). This means that both the inputs and outputs of useQuery are reactive and may change over time.
```

My primary concern with `unwrapDeeply` is that it seems intractable to type. The current typing (`(input: T) => T`) is wrong (and passes type check with a type assertion). You can check the typing of Vue's `reactive` (which implements some kind of deep unwrapping of refs) to see that it is already not trivial. And yet, `unwrapDeeply` does much more, because it also unwrap getters and elements of array (which `reactive` does not).

My secondary concern is that it is quite a  complicated and confusing mechanism, that can be simply achieved by wrapping `variables` in a `reactive` or a `ref` (at least to unwrap the nested refs -- for getters one should use `computed` or simply unwrap a "shallow" getter). A simple unwrapping would be easier to understand, document, and would cover most use case (I'll write a more detailed example below). As a comparaison, this is how `vue-apollo` or `villus` handle variables reactivity.

arkandias:
Here is some expanded explanation about how you can already achieve deep unwrapping with a simple unwrapping (`toValue`) and Vue built-in utils.

1) `reactive` (and `ref`, which uses `reactive` under the hood) already deeply unwraps references (though not for elements of arrays), so if your variables look like
```
const variables = {
  foo: someRef,
  bar: someOtherRef,
  baz: {
    a: yetAnotherRef,
    b: andSoOn,
  },
};
```
you can just pass `reactive(variables)` or `ref(variables)` to your query.

2) If you have nested getters, like
```
const variables = {
  foo: () => someFn(someRef.value),
  bar: () => someOtherRef.value,
};
```
you can use a "shallow getter" instead
```
const variables = () => ({
  foo: someFn(someRef.value),
  bar: someOtherRef.value,
});
```

3) In the case where you'd have both nested refs and getters (for the sake of completeness), just use `computed` references, and wrap everything in a `reactive` or `ref`, e.g.,
```
const variables = reactive({
  foo: someRef,
  bar: computed(() => someOtherRef.value),
});
```

These examples are merely there to show that you can pass deeply nested reactive variables without implementing some deep unwrapping on urql side. I do understand that in some cases you need to prepare a bit the `variables` field (wrapping it in `ref`/`reactive`, or using `computed` for more complex cases), but it seems totally acceptable to me (and this is what other vue-compatible GraphQL clients implement)

kitten:
Ah, I remember this better now. I believe my initial impression here was the same re. the solutions you're mentioning. My main question (as someone who hasn't used Vue since this reactivity system was first added) would be whether that's "ergonomic and established"

If it's not expected for us to deeply unwrap, then we shouldn't do it. As you're saying other clients (Villus?) don't deeply unwrap, right?

However, I suppose that brings us to a very simple conclusion of just a few questions:
- What do other libraries do? (like Villus, Vue Apollo, Tanstack, etc)
- What's expected by most Vue users when they use these bindings?
- What's the difference to Villus apart from this that we may want to adopt?

I'm not against a breaking change, especially since a raised peer dep on `vue` for the other issue already implies a breaking change.
I'm also not against upholding this decision afterwards (we've clearly accidentally flip-flopped on this, but that's not necessarily an issue — we were relatively early with adopting Vue Reactivity imho, before some preferences were established)

The only concern here is documentation, since they're not in a great shape, and not necessarily in a changeable shape. But we can cross that bridge when we get to it

negezor:
Hi! I prefer an approach where we use `MaybeRefOrGetter` only for `variables`. In my code, I always write something like this:  

```ts
useFetchThreadCommentsQuery({
    variables: () => ({
        subjectId: props.subjectId,
        after: afterCursor.value,
        orderBy: orderBy.value,
    }),
});
```

I got used to this approach back in vue-apollo. However, I'm against wrapping `reactive(variables)`, as this would again lead to memory leaks. Since we're using constants anyway, Vue can attach its proxies to them. This caused me problems in SSR, since each new Vue instance would add new getters, more details #3612.

arkandias:
@negezor Thanks for your feedback! I'll look into PR #3612 to understand the issue (and avoid a possible regression on this matter).

Just to clarify: I did not mean to wrap `variables` inside `reactive` on urql side, I just meant that if one wants to unwrap deeply nested refs, one can apply `reactive` before passing it to `useQuery`.

Hebilicious:
@arkandias in the PR I see you are modifying tests like : 

```ts
    const sub = useSubscription({
      query: `{ test }`,
    });
```

to 

```ts
    const sub = reactive(
      useSubscription({
        query: `{ test }`,
      })
    );
```

and that's confusing me a little. I was under the impression that your changes were affecting variables only ? Could you clarify ?

I agree with what @negezor is saying, using a `MaybeRefOrGetter ` for `variables` feels natural and in-line with the most common vue patterns.

That being said, I believe some of the changes here are very good and could be merged much quicker if this PR was split into multiple PRs. Like the renaming of the types to align with vue 3.3 naming.

arkandias:
Hi @Hebilicious, Thanks for taking a look at this! These changes come from the fact that I reverted the tests of `useSubscription` to before commit #068df71f, but I agree: wrapping `useSubscription` in a `reactive` seems odd. Actually, these test are passing just fine as they were, so there is no point reverting them (unlike `useQuery`: I had to revert some of those because they were testing the deep unwrapping behavior).

What about the `query` field? Should it be reactive too? If so, I guess MaybeRefOrGetter too? (I never used a reactive query, but maybe it's not so uncommon.)

There is also the implementation note I talked about [here](https://github.com/urql-graphql/urql/pull/3734#issue-2773060570), if anyone's interested (I had to use a type assertion at some point).

arkandias:
Sorry for the long delay since last update. I just:
- rebased the commits
- made minimal changes to the test files, rather than reverting from previous commit (basically I just removed deep-unwrapping related behaviors from tests)
- added changeset

So now this PR is ready to:
1. Fix type-breaking regression (and removing deep unwrapping)
2. Update utils name following Vue 3.3+ convention

I think @negezor and @Hebilicious comments support my claim about how reactive variables should be handled.

@kitten Could you take a look when you have some time?

JoviDeCroock (approved):

yurks:
The changes looks reasonable for me, supporting of deep unwrapping is quite tricky and maybe non used widely.

However, this is not just `Fix regression breaking `variables` typing`, but breaking change, as we removing ~3yo functionality. Don't we want to mention this in changelog and bump major version?

kitten:
The major version is bumped by a parallel change so I didn't point it out. But this will be released as a major bump.

PR #3755, opened 2025-03-03
chore: Omit minified files and `sourcesContent` in published packages

kitten:
## Summary

We shouldn't publish with minified files due to the source map sizes and increased publish size. They're not meant to be used directly and we mostly relied on them for measurements.

Definitely a trade-off, but we shouldn't include `sourcesContent`. It likely doesn't help debugging much over our "half prettified" outputs and increases size substantially

## Set of changes

- Disable `sourcesContent` when publishing
- Disable minified bundle output when publishing

### DIFF
--- a/.changeset/purple-baboons-nail.md
+++ b/.changeset/purple-baboons-nail.md
@@ -1,0 +1,23 @@
+---
+'@urql/exchange-auth': patch
+'@urql/exchange-context': patch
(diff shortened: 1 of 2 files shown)


JoviDeCroock (approved):

byCedric:
thank-you

Reduces the `@urql/core` package size by 316.8kb / 34% ❤️

PR #3759, opened 2025-03-07
Allow SSR exchange to use custom serializer

Warxcell:
## Summary

Allows SSR exchange to use custom serializer.

### DIFF
--- a/packages/core/src/exchanges/ssr.ts
+++ b/packages/core/src/exchanges/ssr.ts
@@ -36,6 +36,9 @@ export interface SSRData {
   [key: string]: SerializedResult;
 }
 
+export type Serializer = (data: any) => string
+export type Deserializer = (serialized: string) => any
+
 /** Options for the `ssrExchange` allowing it to either operate on the server- or client-side. */
 export interface SSRExchangeParams {
   /** Indicates to the {@link SSRExchange} whether it's currently in server-side or client-side mode.
@@ -74,6 +77,10 @@ export interface SSRExchangeParams {
    * not serialize this data by default, unless this flag is set.
    */
   includeExtensions?: boolean;
+
+  serialize?: Serializer;
+
+  deserialize?: Deserializer;
 }
 
 /** An `SSRExchange` either in server-side mode, serializing results, or client-side mode, deserializing and replaying results..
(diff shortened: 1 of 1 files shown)


kitten:
This has been missing multiple things from our [contributing guide](https://github.com/urql-graphql/urql/blob/main/CONTRIBUTING.md)

We at least need a changeset to be able to merge this, which is a changelog entry file that describes the change for release notes: https://github.com/urql-graphql/urql/blob/main/CONTRIBUTING.md#how-do-i-document-a-change-for-the-changelog

However, the PR also contains no description on _why_ this change is important to you or why you decided to contribute this and no corresponding issue: https://github.com/urql-graphql/urql/blob/main/CONTRIBUTING.md#how-do-i-propose-changes

API additions without an explanation, even if they're straightforward look often without any reasons as to why they're needed like bloat. In this case, while the change is straightforward, exchanges are pluggable/interchangeable, i.e. you can make your own exchange that supports this behaviour separately from the core library, so we'd need a reason to know whether this would be needed by more users

kitten:
> The use-case is to be able to use more effective methods for serialize/deserialize

That's self-evident, but also imprecise. This is basically the [XY Problem](https://xyproblem.info/). I understand already that you want this change and what it does, but it's an API addition, so I'd love to understand what the motivation is.

Sorry for not clarifying this better. The "why" I'm after isn't "Why did you make this change?", as in describing the API and code change, but instead more like "Why did you need this?", so to go to the root of the change request,
- What would serialisation would you plug into these APIs?
- Why did you want to change from pure-JSON serialisation?
- What is the underlying problem you're trying to solve when swapping out serialisation?

> I would prefer not to copy the whole exchange just to change these 2 methods.

The problem to us here is that that's a compelling and documented API surface. Exchanges are basically customisable extensions and we can't feasibly cover all use-cases with all exchanges all the time. The API of each exchange would balloon to tens of options. That's because exchanges are basically almost pure control flow. The same goes for this one.
So, it's a balance, and hence we at least have to keep track of the changes we do want to make and why

kitten:
Two small concerns here:
- If this is run on every SSR request the extra cost of running this may outweigh the benefits in many cases since JSON operations in most JS engines are very efficient if JS scalar parsing is skipped entirely in favour of raw string parsing (which is what we're doing in here)
- I'm a bit concerned that we'd also be applying this to `result.extensions` but haven't looked into this library enough to see if that's safe 

Re. 1) The types do sufficiently enforce and encourage double serialisation which is desirable, so we can probably ignore this.

Re. 2) It may be nice to have a test here, specifically for graphql-crunch so we can keep it working?

Lastly, the changeset would be needed, but if push comes to shove I can add it myself to the PR, so no worries

PR #3767, opened 2025-03-31
fix(core): Fix multibyte character decoding by using `stream` option

i110:
## Summary
Huge thanks for open-sourcing such a great library. In my private frontend project where I'm using urql, I happened to find an issue where multibyte strings fetched via urql occasionally become garbled. After inspecting the network and logic, I found that the root cause was decoding strings without considering UTF-8 character boundaries.

Fortunately, in frontend development, the TextDecoder API provides a useful `stream` option, and I confirmed that simply enabling this resolves the issue. As for the Node.js or other backend dev environment, I believe the problem could be addressed by using StringDecoder or something, but I haven't explored that part yet.

I would like to request this change be merged in order to simplify my vendoring. I created this PR primarily to share the issue for now, but would be happy to receive any suggestions on how to proceed or how to approach testing. Thank you!

## Set of changes

- Just added an option `stream: true` to TextDecoder::decode call

### DIFF
--- a/.changeset/nervous-flies-confess.md
+++ b/.changeset/nervous-flies-confess.md
@@ -1,0 +1,5 @@
+---
+'@urql/core': patch
+---
+
+Fix `fetchSource` not text-decoding response chunks as streams, which could cause UTF-8 decoding to break.
--- a/packages/core/src/internal/fetchSource.ts
+++ b/packages/core/src/internal/fetchSource.ts
@@ -26,7 +26,7 @@
  * The implementation in this file needs to make certain accommodations for:
  * - The Web Fetch API
  * - Non-browser or polyfill Fetch APIs
- * - Node.js-like Fetch implementations (see `toString` below)
+ * - Node.js-like Fetch implementations
  *
  * GraphQL over SSE has a reference implementation, which supports non-HTTP/2
  * modes and is a faithful implementation of the spec.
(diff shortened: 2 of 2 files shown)


kitten:
Nice catch! :raised_hands: I must've forgotten about that when doing some stream refactors. We could consider `TextDecoderStream` here in the future, but I think that might have implications on browser support, so it's definitely not something we'll have to check immediately for this PR.

I don't remember if this also implies that `TextDecoder` needs to be instantiated per stream, but in the past, if I recall correctly, this didn't matter when I tested it, so it looks good to me as is.

kitten (commented):
Before merging, this will need a changeset file, for a changelog entry, e.g. a patch entry that contains:
> Fix `fetchSource` not decoding text as stream, causing UTF-8 stream decoding issues

https://github.com/urql-graphql/urql/blob/main/CONTRIBUTING.md#what-are-the-pr-conventions

i110:
Thank you for your comment, @kitten 

> I don't remember if this also implies that TextDecoder needs to be instantiated per stream, but in the past, if I recall correctly, this didn't matter when I tested it, so it looks good to me as is.

This is a great catch too. IIUC, since TextDecoder maintains internal state, as you rightly pointed out, reading from multiple streams simultaneously could lead to another kind of garbled result.

To address this, we might consider various options - such as creating a separate TextDecoder instance for each call to `streamBody`, or refactoring `streamBody` using TextDecoderStream, etc. What do you think would be the best? I’d love to hear your thoughts.

Also, while I’d definitely be interested in contributing to this project, solving the issue takes priority. So please feel free to close this PR and make any refactors or fixes as you see fit - I truly don’t mind at all :)

JoviDeCroock (approved):

i110:
Thank you!

PR #3769, opened 2025-04-01
fix(react-urql): Upgrade muted warning code for React 19 internals

### LINKED ISSUE #3764: React 19 Cannot update a component (`Component`) while rendering a different component
### Describe the bug

Since react 19, there is an error showing up when trying to mount a component that runs a query that was already executed by another component.

I found this error while updating a website to next15. It has a page that calls a follow button that queries the followage status. The error shows up when navigating to a different page that imports the same button component.

I prepared two examples, one using nextjs, which is closer to my use case, and another one using `npx create-react-router@latest`.

What I found out is that the error only happens when using network-only, and when the two query invocation has the same query name. (There's an extra page on next's example that test it). On nextjs, it happens regardless using network-only or not

create-react-router example:
https://github.com/JoaoBGusmao/urql-r19-bug

nextjs example:
https://github.com/JoaoBGusmao/urql-next15-bug

Both projects, just install the dependencies and run yarn dev

### Reproduction

https://github.com/JoaoBGusmao/urql-r19-bug

### Urql version

urql 4.2.2

### Validations

- [x] I can confirm that this is a bug report, and not a feature request, RFC, question, or discussion, for which [GitHub Discussions](https://github.com/urql-graphql/urql/discussions) should be used
- [x] Read the [docs](https://formidable.com/open-source/urql/docs/).
- [x] Follow our [Code of Conduct](https://github.com/urql-graphql/urql/blob/main/CODE_OF_CONDUCT.md)

kitten:
Resolves #3764

## Summary

Same as before: the warning is irrelevant, and we're looking to mute it. We can make the deferred dispatching permanent, since we don't want to rely on internals

As before, just because the warning is issued that doesn't mean anything is wrong. We're only doing this to mute this warning.

## Set of changes

- Update how `ReactCurrentOwner` is accessed using new internals
- Always defer `setState` dispatch unless there's no synchronous/concurrent update in parallel

### DIFF
--- a/.changeset/rich-melons-play.md
+++ b/.changeset/rich-melons-play.md
@@ -1,0 +1,5 @@
+---
+'urql': patch
+---
+
+Upgrade false-positive circumvention for internal React warning to support React 19
--- a/packages/react-urql/src/hooks/state.ts
+++ b/packages/react-urql/src/hooks/state.ts
@@ -1,4 +1,4 @@
-import * as React from 'react';
+import type { Dispatch } from 'react';
 
 export const initialState = {
   fetching: false,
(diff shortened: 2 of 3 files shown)


ericpoulinnz:
@kitten any timeline for when this PR will get merged and released? With Expo 53 being released this week there will be a lot of projects now on React 19 and getting this error.

sep2:
We use this pnpm patch to appy this PR before it merged

```patch
diff --git a/dist/urql.es.js b/dist/urql.es.js
index c2ef1c7e614e49bad3002c98893ccbc5ed2abc5b..0b5feec9a4631e3827389803d687b573b21c26a5 100644
--- a/dist/urql.es.js
+++ b/dist/urql.es.js
@@ -70,11 +70,21 @@ var hasDepsChanged = (e, r) => {
   return !1;
 };
 
-var p = r.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED;
+
+var p = r.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED || r.__CLIENT_INTERNALS_DO_NOT_USE_OR_WARN_USERS_THEY_CANNOT_UPGRADE;
 
 function deferDispatch(e, r) {
-  if ("production" !== process.env.NODE_ENV && p && p.ReactCurrentOwner && p.ReactCurrentOwner.current) {
-    Promise.resolve(r).then(e);
+  if (!!p && process.env.NODE_ENV !== 'production') {
+    const currentOwner = p.ReactCurrentOwner
+      ? p.ReactCurrentOwner.current
+      : p.A &&
+      p.A.getOwner &&
+      p.A.getOwner();
+    if (currentOwner) {
+      Promise.resolve(r).then(e);
+    } else {
+      e(r);
+    }
   } else {
     e(r);
   }

```

ericpoulinnz:
@kitten I've patched urql with this change in my production app (a little over 10k MAU) for the last two months without encountering any issues. 

I honestly don't understand how anyone using urql with the latest react doesn't do this. Unless I've implemented it incorrectly? The constant warnings made debugging an absolute nightmare.

JoviDeCroock (approved):

PR #3780, opened 2025-05-17
feat: expose query in authExchange

ThaUnknown:
## Summary

Exposes client.query, which wasn't exposed previously unlike mutation

## Set of changes

I honestly just copied the mutation implementation and modified it to line up with how client.query is called, so this should be fine, but I haven't tested it, it isn't breaking it's just a minor feat for those odd-ball gql api's

### DIFF
--- a/exchanges/auth/src/authExchange.ts
+++ b/exchanges/auth/src/authExchange.ts
@@ -46,6 +46,28 @@ export interface AuthUtilities {
     context?: Partial<OperationContext>
   ): Promise<OperationResult<Data>>;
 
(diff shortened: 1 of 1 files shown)


kitten:
As per the conversation on Discord, just because it's possible doesn't mean it's a pattern we want to encourage. You're free to use a copy/fork of the `authExchange` in your app, as for every exchange, which is why we don't prioritise or consider merging API changes for bad practices and/or edge cases all the time.

In this case, your API is doing something I'd consider a bad practice, and it's just as easy/hard to either change the API or to petition the API's maintainer/owners to make a change.

The escape hatch here is to make a manual fetch call or to copy/fork the `authExchange`

PR #3781, opened 2025-05-25
fix: do not overwrite application provided abort signals

ccouzens:
If the application has provided an abort signal to `fetchOptions` when initializing the `Client`, then that abort signal shouldn't be overwritten when `makeFetchSource` sets up its own abort controller.

In my case, this will allow me to easily set timeouts on requests from my web server to my graphql server.

---

I wanted to test this, but I can't see how without intrusive changes to the test suite.

Now we're using `AbortSignal.any` it validates that the arguments are actually abort signals. The mock functions are no longer sufficient. But the best way of getting an abort signal is through AbortControllers. But the test suite removes access to AbortController in a couple different files. For example here
https://github.com/urql-graphql/urql/blob/92a101a5497040641306a783b56d5fc3ee0917b1/scripts/vitest/setup.js#L2

### DIFF
--- a/packages/core/src/internal/fetchSource.ts
+++ b/packages/core/src/internal/fetchSource.ts
@@ -265,11 +265,10 @@ export function makeFetchSource(
   fetchOptions: RequestInit
 ): Source<OperationResult> {
   let abortController: AbortController | void;
-  let userAbortSignal: AbortSignal | null | undefined;
+  const userAbortSignal = fetchOptions.signal;
   let abortAbortController: (() => void) | undefined;
   if (typeof AbortController !== 'undefined') {
     abortController = new AbortController();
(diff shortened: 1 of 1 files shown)


JoviDeCroock:
Hmm, I looked into browser support for `AbortSignal.any` and it's a bit problematic 😅 only 89% https://caniuse.com/?search=AbortSignal.any

ccouzens:
> But I still don't see a good argument for passing `fetchOptions.signal` in the first place, so another option here is to simply restrict the input type to explicitly mark it as unsupported

I've raised the following PR for this.

https://github.com/urql-graphql/urql/pull/3788

I've kept it in draft while I explore some of the other suggestions, but feel free to merge if it's likely to be the preferred option.

PR #3785, opened 2025-06-19
chore: export the getOperationName function

### LINKED ISSUE #3784: RFC: export the getOperationName utility function
## Summary

When logging out errors with a map exchange is useful to get the operation name.
So whether we're doing a query or a mutation, like the following:

```
query empireHero {
  hero(episode: EMPIRE) {
    name
  }
}
```

I'm saying I'd like to have `empireHero`.

I've noticed you can grab it from the ast that the graphql parser generates and you already have a function called getOperationName to extract it correctly.

## Proposed Solution

Either:

- Export the getOperationName function from the codebase (like apollo does https://github.com/apollographql/apollo-client/blob/main/src/utilities/graphql/getFromAST.ts#L55).
- Include it among the operation fields. This one would be quite invasive in terms of code changes and also a bit out of scope for the `Operation` type.

giacomocerquone:
## Summary

Resolves #3784 

## Set of changes

Just exported the getOperationName function

### DIFF
--- a/.changeset/wild-flowers-yawn.md
+++ b/.changeset/wild-flowers-yawn.md
@@ -1,0 +1,5 @@
+---
+'@urql/core': major
+---
+
+export the getOperationName utility function
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -14,4 +14,5 @@ export {
   mergeResultPatch,
   formatDocument,
   makeOperation,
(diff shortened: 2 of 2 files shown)


JoviDeCroock:
Still needs a changeset, see our contributing doc

JoviDeCroock (approved):

PR #3789, opened 2025-06-29
major(core|persisted): use GET for queries by default

JoviDeCroock:
## Summary

I've been thinking about this for quite a while but it's time that this default is switched up, most GraphQL clients force you in this paradigm of always leveraging POST which misses out on a lot of benefits. The settings to actually do it the "right way" HTTP semantically is hidden behind an option.

The downside is that some servers _might_ not support GET, however they are still free to turn this off. I prefer the explicit having to turn it off from having a degraded default experience.

The default should still be persisted operations but at least we take a step forward HTTP semantically wise.

### DIFF
--- a/.changeset/giant-pets-buy.md
+++ b/.changeset/giant-pets-buy.md
@@ -1,0 +1,9 @@
+---
+'@urql/exchange-persisted': major
+'@urql/core': major
+---
+
+By default leverage GET for queries where the query-string + variables comes down to less than 2048 characters.
+When upgrading it's important to see whether your server supports `GET`, if it doesn't ideally adding support for it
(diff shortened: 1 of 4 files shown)


kitten (approved):
I think it's fair to push for this and change the default here.
This also enables us to look at what we can do to make this behaviour more expected in `core`.

Typically, query-level approaches omit the AST. We'll have to look at how we can deal with the lack of the AST, or rather, enforce that it's still there if a cache (or other logic) expects it.

Basically, we can be more explicit about:
- `persistedExchange` behaviour (potentially in core)
- how exchanges deal with a missing document (explicit incompatibilities)
- how exchanges deal with a document + hash (explicitly disallow modifications via freezes in development)

PR #3805, opened 2025-08-07
feat(graphcache): add possibleTypes config for deterministic fragment matching

xuanduc987:
We want to deterministically match fragments, but don't want the downside of schema-aware: security risk, bundle size, partial result on nullable field.

### DIFF
--- a/exchanges/graphcache/src/operations/shared.ts
+++ b/exchanges/graphcache/src/operations/shared.ts
@@ -234,18 +234,24 @@ export class SelectionIterator {
                     fragment,
                     this.typename
                   )
-                : (currentOperation === 'read' &&
-                    isFragmentMatching(
+                : this.ctx.store.possibleTypeMap
+                  ? isSuperType(
+                      this.ctx.store.possibleTypeMap,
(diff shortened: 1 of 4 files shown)


exchanges/graphcache/src/operations/shared.ts
```diff
@@ -290,6 +296,19 @@ export class SelectionIterator {
   }
 }
 
+const isSuperType = (
```
JoviDeCroock, line 299: How would this work for an abstract type which is extending an abstract type i.e. a union of interfaces etc. basically this might need to be recursive

xuanduc987: @JoviDeCroock I think we can only have union of concrete object types
https://spec.graphql.org/draft/#sec-Unions.Type-Validation

And if interface A implements interface B, then an object C implements A must also declare that it implements B.
https://spec.graphql.org/draft/#sec-Objects.Type-Validation

From this, I believe we don't need recursively check for super type here (possibleTypes from introspection should contains all subtypes)

PR #3806, opened 2025-08-07
Use getCurrentScope instead of getCurrentInstance

### LINKED ISSUE #3731: [@urql/vue] Cannot use useQuery inside of vue-router guard
### Describe the bug

The setup: There's a pinia store that is used in a router guard, the store uses `useQuery` composable from `@urql/vue`

Because it tried to access vue instance in router guard the app returns the following error:

```
use* functions may only be called during the `setup()` or other lifecycle hooks.
```

As far as I can see the error comes from https://github.com/urql-graphql/urql/blob/25d114d25807f0676dbf453732602753279ba0db/packages/vue-urql/src/useClient.ts#L104

Am I doing something wrong?

### Reproduction

https://stackblitz.com/edit/vitejs-vite-kudyigng

### Urql version

@urql/vue v1.4.2

### Validations

arkandias:
Resolves #3731 

## Summary

Currently, `useClient` uses `getCurrentInstance()` to check if it is called from a component setup, and throws if not.
This is problematic for two reasons:
1. `getCurrentInstance` is an undocumented internal API, and an anti-pattern for external libs (see [here](https://github.com/vuejs/docs/issues/1422)). Edit: this method won't be available in Vapor mode (see [here](https://github.com/vuejs/core/releases/tag/v3.6.0-alpha.1)). 
2. The client should be accessible from reactive contexts outside components. For example, Vue Router Navigation Guards allow global injections (see [here](https://github.com/vuejs/docs/issues/1422)). They run inside a scope, but no inherited instance:
```typescript
router.beforeEach((to) => {
  console.log('INSTANCE', getCurrentInstance()); // null
  console.log('SCOPE', getCurrentScope()); // not null
});
```
This is what causes #3731 

## Set of changes

In order to use `getCurrentScope`, we need to bump to Vue 3.2+.
This release is already 4 years old, so this seems reasonable.
Also, Vue 2.7 is EOL so it does not make sense to keep it as a peer dependency.

Edit: Regarding Vue 2.7, README already mention "`@urql/vue` has a peer dependency on `vue@^3.0.0` (Not v2) and doesn't currently plan to be backwards compatible to Vue 2.".

I also updated the documentation in `useClient.ts`, and added some tests in `useClient.test.ts`.

Since I bump peer dependencies, this is a breaking change and should be a major bump.

### DIFF
--- a/.changeset/thick-roses-joke.md
+++ b/.changeset/thick-roses-joke.md
@@ -1,0 +1,5 @@
+---
+'@urql/vue': major
+---
+
+Bump Vue to 3.2+ and replace getCurrentInstance with getCurrentScope


arkandias:
Just added a changeset with major bump.

JoviDeCroock (approved):

PR #3812, opened 2025-08-10
fix: use nullish coalescing for get method preference

dargmuesli:
## Summary

Resolves https://github.com/urql-graphql/urql/issues/3811

## Set of changes

Use nullish coalescing instead of `||` for `preferGetMethod` and `preferGetForPersistedQueries` so that `false` is kept if set.

### DIFF
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -549,7 +549,8 @@ export const Client: new (opts: ClientOptions) => Client = function Client(
     fetchSubscriptions: opts.fetchSubscriptions,
     fetchOptions: opts.fetchOptions,
     fetch: opts.fetch,
-    preferGetMethod: opts.preferGetMethod || 'within-url-limit',
+    preferGetMethod:
+      opts.preferGetMethod != null ? opts.preferGetMethod : 'within-url-limit',
     requestPolicy: opts.requestPolicy || 'cache-first',
   };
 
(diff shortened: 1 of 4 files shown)


packages/core/src/client.ts
```diff
@@ -549,7 +549,10 @@ export const Client: new (opts: ClientOptions) => Client = function Client(
     fetchSubscriptions: opts.fetchSubscriptions,
     fetchOptions: opts.fetchOptions,
     fetch: opts.fetch,
-    preferGetMethod: opts.preferGetMethod || 'within-url-limit',
+    preferGetMethod:
+      opts.preferGetMethod !== null && opts.preferGetMethod !== undefined
```
JoviDeCroock, line 553: This change also needs to go into exchange-persisted https://github.com/urql-graphql/urql/blob/main/exchanges/persisted/src/persistedExchange.ts#L141 and a patch changeset for both

```suggestion
      opts.preferGetMethod != null
```

JoviDeCroock (approved):
Let's add a test so we don't regress! Thanks for taking the time to contribute

.changeset/tender-panthers-sip.md
```diff
@@ -0,0 +1,5 @@
+---
+'@urql/core': patch
+---
+
+Use nullish coalescing for `preferGetMethod` so that `false` is kept if set.
```
JoviDeCroock, line 5: This can be combined into one changeset

JoviDeCroock (approved):

PR #3815, opened 2025-08-21
Actual value for timeout

m1212e:
## Summary
There were multiple occurrences of `setTimeout` calls without an actual value for the delay. [According to the Node.js docs](https://nodejs.org/api/timers.html#settimeoutcallback-delay-args), this will default to 1. In the bun runtime this will also apply but issue a warning to notify of the behavior: 

```
TimeoutNaNWarning: NaN is not a number.
Timeout duration was set to 1.
```

## Set of changes
I changed all uses of the `setTimeout` without a specified delay to a delay of 1. This does not change any behavior but will silence the warnings.

### DIFF
--- a/.changeset/puny-beans-allow.md
+++ b/.changeset/puny-beans-allow.md
@@ -1,0 +1,7 @@
+---
+'@urql/exchange-graphcache': patch
+'@urql/exchange-auth': patch
+'@urql/core': patch
+---
+
+Use actual values instead of omitting the setTimeout delay since leaving it out emits a warning in the bun runtime. Replaced all omitted calls with the default value 1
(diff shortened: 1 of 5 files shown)


kitten:
That's intended here. There's nothing wrong with not passing a timeout timing, especially in tests https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#timers

m1212e:
Hey, thats true! It is not wrong but still issues warning in the bun runtime the users might not be able to properly connect to using urql. And since this does not change any behavior, what would be the problem in explicitly providing a value which is equivalent to the default? This would only increase compatibility of urql with the bun runtime while not changing anything in the logic of the library?

Would you suggest only adding this in the actual runtime code instead of the tests too?

I'd be very thankful for some feedback!

PR #3839, opened 2026-01-21
fix(react-urql): keep suspending until data or error arrives

jtomaszewski:
## Summary
- Fixes suspense behavior when graphcache or other exchanges emit unresolved results
- Previously, the `takeWhile` condition would stop early because `result` was truthy, causing `useQuery` to return `{ data: undefined, error: undefined }` instead of staying suspended
- Added condition to keep subscription alive in suspense mode until either `data` or `error` is present

## Test plan
- [x] Added `useQuery.suspense.test.tsx` with tests that verify:
  - Component stays suspended when partial result is emitted
  - Component unsuspends when data arrives
  - Component unsuspends when error arrives
- [x] All existing react-urql tests pass (50 tests)

### DIFF
--- a/packages/react-urql/src/hooks/useQuery.suspense-graphcache.test.tsx
+++ b/packages/react-urql/src/hooks/useQuery.suspense-graphcache.test.tsx
@@ -1566,5 +1566,95 @@ describe('useQuery suspense with graphcache', () => {
         'fetching: false'
       );
     });
+
+    it('should clean up orphaned promise when pausing so unpause creates new subscription', async () => {
+      const network = createTestNetworkExchange();
+      const client = createTestClient(network);
+
+      const query = gql`
+        query TestQuery {
+          author {
+            __typename
+            id
+            name
+          }
+        }
+      `;
+
+      const TestComponent = ({ pause }: { pause: boolean }) => {
+        const [result] = useQuery({ query, pause });
+        if (!pause) {
+          assertValidSuspenseResult(result, pause);
+        }
+        return (
+          <div data-testid="data">
+            fetching: {String(result.fetching)}, data:{' '}
+            {result.data?.author?.name ?? 'none'}
+          </div>
+        );
+      };
+
+      const Fallback = () => <div data-testid="fallback">Loading...</div>;
+
+      const { rerender } = render(
+        <Provider value={client}>
+          <React.Suspense fallback={<Fallback />}>
+            <TestComponent pause={false} />
+          </React.Suspense>
+        </Provider>
+      );
+
+      await waitFor(() => {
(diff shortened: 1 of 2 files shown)


kitten:
We'll need either an issue, reproduction, or more of an explanation here. LLMs are, while powerful, to be wielded like a dangerous weapon, more so than just a tool. This is inscrutable as for the initial intention, and changes likely much much more than needed.

> Fixes suspense behavior when graphcache or other exchanges emit partial results like { fetching: true } synchronously

The suspense behaviour is compliant with how we defined this to behave. Exchanges aren't to issue "empty" results. If they do, that signals they've settled, unless `stale` is set. `hasNext` is excluded since it signifies deferred results. `stale` however is a state we don't consider suspendable, since it explicitly exists to bypass the usual loading states.

The reasoning here also looks generated since it often mentions:
> exchanges emit partial results like `{ fetching: true }`

Exchanges don't emit `fetching`. `fetching` isn't part of `OperationResult` since it's a "semantic" state, rather than an explicit one, for a time period where no `OperationResult` is available.

Hence, this behaviour basically just matches `{ data: undefined, error: undefined, ... }` which is an illegal or irresolvable state and suspending on it risks pending infinitely.

This could also be expressed with a one line change. But without a reproduction or full description, I'm not willing to read through this 😅 Sorry

Moral of this, before we can maybe move to a broader discussion: Please don't submit PRs with content that isn't how you would've written this yourself, or at least not without first reporting an issue separately.

The problem here is:
- description doesn't match changes and changes aren't minimal to demonstrate on their own what's been changed
- the bug condition report doesn't quite match what I'm expecting, since it mentions an `OperationResult` that isn't possible (or has a special meaning, which would imply waiting longer doesn't make it resolvable)
- the changes imply that something's been changed due to a different issue (orphaned promise?) which is not a demonstrated or mentioned bug (and seems unlikely to me since this was used unchanged for quite a while, unless it's React 19 specific)

jtomaszewski:
In very short: I keep getting empty result (that has no data nor error) when using suspense. That should never happen. If suspense is used, I shouldn't get anything. I expect rendering to be suspended, by urql throwing a promise, which upon resolving, causes result to either have `.data` or `.error` present.

Could we agree on that desired behaviour at least?

For now, I'm still fighting on giving a correct fix to it, but **I've already delivered a few test scenarios (seen in this PR) that are failing on today's urql** (without my fixes to the implementation).

jtomaszewski:
I've submitted the PR because I was pretty sure I nailed it at the first go, but my `useQuery()` keeps returning an empty result when it shouldn't, even with those fixes. So yeah, the PR isn't ready, but I'd like some feedback from you, that it is something we'd like to fix.

PR #3860, opened 2026-04-01
fix: (react) explicitly overwrite error variable in state

BelgianNoise:
## Issue

The `error` variable on the internal state would not clear once the query was refetched and did not receive any errors -> once an error occured, it would live in the `useSuery` state until it was overwritten by another error, but `error` would be defined indefinitely.

## Summary

This pull request introduces a small update to the `computeNextState` function in `state.ts`. The change ensures that the `error` property from the result is now explicitly included in the computed next state object.

## Set of changes

- set the `error` variable explicitly on the next state

### DIFF
--- a/.changeset/bright-jobs-shout.md
+++ b/.changeset/bright-jobs-shout.md
@@ -1,0 +1,5 @@
+---
+'urql': patch
+---
+
+don't persist error state of old gql responses after refetch in react state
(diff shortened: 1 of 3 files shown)


JoviDeCroock (approved):

ohanslik-vendavo:
This has been released as patch 5.0.2 but it broke one of our tests so it was probably a breaking change.

PR #3868, opened 2026-05-03
feat: add broadcast api based cross tab sync

m1212e:
## Summary
This is an attempt of syncing the cache between tabs via the Broadcast Channel API. I found it useful while operating an app that uses two tabs as a control and presentation view. Without, the optimistic mutations made in the control view do not immediately show up in the presentation view. This is more a proof of concept to hear your opinion on this strategy, looking forward for your feedback! It is still work in progress.

### DIFF
--- a/.changeset/full-plants-bathe.md
+++ b/.changeset/full-plants-bathe.md
@@ -1,0 +1,5 @@
+---
+'@urql/exchange-graphcache': patch
+---
+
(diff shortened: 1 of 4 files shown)


JoviDeCroock:
This should probably be its own exchange that listens for cross-tab interactions and replays results imho

m1212e:
https://github.com/m1212e/urql-crosstab-sync

